PR #431, opened 2018-10-10
Have Mongoose sync indexes with Mongo on boot

jesstelford:
This is a PR in two acts:

### Act One: The meat 🥩

https://github.com/keystonejs/keystone-5/pull/431/commits/8ddb5e32f8e2d404d0931aa3f1ba1cac4c95cd3d?w=1

Here we setup Mongoose with a specific magic incantation:
- Enable `autoCreate: true` on collections. This ensures that Mongoose creates the collections at `.init()` time, rather than on first write. This is required so subsequent calls to `listIndexes()` don't fail (because there's no collection to read indexes from)
- Upon list creation, we schedule a call to Mongoose's `.init()` call (a _setup task_), which will happen after mongoose has connected to the database. Within the `.init()` call, collections will be created, and indexes initialized where necessary. NOTE: This will only ever create new indexes. It does _not_ update indexes no matter how much they've changed since Mongoose was last run.
- We chain onto the `.init()` a call to `.syncIndexes()` which triggers Mongoose to do a diff of indexes, and forcibly update them (ie; drop them and recreate them). This is necessary for production databases where, for example, we add a `unique` constraint.

### Act Two: The memory test 🤔

https://github.com/keystonejs/keystone-5/pull/431/commits/b39525fa453136df44dd4347aeb820900fda523c?w=1

The above Mongoose process highlighted issues with our test suite involving timing issues around adding/updating/dropping data from databases shared across multiple tests.

When Jest runs, it tries to parallelize as many tests as possible, which means it's possible for multiple tests to try add/update/drop data from the same DB, and one or the other test will fail (depending on which one calls `dropDatabase()` first).

The big change here is to ensure all tests that deal with a database connection have their own unique database they can do whatever they want with. To avoid thrashing the disc, we do this with `mongo-memory-server` which spins up an actual instance of `mongod`, but uses an in-memory storage engine.

## Notes

1. Whilst Mongoose does its best to sync indexes, if you try to set a `unique` constraint on a field which has duplicate data within a collection, then _Mongo_ will silently ignore your attempts, and there's nothing Mongoose can do about that. In that case, you need to do a [data migration](https://github.com/keystonejs/keystone-5/issues/299) first, then enable the `unique` constraint.
2. This diff is best viewed with [whitespace disabled](https://github.com/keystonejs/keystone-5/pull/431/files?w=1) (_thanks Prettier!_).
1. After this PR is merged, it's safe to cleanup your local test databases with a `mongod` script like so:

```
$ mongo
> use admin
> db = db.getSiblingDB("admin");
> dbs = db.runCommand({ "listDatabases": 1 }).databases;
> dbs.forEach(function(database) { if (database.name.includes('ks5-testdb-c')) { print(database.name); db = db.getSiblingDB(database.name); db.dropDatabase() } });
```

### DIFF
--- a/packages/adapter-mongoose/index.js
+++ b/packages/adapter-mongoose/index.js
@@ -88,7 +88,6 @@ class MongooseAdapter extends BaseKeystoneAdapter {
     if (debugMongoose()) {
       this.mongoose.set('debug', true);
     }
-    this.mongoose.set('autoCreate', true);
     this.listAdapterClass = this.listAdapterClass || this.defaultListAdapterClass;
   }
 


molomby:
> .. if you try to set a unique constraint on a field which has duplicate data within a collection, then Mongo will silently ignore your attempts, and there's nothing Mongoose can do about that.

I'm _pretty_ these errors come out as events that you can create listeners for. That's [what we did in KS4](https://github.com/keystonejs/keystone/blob/master/lib/list/register.js#L70) (although in that case, by "capture" I mean "log then forget about")..

Issue #744, opened 2019-02-27
Performance Audit


jossmac:
Our lighthouse score is pretty bad...

![Lighthouse Performance Insights](https://user-images.githubusercontent.com/2730833/53531290-ea875d00-3b46-11e9-8cd5-50d3f0f12c3f.png)

PR #749, opened 2019-03-01
[Proposal] Lazy load views with suspense

mitchellhamilton:
Related to #744 

Note that this PR should be merged after ~~#748~~ #751


mitchellhamilton:
I found out why some of the access control tests were failing, the tests were looking for the plural version of the list name in the title on the list page and before this PR, it found the plural version because it checked the title before the data loaded but I think this PR subtly changed timing things which can't be be seen by people but made the data finish loading before cypress could find the title since once the data loads, the title changes to `1 singular list name`. So I fixed it by changing the tests from searching for the plural to the singular(and if it found the plural, it would still work).

PR #874, opened 2019-03-19
[Proposal] Alternate API for configuring Keystone

JedWatson:
I've been wanting to take a look at how Keystone is configured and started, so I had a go at refactoring the blog demo project against a "fantasy API".

Treat this as a conversation starter 🙂 

My thinking:

* It moves away a bit from the magic CLI interface. I'm not sure where to land on that, there were a few things that weren't obvious to me with the way that works that I think we can make a bit more explicit without reintroducing boilerplate.
* I separated out the lists into their own schema file, because I think that makes sense. But the initialisation and server config stuff now lives together.
* I also pulled (some of) the config out into separate file. Having the CLI look at `index.js` and also importing it to get the static paths really confused me for a bit, so I think this is clearer.
* This includes a separate public graphQL API that would only expose a subset of the schema to users who weren't authenticated and enforces some permissions. I think that could be taken further, but also don't like the duplication. We need to do something here though; right now the client app is using the Admin schema and there's no auth.
* I don't like how unclear it is what happens with the authStrategy (related to #863). It obviously blocks the Admin UI, but what does this mean for using the same session in the next app? This could use some work.
* Related to that, this whole setup puts the static resources pretty low in the middleware stack, which is super inefficient


JedWatson:
Most of this has been solved in the blog demo project, or captured elsewhere now.

PR #907, opened 2019-03-22
make twitter auth work again

gautamsi:
closes #801 

Looks like it broke my fork and its upstream link when this repo went public. I can not even open that pull request in browser, says error 500. 

adding again.

This PR only fixes existing twitter auth functionality. I have also removed `oauth-libre` the only GPL dependency for this repo.


jesstelford:
So happy this landed! Thank you for the hard work @gautamsi ♥️

PR #909, opened 2019-03-22
Class-up the create-keystone-app CLI output

jesstelford:
## What it looks like now

create-ks-app-improvements

### With an error

Screen Shot 2019-03-23 at 12 37 31 am

## What it looked like before

create-ks-app-old

### With an error

Screen Shot 2019-03-23 at 12 44 33 am


packages/create-keystone-app/bin/cli.js
```diff
@@ -93,31 +100,31 @@ if (args._.length === 0 || name === '') {
...
+  .then(({ projectDir, hasYarn }) => {
     console.log();
+    console.log(endent`
+      🎉 KeystoneJS app created in ${chalk.bold(projectDir)}
+
+      ${chalk.bold('Get started:')}
+
+      ${chalk.yellow.bold(endent`
+        cd ${projectDir}
+        ${hasYarn ? 'yarn' : 'npm run'} start
```
gautamsi: do we need `npm run` for `start`?

gautamsi: LGTM

PR #910, opened 2019-03-22
Put clickable URLs from quickstart in console

wesbos:
Feel free to close this one, but having the URLs clickable from your terminal is really helpful and I think it really shows the magic of how fast you can get up and running with keystone. 

Screen Shot 2019-03-22 at 9 56 43 AM

### DIFF
--- a/demo-projects/todo/server.js
+++ b/demo-projects/todo/server.js
@@ -12,7 +12,13 @@ keystone
     return server.start();
   })
   .then(({ port }) => {
-    console.log(`Listening on port ${port}`);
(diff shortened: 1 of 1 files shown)


jesstelford:
Yes! This is an excellent idea!

We do actually have this kind of output, but it's disabled by default on the demo project because we've got `DISABLE_LOGGING=true` set in the `package.json`:

Screen Shot 2019-03-23 at 1 16 36 am

wesbos:
Oh that is even better :)

PR #912, opened 2019-03-22
Social login improvements with facebook and twitter both fixed

gautamsi:
continuation to #164 

Fixed Facebook login.
This includes fixed Twitter login from - #907 (original #801)

tested working with facebook and twitter. 

instead of master PR to `social-login-improvements` branch so that any cleanup needed can be done

Also merged latest master

link to diff with master: https://github.com/keystonejs/keystone-5/compare/master...gautamsi:social-login-improvements?expand=1


gautamsi:
Will open another PR with facebook only with master

PR #927, opened 2019-03-25
Rearranging field views

JedWatson:
I moved the field views out of `src` into their entrypoints, which simplifies the directory structure a bit and hopefully makes it easier for contributors to get into.

A few limitations here, which I've talked to @mitchellhamilton about and can wait for more signficiant updates to [preconstruct](https://github.com/preconstruct/preconstruct#readme) but to call them out:

* I had to disable the `import/no-extraneous-dependencies` eslint rule because it gets confused by the entrypoint package.json
* It would be nice if we could move the `dist` files up and out of each entry point directory into the package-level one
* I remain convinced we could automate a lot of the stuff going on here 🙂


jesstelford:
To help get it straight in my head, do you have a before/after of the folder structure? (Screenshots are fine)

packages/fields/types/CalendarDay/views/Cell/Cell.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
mitchellhamilton: Maybe we should use [ESLint's overrides option](https://eslint.org/docs/user-guide/configuring#disabling-rules-only-for-a-group-of-files) to disable the rule for all the field type views? Using preconstruct also essentially has the same effect as using this rule since it'll throw an error if any node module is imported that isn't specified in its packages' dependencies or peerDependencies

mitchellhamilton:
I think an API like what's shown below would work. It would probably mean creating a fork of preconstruct that would just be for building field types since it would have to customise rollup in ways that I don't want to let consumers of preconstruct do.

This is an example of what it would look like with a single field type package, outside a monorepo, wdyt?
```jsx
// src/index.js
import { importView } from '@keystonejs/build-field-types'

export default {
  ...otherFieldTypeStuff,
  views: {
    // some magic babel stuff would be involved with this
    // this would create a chunk in the dist folder
    // and from an api perspective, it returns the path to the field
    Field: importView('./views/Field')
  }
}
// src/views/Field.js

export default () => {
  return 'some cool react component stuff'
}

// package.json
{
  "name": "my-cool-field-type",
  "main": "dist/my-cool-field-type.cjs.js",
  "module": "dist/my-cool-field-type.esm.js",
  "scripts": {
    "build": "field-types build"
  },
  "dependencies": {
    "@keystonejs/build-field-types": "*"
  }
}
```

packages/fields/types/CalendarDay/views/Cell/Cell.js
JedWatson: I'd actually rather leave those comments in-place until we decide whether we will disable the rule permanently for these files and depend on preconstruct's errors, or work out a way to make the rule compatible with how preconstruct does entrypoints.

They're a bit in-your-face, and I think that's a good thing.

JedWatson:
@jesstelford 

image

became:

image

@mitchellhamilton that looks 😍 I love it. Although I'm not sure what you mean by

> I’m not sure how this would work with pages or etc. since it relies on being able to use our build system on node code so maybe it’s not the right approach.

I think it's fine to have a highly specialised solution for field types, they're pretty wack because of all the constraints we have on them (including internal per-type, per-view code splitting). Pages and other Admin UI extensions hopefully we can make much simpler.

In the meantime I'd like to land this PR asap if you're happy to approve it, because it touches a _lot_ of files and will go stale really easily; it's a good incremental improvement to the current setup imo while we work on a longer-term solution.

Issue #842, opened 2019-03-15
[RFC] Delay List initialization until keystone.conenct(), Helps plugin/extension scenario


gautamsi:
## Custom Field option to be used by plugins/extension

Currently there is no documented way to add custom field outside or `keystone.createList` probably you can not do that either due to how they are tightly coupled with schema generation.

While working on Twitter login fix (in #801) I was thinking about how to achieve the scenario on #331.

A plugin/extension point needed for such scenario. One of them is creating custom fields in lists. For example, we may not need to code twitter (or facebook e.t.c) specific fields in User list, they could be generated by  specific `AuthStrategy`. (This scenario may even cover the `PasswordAuthStrategy` to create password and username fields)

### Caveat in current code:
List fields are initialized inside Keystone `createList` method. Once created, you may not (or should not) change the list fields, all the meta information is finalized during `list.initFields`. 

If we can delay the execution of this until `keystone.conenct()`/`server.start()` then any plugin/extension can have the opportunity to add related fields information in the list.

I tried to make twitter login work again in #801. went ahead and added another commit 08375b5d5cd49bdbf93101ebaaa5b795966d599d and crated a [draft pull request  in my own fork for reference](https://github.com/gautamsi/keystone-5/pull/1) . 

### How did I achieve this?
- move `list.initFields` call inside `keystone.connect`
- Had to make WebServer package delay the app.use commands which needs initialized fields, moved it to another function and called it right after `keystone.connect()` inside `server.start()`
- Also had to create a processing queue for `app.use` in web server. this is needed to correct the order of middlewares and routes. 
- queued app.use call instead of calling it directly inside Twitter Auth Strategy and twitter.js in test project. 
- removed twitter fields from "User" list, also added a custom text field inside server.js of test project.

I am sure someone can figure out smarter way to achieve this. I am just putting this here for starting discussion in this direction. 

here is the result of this experiment:
- twitter auth still works, 
- fields now generated and added by TwitterAuthStrategy rather than inside User list creation.
- new fields are visible in admin ui and graphiql

## Possibility
1. Plugin/extension can be created in userland
2. Plugins can have their own workflow like "Publishing" scenario"
3. Plugins can create custom field type and corresponding view bundled with them

### Admin Ui enhancement needed
1. Add marking that this is plugin generated field (tooltip or info button which can pop out plugin information)
2. plugin may have option to make it read only

**This will also help in plugins adding their own custom field Type and views**

### Screenshots:

[...]

PR #929, opened 2019-03-25
Adding TinyMCE Wysiwyg Field Type

### LINKED ISSUE #928: Add TinyMCE Wysiwyg Field
Separate to #616, I want to add a WYSIWYG html field for the Admin UI.

It'll be useful while the content field is under development, and longer term helpful for transitioning from Keystone 4 or any other CMS that stores content in HTML.

The initial version should be a basic implementation of [TinyMCE 5](https://www.tiny.cloud/) and we should skin it a bit to fit in with the Admin UI style like we did in Keystone 4.

JedWatson:
### Ready for Review

Closes #928 

This adds a new field type package that renders a TinyMCE Editor in the Admin UI. Behind the scenes it inherits the implementation from the `Text` field.

There are a few cleanup tasks, which I'll open a separate issue for.

---

### Original Content

@mitchellhamilton I'm having trouble getting the `preconstruct` setup here to work. Would you mind taking a look to see if you can work out what I've missed, or if there's an underlying issue?

I get this error:

```
$ preconstruct build
🎁 info building bundles!
🎁 error @keystone-alpha/fields-wysiwyg-tinymce/views/Field main field is invalid
```

However I **think** I've set it up the same way you have for the entry points in the `fields` package. The paths look valid:

```json
{
  "main": "dist/fields.cjs.js",
  "module": "dist/fields.esm.js",
  "preconstruct": {
    "source": "./Field"
  }
}
```


mitchellhamilton:
The dist file names are derived from the package name(it's the package name but if it has a scope, the scope is removed) so for this, the filenames would be `dist/fields-wysiwyg-tinymce.cjs.js` and `dist/fields-wysiwyg-tinymce.esm.js`

This problem can also be fixed by running `yarn preconstruct fix` but I just checked to make sure it would work and noticed that it added a newline to every entrypoint package.json(even though it didn't change them in any other way) which isn't a great experience so I fixed that and did a release so if you update to `preconstruct@0.0.51` and then run `yarn preconstruct fix`, it'll fix it for you.

There are two things that I want to do in preconstruct to make this a better experience in the future for other people.

1. Show the incorrect value for the field and what the value should be
2. For any errors that are fixable with `preconstruct fix`, say that the error is fixable with preconstruct fix below the error message

packages/fields-wysiwyg-tinymce/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@keystone-alpha/fields-wysiwyg-tinymce",
```
jesstelford: Bikeshed: should the name be singular, since there is only one field in this package?

```
"name": "@keystone-alpha/field-wysiwyg-tinymce",
```

packages/fields-wysiwyg-tinymce/views/Field/Field.js
```diff
@@ -0,0 +1,51 @@
...
+            justifyContent: 'space-between',
+          }}
+        >
+          {field.label}{' '}
+          {!canRead ? (
+            <ShieldIcon title={error.message} css={{ color: colors.N20, marginRight: '1em' }} />
+          ) : null}
+        </FieldLabel>
+        <FieldInput>
+          <Input
```
jesstelford: This is where you'll put tinyMCE, right?

gautamsi:
https://github.com/keystonejs/keystone-5/issues/928#issuecomment-476346462
> How about a provider/adapter system with this field type? I am sure people would want to use something like Trumbowyg and Quill     
if there is adapter/provider system, dev can add their own editor without creating whole new FieldType

packages/fields-wysiwyg-tinymce/package.json
JedWatson: Good question. I went back and forth myself on the package name a few times.

Settled on `fields` partly just becuase it makes the structure read more coherently in the `packages` directory. Also because some packages may contain multiple field types (e.g I expect `CloudinaryImage` and `CloudinaryImages` would go in the same `fields-cloudinary` package) so it might be nice to keep the naming convention consistent.

JedWatson:
> The dist file names are derived from the package name(it's the package name but if it has a scope, the scope is removed) so for this, the filenames would be dist/fields-wysiwyg-tinymce.cjs.js and dist/fields-wysiwyg-tinymce.esm.js

Right! ok that makes sense. If you can detect this is why it failed, might be nice to explain that in the error itself (or link to docs, etc)... was a bit of a dead end otherwise. Maybe even hint to run `yarn preconstruct fix`?

Related: I actually nearly went through renaming the dist file names for the entries in the `fields` package so they matched the field type and entry names (e.g `Text.Cell.esm.js` would make more sense than `fields.esm.js` for the `fields/types/Text/views/Cell` entry. 🤷‍♂️ maybe we should relax this restriction or something?

Anyway, thanks! looks like it worked. Also, this was really nice:

```
🎁 ?    @keystone-alpha/fields-wysiwyg-tinymce Babel helpers (functions inserted by babel transforms) should be imported from a @babel/runtime {...}
```

mitchellhamilton:
>Right! ok that makes sense. If you can detect this is why it failed, might be nice to explain that in the error itself (or link to docs, etc)... was a bit of a dead end otherwise. Maybe even hint to run yarn preconstruct fix?

Yup, that's the plan.

> Related: I actually nearly went through renaming the dist file names for the entries in the fields package so they matched the field type and entry names (e.g Text.Cell.esm.js would make more sense than fields.esm.js for the fields/types/Text/views/Cell entry. 🤷‍♂️ maybe we should relax this restriction or something?

I definitely think there needs to be some way to have different filenames, especially between entrypoints in a package. I don't think it should just be letting users customize it directly though since enforcing it enables really awesome things like auto fixing and initializing, I'm not entirely sure what it should look like yet though. Right now, what I'm thinking of is the name that it is currently + path to entrypoint but `/` replaced with `-`, so for example for `fields/types/Text/views/Cell` making it, `fields-types-Text-views-Cell.esm.js`. The other thing I'm thinking of is letting people specify a name in entrypoints just like you can in packages and using that name for the file name rather than the package's name so the name could be `text.cell`(lowercase to follow npm package naming requirements to make things consistent with when you have a single entrypoint which is the root of the package and so the name comes from the package's name) and a filename would be `text.cell.esm.js`

JedWatson:
@gautamsi 

> How about a provider/adapter system with this field type? I am sure people would want to use something like Trumbowyg and Quill. If there is adapter/provider system, dev can add their own editor without creating whole new FieldType

At this point in the PR, I have TinyMCE working as a field in the Admin UI. If you take a look at the code, you'll see that there's _very little to this package that isn't specific to implementing TinyMCE_.

That's how we've designed field types; easy to change just the minimum when creating you own, and easy to inherit from other types. In this case I'm using most of the functionality of the `Text` field and just customising its view to render a TinyMCE editor.

So I think that's what you're actually asking for here, and adding an adapter/provider pattern would just introduce an unnecessary additional layer of complexity.

Do let me know if you think it could be simplified further though!

JedWatson:
So this is ready for review!

It works, and presents pretty good basic functionality for WYSIWYG HTML in Keystone.

image

There are a few cleanup tasks I'd like to do (like making the toolbar and plugins customisable through options) that I'll create a followup issue for.

packages/fields-wysiwyg-tinymce/README.md
```diff
@@ -0,0 +1,32 @@
...
+In addition to using this field type in a List, for it to work in the Admin UI you'll need to serve the TinyMCE assets from your webserver.
+
+The Field includes a function that sets up the middleware for you. Configure it in your custom server like this:
+
+```js
+const keystone = require('@keystone-alpha/core');
+const WysiwygField = require('@keystone-alpha/fields-wysiwyg-tinymce');
+
+keystone.prepare().then(async ({ server }) => {
+  WysiwygField.bindStaticMiddleware(server);
```
mitchellhamilton: Why is this necessary? It feels really annoying to have to do this because we're breaking out of the API of field types which people are used to.

packages/fields-wysiwyg-tinymce/package.json
jesstelford: Even in the `Cloudinary` example, it still feels singular to me; "This field is related to `Cloudinary`.

BUT! I don't really care all that much. We just need to make sure we're consistent going forward 👍

packages/fields-wysiwyg-tinymce/README.md
jesstelford: I'm ok with this. It's not quite a _field_, but it's a part that's necessary for the field to work. Really, fields can export whatever they like, as long as their main export includes the minimum stuff for KS5 to understand it. If it exports more; so be it 👍

packages/fields-wysiwyg-tinymce/README.md
```diff
@@ -0,0 +1,32 @@
+---
+section: field-types
+title: Wysiwyg
+---
+
+# WYSIWYG HTML Field powered by TinyMCE Editor
+
+This field inserts a string path into your schema based on the `Text` field type implementation, and renders a WYSIWYG editor in the Admin UI using [TinyMCE](https://www.tiny.cloud/)
```
jesstelford:
```suggestion
This field inserts a string path into your schema, and renders a WYSIWYG editor in the Admin UI using [TinyMCE](https://www.tiny.cloud/)
```

The internal implementation probably isn't really important to most users.

packages/fields-wysiwyg-tinymce/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+keystone.prepare().then(async ({ server }) => {
+  WysiwygField.bindStaticMiddleware(server);
+  await server.start();
+});
+```
+
+```DOCS_TODO
+TODO
+```
```
jesstelford: Can remove this `TODO` (the docs look pretty complete to me based on the current implementation 👍 )

packages/fields-wysiwyg-tinymce/index.js
```diff
@@ -0,0 +1,20 @@
+const { Text } = require('@keystone-alpha/fields');
+const path = require('path');
+
+function bindStaticMiddleware(server) {
+  const tinymce = require.resolve('tinymce');
+  const tinymcePath = tinymce.substr(0, tinymce.lastIndexOf('/'));
+  server.app.use('/tinymce', server.express.static(tinymcePath));
```
jesstelford: This doesn't take into account any `adminPath` the user may have set (for example; in the blog project, it's set to `/admin` (the default)).

This is non-blocking. Have added it to the list of follow-up tasks in #936.

packages/fields-wysiwyg-tinymce/views/Editor.js
```diff
@@ -0,0 +1,117 @@
...
+
+// Tell TinyMCE where its assets are
+tinymce.baseURL = '/tinymce';
+
+const defaultOptions = {
+  // min_height: 80,
+  autoresize_bottom_margin: 20,
+  branding: false,
+  menubar: false,
+  // toolbar_drawer: 'floating',
```
jesstelford:
```suggestion
```

packages/fields-wysiwyg-tinymce/views/Editor.js
```diff
@@ -0,0 +1,117 @@
+import React, { useEffect, useRef } from 'react';
+import { Global, css } from '@emotion/core';
+import PropTypes from 'prop-types';
+import tinymce from 'tinymce/tinymce';
+
+// Tell TinyMCE where its assets are
+tinymce.baseURL = '/tinymce';
+
+const defaultOptions = {
+  // min_height: 80,
```
jesstelford:
```suggestion
```

packages/fields-wysiwyg-tinymce/views/Editor.js
```diff
@@ -0,0 +1,117 @@
...
+          editor.on('change keyup setcontent', handleChange);
+        });
+      },
+    };
+    tinymce.init(options);
+
+    return () => {
+      tinymce.remove(editorRef.current);
+    };
+  }, []);
```
jesstelford:
```suggestion
  }, [this.props.autoFocus, this.props.plugins, this.props.toolbar, this.props.value, this.props.isDisabled]);
```

These are external variables which the editor relies on - we'd want to remount it if they change.

packages/fields-wysiwyg-tinymce/views/Field/Field.js
```diff
@@ -0,0 +1,50 @@
...
+            display: 'flex',
+            flexDirection: 'row',
+            justifyContent: 'space-between',
+          }}
+        >
+          {field.label}
+        </FieldLabel>
+        <div css={{ display: 'flex', flex: 1 }}>
+          <Editor
+            value={canRead ? value : undefined}
```
jesstelford:
```suggestion
            value={value}
```

Line 25 means this can never be false at this point.

packages/fields-wysiwyg-tinymce/views/Editor.js
```diff
@@ -0,0 +1,117 @@
...
+  useEffect(() => {
+    if (!editorIsReady() || typeof value !== 'string' || value === contentRef.current) return;
+    editorRef.current.setContent(value);
+  }, [value]);
+
+  // Update the Editor mode when the isDisabled prop changes
+  useEffect(() => {
+    if (!editorIsReady()) return;
+    editorRef.current.setMode(isDisabled ? 'readonly' : 'design');
+  }, [isDisabled]);
```
jesstelford:
```suggestion
  }, [isDisabled, editorRef.current]);
```

If the editor remounts, we want to re-apply the `isDisabled` option.

packages/fields-wysiwyg-tinymce/views/Editor.js
```diff
@@ -0,0 +1,117 @@
...
+    return () => {
+      tinymce.remove(editorRef.current);
+    };
+  }, []);
+
+  // Update the Editor content when the value prop changes
+  useEffect(() => {
+    if (!editorIsReady() || typeof value !== 'string' || value === contentRef.current) return;
+    editorRef.current.setContent(value);
+  }, [value]);
```
jesstelford:
```suggestion
  }, [value, contentRef.current, editorRef.current]);
```

If the editor was remounted, the ref would change, but the value would not get updated correctly (I think). This change should fix that.

packages/fields-wysiwyg-tinymce/views/Field/Field.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/** @jsx jsx */
+
+import { jsx } from '@emotion/core';
+import { Component } from 'react';
+
+import { FieldContainer, FieldLabel } from '@arch-ui/fields';
+import { ShieldIcon } from '@arch-ui/icons';
+import { colors } from '@arch-ui/theme';
```
jesstelford:
```suggestion
```

This isn't used below.

packages/fields-wysiwyg-tinymce/views/Field/Field.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable import/no-extraneous-dependencies */
+/** @jsx jsx */
+
+import { jsx } from '@emotion/core';
+import { Component } from 'react';
+
+import { FieldContainer, FieldLabel } from '@arch-ui/fields';
+import { ShieldIcon } from '@arch-ui/icons';
```
jesstelford:
```suggestion
```

This isn't used below.

packages/fields-wysiwyg-tinymce/README.md
JedWatson: So this was really tricky because TinyMCE basically *needs* to be able to load files from its package at runtime. We either have to get users to sign up for a `tiny.cloud` account and provide an API key, or we serve them ourselves from _somewhere_.

Basically you just can't vendor this stuff from an npm module using webpack, there are just too many issues that are fundamental to TinyMCE's architecture and history that prevent that from being possible - at least as far as they've got in v5. I spent _hours_ trying not to need this and tl;dr failed.

(to be clear; their official recommendation in the docs is to _literally copy the files when you download the package, into somewhere in your repo that static assets are hosted_. This is not something I want to do, so consider the solution here a middleground.)

Now that the field is decoupled from keystone, the admin ui, and the webserver, it's harder than it was in Keystone 4's monolith (in a good way) to serve random static resources. I was actually really happy with this solution, because it's a neat solution to an otherwise horrid problem.

If the field package exports its own bindings, and they're easy to plug into the server, it means everything is still encapsulated and we don't have issues with updating tinymce / etc like we have in the past.

packages/fields-wysiwyg-tinymce/index.js
gautamsi: Should we make it easy for migration, use `static` path from config? (similar to v4)

packages/fields-wysiwyg-tinymce/README.md
mitchellhamilton: Could you add a comment basically saying what you said here for people looking at this in the future?

packages/fields-wysiwyg-tinymce/README.md
JedWatson: I actually think this is useful information - and I'd like to link to the `Text` documentation because that will eventually describe the graphql schema / filters / etc that are relevant for this field as well.

packages/fields-wysiwyg-tinymce/README.md
JedWatson: I want to add a few more things, so leaving this in place for now (task in #936)

packages/fields-wysiwyg-tinymce/views/Editor.js
JedWatson: I'm OK with this for a few reasons:

* `autoFocus` only matters when the component first mounts
* `value` and `isDisabled` have separate effect hooks that handle when they change
* have added `plugins` and `toolbar` for completeness, but fwiw those are configured in the options of the field when the list is created. so they can never change without a server restart.

packages/fields-wysiwyg-tinymce/views/Editor.js
JedWatson: 👌 nice catch!

packages/fields-wysiwyg-tinymce/views/Editor.js
JedWatson: If the editor re-mounts, the content is set when it initialises. And the synchronisation between content and props is pretty delicate, this introduces an unnecessary update. Going to leave as-is

JedWatson:
Thanks for the comprehensive review @jesstelford!

I think this is good to go now?

packages/fields-wysiwyg-tinymce/README.md
gautamsi: My concept in #842  would help here.  basically postponing field initialization and server preparation to the very end of processing queue. This way we can have `keystone`/`server` instance passed to Field type at the initialization, which would remove all this complexity.

We have to pass keystone/server to `list.initFields()`

packages/fields-wysiwyg-tinymce/README.md
jesstelford: > that will eventually describe the graphql schema / filters / etc that are relevant for this field as well.

Great point 👍

JedWatson:
I _believe_ that the tests here are failing because of an unrelated reason. It's a cypress test that's also failing on master, so I'm going to merge this in 👍

JedWatson:
Scratch that, I ran them again and they went green!

We have test gremlins 😬

PR #930, opened 2019-03-25
Remove getAuxQueryResolvers method from keystone

timleslie:
This method was a bit of a weird outlier that didn't serve much purpose. I've moved it inline to reduce the public API surface area.

(Also brings in the result of `yarn format` on some changelogs).


jesstelford:
Good code-health clean up 👍

PR #931, opened 2019-03-25
[WIP] Add support for multiple ApolloServers

timleslie:
This is an experimental branch to investigate what would be required to implement the ideas in #874.

I'll be playing around with different ideas for how to put all the plumbing in place. Once I have *something* working and we have a better feel for the surface area of the impact of this change I'll loop back around and work out what the final APIs should look like and how they should behave.


MadeByMike:
@timleslie what do you want to do with this? Still current?

timleslie:
This is finally obsolete :-)

PR #934, opened 2019-03-26
add facebook login from #164

gautamsi:
Closes #164 

fixed facebook auth from #164 similar to #907


gautamsi:
anything missing in this before merging?

PR #935, opened 2019-03-26
improve twitter login from #164 code.

gautamsi:
it brings similarities with facebook as in #934 
added signout example.
some more comments to help understand for anyone.


test-projects/twitter-login/server.js
```diff
@@ -33,6 +33,8 @@ keystone
       });
       await keystoneApp.createItems(initialData);
     }
+
+    console.log(`Listening on port ${port}`);
```
timleslie: We don't need this message anymore, now that `createGraphQLMiddleware` and `createDevMiddleware` do a better job of logging the routes they set up.

PR #940, opened 2019-03-27
Add contextual intro text to todo app demo

simonswiss:
- Adds some contextual text to the `todo` demo app, with a link to the admin UI.
- Updates the image in the `quickstart` page from docs to match how the todo app looks

image


simonswiss:
> Looks great! Do you mind cropping the screenshot so it's centred / roughly the same res as the others (even if it cuts off one of the example TODOS)?

@jesstelford I thought i did the same ratio? `1222x836` for the main image, `300x205` for the thumb?

I'll add some padding left just so i can make a centred screen grab :)

simonswiss:
Nevermind @jesstelford - I see what you mean by res - the font size of body copy should be similar so there is more consitency between the screenshots. Updating now, will also do the changes requested by @mitchellhamilton

.changeset/825aeca5/changes.md
```diff
@@ -0,0 +1 @@
+- Updating the template to match the changes in the todo demo app
```
mitchellhamilton: Not really something that's worth updating since the changelogs aren't hugely relevant right now since ks-5 isn't widely used but this comment feels more like a commit message rather than a changelog entry since you can't understand part of the changes that happened to the package unless you have more context surrounding this change. e.g. If I was just a consumer of this package, I would think, what were the changes in the todo demo app?

jesstelford:
Screenshot_20190327-200916

♥️

PR #941, opened 2019-03-27
Fix heroku deployment and add app.json necessary for review apps

mitchellhamilton:
@JedWatson I tried to deploy the blog demo project but it used too much memory so I've left it as the basic test project. We can probably switch to the blog when we statically build the admin ui.


demo-projects/todo/server.js
```diff
@@ -9,7 +9,7 @@ keystone
   })
   .then(({ server }) => {
     server.app.use(server.express.static(path.join(__dirname, 'public')));
-    return server.start();
+    return server.start(process.env.MONGODB_URI);
```
JedWatson: So I *really* want to review this API, but will open a separate issue for that

packages/admin-ui/server/getWebpackConfig.js
```diff
@@ -18,7 +18,15 @@ module.exports = function({ adminMeta, entry }) {
...
       test: /\.js$/,
-      exclude: [/node_modules(?!\/@keystone-alpha\/admin-ui)/],
+      exclude: [
+        /node_modules(?!\/@keystone-alpha\/admin-ui)/,
+        // this only affects things while developing in the monorepo
+        // we do this so we can use less memory on heroku, this uses less memory
+        // since it doesn't have to compile these things with babel
+        // note that the preconstruct aliases are also disabled on heroku to enable this
+        // when we have static builds for the admin ui, we can remove this
+        ...(process.env.HEROKU === 'true' ? [/@keystone-alpha\/field/, /@arch-ui/] : []),
```
JedWatson: Nice. I'd suggest a slight improvement by creating a function like

```js
isHerokuEnv = () => process.env.HEROKU
```

Makes the encapsulation and intent slightly clearer

JedWatson:
Also, maybe squash the git history here? it's pretty wild 😜

mitchellhamilton:
> Also, maybe squash the git history here? it's pretty wild 😜

GitHub's squash and merge is a thing so we don't have to do it ourselves.

packages/admin-ui/server/getWebpackConfig.js
mitchellhamilton: I extracted out `process.env.HEROKU === 'true'` into a constant `isHerokuEnv`, is that what you want? or is there a reason it should be function?

packages/admin-ui/server/getWebpackConfig.js
```diff
@@ -4,6 +4,8 @@ const path = require('path');
 
 const { enableDevFeatures, mode } = require('./env');
 
+let isHerokuEnv = process.env.HEROKU === 'true';
```
JedWatson: `const`?

sorry, being super nit-picky

PR #943, opened 2019-03-27
Admin Tinkering

jossmac:
- more subtle navigation aesthetic


packages/arch/packages/navbar/src/PrimaryNav.js
```diff
@@ -87,11 +86,12 @@ export const PrimaryNavItem = styled(ItemElement)(({ depth, isSelected }) => {
 
     ':hover, :focus': {
       backgroundColor: colors.B.A5,
-      color: colors.N100,
+      // color: colors.primary,
```
jesstelford:
```suggestion
```

packages/arch/packages/navbar/src/PrimaryNav.js
```diff
@@ -87,11 +86,12 @@ export const PrimaryNavItem = styled(ItemElement)(({ depth, isSelected }) => {
...
       textDecoration: 'none',
     },
     ':active': {
       backgroundColor: colors.B.A10,
+      // color: colors.N100,
```
jesstelford:
```suggestion
```

packages/admin-ui/client/components/Nav/index.js
```diff
@@ -81,50 +79,31 @@ const GrabHandle = styled.div({
...
 
   return {
     alignItems: 'center',
-    background: 'white',
+    background: 0,
     border: 0,
     borderRadius: '50%',
-    boxShadow,
-    color: colors.N60,
+    // boxShadow,
```
jesstelford:
```suggestion
```

jesstelford:
Got a screenshot of the changes?

JedWatson:
@jossmac I like where you're going with this.

One little thing I noticed: at smaller widths, the "back to lists" arrow and the navigation collapse icon get uncomfortably close -

image

JedWatson:
On a related note, while you're tinkering:

I don't think the "dots" button on the right of the list search/toolbar row adds a lot of value:

image

We already have a "clear all" button when more than one filter has been added, we don't need two.

And I now that the nav is left-aligned we might as well just always take up the full width with lists and anchor the form to the left-hand side of the screen imo.

Running these ideas by you instead here of just changing it myself in case you disagree 🤷‍♂️

jossmac:
Loads of stuff was frustrating me so I just started tinkering with the code directly. This included opening Sketch and struggling to get anywhere.  Basically, I didn't feel that the interface was cohesive; everything felt _similar_, but didn't tell the same story. Here's some of the changes and an attempt to explain why:

- the sidebar/navigation was overwhelming. i've lightened up the sidebar and wrapped the list table in a white card-like element to let the user know that it's the most important content.
- when looking at tabular data users are likely to want as much horizontal real estate as possible. there's no longer an option to "collapse" the table into a container
- search and filters are related, but columns and create aren't. i think they were just stuck together because it was easy.
  - the columns popout can be accessed from the "more" icon on the right of the column headers.
  - the create button has been moved to the top right, which is the same as the Item view. i've tried to allow a single "call to action" for a given view. the solid green create button added too much noise.
  - the search input was unnecessarily wide. once i fixed that (and the other buttons were moved) there was no need to stack filters underneath. filters still wrap if they need to.
- i worried that not all users would attempt to right-click in hopes of a context menu so it can also be found on the right of any row under the "more" icon.
- pagination is moved out of the way to below the table
- the management interface (update/delete) slides in at the top of the table when a row is selected

There's still quite a bit of cleanup to bring some elements into line after this change, but i have a good idea of how that can be done.

### Before

List View 0


### After

Tinkering

* * * 

@JedWatson @jesstelford @mitchellhamilton @timleslie 

Would love some feedback. If everyone's happy with this direction, I'll continue to cleanup my branch -- as you can imagine there's a bit of mess from moving everything around.

gautamsi:
We may want to move the modified `blog` demo code from this PR into `test-projects` or we may want to implement `examples` folder soon preserving the `blog` and `todo` projects before merging this.

timleslie:
@jossmac Is this PR still active? If it's now obsolete could you close it. Otherwise it would be good to get it up to date with `master` so we don't end up forked too far.

jossmac:
Closing in favour of #1023

PR #944, opened 2019-03-28
Add audiences to the session to help restrict access to different APIs.

timleslie:
This PR introduces the concept of a session `audience`.

In order to avoid schema leakage, we want to be able to restrict access to different routes based on currently authenticated "user". This consideration operates at a level above the schema level access control mechanisms.

There are two sides to this implementation. On the first side, we want to make sure `audiences` can be attached to sessions. `startAuthedSession()` now takes an `audiences` parameter and attaches it to the `req.session`. This value should be an array of strings. This in turn is handled by `populateAuthedItemMiddleware` which makes `req.audiences` available to downstream middlewares. `createSessionMiddleware` has been updated to take an `audiences` argument and use this in its calls to `startAuthedSession()`. The admin UI in turn passes in an `audiences` list of [`admin`], meaning that anyone who has logged in via the admin UI session routes will have that `audience` enabled.

The second side of the implementation involves putting in place middleware to check the session `audiences` and deny access unless the logged in user is part of a specific audience. `restrictAudienceMiddleware()` takes `{ public, audiences }` as an argument. If `public` is true, then everyone is passed through. If `audiences` is specified, then the current session must have at least one `audience` in common with this argument. e.g. if `req.audiences = ['a', 'b', 'c']` and `audiences = ['c', 'd', 'e']` then the user would be let in, because they are a member of the `'c'` audience, which is one of the valid audiences.

The `restrictAudienceMiddleware` middleware is used by `createGraphQLMiddleware` to protect all of the graphql endpoints which it sets up. `createGraphQLMiddleware` now also takes `{public, audiences }` as an argument, which it passes directly to `restrictAudienceMiddleware`. We currently call `createGraphQLMiddleware(..., { public: true} )` when setting up the single graphql endpoint supported by Keystone.

This mechanism will allow us to set up multiple graphql endpoints with different session level access control. The exact final API to be exposed in userland to support this is yet to be finalised.

c.f. #931 #874

### DIFF
--- a/packages/server/WebServer/graphql.js
+++ b/packages/server/WebServer/graphql.js
@@ -4,6 +4,7 @@ const playgroundPkg = require('graphql-playground-react/package.json');
 const chalk = require('chalk');
 const falsey = require('falsey');
 const terminalLink = require('terminal-link');
+const { restrictAudienceMiddleware } = require('@keystone-alpha/session');
 
 const { addDevQueryMiddlewares } = require('./devQuery');
 
@@ -29,8 +30,13 @@ const ttyLink = (text, path, port, version) => {
   }
 };
 
-module.exports = function createGraphQLMiddleware(server, { apiPath, graphiqlPath, port }) {
+module.exports = function createGraphQLMiddleware(
+  server,
+  { apiPath, graphiqlPath, port },
+  { isPublic, audiences }
+) {
   const app = express();
+  const restrict = restrictAudienceMiddleware({ isPublic, audiences });
 
   if (graphiqlPath) {
     if (process.env.NODE_ENV !== 'production') {
@@ -39,16 +45,19 @@ module.exports = function createGraphQLMiddleware(server, { apiPath, graphiqlPat
 
       if (falsey(process.env.DISABLE_LOGGING)) {
         // NOTE: Must come before we setup the API below
+        app.use(devQueryPath, restrict);
         addDevQueryMiddlewares(app, apiPath, graphiqlPath, devQueryPath);
       }
     }
     ttyLink('GraphQL Playground:', graphiqlPath, port, playgroundPkg.version);
+    app.use(graphiqlPath, restrict);
     app.use(graphiqlPath, graphiqlMiddleware(apiPath));
   }
 
   ttyLink('GraphQL API:', apiPath, port);
   // { cors: false } - prevent ApolloServer from overriding Keystone's CORS configuration.
   // https://www.apollographql.com/docs/apollo-server/api/apollo-server.html#ApolloServer-applyMiddleware
+  app.use(apiPath, restrict);
   server.applyMiddleware({ app, path: apiPath, cors: false });
 
   return app;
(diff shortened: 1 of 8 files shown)


packages/server/WebServer/graphql.js
```diff
@@ -29,8 +29,13 @@ const ttyLink = (text, path, port, version) => {
   }
 };
 
-module.exports = function createGraphQLMiddleware(server, { apiPath, graphiqlPath, port }) {
+module.exports = function createGraphQLMiddleware(
+  server,
+  { apiPath, graphiqlPath, port },
+  { public, audiences }
```
jesstelford: `public` 💥 

`isPublic`? It would then match the `isUnique` / `isRequired` API used elsewhere.

timleslie: Yep, `isPublic` is a much better idea 😅

packages/session/index.js
```diff
@@ -160,17 +161,38 @@ function populateAuthedItemMiddleware(keystone) {
...
 
-function startAuthedSession(req, { item, list }) {
+function restrictAudienceMiddleware({ public, audiences }) {
+  return (req, res, next) => {
+    if (public) {
+      // If the session restriction is marked public, we let everything through.
+      next();
+    } else if (
+      req.audiences &&
+      audiences &&
```
jesstelford:
```suggestion
      audiences &&
      Array.isArray(audiences) &&
```

packages/session/index.js
```diff
@@ -160,17 +161,38 @@ function populateAuthedItemMiddleware(keystone) {
...
+      req.audiences.some(audience => audiences.includes(audience))
+    ) {
+      // Otherwise, if one of the session audiences matches one of the restriction audiences, we let them through.
+      next();
+    } else {
+      // If the don't make it through, we simply respond with a 403 Permission Denied
+      res.status(403).send();
```
jesstelford: Bikeshed: A 403 here can possibly leak information about an endpoint existing, etc. It would be good to fall through to a generic 404 page instead.

packages/session/index.js
```diff
@@ -115,7 +115,8 @@ const session = (req, res) => {
 
 const createSessionMiddleware = (
   { signinPath, signoutPath, sessionPath, successPath },
-  authStrategy
+  authStrategy,
+  audiences
```
jesstelford: What's the point of passing this in here? It'd be hard-coded on app start-up, which means it'd be the same for everyone.

Perhaps this is placeholder until we bottom out on an API for specifying what users are in what audiences?

If so, I'd say we don't expose it at this level, instead, we leave it off, and just pass `[]` below in the `signin()` call.

timleslie: `createSessionMiddleware` can potentially be called multiple times by multiple independent concerns. We currently only call it once, but that will change in the very near future.

timleslie:
> I like the concept!
> 
> We could add a setting to the auth strategy constructor which accepts a `audiences` field to look for on the list item, like the `identityField`/`secretField` options:
> 
> https://github.com/keystonejs/keystone-5/blob/master/packages/keystone/auth/Password.js#L12-L13

I've got some thoughts on why this might not be the right idea, but let's discuss them offline when I can walk you through a few bits of the code. 👍

JedWatson:
I'm unclear about the desired solution / API here. Would be great to see a "fantasy spec" for it that illustrates how it gets used, which we can review and discuss.

Not meaning to block, because I don't think I understand this properly, but I'm really uneasy about building something like audiences into the core API if we can achieve the goal with a separate package and higher order functions / plugins.

PR #948, opened 2019-03-28
Demo blog improvements

JedWatson:
I've brought forward some parts of #874 that work now and that I think make the code setting up the blog project clearer.

Hopefully uncontroversial, only thing I'm not totally sold on is the duplication of exporting lists and creating them:

```
const { User, Post, PostCategory, Comment } = require('./schema');

keystone.createList('User', User);
keystone.createList('Post', Post);
keystone.createList('PostCategory', PostCategory);
keystone.createList('Comment', Comment);
```

... it's a bit of repetition, but I think it's ok to be this explicit about all the configuration of keystone in the `index.js` file rather than having a "initLists" function exported from `schema.js`.

### DIFF
--- a/demo-projects/blog/README.md
+++ b/demo-projects/blog/README.md
@@ -8,8 +8,9 @@ The Blog is a great example and boilerplate for more complex, real-world impleme
 
 To run this project, open your terminal and run `bolt` within the Keystone project root to install all required packages, then run `bolt start blog` to begin running Keystone.
 
-The Keystone Admin UI is reachable from `localhost:3000/admin`. To log in, use the following credentials...
-Username: `admin@keystone.project`
+The Keystone Admin UI is reachable from `localhost:3000/admin`. To log in, use the following credentials:
+
+Username: `admin@keystonejs.com`
 Password: `password`
 
 To see an example Next.js app using Keystone's GraphQl APIs, head to `localhost:3000`.
--- a/demo-projects/blog/config.js
+++ b/demo-projects/blog/config.js
@@ -1,0 +1,5 @@
+const path = require('path');
+
+exports.port = process.env.PORT || 3000;
+exports.staticRoute = '/public'; // The URL portion
+exports.staticPath = path.join(process.cwd(), 'public'); // The local path on disk
--- a/demo-projects/blog/index.js
+++ b/demo-projects/blog/index.js
@@ -1,26 +1,11 @@
 //imports for Keystone app core
 const { AdminUI } = require('@keystone-alpha/admin-ui');
 const { Keystone } = require('@keystone-alpha/keystone');
-const {
-  File,
-  Text,
-  Relationship,
-  Select,
-  Password,
-  Checkbox,
-  CalendarDay,
-  DateTime,
-} = require('@keystone-alpha/fields');
-const Wysiwyg = require('@keystone-alpha/fields-wysiwyg-tinymce');
-const { LocalFileAdapter } = require('@keystone-alpha/file-adapters');
 const PasswordAuthStrategy = require('@keystone-alpha/keystone/auth/Password');
 const { MongooseAdapter } = require('@keystone-alpha/adapter-mongoose');
 
-// config
-const path = require('path');
-const staticRoute = '/public'; // The URL portion
-const staticPath = path.join(process.cwd(), 'public'); // The local path on disk
-const getYear = require('date-fns/get_year');
+const { staticRoute, staticPath } = require('./config');
+const { User, Post, PostCategory, Comment } = require('./schema');
 
 const keystone = new Keystone({
   name: 'Keystone Demo Blog',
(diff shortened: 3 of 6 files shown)


mitchellhamilton:
> Hopefully uncontroversial, only thing I'm not totally sold on is the duplication of exporting lists and creating them:
> 
> ```
> const { User, Post, PostCategory, Comment } = require('./schema');
> 
> keystone.createList('User', User);
> keystone.createList('Post', Post);
> keystone.createList('PostCategory', PostCategory);
> keystone.createList('Comment', Comment);
> ```
> 
> ... it's a bit of repetition, but I think it's ok to be this explicit about all the configuration of keystone in the `index.js` file rather than having a "initLists" function exported from `schema.js`.
> 

What if we removed createList and made the Keystone class accept schema as an option which would be an object with the keys being list names and values being the list configs?

JedWatson:
So it would look like

```js
const lists = require('./schema');

  name: 'Keystone Demo Blog',
  adapter: new MongooseAdapter(),
  lists,
});
```

..? I kind of like that. We could also add a `createLists(lists)` shortcut in case for some reason you didn't want to do it in the `init`.

I do kind of like the explicit creation of lists here in the demo though, makes it really clear what's going on. I think I'd be happy to support all the above scenarios. What do you think? is the current state of this PR too much boilerplate?

JedWatson:
Looking at #874 again, this is also exported from `schema.js`:

```js
exports.createPublicAPI = ({ keystone }) => {
  const read = true;
  const create = true;
  return keystone.createGraphQLMiddleware({
    lists: {
      User: { fields: { name: { read }, avatar: { read } } },
      Post: { access: { read: { status: 'published' } } },
      PostCategory: { access: { read } },
      Comment: { access: read, create },
    },
  });
};
```

I like this being colocated with the rest of the schema fwiw. Duplicating fields is bad enough, it should all be put in the same place imo. So we wouldn't just blindly create lists for everything the module exports.

This is all really just userland stuff, but on the other hand it's going to be a reference most people copy when structuring their apps (and we would document it in the guides etc)

mitchellhamilton:
> So it would look like

Yeah, something like that

> I think I'd be happy to support all the above scenarios.

I'm not a fan of doing this, I don't particularly mind what we do specifically but I think with primarily stylistic things like this, we should have a single way to do it so people don't have to think about all the different ways they can do it and force them to make decision that is inconsequential but costs every consumer effort to think about. a la when using prettier, you don't have to bear the cost of making inconsequential stylistic decisions when writing code.

> What do you think? is the current state of this PR too much boilerplate?

I don't think the boilerplate is necessarily that bad but it just feels it's like unnecessary since the information is already communicated. I don't feel that strongly about it though, it's not really boilerplate that I feel is bad anyway, I more just feel like it's a nicer API to pass in the lists. (that's not really based on anything, it just _feels_ like a better API)

> Looking at #874 again, this is also exported from schema.js:
> 
> ```jsx
> exports.createPublicAPI = ({ keystone }) => {
>   const read = true;
>   const create = true;
>   return keystone.createGraphQLMiddleware({
>     lists: {
>       User: { fields: { name: { read }, avatar: { read } } },
>       Post: { access: { read: { status: 'published' } } },
>       PostCategory: { access: { read } },
>       Comment: { access: read, create },
>     },
>   });
> };
> ```
> 
> I like this being colocated with the rest of the schema fwiw. Duplicating fields is bad enough, it should all be put in the same place imo. So we wouldn't just blindly create lists for everything the module exports.

This is what I'm thinking it could look like. I'm not sure if this API is exactly perfect but I like the "make some objects describing the stuff and then pass it in" over "call all the things to add all the stuff".
```jsx

const read = true;
const create = true;

exports.GraphQLMiddleware = {
  lists: {
    User: { fields: { name: { read }, avatar: { read } } },
    Post: { access: { read: { status: "published" } } },
    PostCategory: { access: { read } },
    Comment: { access: read, create }
  }
};

exports.lists = {
  User: {
    fields: {
      // ...
    }
  },
  Post: {
    fields: {
      // ...
    }
  },
  PostCategory: {
    fields: {
      // ...
    }
  },
  Comment: {
    fields: {
      // ...
    }
  }
};

// index.js

const schema = require("./schema");

  name: "Keystone Demo Blog",
  adapter: new MongooseAdapter(),
  schema
});

```

gautamsi:
I have done something similar to this but more inline with v4 but with explicit initialization. 

`models/User.js` can also call it `schemas/User.js`
```js
module.exports = function(keystone) {
  keystone.createList('User', {
    fields: {
      name: { type: Text },
      email: { type: Text, isUnique: true },
      dob: {
        type: CalendarDay,
        format: 'Do MMMM YYYY',
        yearRangeFrom: 1901,
        yearRangeTo: getYear(new Date()),
      },
      password: { type: Password },
      isAdmin: { type: Checkbox },
      avatar: { type: File, adapter: avatarFileAdapter },
    },
    labelResolver: item => `${item.name} <${item.email}>`,
  });
};
```

other models/schemas similarly in their own file. 

`models/index.js`
```js
const User = require('./User');
const Post = require('./Post');
const PostCategory = require('./PostCategory');
const Comment = require('./Comment');

exports.initModels = function(keystone) {
  [User, Post, PostCategory, Comment].forEach(model => model(keystone));
};
```

somewhere in root `index.js`
```js
// other codes
const models = require('./models');
  name: 'Keystone Demo Blog',
  adapter: new MongooseAdapter(),
});
models.initModels(keystone);
// rest of the code
```

jesstelford:
Thinking about this more: this refactor is 100% a thing I can imagine a new user doing, and would encourage it as they build it out into a more full featured application!

And it's definitely a thing I can imagine someone doing as a way to build their understanding of a Keystone app / how best to structure their own app.

But I think it's something the developer should do themselves, not something we should do for them. There's a limit here: if we add more features to the blog demo, the `index.js` file might just get too big and hard to follow. At that point we should think about either taking Zeit's next/examples method where it's a seperate folder per small feature, or break this project out into its own repo with its own documentation and leave it as _one way_ to build a KS5 app.

Issue #827, opened 2019-03-14
react-select 2.4.x causes cypress test failures


timleslie:
We have pinned our dependency to 2.3.0 until we can resolve this issue.

JedWatson:
We're on 3.0.x now as of #1344

@mitchellhamilton should we close this out?

PR #959, opened 2019-03-29
Update react-select to ^2.4.0

timleslie:
#827


dcousens:
(@timleslie why closed?)

Issue #942, opened 2019-03-27
Review server.start args and environment variable handling


JedWatson:
Reviewing #941 I noticed that our `server.start` method takes the MongoDB connection URL as its first argument.

This struck me as really obscure, I never would have expected that. I'd like to see how we can make it clearer - we use a lot of argument destructuring in Keystone so options are self-documenting, and this really breaks that clarity.

In addition, in Keystone 4 we detected common environment variables like `PORT` and `MONGODB_URI`, I think we're not doing that yet in Keystone 5. Unless there's some new opinion about not doing that, I think it's a nice touch that removes boilerplate, and we should start doing it.

Documented here: https://keystonejs.com/documentation/configuration

Specifically:

* `env: process.env.NODE_ENV || 'development'`
* `port: process.env.PORT || 3000`
* `host: process.env.IP || '127.0.0.1'`
* `mongo: process.env.MONGO_URI || 'mongodb://localhost/{db-name}'` and also [all of these](https://github.com/keystonejs/keystone/blob/master/lib/core/initDatabaseConfig.js) (we can probably seriously cull this list, but should support the heroku default at minimum)

JedWatson:
Also probably related to #926

w01fgang:
Yes, I miss them too. Having env variables would be great.
I just played with the demo projects. I was looking for the option `mongo` or similar.
Also, I found that mongoose errors are lost because of unhandled rejection in `packages/utils/src/index.js`. After adding `.catch` after `.then` I found that the database name is taken from the app name even if I explicitly pass it with the mongo connection string to `.start` method.

jesstelford:
We now have `CONNECT_TO` env variable, and `--connect-to` CLI switch.

Issue #926, opened 2019-03-25
Use a global config for capturing CLI args when using a custom server


jesstelford:
We currently log a warning, and ask the user to pass the config variables into the `prepare()` function call:

https://github.com/keystonejs/keystone-5/blob/8b6734a/packages/keystone/bin/commands/dev.js#L24-L40

We should retain the ability to pass the config variables into the `prepare()` function call, but we should also add the ability to specify them via the CLI.

This could take the form of global variables inside `@keystone-alpha/core` with getter/setters and warnings for when there are multiple instances of `@keystone-alpha/core`:

```javascript
let config;
module.exports = {
  setConfig(userConfig) {
    if (config) {
      throw new Error('Unable to set config twice. This can happen when X');
    }
    config = userConfig;
  },
  getConfig() {
    if (!config) {
      throw new Error('Unable to get config. This can happen when two instances of `@keystone-alpha/core` are included in your build. Make sure to Y');
    }
    return config;
  },
};
```

PR #960, opened 2019-03-29
Fantasy blog configuration

JedWatson:
In the spirit of "Fantasy API Proposals", this is an updated configuration for the blog project that I think is much clearer than what we've got now.

One of the main things I'm addressing here is: it keeps annoying me that everything you set up for keystone is really obvious _except the web server_.

I've reduced a lot of boilerplate here by adding `static` and `next` options to the web server config, which would set up express static and next.js middleware respectively.

I think this is a lot less magic, and clean - the only thing I'm not solving for is exporting keystone's config to enable a static admin build. So I'd like to talk about ways we could achieve that... and also what the CLI would do if setting up keystone config looks like this.

@jesstelford @timleslie @mitchellhamilton keen for your feedback on what I'm proposing here. Love it? Hate it? I'm stuck at the airport in Canberra 😭so let me know!


demo-projects/blog/index.js
```diff
@@ -41,9 +45,31 @@ const admin = new AdminUI(keystone, {
...
-  admin,
-};
+const server = new WebServer(keystone, {
+  adminUI,
+  port,
+  static: {
+    [staticRoute]: staticPath,
+    [WysiwygField.staticRoute]: WysiwygField.staticRoute,
+  },
+  next: next({
```
mitchellhamilton: I don't get this bit, would this mean we would have an opinion on what keystone users should use to build their front end? I was under the impression that this was something that keystone-5 was explicitly not trying to do?

mitchellhamilton:
Personally, I feel like the biggest problem we have here isn't that creating the web server isn't obvious because I think the benefits that it brings with making statically building the admin ui easy and targeting serverless nicely, It's more that we're always forcing people to create a custom server file for basically all use cases (I haven't seen a project without one/is it even possible?). Though this is something that's not clear to me, @jesstelford would deploying keystone to a serverless target use the server file?

So, in short, I think we should address more use cases that currently require digging into the server file rather than remove the server file altogether.

demo-projects/blog/index.js
```diff
@@ -41,9 +45,31 @@ const admin = new AdminUI(keystone, {
...
+    dir: 'app',
+    distDir: 'build',
+    dev: process.env.NODE_ENV !== 'production',
+  }),
+});
+
+server
```
jesstelford: Introducing the server here means we're not able to do any actions with the schema which don't require a server, such as:

- Creating a static Admin UI build
- Doing a `production` bundle of the server/admin UI
- Dumping the generated GraphQL schema (eg, I use this to push the schema to Apollo Engine for stats tracking)

jesstelford: It also reduces our ability to evolve the set of things which we can do between initialising the Keystone data and starting the server.

demo-projects/blog/index.js
jesstelford: I'm also confused why we would create an API such as this when exposing the express app enables all this (and unlimited other things) with less learning on behalf of the developer.

jesstelford:
@mitchellhamilton articulated it well: We might want to expose one or two basic options that can be exported from `index.js` to cover the most common cases where using a custom server is necessary but _trivial_.

A good candidate there is serving some static files. Perhaps we do something like:

```
// index.js
module.exports = {
  keystone,
  admin,
  staticAssets: {
    folder: './public's,
    path: '/',
  },
};
```

jesstelford:
> it keeps annoying me that everything you set up for keystone is really obvious except the web server.

Thinking about this in context of something you mentioned earlier that was along the lines of _"It's confusing that we're importing a `keystone` in `index.js`, exporting an instance of that keystone from `index.js`, then importing a completely different `keystone` inside `server.js` which is neither the same import, nor the instance from `index.js`"_.

Might it clear things up for you if we were to rename a couple of things like so?

```javascript
// index.js
const { createKeystoneSchema } = require('@keystone-alpha/keystone');

const schema = createKeystoneSchema({
  name: 'Keystone Demo Blog',
  adapter: new MongooseAdapter(),
});

schema.createList('User', { /* ... */ });

const admin = new AdminUI(schema);

module.exports = {
  schema,
  admin,
};
```

```javascript
// server.js
const keystone = require('@keystone-alpha/core');

keystone.prepare().then(({ server }) => server.start());
```

demo-projects/blog/index.js
JedWatson: Yeah I was surprised when I did this too. But I tried it out and... actually kind of liked it.

I don't mind supporting a few popular "golden paths" of which integrating a next.js app could be one. _I think_. There's an argument to be made against this though.

It's sort of OK if we keep the supported integrations to a reasonable minimum, they introduce no new dependencies, and we're abstracting boilerplate (this next support passes all those tests for me)

> I'm also confused why we would create an API such as this when exposing the express app enables all this

Partly because if you're using next, supporting this and static assets removes the need to customise the server entirely.

Also I'm not as sold that our current solution doesn't require learning. I _myself_ get confused about things like

- Is the session middleware included in the stack before the app is available for customisation?
- How do I add express middleware _before_ the Admin UI middleware? (e.g static assets should almost never come after session initialisation, but that's the only option we have currently)

... etc. Right now I think we're actually in a "worst of both worlds" scenario, in that what _could be config isn't, but developers don't get full control_.

(at least in our demo configurations)

Along the same theme, I'd like to make it _more_ explicit how the middleware is all being set up if you're not using config. i.e if you're going to completely configure your own server, you should probably be responsible for binding the Admin UI middleware to it in the same way you're responsible for binding the Next middleware in the current demo project.

Sorry if I'm coming across as critical of what we've got; I'm just having a hard time working through the mix of static configuration vs. imperative setup that we're currently doing.

JedWatson:
Some really good comments here @jesstelford and @mitchellhamilton, thanks 🙏 

Thinking about this more, I think part of what's getting to me is that we're currently effectively creating a new configuration API for keystone, that the CLI reads. You could really consider the exports from `index.js` the "top level configuration schema" for keystone.

_but_ that includes something called `keystone`. and when I create the _real_ keystone object, I do that by importing `keystone/core` and preparing it.

So I think there's merit in your suggestion to rename some stuff, Jess 🙂 

As noted in #942 and #926, our config setup is a bit of a mess and I think we should prioritise cleaning it up now Keystone 5 is public 😬 

When I see this:

```
module.exports = {
  keystone,
  admin,
  staticAssets: {
    folder: './public's,
    path: '/',
  },
};
```

I get freaked out because _what even is this API_? it feels like just a random set of _things_. There's some webserver config, keystone itself, and the Admin UI all bundled in there. Is this arguments for the webserver? I mean, it effectively is, but it's also arguments for the build.

Renaming `keystone` to `schema` helps a bit, but that object also owns the auth strategy which is not part of the schema; it's related to the Admin UI and the webserver. Maybe part of the problem here is that we need to reshuffle some things internally 🤔

> we're always forcing people to create a custom server file for basically all use cases (I haven't seen a project without one/is it even possible?)

Yup, that's part of what I'm suggesting here with the new `WebServer` options.

Maybe the best next step here is to write up a bunch of escalating use-cases / examples, going from "super simple no config" to "here's how you do everything yourself with no magic or help from us"

demo-projects/blog/index.js
JedWatson: Creating a static build of the Admin UI or generating the GraphQL schema _without starting the server_ is still something I'm trying to work out how to make co-exist with the sort of code I really want to see in `index.js`.

What else is on that list of things we'd want to do?

re:

> Doing a production bundle of the server/admin UI

imo that's something we should be doing automatically when the server gets started (switching some flags based on `process.env.NODE_ENV`). we can hash the config and compare it to the previous build, to know whether we should be regerated or not.

demo-projects/blog/index.js
mitchellhamilton: I agree that we need to have "golden paths" but I’m not sure if having next as one one of them is right because in actual production apps, people will probably deploy their front end independently of keystone.

jesstelford:
Again, I think @ articulated it the best:

> in actual production apps, people will probably deploy their front end independently of keystone.

This is going to be the 75% case. For the others, we have the escape hatch. And once you're at the point of using the escape hatch, you're already familiar with how Keystone works (`index.js` exports certain things / you don't have to create the express server yourself) that the `server.js will not be much of a learning curve (I think).

jesstelford:
> what even is this API? it feels like just a random set of things.

This sounds like a documentation problem to me.

jesstelford:
> that object also owns the auth strategy which is not part of the schema

That's changed since #861 - it's now passed to the Admin UI constructor directly.

jesstelford:
> I myself get confused about things like
> 
> - Is the session middleware included in the stack before the app is available for customisation?
> - How do I add express middleware before the Admin UI middleware? (e.g static assets should almost never come after session initialisation, but that's the only option we have currently)

Thanks to some recent work by @timleslie, we could have a Custom Server API like so:

```javascript
// server.js
const keystone = require('@keystone-alpha/core');

keystone.prepare().then(({ server, keystoneMiddleware }) => {

  // want to add middleware _before_ KS? Do it here.

  server.app.use(keystoneMiddleware);

  Add middleware _after_ KS here.

  server.start();
});
```

That's essentially what we do under the hood anyway.

timleslie:
> Thanks to some recent work by @timleslie, we could have a Custom Server API like so:

Also perhaps worth noting here, the work I did in this area last week deliberately worked from the inside out and didn't really touch the user facing APIs. The various points *inside* the webserver are now much more neatly boxed so that if we want to use them more flexibly we can, but none of the actual internal behaviour or external APIs have changed.

JedWatson:
@jesstelford a few things:

> This sounds like a documentation problem to me.

I was thinking the same at first, but I also think APIs should ideally be as intuitive as possible. IMO if we keep bashing at this ideally we'll come up with something that's clear enough without _requiring_ docs to make sense. I hope 🤞 

> This is going to be the 75% case. For the others, we have the escape hatch.

I don't agree with that %, but maybe my head is still a bit too much in Keystone 4's world. I wrote up #966 so we can talk through all our different expectations here.

> And once you're at the point of using the escape hatch, you're already familiar with how Keystone works (index.js exports certain things / you don't have to create the express server yourself)

I'm seeing people using the `server.js` escape hatch for _most_ use-cases at the moment. Particularly with our current projects. Adding support for static file serving and additional GraphQL endpoints without creating a custom `server.js` will certainly help with that, but I think that's inline with what I'm arguing for here:

* The config should be _more powerful_ (e.g include support for whatever is needed for common use-cases)
* When you bust out of the escape hatch, it should be _less magical_

I think one of the things we're getting stuck on is that I keep trying to start a web server instead of exporting configured instances. So I'll push an update to this branch that includes web server config in `index.js` but doesn't start it - see what you think of that.

demo-projects/blog/index.js
molomby: > Along the same theme, I'd like to make it more explicit how the middleware is all being set up if you're not using config.

Yes to this.

I'm not especially familiar/up to date with this area of Keystone right now so this may make no sense -- could we split the configuration stuff into a separate package? Ie. out of the box, the dev just gets the building blocks (server, session middleware, etc.) and has to put them together themselves, very explicitly, or they can pull in the  "server configurator" package and have a much "nicer" but less powerful interface to get started.

If it worked, this would make it really obvious that the building blocks can be used directly. If you wanted to understand how they were usually used, you'd only need to understand a small amount of code (the configurator).

jesstelford:
@JedWatson @timleslie @molomby & I spoke offline. We didn't fully bottom out on an answer here, but we are moving in the right direction 👍 

Here's one possible answer which @JedWatson & I just landed on (documenting here to open the discussion):

1️⃣ Maintains separation between `index.js` & `server.js` (_"Custom Server"_).
2️⃣ You'd setup UI's/Servers in one _or_ the other of `index.js` or `server.js`. Ie; If you have a Custom Server, you should not export `servers:` from `index.js` (technically you could, it just wouldn't do anything).
3️⃣ It reframes the Admin UI from being "something special" to "Just another frontend"
4️⃣ All the UI's / Servers follow a common API (very similar to how we already do _fields_!)
5️⃣ The custom server is still un-opinionated, and you can really do whatever you want there.
6️⃣ All the servers need a `build` step for preparing for a production deploy
7️⃣ Without a Custom Server, the CLI would become just `index.servers.forEach(server => app.use(server.middleware()))`

### `index.js`
```javascript
const { Keystone } = require('@keystone-alpha/keystone');
const AdminUI = require('@keystone-alpha/serve-admin');
const NextUI = require('@keystone-alpha/serve-ui');
const StaticUI = require('@keystone-alpha/serve-static');

const keystone = new Keystone(/* ... */);

const admin = new AdminUI({
  path: '/admin',
  keystone,
  session: authStrategy.sessionHandler(),
});

const next = new NextUI({
  path: '/'
  session: authStrategy.sessionHandler(),
  nextConfig: { /* ... */ },
});

const static = new StaticUI({
  path: `${__dirname}/static`,
  route: '/public',
});

module.exports = {
  keystone,
  servers: [
    static,
    admin,
    next
  ],
};
```

> _- OR -_

### `server.js`
```javascript
const keystone = require('@keystone-alpha/core');
const path = require('path');
const PORT = process.env.PORT || 3000;
const AdminUI = require('@keystone-alpha/serve-admin');
const NextUI = require('@keystone-alpha/serve-ui');

const { keystone } = require('./index');

keystone.prepare().then(() => {
  const app = express();

  const admin = new AdminUI({
    path: '/admin',
    keystone,
    session: authStrategy.sessionHandler(),
  });

  const next = new NextUI({
    path: '/'
    session: authStrategy.sessionHandler(),
    nextConfig: { /* ... */ },
  });

  app.use(admin.middleware());
  app.use(next.middleware());

  app.listen(3000);
}).catch(error => {
  console.error(error);
  process.exit(1);
});
```

### `@keystone-alpha/next-ui`
```javascript
const next = require('next');
module.exports = class {
  constructor(nextConfig) {
    this.config = nextConfig;
  }

  middleware(server) {
    const nextApp = next(nextConfig);
    server.use(nextApp.requestHandler());
  }

  build() {
    this.next.build({ out: this.config.distDir });
  }
}
```

### `@keystone-alpha/static-ui`

```javascript
module.exports = class {
  constructor(config) {
    this.config = config;
  }

  middleware(server) {
    server.use(express.static(this.config.route, this.config.path));
  }

  build() {
    // noop
  }
}
```

### `@keystone-alpha/cra-ui`

```javascript
module.exports = class {
  constructor(craConfig) {
    /* ... */
  }

  middleware(server) {
    server.use(express.static(this.config.path, this.config.craOutDir));
  }

  build() {
    createReactApp.build({ out: this.config.craOutDir });
  }
}
```

PR #961, opened 2019-03-29
List: Data Hooks

jossmac:
Recently I was tinkering in the `admin-ui` package and it was a nightmare disconnecting UI from logic. 

This would be the first step to making that easier. Super draft; really looking for input as to whether this is a good idea.


jossmac:
> there is a lot of unrelated logic bundled together

when i started unbundling it, loads of stuff _was actually_ related. not saying it needs to stay that way, just that this is the first step of teasing out related logic 🙂

jossmac:
@jesstelford further to above: i want a single call that memoize a list's data/handlers and passes on that slice to its consumers.

jossmac:
> in most cases their quite well isolated

this was not my experience; it was uncomfortably difficult to move elements

jossmac:
> I don't mind if there's logic mixed with UI

likewise! i just want it to be _the right slice_ of logic. i think this will be evident once the *blocks* have been carved out

jossmac:
> Note also that this will conflict with some refactors I did in the next.js branch to get things SSR ready.

What's happening with that? I feel like maybe we shouldn't do anything in the admin-ui package until it's merged.

> In both the <DataProvider> pattern, and the useListData() hook, there is a lot of unrelated logic bundled together, which I think is the main source of the coupling.

Yup, agreed -- this was the first pass at teasing out the right logic. I needed working hooks to get the patterns for decoupling.

jossmac:
@mitchellhamilton I'm not very proficient with hooks yet. Any chance you could help out with this?

packages/admin-ui/client/pages/List/dataHooks.js
```diff
@@ -0,0 +1,392 @@
...
+
+  if (!list) {
+    throw new Error(`No list matching key "${listKey}"`);
+  }
+
+  return list;
+};
+
+/**
+ * URL State Hook
```
jossmac: Now we're getting somewhere 🙂

mitchellhamilton:
> @mitchellhamilton I'm not very proficient with hooks yet. Any chance you could help out with this?

Is there anything in particular you need help with?

jossmac:
> Is there anything in particular you need help with?

I suspect that the current setup is wildly inefficient. Basic review would be a good start 🙂

packages/admin-ui/client/pages/List/dataHooks.js
```diff
@@ -0,0 +1,396 @@
...
+
+  useEffect(() => {
+    if (data[list.gqlNames.listQueryName]) {
+      setItems(data[list.gqlNames.listQueryName]);
+      setErrors(deconstructErrorsToDataShape(data.error)[list.gqlNames.listQueryName]);
+    }
+    if (data[list.gqlNames.listQueryMetaName]) {
+      setCount(data[list.gqlNames.listQueryMetaName].count);
+    }
+  }); // FIXME this gets called SO OFTEN. needs deps e.g. `listKey`, but then we don't get data in time...
```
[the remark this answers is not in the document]
jossmac: 🤷‍♂️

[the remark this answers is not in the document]
jossmac: > something like @keystone-alpha/apollo-helpers

That sounds great

> why flow -> jsdoc?

It was getting late and I needed the information out of my head. Didn't want to get lost when i picked it up again.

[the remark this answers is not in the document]
jossmac: > things will be subtly broken

After replacing the `DataProvider` i'm seeing this happening 😞

jossmac:
Get tests passing (004f6be ) by reverting to `Query` component. I think this is okay for now, perhaps [react-apollo-hooks](https://github.com/trojanowski/react-apollo-hooks) isn't ready for production.

package.json
```diff
@@ -166,6 +166,7 @@
     "react-addons-text-content": "^0.0.4",
     "react-adopt": "^0.6.0",
     "react-apollo": "2.4.0",
+    "react-apollo-hooks": "^0.4.4",
```
mitchellhamilton: This isn't necessary if we're not using it right?

packages/admin-ui/client/index.js
```diff
@@ -1,6 +1,7 @@
 import React, { Suspense, useMemo } from 'react';
 import ReactDOM from 'react-dom';
 import { ApolloProvider } from 'react-apollo';
+import { ApolloProvider as ApolloHooksProvider } from 'react-apollo-hooks'; // FIXME: Use the provided API when hooks ready
```
mitchellhamilton: Is this still being used?

packages/admin-ui/client/pages/List/Management.js
```diff
@@ -29,80 +29,70 @@ type Props = {
...
-  state = { showDeleteModal: false, showUpdateModal: false };
 
-  closeDeleteModal = () => this.setState({ showDeleteModal: false });
-  openDeleteModal = () => this.setState({ showDeleteModal: true });
-  closeUpdateModal = () => this.setState({ showUpdateModal: false });
-  openUpdateModal = () => this.setState({ showUpdateModal: true });
+export default function ListManage(props: Props) {
+  const { onDeleteMany, onUpdateMany, selectedItems } = props;
+  const [deleteModalIsVisible, toggleDeleteModal] = useState(false);
+  const [updateModalIsVisible, toggleUpdateModal] = useState(false);
```
mitchellhamilton: `set*` would probably be better than `toggle*`

packages/admin-ui/client/pages/List/NoResults.js
```diff
@@ -0,0 +1,69 @@
...
+    }}
+    {...props}
+  >
+    <InfoIcon css={{ height: 48, width: 48, marginBottom: '0.5em' }} />
+    {children}
+  </div>
+);
+
+export const NoResults = ({ currentPage, filters, itemCount, list, search }) => {
+  const { onReset } = useListPagination(list.key);
```
mitchellhamilton: I just saw above this that there's `useReset` which returns `onReset` as well(i assume). I'm not a fan of having a bunch of different ways to do the exact same thing because it raises questions like "why did the author use this hook over here and that hook over there?" and "what hook should i use for this new code?"

packages/admin-ui/client/pages/List/Search.js
```diff
@@ -0,0 +1,83 @@
...
+import { A11yText } from '@arch-ui/typography';
+import { LoadingSpinner } from '@arch-ui/loading';
+import { colors } from '@arch-ui/theme';
+
+import { useListSearch } from './dataHooks';
+
+export default function Search({ isLoading, list }) {
+  const { searchValue, onChange, onClear, onSubmit } = useListSearch(list.key);
+  const [value, setValue] = useState(searchValue);
+  const ref = useRef();
```
mitchellhamilton: I'd call this `inputRef` so it's clear what element this ref is for.

packages/admin-ui/client/pages/List/SortSelect.js
```diff
@@ -1,12 +1,75 @@
...
+      direction = invertDirection(sortValue.direction);
+    } else if (altIsDown) {
+      direction = 'DESC';
+    }
+
+    handleSortChange({ field, direction });
+    popoutRef.current.close();
+  };
+
+  const cachedOptions = useMemo(() => list.fields.map(({ options, ...field }) => field));
```
mitchellhamilton:
```suggestion
  const cachedOptions = useMemo(() => list.fields.map(({ options, ...field }) => field), [list]);
```

jossmac:
[DeepScan](https://deepscan.io/dashboard/#view=project&tid=3252&pid=4799&bid=38273&subview=pull-request&prid=95938) seems a bit confused regarding `jsx`.

@timleslie @mitchellhamilton any ideas?

packages/admin-ui/client/pages/List/SortSelect.js
jossmac: Not sure on the nuances of deps here, but is using an object the right way to go? What about `list.key`?

mitchellhamilton: Using an object is completely fine. Using `list.key` would technically work in this exact case but you should use whatever values are used in the function. `list.fields` would also work but since we know that `list.fields` will only change when `list` changes, using `list` is probably better since it's shorter.

packages/admin-ui/client/pages/List/NoResults.js
jossmac: That's fair enough. The `useReset` hook returns a function that resets all search, filters, columns, sort.

I'll remove `onReset` from pagination --> folks can just write their own from `onChange`.

packages/admin-ui/client/index.js
jossmac: It is, not sure if that's a problem, though. There's a chain of hooks:

```
useListSearch/useListPagination > useListItems > useListQuery > useQuery
```

I'd like to move away from the render component at some point and this seems to work atm.

packages/admin-ui/client/pages/List/Management.js
jossmac: Sure, no worries

packages/admin-ui/client/pages/List/Search.js
jossmac: Sure, no worries

PR #963, opened 2019-03-31
Update external dependencies to silence warnings at install

timleslie:


JedWatson:
@timleslie haha was writing a comment about the eslint errors, and before I finished you pushed a change fixing them. nice 👍

timleslie:
Good call.

JedWatson:
@timleslie I added some commits to fix the errant `FilterTests` case because that'll actually fail to be run on a case-sensitive FS. Good to go now 👌

PR #964, opened 2019-03-31
[RFC][WIP] Restructure packages

timleslie:
The internal file structure of our various packages in the `packages` directory is ... colourful :-)

I propose that we move towards a common file structure (where appropriate) within packages.

Screen Shot 2019-03-31 at 1 54 17 pm

This will make it easier to find our way when working across multiple packages and also make it easier to identify if something (e.g. `tests`) is missing from a particular package.

I also propose that all of our packages use named exports within their top level `index.js` file. This will provide consistency, as well as the ability to expand our APIs as required.

Furthermore, I propose that the top level `index.js` files should **only** contain the `module.exports` statement, and no functional code. This will make it easier to identify the API that each package is exposing.

The code in this PR gives a rough cut of the changes which would be required. There are a bunch of quirky corner cases which need to be resolved for some of the packages to work. As such, I would want to create a separate PR for each package, in order to make the changes easily reviewable, and also to make sure we get our changesets right, as a number of packages will require major bumps.

There are a few packages (`utils`, `field-view-loader`, etc) which I haven't really changed in this PR as they have potentially interesting interactions with `preconstruct`, but the aim would be to have them all fit the same structure eventually.


mitchellhamilton:
I like this, big +1 to the consistency. If we're going to restructure the server side packages, I feel like we should discuss whether we should build all the server side packages with preconstruct if we're going to build the utils package with it. I don't 100% think it's the right thing to do but I think it's worth discussing.

The reasons why I think it could be a good idea:

1. Consistency with front end packages
2. Ability to use ESM and new syntax stuff that isn't in all the versions of Node that we support.

The reasons why I think it could be a bad idea:

1. It could add complexity for the server side packages since people would have to think about a build system when working on the server side packages(though I feel like we already have that complexity since we're already building one server side package)

mitchellhamilton:
> I also propose that all of our packages use named exports within their top level index.js file. This will provide consistency, as well as the ability to expand our APIs as required.

I also strongly like this.

timleslie:
>If we're going to restructure the server side packages, I feel like we should discuss whether we should build all the server side packages with preconstruct if we're going to build the utils package with it.

I agree this a conversation worth having, and with your pros/cons. I think we should hold off having it until after the restructure is complete, mostly for the sake of not muddying the waters too much in this particular block of work. I think once the restructure is complete it will be easier to envisage what the implications of using `preconstruct` would be in the context of more neatly organised packages.

JedWatson:
👌 nice, big fan of this!

> I propose that the top level index.js files should only contain the module.exports statement, and no functional code.

Strongly agree.

> If we're going to restructure the server side packages, I feel like we should discuss whether we should build all the server side packages with preconstruct

Strongly uncertain. We need to look at the implications for the dev loop in detail. I'm also open to introducing TypeScript (or at least, discussing the pros and cons) which would make something like this less artificial for the server-side code, but that's a whole different conversation.

Might be best to leave that whole conversation (including preconstruct) out of scope for this re-org.

Given not all package are consistent (in terms of build, etc) right now, I think it would be OK to have _more than one_ type of package structure (but all types should be defined, and it should be clear which type each package is consistent with)

Specifically, having a `src` directory implies to me that the files are "source", and need to be built for use by preconstruct and they'll end up in a "dist" directory. That's not true of our Node.js packages, but is true of our client-side or universal packages.

If you want an equivalent of `src` for server-side packages, `lib` is commonly used in Node.js projects for containing modules that are imported directly.

We also have some packages (like the field types and Admin UI) which contain two or more groups of files, some of which target Node and some of which are client-side. Where it makes file organisation easier to understand on a per-package basis I think it's OK to have exceptions.

Off the top of my head, the groups of packages would roughly be:

* Node.js packages (like `keystone`, `session`, `access-control`)
* Universal or client-side packages (like `apollo-helpers`, `utils`)
* Field type packages (like `fields`, `fields-wysiwyg-tinymce` - note there are two )
* Exceptions (the `admin-ui` package comes to mind here)

There's a tool we're working on at Atlassian called [stricter](https://www.npmjs.com/package/stricter) that we could look at integrating in the future to keep our package structure enforced at a linter level.

timleslie:
> If you want an equivalent of `src` for server-side packages, `lib` is commonly used in Node.js projects for containing modules that are imported directly.

This works for me

timleslie:
> There's a tool we're working on at Atlassian called stricter that we could look at integrating in the future to keep our package structure enforced at a linter level.

Also agree that something similar to this would be 👍 in order to keep things in line.

timleslie:
> We also have some packages (like the field types and Admin UI) which contain two or more groups of files, some of which target Node and some of which are client-side. Where it makes file organisation easier to understand on a per-package basis I think it's OK to have exceptions.

Yep, agreed. I'm hoping that going through the process of consolidating the structure will shake out these interesting cases to the point where we can identify the most appropriate structure. This is also part of the reason for wanting to do separate PRs for each package.

timleslie:
**Checklist**

 - [ ]  access-control
 - [ ] adapter-knex
 - [ ] adapter-mongoose
 - [ ] admin-ui
 - [ ] apollo-helpers
 - [ ] arch
 - [ ] core
 - [ ] create-keystone-app
 - [ ] email
 - [ ] field-views-loader
 - [ ] fields
 - [ ] fields-wysiwyg-tinymce
 - [ ] file-adapters
 - [ ] keystone
 - [ ] logger
 - [ ] mongo-join-builder
 - [ ] server
 - [ ] session
 - [ ] test-utils
 - [ ] utils

JedWatson:
@timleslie much of this has been done now, I can see you're tracking the list of packages above but if you're not going to merge this PR, maybe close it and move that list of packages to transition into an issue?

PR #965, opened 2019-03-31
Restructure access-control package

timleslie:
c.f. #964


packages/access-control/tests/access-control.test.js
```diff
@@ -3,7 +3,7 @@ import {
   parseFieldAccess,
   validateListAccessControl,
   validateFieldAccessControl,
-} from './index.js';
+} from '../lib/access-control.js';
```
mitchellhamilton: Why `../lib/access-control.js` instead of `../index` or `../`?

timleslie: Fair question, I actually went back and forth on this, and I'm not convinced I've landed on the right (or wrong) answer in this case. It's the philosophical question of are we testing the public API, or the internal implementation of individual functions? In this instance those two things happen to be the same, so in a certain sense it's all the same. I think perhaps once this round of changes lands we should take a look across the board and decide how best to answer the question in a consistent way. If you have strong feelings on which way to approach this I'm open to ideas. I think in this specific instance it's neither here nor there (at least for now). Classic hedging :-)

mitchellhamilton: I have strong feelings towards testing public APIs. It means that if you change the implementation and the behavior doesn't change, your tests will still work. Of course, there will always be some tests where it's impractical to only test public APIs so it's not something that I think should be followed 100% of time absolutely but the general principle should be kept in mind IMO.

Totally agree that for this specific instance it doesn't really matter very much but just to follow the general principle, I'd make it `../index`.  If we were using yarn workspaces, I'd say we should even import by the package's name since yarn workspaces links packages at the root so you can access them anywhere.

timleslie: Yep, I think we're in agreement here. I'll change it to `../` (as that matches what we have in a couple of other similar packages). Once these restructures land I'll go back and revisit all of our test directories in a more holistic way so that we can more easily identify which tests are testing against the public API and which are testing internal implementations.

PR #973, opened 2019-03-31
Restructure keystone package

timleslie:
c.f. #964


packages/keystone/index.js
```diff
@@ -1 +1,15 @@
...
+const PasswordAuthStrategy = require('./lib/auth/Password');
+const TwitterAuthStrategy = require('./lib/auth/Twitter');
+const { FacebookAuthStrategy } = require('./lib/auth/Facebook');
+const { BaseKeystoneAdapter, BaseListAdapter, BaseFieldAdapter } = require('./lib/adapters');
+
+module.exports = {
+  Keystone,
+  PasswordAuthStrategy,
+  TwitterAuthStrategy,
+  FacebookAuthStrategy,
```
mitchellhamilton: I'm not a fan of exporting the auth strategies from `@keystone-alpha/keystone`, could we move them to a separate package?

timleslie: Yep, 100% agree that they should be moved to a different package, just not in this PR. For now I want to make them explicit exports of this package, which is an improvement on the existing implicit exports from internal modules.

PR #977, opened 2019-03-31
Restructure logger package

timleslie:
c.f. #964


packages/keystone/lib/List/index.js
```diff
@@ -0,0 +1,1287 @@
+/* eslint-disable no-shadow */
```
mitchellhamilton: What's this file? I don't see a file that this is replacing?

timleslie: Good question, should definitely not be here in this branch!

PR #978, opened 2019-03-31
Update react in todo demo

JedWatson:
Minor fix: when we built this, react 16.8 was still in alpha.


JedWatson:
For the record, the CI fail here was just cypress failing for no apparent reason 😞

PR #980, opened 2019-04-01
Social login refactor

gautamsi:
Working on creating abstract implementation of `PassportStrategy` which can be base for all oAuth 2.0 related strategy.

Goal for this:

- [x] Create base class for Passport Authentication oAuth 2.0
~~- [ ] Add configurable authRoot option for mounting express routes.~~ not needed, no route binding inside PassportStrategy
- [x] Add GitHub auth Strategy be inheriting this base
- [x] refactor Facebook to be sub class of base passport strategy
- [x] Add Google auth Strategy be inheriting this base
- [x] Update Auth Readme on how to add/use more strategy by simple methods
- [x] Refactored Twitter to use the base PassportStrategy this can also be example for little complex Strategy with maximum code reuse.
- [x] Moved PassportJS related Auth Strategies into new package `@keystone-alpha/passport-auth`

Also added an example `WordPressAuthStrategy` in `social-login` test project (renamed from `facebook-login`). Anyone can extend inbuilt `PassportAuthStrategy` from `@keystone-alpha/passport-auth`. 

~~TODO:~~
~~* refactor auth routes to make them individuals functions, should easily be overridden by dev~~
~~* refactor multi step signup to enable custom routes and field capture.~~


gautamsi:
looks like it closed when I forced branch with master after #934

gautamsi:
@timleslie @jesstelford  can you take a look at it.

JedWatson:
@gautamsi this looks great. I'm going to leave it for @timleslie or @jesstelford to review in more detail because they're more familiar with this area than I am, but nice work!

One thing I'd like to propose: let's put everything but the basic password auth strategy into a separate package called `@keystone-alpha/passport-auth`. I'd like to keep the dependencies on the core keystone package lean, and pulling in everything from passport and the various social integrations if you're not using them is not great.

Creating a new package should be as simple as adding a new directory under `packages` and following the conventions in the core as you have here, but if something goes wrong your you want help let us know!

gautamsi:
I did propose the package split in https://github.com/keystonejs/keystone-5/issues/878#issuecomment-475331249

JedWatson:
You did too!

It made sense to me reviewing this that all these are implementations of Passport auth, so that seems like a logical grouping for a package called `passport-auth`. @timleslie any reason not to separate that out now?

Something more generic like `auth` or `auth-ui` is, while probably a good idea down the track, possibly a bit harder to get our heads around without thinking about the long term plan a bit more. Totally guessing here, but I think that might be what got in the way 🤷‍♂️

gautamsi:
@timleslie do you have any comment on creation of new package @JedWatson  has suggested? I will refactor and move to next improvements.

gautamsi:
moved all passport related strategies to `@keystone-alpha/passport-auth`. `@keystone-alpha/keystone` now only contains `PasswordAuthStrategy`.

@timleslie  I think I am done here for this one. Will think through later to add more improvements.

timleslie:
Thanks @gautamsi, I'll review this now. From a first pass it all looks really good 👍

.changeset/93e02260/changes.json
```diff
@@ -0,0 +1,15 @@
+{
+  "releases": [{ "name": "@keystone-alpha/passport-auth", "type": "major" }],
+  "dependents": [
+    {
+      "name": "@keystone-alpha/cypress-project-social-login",
+      "type": "patch",
+      "dependencies": ["@keystone-alpha/passport-auth"]
+    },
+    {
+      "name": "@keystone-alpha/cypress-project-twitter-login",
```
timleslie: Since `cypress-project-twitter-login` has been removed, we should remove this entry from the `dependents` list here.

.changeset/93e02260/changes.md
```diff
@@ -0,0 +1,5 @@
+- First release of @keystone-alpha/passport-auth. 
+- Refactored Passport Strategies into this package. 
+- Created base strategy for PassportAuthStrategy. 
+- Refactored  Twitter and Facebook to use base Passport Strategy.
```
timleslie: Two spaces between `Refactored` and `Twitter`

README.md
```diff
@@ -265,7 +265,7 @@ list used for authentication in `index.js`:
 
 <!-- prettier-ignore -->
 ```javascript
-const { Keystone, PasswordAuth } = require('@keystone-alpha/keystone');
+const { Keystone, PasswordAuthStrategy } = require('@keystone-alpha/keystone');
```
timleslie: Good catch! 👍

packages/keystone/index.js
```diff
@@ -1,14 +1,10 @@
...
-const TwitterAuthStrategy = require('./lib/auth/Twitter');
-const { FacebookAuthStrategy } = require('./lib/auth/Facebook');
 const { BaseKeystoneAdapter, BaseListAdapter, BaseFieldAdapter } = require('./lib/adapters');
 
 module.exports = {
   Keystone,
   PasswordAuthStrategy,
-  TwitterAuthStrategy,
-  FacebookAuthStrategy,
   BaseKeystoneAdapter,
```
timleslie: Because we are removing these from the `@keystone-alpha/keystone` package we will need to add a `changeset` with a `major` version bump.

packages/keystone/package.json
```diff
@@ -25,9 +25,6 @@
     "graphql-tag": "^2.10.1",
     "graphql-type-json": "^0.2.1",
     "p-waterfall": "^1.0.0",
-    "passport": "^0.4.0",
-    "passport-facebook": "^3.0.0",
-    "passport-twitter": "^1.0.4",
```
timleslie: It's nice to no longer depend on these here :-)

packages/passport-auth/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# @keystone-alpha/passport-auth
+
+## 1.0.0
```
timleslie: The `CHANGELOG.md` will get generated with this content by the `changeset` tool next time we bump versions, so this file should just contain:

```
# @keystone-alpha/passport-auth
```

packages/passport-auth/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@keystone-alpha/passport-auth",
+  "description": "Provides Authentication Strategies based on PassportJS.",
+  "version": "1.0.0",
```
timleslie: The version should be set to `0.0.0` here, it will be bumped to `1.0.0` by the `changeset` tool when we next bump all our package versions.

.changeset/93e02260/changes.json
gautamsi: I am consolidating this with changes for keystone package.

packages/keystone/index.js
gautamsi: consolidating this with previous one in a new `changeset`

packages/passport-auth/CHANGELOG.md
gautamsi: new to this `changeset` stuff, still learning. will fix this

packages/passport-auth/package.json
gautamsi: 👍

test-projects/social-login/server.js
```diff
@@ -0,0 +1,89 @@
...
+const { configureGoogleAuth } = require('./google');
+const { configureWPAuth } = require('./wordpress');
+
+const initialData = require('./data');
+
+keystone
+  .prepare({ port })
+  .then(async ({ server, keystone: keystoneApp }) => {
+    if (facebookAuthEnabled) {
+      configureFacebookAuth(keystoneApp, server);
```
timleslie: These `configure*` functions will no longer need a `server` parameter.

test-projects/social-login/server.js
```diff
@@ -0,0 +1,89 @@
...
+    if (googleAuthEnabled) {
+      configureGoogleAuth(keystoneApp, server);
+    }
+
+    if (wpAuthEnabled) {
+      configureWPAuth(keystoneApp, server);
+    }
+
+    await keystoneApp.connect();
+
```
timleslie: `setupAuthRoutes()` should be called here.

test-projects/social-login/setupAuthRoutes.js
```diff
@@ -0,0 +1,98 @@
+const { startAuthedSession } = require('@keystone-alpha/session');
+
+module.exports = ({
+  strategy,
+  authRoot = '/auth',
+  successRedirect = '/api/session',
+  failureRedirect = '/',
+}) => {
+  // Hit this route to start the auth process for service
```
timleslie: Adding `const basePath = `${authRoot}/${strategy.authType}`;` here will allow a bunch of code to be simplified in this file and make it more clear where the other aspects of the `strategy` object are being used.

test-projects/social-login/setupAuthRoutes.js
```diff
@@ -0,0 +1,98 @@
+const { startAuthedSession } = require('@keystone-alpha/session');
+
+module.exports = ({
+  strategy,
```
timleslie: This function should take a `server` parameter, since the strategy itself will no longer have a `server` config option.

packages/passport-auth/lib/Passport.js
```diff
@@ -0,0 +1,306 @@
...
+
+    this.createSessionList();
+
+    PassportAuthStrategy.initializePassport(this.config.server);
+
+    this.passportStrategy = this.getPassportStrategy();
+    passport.use(this.passportStrategy);
+  }
+
+  static initializePassport(server) {
```
timleslie: `initializePassport` should be removed, and the onus should be on the system builder to call `server.app.use(passport.initialize());` at the appropriate place in their middleware stack.

In the future we can add helper functions around that, but we shouldn't be forcing our way into the middleware stack right here.

Removing this method means we no longer need to pass in a `server` to `PassportAuthStrategy`.

test-projects/social-login/server.js
gautamsi: still needs it to pass down to `setupAuthRoutes`

test-projects/social-login/server.js
gautamsi: moved all call to `configure*()` after `keystoneApp.connect()`,
`setupAuthRoutes()` is per service, so it has to be called via `configure*()`

packages/passport-auth/lib/Passport.js
gautamsi: added a helper function to `PassportAuthStrategy` which is exposed on `@keystone-alpha/passport-auth`

gautamsi:
I faced this challenge as to split the  calling code into two part one that runs before `keystone.connect()` which calls `keystone.createAuthStrategy({...})` and 2nd which set ups the routes. 

anyone who is using this or any other similar scenario in which use of middleware is requires, those calls must be done after `keystone.connect()`

test-projects/social-login/server.js
gautamsi: removed it now. removed `setupAuthRoutes` form `configure*()`

packages/passport-auth/README.md
```diff
@@ -0,0 +1,219 @@
+---
```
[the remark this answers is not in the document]
gautamsi: Take a look at the verbiage, I am sure you want to refine it little bit.

gautamsi:
@timleslie failing test seems irrelevant

packages/passport-auth/lib/Passport.js
```diff
@@ -0,0 +1,304 @@
...
+      keystoneSessionIdField: 'keystone_passport',
+      scope: [],
+      ...config,
+    };
+    this.ServiceStrategy = ServiceStrategy;
+
+    this.createSessionList();
+
+    this.passportStrategy = this.getPassportStrategy();
+    passport.use(this.passportStrategy);
```
timleslie: We can remove this line now.

test-projects/social-login/server.js
```diff
@@ -0,0 +1,105 @@
...
+    if (wpAuthEnabled) {
+      socialLogins.push(configureWPAuth(keystoneApp));
+    }
+
+    await keystoneApp.connect();
+
+    socialLogins.forEach(strategy => setupAuthRoutes({ strategy, server }));
+
+    if (
```
timleslie: This if statement can be simplified down to

```js
if (socialLogins.length > 0) {
  ...
}
```

test-projects/social-login/server.js
```diff
@@ -0,0 +1,105 @@
...
+      facebookAuthEnabled ||
+      githubAuthEnabled ||
+      googleAuthEnabled ||
+      twitterAuthEnabled ||
+      wpAuthEnabled
+    ) {
+      InitializePassportAuthStrategies(server.app);
```
timleslie: I believe this block should be moved above `setupAuthRoutes`, as `passport.initialize()` should come before other middlewares.

test-projects/social-login/auth/facebook.js
```diff
@@ -0,0 +1,16 @@
+const { FacebookAuthStrategy } = require('@keystone-alpha/passport-auth');
```
timleslie: Could we move all the strategies into a single `strategies.js` file? I think it makes sense in the library for them to be in separate files, but in the context of the example project I think it would be useful to see them all in one place to make it easier to understand how they are all similar/different to each other.

packages/passport-auth/lib/Passport.js
gautamsi: We would have to call it sometime, Isn't it? Where should I add this then?

gautamsi: we can not remove this line `passport.use(this.passportStrategy);` passportjs would fail. I have to add it somewhere. In my tests it has no impact on whether it is called before `passport.initialize` or after

gautamsi: this does not affect `express` middleware, I think it should be fine here.

PR #981, opened 2019-04-01
Update test-utils to not use WebServer

timleslie:
The `test-utils` package provides a bunch of helper code to support the `api-tests` tests. These tests exist to ensure that our graphQL resolvers are behaving correctly. The current system exercises the resolvers via an HTTP query an established `WebServer`. Recent changes have made it so that we can set up an `ApolloServer` object without actually starting a webserver. We can use this object to exercise graphQL queries locally without needing to go via HTTP.

This PR makes the changes required to remove the HTTP layer from our test framework. This involves editing an awful lot of files, but the meaty changes are confined to `test-utils/index.js`.

### DIFF
--- a/api-tests/queries-access-control/meta.test.js
+++ b/api-tests/queries-access-control/meta.test.js
@@ -40,7 +40,7 @@ multiAdapterRunners().map(({ runner, adapterName }) =>
       test(
         `'access' field returns results`,
         runner(setupKeystone, async ({ keystone }) => {
-          const { data } = await graphqlRequest({
+          const { data, errors } = await graphqlRequest({
             keystone,
             query: `
           query {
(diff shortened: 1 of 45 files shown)


api-tests/queries-access-control/meta.test.js
```diff
@@ -56,9 +56,8 @@ multiAdapterRunners().map(({ runner, adapterName }) =>
       `,
           });
 
-          expect(query.body).not.toHaveProperty('errors');
```
jesstelford: 🤔  It's possible for a query to have _both_ `.errors` _and_ `.data`. `.data` could appear to be correct, but it's also important that we check there aren't any `.errors`.

Or is this handled elsewhere now?

PR #982, opened 2019-04-02
Create a build for the Controller

mitchellhamilton:
This fixes a problem that @ticidesign found when creating custom field types outside the monorepo.

What happened is that the Controller file which field type Controllers extend isn't compiled since the webpack config(intentionally) won't compile anything in node_modules so this adds another preconstruct entrypoint to the fields package for the Controller file.

Note that this has a major changeset to the fields package just in case people imported `@keystone-alpha/fields/Controller.js`(explicitly with the `.js`).


mitchellhamilton:
> Why packages/fields/Controller-src.js and not packages/fields/Controller/index.js?

That would work but I think it would be slightly confusing because people would have to know the specifics of the order which things happen in the node module resolution algorithm and that for the fields package, people would have to import `../some number of ../Controller/index` and `../some number of ../Controller` wouldn't work and for outside the fields package, the opposite would be true. Whereas having `src` in there something, it's more clear what should be imported where.

Do you think packages/fields/Controller/src/index.js would be better?

gautamsi:
is this because of `Flow`? Does Flow support different file extension? like `.flow` or `.flow.js` so that we do not have to have `controller-src.js` and compiled `controller.js`.

I did bring `TypeScript` in past. I know that is not favorable here.   TypeScript has its extension `.ts`. 
`TypeScript` also works like Flow too with purely based on jsdoc comments in js files without the need of compiling them again (no typescript specific stuffs in js file). It does work with `babel 7` so no typescript step needed outside of existing compile step. 

There are several issues related to file extension in flow repo.  one of them is https://github.com/facebook/flow/issues/4233

jesstelford:
> _ and for outside the fields package, the opposite would be true. Do you think packages/fields/Controller/src/index.js would be better?_

It's already different between the internal and external imports. I'd be happier with `packages/fields/Controller/src/index.js` - it's more explicit, and I've seen that general pattern a lot in the node ecosystem, so is easy to understand.

packages/fields/Controller/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "main": "dist/fields.cjs.js",
+  "module": "dist/fields.esm.js",
+  "preconstruct": {
+    "source": "../Controller-src.js"
```
jesstelford:
```suggestion
    "source": "../Controller/src/index.js"
```

packages/fields/types/CalendarDay/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Checkbox/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/DateTime/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Decimal/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/File/views/Controller/Controller.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Float/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Integer/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Password/views/Controller/Controller.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Relationship/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Select/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

packages/fields/types/Text/views/Controller/Controller.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
-import FieldController from '../../../../Controller';
+import FieldController from '../../../../Controller-src';
```
jesstelford:
```suggestion
import FieldController from '../../../../Controller/src/index';
```

jesstelford:
@mitchellhamilton I've gone ahead and made those changes on this branch 👍

PR #988, opened 2019-04-03
remove arch website

Noviny:


gautamsi:
Where is new home for arch ui?

JedWatson:
@gautamsi we're removing it for simplicity right now (it was in a half-finished state without a solid plan to take it forward) but will revisit when we can focus on it.

Might be incorporated into the main Keystone site, we'll see. In the meantime there's also a usage demo page in the Admin UI.

Issue #214, opened 2018-08-13
Clarify Node / Browser support, JS standards


JedWatson:
Follow-on task from my comment on #203 

Document:

* Supported Node version
* Supported Browsers
* Any features that we want to block (should be protected by ESLint where possible)

An example of the last is, if we're supporting IE 11 but not supporting async functions in the client-side code, we need to block that with an ESLint rule and document why.

Conversely, if we are, we should start building a guide on our code standards; i.e encouraging async functions.

MadeByMike:
I'm not sure documenting this is required in 2019. We have a babel config that builds for the last 2 versions of major browsers. I think that is adequate. If specific bugs are raised that relate to a required browser feature we can say, "The AdminUI requires a browser that supports X" where X is the feature. 

For node versions I'm not as familiar with how this is normally done but I took a look at the readme of a handful of major OS projects and I don't see any explicitly listing node versions.... let's treat bugs as they come and add a note about versions if the need arises.

PR #993, opened 2019-04-05
Field type build tool

mitchellhamilton:
This is a continuation of things discussed in #927.

The tool is essentially a fork of preconstruct, there's a bunch of stuff I'm going to rip out though that this tool won't need so I wouldn't recommend reviewing the implementation of this right now.

Something that I would like to hear thoughts on is the API, I've converted `@keystone-alpha/fields-wysiwyg-tinymce` and it works. Below is essentially the most important part of the API, to import views, you call `importView` which you import from `@keystone-alpha/build-field-types`
https://github.com/keystonejs/keystone-5/blob/7eddd38ca4de25aca4f6754e99670e3ccb1c86ce/packages/fields-wysiwyg-tinymce/src/index.js#L1-L20

## TODO

- [x] figure out how dev stuff should work inside and outside the monorepo
- [x] build ESM bundles and make chunks resolve to directories that are automatically created in the dist folder with the correct package.json files and bundles
- [x] probably a bunch of other stuff
- [x] get all the tests working


packages/fields-wysiwyg-tinymce/src/index.js
```diff
@@ -1,19 +1,19 @@
...
 
-module.exports = {
+export let Wysiwyg = {
   type: 'Wysiwyg',
   implementation: Text.implementation,
   views: {
-    Controller: '@keystone-alpha/fields/types/Text/views/Controller',
-    Field: path.join(__dirname, './views/Field/Field'),
-    Filter: '@keystone-alpha/fields/types/Text/views/Filter',
+    Controller: Text.views.Controller,
```
JedWatson: This is really interesting, if it works 😍

JedWatson:
I really like what this does to the Keystone fields API.

Can you share what it would look like (in terms of file structure, screenshot or comment here would do) for the main `fields` package?

Also, do you think this changes anything for other views (e.g. for custom pages in the Admin UI)? There aren't any chances to the fields-loader, which we've talked about extending, just double-checking that's a separate thing.

Thoughts unrelated to the API:

* The fields introduces flow for server-side code for the first time. I'm ... mixed on that? not sure what it's going to do to the dev loop. Probably fine since `build-field-types` is a CLI package, so the main Keystone experience isn't affected in any way 🤷‍♂️
* The wysiwyg field's index.js also moves behind the build step... which I think is OK since that's now consistent with other packages like `utils` that preconstruct builds for universal use anyway. But it seems a bit weird, since there's a hard boundary between the server and client parts of each field. Is that change necessary?

mitchellhamilton:
> Can you share what it would look like (in terms of file structure, screenshot or comment here would do) for the main fields package?

field structure


> Also, do you think this changes anything for other views (e.g. for custom pages in the Admin UI)? There aren't any chances to the fields-loader, which we've talked about extending, just double-checking that's a separate thing.

It shouldn't change anything about that.

>The wysiwyg field's index.js also moves behind the build step... which I think is OK since that's now consistent with other packages like utils that preconstruct builds for universal use anyway. But it seems a bit weird, since there's a hard boundary between the server and client parts of each field. Is that change necessary?

Having the server side parts of field types behind a build step is what enables `importView` so, yes, it's necessary.

👇  is an explanation of how importView will work which might answer any other questions you might have.

We have a file like this,
```jsx

export let Wysiwyg = {
  ...otherStuff
  views: {
    ...otherViewsStuff
    Field: importView('./views/Field'),
  },
};
```
It goes through a babel plugin which turns it into this
```jsx
export let Wysiwyg = {
  ...otherStuff
  views: {
    ...otherViewsStuff
    Field: ___KS_BUILD_SYSTEM_INTERNAL___importView(import('./views/Field')),
  },
};
```

Rollup does its thing and we have this for CommonJS builds. (and a different thing for ESM but I'm going to ignore that for this explanation) It's going to be written to the dist folder.

```jsx
export let Wysiwyg = {
  ...otherStuff
  views: {
    ...otherViewsStuff
    Field: ___KS_BUILD_SYSTEM_INTERNAL___importView(Promise.resolve(require('./chunk-hash'))),
  },
};
```

We then transform it with another babel plugin into this

```jsx
const _join = require('path').join

export let Wysiwyg = {
  ...otherStuff
  views: {
    ...otherViewsStuff
    Field: _join(__dirname, './chunk-hash'),
  },
};
```

And rollup will write the ESM and CJS versions of that chunk to `dist/chunk-hash/not-sure-about-the-filename-yet.cjs.js` and `dist/chunk-hash/not-sure-about-the-filename-yet.esm.js` and build-field-types will write a package.json with main and module fields to `dist/chunk-hash/package.json`

JedWatson:
❤️ it @mitchellhamilton

packages/keystone/bin/cli.js
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env node
+
 const arg = require('arg');
 const path = require('path');
 const globby = require('globby');
+try {
+  require('@keystone-alpha/build-field-types').hook(path.join(__dirname, '..', '..', '..'));
```
JedWatson: What is this doing?

mitchellhamilton: It sets up require hooks so when you import from an entrypoint built with `build-field-types` it will import the source file and run Babel on it and it's dependencies. This means that you get live reload for the client-side stuff and for the server-side stuff, you don't have to run a build or watch process.

I'm not 100% sure if it's a good idea because it feels a bit hacky but it seems to work pretty well so maybe it's okay ¯\\\_(ツ)_/¯.

I had another idea to sort of solve this problem. adding a command called `dev` to build-field-types/preconstruct which would create symlinks for all the entrypoints from where the dist files would be to the appropriate source file. I think it might be nice idea since then you don't need tool-specific aliases and everything just works.

Babel still needs to be run on the source files.

(this is just my process of thinking through this)

Maybe, instead of symlinks, actual files could be written with something like this

```jsx
// babel require hook (not @babel/register, a custom one because we need to be able to specifically unregister the one we register which @babel/register doesn't allow)
module.exports = require('../source-file')
// unregister babel require
```

Though then it still uses a require hook and that would be a worse experience for Jest and it wouldn't work in webpack.(though i suppose the module dev file could just re-export or be a symlink and then it would work for webpack)

So maybe the best thing to do would be symlinks + a require hook for the node use case so the tool-specific aliases are unnecessary.

mitchellhamilton:
I've been experimenting and I think I've found a good solution to all of this in dev which also solves some other problems.

I've added a command called `dev`(I'm not 100% sure about the name) to `build-field-types` and `preconstruct`.

For each entrypoint, the command creates a symlink from where the ESM dist files are to the appropriate source file and for CJS it creates a file like below that registers a require hook, re-exports the source file and unregisters the require hook.
```jsx
'use strict';

let unregister = require('/absolute/path/to/a/thing/in/build-field-types').___internalHook();

module.exports = require('/absolute/path/to/the/source/file');

unregister();
```

For the ESM file, we basically know the file will be compiled using webpack and Babel so we don't need to worry about how that file will be compiled.

For the CJS file, since we basically know that it'll be running in Node, we can use a require hook to handle running Babel.

Problems this solves:
- We can run `preconstruct dev` and `build-field-types dev` in a postinstall hook
- Each tool doesn't need their own special aliases
- You don't need to run a build command after changing any source files (including the utils package)
- Probably other problems I can't think of right now because this is second time I've written this comment because I accidentally closed the tab and lost it.

Before I convert the fields package to use build-field-types, I'd like to make sure that people(mainly @timleslie, @jesstelford and @JedWatson) are good with this and everything else in this PR so I don't have to update this all the time and this doesn't become merge conflict hell.

JedWatson:
@mitchellhamilton I don't have any major concerns about what you're proposing. looks like it solves quite a few problems so gets a 👍 from me!

generally speaking, I think this whole PR is a big step forward for fields so on the condition that it actually works, I'm in favour of getting it in and then continuing to iterate as necessary.

babel.config.js
```diff
@@ -1,6 +1,19 @@
...
+      '@babel/env',
+      {
+        targets: {
+          node: 10,
+          browsers: [
+            'last 2 chrome versions',
+            'last 2 firefox versions',
+            'last 2 safari versions',
+            'last 2 edge versions',
+          ],
```
mitchellhamilton: I want to call out this change in case anyone has any concerns. From #214 it seems like we're okay with not supporting IE 11.

The reason I did this isn't really for smaller bundles but that's a nice benefit, I did this is because there's a bug in Babel which means Babel compiled classes can't extend native classes. (https://github.com/babel/babel/issues/9367)

packages/admin-ui/server/getWebpackConfig.js
```diff
@@ -109,13 +109,6 @@ module.exports = function({ adminMeta, entry }) {
         // which depends on the version of react that keystone uses
         react$: require.resolve('react'),
         'react-dom$': require.resolve('react-dom'),
-        ...(() => {
-          try {
-            // see the comment in the babel-loader exclude option for why this is disabled on heroku
-            if (isHerokuEnv) return;
-            return require('preconstruct').aliases.webpack(path.join(__dirname, '..', '..', '..'));
-          } catch (e) {}
-        })(),
```
jesstelford: :+1:

packages/fields/src/types/CloudinaryImage/Block.js
```diff
@@ -27,17 +30,17 @@ class CloudinaryBlock extends Block {
       auxList = createAuxList(auxListKey, {
         fields: {
           // We perform the requires here to avoid circular dependencies
```
jesstelford:
```suggestion
```

PR #987, opened 2019-04-02
Updates based on feedback from deepscan.io

VinayaSathyanarayana:
Only Items marked High are covered/fixed here.

Check done by deepscan.io

PR #994, opened 2019-04-05
Fix most issues identified by deepscan

timleslie:
Inspired by #987

### DIFF
--- a/.changeset/7107db33/changes.json
+++ b/.changeset/7107db33/changes.json
@@ -1,0 +1,33 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/api-tests", "type": "patch" },
+    { "name": "@keystone-alpha/adapter-mongoose", "type": "patch" },
+    { "name": "@arch-ui/modal-utils", "type": "patch" },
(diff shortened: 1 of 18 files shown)


packages/adapter-mongoose/tokenizers/relationship-path.js
```diff
@@ -29,15 +29,16 @@ module.exports = function getRelatedListAdapterFromQueryPathFactory(listAdapter)
       }
 
       // Then follow the breadcrumbs to find the list adapter
+      const currentKey = foundListAdapter.key;
       foundListAdapter = fieldAdapter.getRefListAdapter();
 
       if (!foundListAdapter) {
```
mitchellhamilton: I don't get this, if `foundListAdapter` is falsy then trying to get `foundListAdapter.key` will throw so this branch will never happen.

timleslie: We are pulling out `foundListAdapter` just before we re-assign its value. If the re-assigned `foundListAdapter` is falsey then we can still through a useful error message containing the `key` name of the last good `foundListAdapter`.

Issue #184, opened 2018-08-03
Add special fields `updatedAt` and `createdAt` of type DateTime


jesstelford:
Very useful fields when doing sorting (eg, for a news feed)

## `createdAt`

Is set when a new entry is made to the current datetime. It cannot be updated, only read.

## `updatedAt`

Is set every time any values in the item are changed. It cannot be directly updated, only indirectly via an `update` mutation. It can be read.

Initial value == `createdAt`

jesstelford:
Two possible APIs:

```javascript
import { withTrackingFields } from '@keystone-alpha/tracking-fields';
keystone.createList('Post', withTrackingFields({
  fields: {
    title: { type: Text },
  },
}));
```

```javascript
import { buildTrackingFields } from '@keystone-alpha/tracking-fields';
keystone.createList('Post', {
  fields: {
    title: { type: Text },
    ...buildTrackingFields(),
  },
});
```

I like the first (`withTrackingFields`) a bit better as it can do more things, like injecting `hooks` / `access` at the list level rather than only at the field level.

gautamsi:
> For prior art, see next's way of modifying their config object.

I am not against higher order functions but there is more complexity here than React or Next.js. 
1. it will look ugly if we have to use multiple higher order functions with multiple arguments for them. 
2. It will be limited to narrow use case. with narrow I mean, it will be focused with specific tasks but not able to create a bigger scenario of extension/add-ins

I propose api like this:

```js
import { withTrackingFields, withPublishingFields } from '@keystone-alpha/tracking-fields';
keystone.createList('Post', {
  fields: {
    title: { type: Text },
  },
},
modifiers: [withTrackingFields, withPublishingFields], 
// or
extensions: [withTrackingFields, withPublishingFields],
); 
```

or using extension with keystone instance, I should use,

```js
import { PublishingExtension, TrackingExtension } from '@keystone-alpha/extensions';
const keystone = new Keystone({
  name: 'Keystone Demo Blog',
  adapter: new MongooseAdapter(),
  extensions: [PublishingExtension, TrackingExtension], // I would prefer instance here so that custom parameters can be passed
});
```

## Change needed

Keystone should run list options (1st and 2nd arguments) with modifiers/extensions internally. This way it is clean reusable with multiple modifiers.

in Extension scenario, we can have conventions to have specific methods (static or instance) on the Extension so that keystone can run each of them with required parameters at specific times like before calling `initFields` we run field option with `Extension.initFields`.

[...]

PR #998, opened 2019-04-06
POC for #184 createdAt and updatedAt

gautamsi:
Discussed in #184


gautamsi:
seeing no activity from maintainers, I created #1291 without the friction of adding new features. Hopefully someone looks at that.

PR #1000, opened 2019-04-07
Update dev dependencies

timleslie:


.changeset/248c86e9/changes.md
```diff
@@ -0,0 +1 @@
+- Update dev devependencies
```
mitchellhamilton:
```suggestion
- Update dev dependencies
```

timleslie: hah, that is a weird typo to make!

PR #1012, opened 2019-04-09
Correctly handle async .getValue() calls when saving

jesstelford:


packages/admin-ui/client/pages/Item/index.js
```diff
@@ -112,18 +130,17 @@ const ItemDetails = withRouter(
...
+        getInitialValues(fields, initialData),
+        getCurrentValues(fields, item),
+      ]).then(([initialValues, currentValues]) => (
         // Don't try to update anything that hasn't changed.
         // This is particularly important for access control where a field
         // may be `read: true, update: false`, so will appear in the item
         // details, but is not editable, and would cause an error if a value
         // was sent as part of the update query.
-        arrayToObject(
-          fields.filter(field => !isEqual(field.getValue(initialData), field.getValue(item))),
```
jesstelford: calling `field.getValue()` in the `.filter()` call could potentially return a Promise, so the `isEqual()` would be comparing a `new Promise()` vs `new Promise()` which will never be equal.

JedWatson:
Nice catch 👌

PR #1013, opened 2019-04-09
Content Field shouldn't throw when no blocks specified or server data is corrupt

jesstelford:
I ran into both situations when attempting to add a `Content` field to an existing database. This PR should stop those errors from occurring.

### DIFF
--- a/packages/fields/types/Content/Implementation.js
+++ b/packages/fields/types/Content/Implementation.js
@@ -25,7 +25,9 @@ class Content extends Text {
       output: type,
     };
 
-    this.complexBlocks = this.config.blocks
+    this.blocks = Array.isArray(this.blocks) || [];
+
+    this.complexBlocks = this.blocks
       .map(blockConfig => {
         let Impl = blockConfig;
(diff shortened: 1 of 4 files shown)


packages/fields/types/Content/Implementation.js
```diff
@@ -25,7 +25,9 @@ class Content extends Text {
       output: type,
     };
 
-    this.complexBlocks = this.config.blocks
+    this.blocks = Array.isArray(this.blocks) || [];
```
mitchellhamilton: Wouldn't `Array.isArray(this.blocks)` always evaluate to false since `this.blocks` wouldn't be defined at that point?

jesstelford: Whoa, great catch! That whole line of code is all sorts of messed up! Not sure what I did there, but it's totally wrong 😓

jesstelford:
```suggestion
    this.blocks = Array.isArray(this.config.blocks) ? this.config.blocks : [];
```

PR #1017, opened 2019-04-10
Initialise Block Constructors inside Field Controller

jesstelford:
This is mostly a move-and-refactor, no functionality is changed (except for one minor improvement noted inline below).

This enables access to the block config during Controller actions (eg; inside `Controller#getValue()`), which I'll be implementing in a future PR.

### DIFF
--- a/packages/fields/types/Content/views/src/Field.js
+++ b/packages/fields/types/Content/views/src/Field.js
@@ -1,57 +1,13 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { useState, useMemo } from 'react';
+import { useState } from 'react';
 import Editor from './editor';
 import { Value } from 'slate';
 import { initialValue } from './editor/constants';
-import * as paragraph from './editor/blocks/paragraph';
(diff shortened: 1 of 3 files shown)


packages/fields/types/Content/views/src/Field.js
```diff
@@ -1,57 +1,13 @@
...
-    function pushBlocks(blocksToPush) {
-      blocksToPush.forEach(block => {
-        let existingItem = flatBlocks.find(({ type }) => type === block.type);
-        if (existingItem === undefined) {
-          let { dependencies, ...blockToInsert } = block;
-          flatBlocks.push(blockToInsert);
-          if (dependencies !== undefined) {
-            pushBlocks(dependencies);
-          }
-        } else if (existingItem.Node === undefined) {
```
jesstelford: This condition would previously only ever be hit when a block is pushed _more than once_. Which made it possible for a block without the required `.Node` property to be pushed as long as it wasn't pushed twice. I have fixed this in my refactor.

PR #1018, opened 2019-04-10
Ensure Content Block views are always loaded

jesstelford:
Previous behaviour was to rely on manually importing a block for it to
be included (eg; `unordered-list` imports the `list-item` block). While
this is currently always the case, it doesn't necessarily have to be.

One example might be a 3rd party block which wants to create
heading+subheading pair, and set a dependency on the `heading` block
without ever importing it (instead; referencing it by its type name from
within the 3rd party block).

By surfacing the type on the server / during webpack build we're able to
do a couple more checks such as for duplicates and also replace the
sparse config array with a keyed object.

Finally, we can pull up the default `paragraph` block into the pre-build
step and remove that logic from the client code, making it a bit
cleaner.

### DIFF
--- a/packages/fields/types/Content/Implementation.js
+++ b/packages/fields/types/Content/Implementation.js
@@ -1,7 +1,17 @@
 const { MongoTextInterface, KnexTextInterface, Text } = require('../Text/Implementation');
-const { flatMap } = require('@keystone-alpha/utils');
+const { flatMap, unique } = require('@keystone-alpha/utils');
+const paragraph = require('./blocks/paragraph');
 
 const GQL_TYPE_PREFIX = '_ContentType';
+
+const DEFAULT_BLOCKS = [paragraph];
+
+function flattenBlockViews(block) {
+  return [
+    block.viewPath,
+    ...(block.dependencies ? flatMap(block.dependencies, flattenBlockViews) : []),
+  ];
+}
 
 class Content extends Text {
   constructor(path, config, listConfig) {
@@ -27,6 +37,28 @@ class Content extends Text {
 
     this.blocks = Array.isArray(this.config.blocks) ? this.config.blocks : [];
 
+    this.blocks.push(
+      ...DEFAULT_BLOCKS.filter(
+        defaultBlock =>
+          !this.blocks.find(
+            block => (Array.isArray(block) ? block[0] : block).type === defaultBlock.type
+          )
+      )
+    );
+
+    // Checking for duplicate block types
+    for (let currentIndex = 0; currentIndex < this.blocks.length; currentIndex++) {
+      const currentBlock = this.blocks[currentIndex];
+      const currentType = (Array.isArray(currentBlock) ? currentBlock[0] : currentBlock).type;
+      for (let checkIndex = currentIndex + 1; checkIndex < this.blocks.length; checkIndex++) {
+        const checkBlock = this.blocks[checkIndex];
+        const checkType = (Array.isArray(checkBlock) ? checkBlock[0] : checkBlock).type;
+        if (currentType === checkType) {
+          throw new Error(`Encountered duplicate Content block type '${currentType}'.`);
+        }
+      }
+    }
+
     this.complexBlocks = this.blocks
       .map(blockConfig => {
         let Impl = blockConfig;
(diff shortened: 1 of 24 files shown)


packages/fields/types/Content/Implementation.js
```diff
@@ -65,18 +92,36 @@ class Content extends Text {
...
   }
+
   // Add the blocks config to the views object for usage in the admin UI
   // (ie; { Cell: , Field: , Filters: , blocks: ...})
   extendViews(views) {
     return {
       ...views,
-      blocks: this.blocks.map(block => (Array.isArray(block) ? block[0] : block).viewPath),
+      blocks: unique(
+        flatMap(this.blocks, block => flattenBlockViews(Array.isArray(block) ? block[0] : block))
```
jesstelford: Ensuring all the dependency's views are included in this list so they're always bundled even if they're not imported by another block.

packages/fields/types/Content/blocks/blockquote.js
```diff
@@ -0,0 +1,6 @@
+const path = require('path');
```
jesstelford: I started out with all this in the `index.js`, but broke them out into individual files like this once I realised I need the `paragraph.js` for setting the default in `Implementation.js`.

I don't mind them being split out into separate files like this; it's reminiscent of the field types themselves. OTOH; I'm not against moving them back into `index.js`, it just means we'll end up with duplicate definitions for `paragraph.js` which is a bit icky.

packages/fields/types/Content/views/blocks/caption/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
jesstelford: Making sure that _every_ block type is bundled, even the dependencies, in case they're not imported directly.

packages/fields/types/Content/views/src/Controller.js
```diff
@@ -6,36 +6,26 @@ import memoizeOne from 'memoize-one';
...
+      throw new Error(`Unable to load Content block '${block.type}': no 'Node' export found.`);
     }
 
-    const { dependencies, ...blockToInsert } = block;
-
-    outputBlocks[block.type] = blockToInsert;
-
-    if (dependencies !== undefined) {
-      // Recurse on the dependencies
-      flattenBlocks(dependencies, outputBlocks);
```
jesstelford: Dependencies are already flattened pre-build in `Implemenation.js`, so there's nothing to do here.

JedWatson:
@jesstelford looking good!

One thing that occurred to me; if the blocks are all loaded, always, is it worth having separate entrypoints for them? or could a higher level `index` just `export { all, block, types }`

I do like the individual files for block types (esp. that they are following patterns established by field types) but couldn't tell how having separate entrypoints interacts with your point about having extracted them all from the index and the reduced duplication.

Feel free to act / not act on this feedback, as-is looks good ✅ 

@mitchellhamilton this might be something to consider for #993, if that could work for block types as well as field types?

jesstelford:
> if the blocks are all loaded, always, is it worth having separate entrypoints for them?

All the blocks _for that field_ are loaded.

For example;

```javascript
Keystone.createList('User', {
  fields: {
    bio: {
      type: Content,
      blocks: [
        Content.blocks.link
      ]
    }
  }
});
```

Would load the `link` and the default `paragraph` blocks when landing on the `Field` view for the `User` list (ie; editing or adding a `User`).

```javascript
Keystone.createList('Post', {
  fields: {
    bio: {
      type: Content,
      blocks: [
        Content.blocks.header,
        Content.blocks.unorderedList,
      ]
    }
  }
});
```

Would load the `unorderedList`, it's dependence of `listItem`, and the default `paragraph` blocks when landing on the `Field` view for the `Post` list (ie; editing or adding a `Post`).

This way it only loads exactly what is required per field 👍

> how having separate entrypoints interacts with your point about having extracted them all from the index and the reduced duplication.

The main one was the `paragraph` type; it's a dependency for a couple of the others and is reference in the default list. Once I dedicated that by putting it in its own file, I figured why not do it for the rest too.

jesstelford:
Looks like I broke the tests - will fix them up then merge.

mitchellhamilton:
> @mitchellhamilton this might be something to consider for #993, if that could work for block types as well as field types?

It'll 100% work for block types.

PR #1019, opened 2019-04-10
Custom Admin Pages

### LINKED ISSUE #7: Add support for custom pages / views in the Admin UI
@lukebatchelor thought this one might interest you.

I've been thinking we should make it so you can plug in custom routes/pages/views into the Admin UI. While there's a bit to design / spec for this, a good place to start would be moving the new `style-guide` that @jossmac has been working on in the Admin UI into the `test-project` as a proof of concept.

As a _really_ draft idea (feel free to challenge / iterate on this):

```js
const admin = new AdminUI(keystone, {
  views: [
    {
      route: '/style-guide' // ends up being mounted at /admin/style-guide
      label: 'Style Guide', // label in the top nav bar
      view: './admin/views/StyleGuide', // path to the React component that we inject
    }
  ]
});
```

Another thing to consider would be how we can plug in custom _pieces_ of UI into existing views, like the Home Page / Items List / Item Details. Ideally, we could add a button to the homepage linking to the style guide... not sure the best way to design the plugin API for that yet though. Maybe some "slots" concept...

In any case the API could change pretty easily, getting something in place that integrates with the build process is a good first step.

JedWatson:
**EDIT** Maybe not an experiment, see comment below. Closes #7 

I created this while reviewing #933, primarily to understand the difference between what it looks like to add with our current architecture vs. the Next.js branch. Primarily an experiment!

Note this lacks SSR support, and I didn't put as much effort into the examples, but it's otherwise (I think) identical in functionality.

@jesstelford I'd like to land a few changes I came up with while working on this (including a fair bit of refactoring in the `field-views-loader` package) but am concerned with creating more potential conflicts with your work so let's pick up that conversation soon 🙂


JedWatson:
So I've cleaned up the failing tests and would like to merge this to unblock some work on a demo project with custom pages and workflows. The actual API for setting up custom pages is consistent with what we've discussed elsewhere and the implementation seems trivial enough to update later with any other changes we make to how the Admin UI is built and served.

@jesstelford @timleslie thoughts?

packages/field-views-loader/index.js
```diff
@@ -62,60 +86,66 @@ module.exports = function() {
...
       return `'${path}': () => import('${path}').then(interopDefault)`;
     })
     .join(',\n')}\n}`;
 
-  return `
-  let promiseCache = new Map();
-  let valueCache = new Map();
+  const source = endent`
+    let promiseCache = new Map();
+    let valueCache = new Map();
```
mitchellhamilton: Not a thing about this PR but we should probably move all this suspense stuff into the Admin UI since it doesn't really need to be in the loader and we can have prettier, eslint and etc. on it.

packages/field-views-loader/index.js
```diff
@@ -62,60 +86,66 @@ module.exports = function() {
...
+  let pageComponents = findPageComponents(adminMeta.pages);
+
+  let allViews = Object.entries(adminMeta.lists).reduce(
+    (obj, [listPath, { views }]) => {
       obj[listPath] = views;
       return obj;
-    }, {}),
-    allPaths
+    },
+    { __pages__: pageComponents }
```
mitchellhamilton: Could the pages be another export? Or maybe the same object but with `pages` and `lists` keys?

JedWatson: I'm planning to do a more significant refactor of the loader, including getting rid of the lists export altogether and exporting views per field type instead which will reduce the bundle size quite a bit for projects with lots of lists and fields.

I'd rather do that refactoring with another PR, and keep this to just the addition of custom pages into the current export.

mitchellhamilton: > including getting rid of the lists export altogether and exporting views per field type instead which will reduce the bundle size quite a bit for projects with lots of lists and fields.

I just thought about this more and realized that it has a problem, the views of a field type can change for each for each usage of a field type. While this isn't true for most field types, it's true for one, the Content field type since the blocks defined for a specific usage of the field type determine the views.

Issue #139, opened 2018-06-28
`keystone.createItems` should validate data against the schema before insertion


jesstelford:
The GraphQL endpoints do this validation for us. Here, we are circumventing those checks and blindly pushing data into the database. We should be more strict about validating it before we do.

JedWatson:
@gautamsi I was thinking we could come up with a package that contains similar functionality to keystone 4's `View`, especially to help with transitions from Keystone 4. It will hopefully be fairly simple now that everything's decoupled.

gautamsi:
Working on it now, can you assign this issue to me.

JedWatson:
@gautamsi thanks! assigned 👍

gautamsi:
@JedWatson I have PR #1024 in which I have focused on view compatibility. but it seems it needs more than just view to implement full features of v4 view with backward compatibility. 

Should I create a separate PR for each of following or just update PR #1024 with additional features? All of them related to v4 comatibility

1. keystone.updateHandler
2.  route importer
3. keystone.populaterelated + list model schema methods (populateRelated and getRelated)
4. List.updateItem

I am inclined to add another package `v4-compat` (or suggest better name) which wraps Keystone and List with helper methods to help upgrade in place for v4 project with minimal changes. This will obviously focused with mongoose adapter. to be included in pr #1024. I am planning to test this with `sydjs-site`.

thoughts

MadeByMike:
Maybe solution is to remove createItems. Re-document the createItems\Initial data with `executeQuery`

molomby:
Should be removed; see #1756.

MadeByMike:
Closing issues related to `createItems` as we're not going to work on it further.

PR #1024, opened 2019-04-12
Keystone v4 compatibility: View

gautamsi:
ref #139 
### View
This is port of Keystone v4 View which enables migration from v4 easier

### API difference:

at present, you have to set View path and view engine with express
```js
server.app.set('views', './templates');
server.app.set('view engine', 'pug');
```

#### `constructor`
v4
```js
const view = new keystone.View(req, res);
```
v5
> v4 compatible querying of mongoose can also be enabled when https://github.com/keystonejs/keystone-5/issues/1002#issuecomment-485238089 completes
```js
const View = require('@keystone-alpha/view');
// n v5 everything is decoupled, need to pass keystone instance
const view = new View(keystone /*Keystone instance*/, req, res);
```

#### `query`
v4
```js
    const Meetup = keystone.list("Meetup");
    view.query("upcomingMeetup",
        Meetup.model.findOne()
            .where("state", "active")
            .sort("-startDate")
        , "talks[who]");
```
v5
```js
view.query("upcomingMeetup", `query {
  allPosts {
    title
    author {
      email
      isAdmin
      password_is_set
      id
    }
    posted
    status
    id
  }
}`);
```

#### All the callbacks are `async` and awaited instead of `next()` call

v4
```js
    // Load Organisers
    view.on("init", function (next) {
        User.model.find()
            .sort("name.first")
            .where("isPublic", true)
            .where("isOrganiser", true)
            .exec(function (err, organisers) {
                if (err) res.err(err);
                locals.organisers = organisers;
                next();
            });
    });
```
v5
```js
    // Load Organisers
    view.on("init", async () {
        await <return promise or do something>
    });
```
### Mutations
`view.query` can also be used with mutations, the path will be populated with what is returned after the mutation.

~~ ### TBD ~~

~~- [ ] populate related~~
~~- [ ] tests~~


packages/view/lib/view.js
```diff
@@ -0,0 +1,383 @@
...
+    for (let i = 0; i < parts.length; i++) {
+      if (!locals[parts[i]]) {
+        locals[parts[i]] = {};
+      }
+      locals = locals[parts[i]];
+    }
+
+    this.queryQueue.push(async () => {
+      const callbacks = chain.callbacks;
+      const context = this.keystone.getAccessContext('admin', this.req);
```
gautamsi: is `admin` always going to be schemaName?

packages/view/lib/view.js
```diff
@@ -0,0 +1,383 @@
...
+      try {
+        const results = await list.listQuery(query, context, list.gqlNames.listQueryName);
```
gautamsi: @jesstelford @JedWatson: is `list.listQuery` right way to call this from within server side code? Can I use mutations similarly (`createMutation`, `updateMutation` and so on)?

I could not find any better way to do this. Still feel there can be improvement. In gql query you must specify the fields you want to retrieve. How do I do it here? filter from result? Let me know if there is better way to use the gql layer properly, I would also want to use the gql layer with `skipAccessControl` as my server side logic may already handle more complex scenario sometime.

packages/view/lib/view.js
timleslie: `admin` will be the default schema name used when creating the apollo server used by the Admin UI. At some point this will probably become configurable. Also, we will soon be able to spin up multiple apollo servers with different schemas, each of which will be accessible via a different schema name. As such, it probably makes sense to make the `schemaName` a config option for the view, with a default of `"admin"`.

packages/view/lib/view.js
timleslie: I think the most useful thing to expose here would be the apollo server query API. That way you can get the full power of our graphQL queries and don't have to reinvent wheels like selecting fields, filtering results, etc.

```javascript
      const schemaName = 'admin';
      const queryFn = this.keystone._graphQLQuery[schemaName];
      const context = this.keystone.getAccessContext(schemaName, this.req);
      try {
        const { data, errors } = queryFn(query, context);
        ....
```
if you want to turn off access control, you can do a similar thing to what we do in hooks, which is:

```javascript
        if (skipAccessControl) {
          passThroughContext = {
            ...context,
            getListAccessControlForUser: () => true,
            getFieldAccessControlForUser: () => true,
          };
        }
        ...
        const { data, errors } = queryFn(query, passThroughContext);
```

At the moment `_graphQLQuery` isn't technically part of the public API, but we might want to change this in the future. For now feel free to work as if it were.

gautamsi: thanks, this greatly helps. 

updated.

packages/view/lib/view.js
gautamsi: adding this as part of View constructor now.

gautamsi:
added v4 style query support in https://github.com/keystonejs/keystone-5/pull/1024/commits/1bcff27e722d3303bfb2a201afc1e06fa22055db, you can now also do this similar to v4. still pending how to populate related. 

```js
const Post = keystone.getListByKey('Post').adapter;
view.query('q', Post.model.find({ id:.... }))
        .then((err, result) => {
          console.log(result);
        })
        .err(err => {
          console.log(err);
        });
```
Once proposed `v4-compat` package lands, following should be possible
```js
const Post = keystoneCompat.list('Post');
view.query('q', Post.model.find({ id:.... }),"author[name]")
        .then((err, result) => {
          console.log(result);
        })
        .err(err => {
          console.log(err);
        });
```

packages/view/lib/view.js
```diff
@@ -0,0 +1,475 @@
...
+    if (typeof query === 'string') {
+      return this._queryGql(key, query, options);
+    }
+    return this._queryLegacy(key, query, options);
+  }
+
+  _queryGql(key, query, { unwrap = true, ...options } = {}) {
+    let locals = this.res.locals;
+    const parts = key.split('.');
+    const chain = new QueryCallbacks(typeof options === 'string' ? null : options);
```
w01fgang: `options` is a result of destructuring, so it will always be an object.

gautamsi:
Added `examples/blog-view` which is ported from `demo-projects/blog` using `view` and pug view engine.

gautamsi:
I have limited availability for next few weeks, will limit this PR for View, later I can pickup other v4 compatibility items.

gautamsi:
This PR is good for review now

jesstelford:
@gautamsi is this still working now that #1109 has landed (and no longer returns a `server` instance from the `keystone.prepare()` call)?

gautamsi:
I have to check, I am away until friday (IST)

gautamsi:
@jesstelford I have added console links similar to how `keystone start` does after the server listen.

I think we should export supporting function to be used with custom server.

gautamsi:
@jesstelford  not sure what these failing tests are.

gautamsi:
@jesstelford this one is now looking good.

ready for review and merge.

MadeByMike:
Hey @gautamsi I think PR is a perfect candidate for a standalone package. Perhaps we can publish this outside the keystone-5 repo? if we did, would you be interested in maintaining it? 

This would give you more flexibility to take it in whatever direction you want and it would help us grow and ecosystem of Keystone 'plugins\libraries'. Maybe call it `keystone-compatibility-views`? I think this is convention for things like Gatsby and similar?

This would also be a chance for us to test how we can better support developers working outside the mono repo. I hope to see more of that so I'd be interested in learning what we can do to make it easier.

What are your thoughts?

gautamsi:
I already had fallback plan to publish this in `@keystone-js` org on npm. I was waiting for formal view from core team :)

MadeByMike:
Go for it ;)

gautamsi:
[@keystone-js/view](https://www.npmjs.com/package/@keystone-js/view) is live on npm.

we can close this now.

How do we go about showcasing community projects on https://v5.keystonejs.com

MadeByMike:
Since this is now published closing this PR. Thanks @gautamsi

PR #1027, opened 2019-04-12
Admin UI client bugfixes

### LINKED ISSUE #1026: AdminUI: State update on unmounted component on first item creation
## Steps to reproduce

1. `yarn create keystone-app myApp`
2. in myApp `yarn start`
3. Navigate to AdminUI in browser of choice
4. Create a new Todo

## Expected
No console errors

## Actual
`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

thiagodebastos:
Prevent `setState` from being called after component unmounts.


packages/admin-ui/client/pages/Item/index.js
```diff
@@ -77,7 +77,7 @@ const ItemDetails = withRouter(
           }
           history.push(`${adminPath}/${list.path}`);
 
-          toastItemSuccess(this.props.toast, item, 'Deleted successfully');
+          toastItemSuccess(this.props.toastManager, item, 'Deleted successfully');
         })
         .catch(error => {
           toastError(this.props.toast, error);
```
mitchellhamilton:
```suggestion
           toastError(this.props.toastManager, error);
```

thiagodebastos: This didn't fix the issue, I am investigating further

thiagodebastos: Hmm I've been looking into it, and it looks like `toastManager` is not being passed to `toastSuccess` when deleting an item.

In `toastItemSuccess`, `toast` is undefined.

thiagodebastos: I have a temporary fix which just creates toast content similar to `onSave`

mitchellhamilton: I think part of the problem might be here since I don't think .addToast is actually a thing in react-toast-notifications
https://github.com/keystonejs/keystone-5/blob/a15601ee6b551eae04f41289c339dbef8315843a/packages/admin-ui/client/util.js#L39-L42
I think this might fix it (along with passing in `this.props.toastManager` + applying the same fix for the error state)
```jsx
toast.add(toastContent, { 
  autoDismiss: true, 
  appearance: 'success', 
}); 
```

thiagodebastos:
Oh and I’ll add the change set after lunch. And I think I’ve tracked the problem down

With Optimism,

Thiago de Bastos

thiagodebastos:
Ah cool you found it too :)

With Optimism,

Thiago de Bastos

packages/admin-ui/client/pages/Item/index.js
thiagodebastos: Ah awesome, i arrived at the same thing. I haven't used react-toast-notifications before, but am guessing that if this worked properly before, the API must have changed?
By changing it to `toast.add` and not immediately invoking the function, everything goes back to working as expected.

thiagodebastos:
I'm not sure why the tests are failing, I will have to look into that at another time, perhaps tomorrow morning.

mitchellhamilton:
I'm guessing they're failing because they're asserting that the label of the deleted item isn't in the body but now it is since the toast is working so a solution would probably be to assert that it doesn't exist within `#app` rather than `body` (since the toasts are in a portal iirc)

thiagodebastos:
Thanks that makes sense! I’ll have another look tomorrow.

thiagodebastos:
Hey guys, I will look for time this week to get back onto this task.

packages/admin-ui/client/pages/Item/index.js
jossmac: My bad, sorry folks --> I updated the API.

.changeset/15951916/changes.md
```diff
@@ -0,0 +1 @@
+- Internal changes only. Fixes bugs with toast notifications on item deletion.
```
jesstelford:
```suggestion
- Fix bugs with toast notifications on item deletion.
```

timleslie:
@thiagodebastos Hey Thiago, just checking in to see what the status of this PR is.

thiagodebastos:
> @thiagodebastos Hey Thiago, just checking in to see what the status of this PR is.

Hi @timleslie, I will look for time to look into this again this week. If I can't figure it out I'll come into the office and hopefully someone can walk me through the bits I don't comprehend.

MadeByMike:
@thiagodebastos Is this PR stale?

thiagodebastos:
> @thiagodebastos Is this PR stale?

Unfortunately yes Mike, I've not had a chance to look into it again.

PR #1023, opened 2019-04-12
Admin revamp

jossmac:
image

image

PR #1034, opened 2019-04-16
Add missing arch-ui deps to admin-ui

mitchellhamilton:
These dependencies were missing from the admin-ui package but preconstruct was covering up the fact that this would be broken when published. I found this when I was working on #993 because it's using a different version of preconstruct which doesn't cover up these problems.

I also didn't add a changeset since this package will already be released because of the changeset in #1023


Noviny:
> I also didn't add a changeset since this package will already be released because of the changeset in #1023

I don't want to be the changeset police, but am wanting to understand the reasoning behind not having a changeset because something else will release it? I would generally still want a changeset, just so the change in this PR ('add missing dependencies') can be included in the changelog for the package.

mitchellhamilton:
> I don't want to be the changeset police, but am wanting to understand the reasoning behind not having a changeset because something else will release it? I would generally still want a changeset, just so the change in this PR ('add missing dependencies') can be included in the changelog for the package.

1. I wanted to this quickly and I couldn't be bothered adding a changeset
2. I would argue something like this _shouldn't_ be in the changelog since the thing that it's fixing hasn't been released yet so it doesn't affect consumers. I don't really feel that strongly about this point though

jossmac:
Hmm, #1023 was released before this...

mitchellhamilton:
> Hmm, #1023 was released before this...

Huh? Do you mean it was merged before this? Because we haven't done a release since that PR.

jossmac:
> Because we haven't done a release since that PR.

Okay cool. I assumed anything that went into master was released.

PR #1036, opened 2019-04-16
Add some delight to the Popout component

jossmac:
- Fancy rounded arrow
- More "pop" on the animation

popout-delight


gautamsi:
Is the animation configurable, would I be able to turn off the animations from config?

packages/arch/packages/popout/src/index.js
```diff
@@ -127,8 +133,9 @@ class Popout extends Component<Props, State> {
   render() {
     const { children, getModalRef, style, width } = this.props;
     let { leftOffset, topOffset, arrowLeftOffset } = this.state;
+    const attachTo = ((document.body: any): HTMLElement); // NOTE: flow junk
 
-    return document.body
+    return attachTo
       ? createPortal(
```
mitchellhamilton: Couldn't you remove the `attachTo ? createPortal(...) : null` and just make it `createPortal(...)`?

packages/arch/packages/modal-utils/src/transitions.js
```diff
@@ -197,3 +197,39 @@ export const ZoomInDown = (props: TransitionProps) => {
...
+
+export const SpringDown = (props: TransitionProps) => {
+  const base = {
+    transformOrigin: 'top',
+    transitionProperty: 'opacity, transform',
+    transitionDuration,
+    transitionTimingFunction: 'cubic-bezier(0.2, 0, 0.16, 1.6)',
+  };
+  return (
+    <TransitionReducer
```
mitchellhamilton: Definitely not all something that we should do in this PR or necessarily soon but I'd love to remove this `TransitionReducer` component at some point and have the api for a transition be `transitionState => styles`

packages/arch/packages/popout/src/index.js
jossmac: Flow still gets cranky...

packages/arch/packages/modal-utils/src/transitions.js
jossmac: God yes. TBH, most of Arch needs an overhaul.

jossmac:
> Is the animation configurable, would I be able to turn off the animations from config?

@gautamsi No. But you couldn't before either...

gautamsi:
@jossmac I have opened #1037, once that get any traction we should have some option later.

PR #1038, opened 2019-04-16
isMultiline support for Text fields

JedWatson:
@dcousens mentioned this so I fixed it.

Our blog demo specifies the `isMultiline` option for the content Text field in the Comments lists, but the field wasn't passing the value through to the Arch component.

### DIFF
--- a/packages/arch/packages/input/src/Input.js
+++ b/packages/arch/packages/input/src/Input.js
@@ -35,13 +35,14 @@ export const inputStyles = (props: InputProps = {}) => ({
   ...(props.isMultiline
     ? {
         lineHeight: 'inherit',
-        height: 'auto',
+        height: 100,
       }
     : undefined),
 });
 
 type InputProps = { isMultiline?: boolean, disabled?: boolean };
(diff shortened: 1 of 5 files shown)


packages/arch/packages/input/src/Input.js
```diff
@@ -35,13 +35,14 @@ export const inputStyles = (props: InputProps = {}) => ({
   ...(props.isMultiline
     ? {
         lineHeight: 'inherit',
-        height: 'auto',
+        height: 100,
```
jossmac: Limit to vertical resize:
```suggestion
        height: 100,
        resize: 'vertical',
```

jossmac:
Would a `Textarea` field type make more sense? I'm not sure that we should let presentational component props inform the config for our types.

jesstelford:
@jossmac 

>  I'm not sure that we should let presentational component props inform the config for our types.

That's a good point! I've opened an issue to continue the discussion: https://github.com/keystonejs/keystone-5/issues/1039

PR #1043, opened 2019-04-17
Skip flakey test

mitchellhamilton:
This test is very flakey and it's getting really annoying to have to retry the tests a bunch of times to get them passing

### DIFF
--- a/test-projects/basic/cypress/integration/add-edit-delete-data.js
+++ b/test-projects/basic/cypress/integration/add-edit-delete-data.js
@@ -109,7 +109,8 @@ describe('Editing data', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


test-projects/basic/cypress/integration/add-edit-delete-data.js
```diff
@@ -109,7 +109,8 @@ describe('Editing data', () => {
     });
   });
 
-  it(`Updates relationship items`, () => {
+  // this test is very flakey so we're skipping it for now
```
jossmac:
```suggestion
// FIXME: this test is very flakey so we're skipping it for now
```

PR #1049, opened 2019-04-21
Explicit config options in Keystone

timleslie:
We pass in a `config` parameter to `Keystone`, `List` etc and it's really not clear what the possible config values are. This makes it difficult to a) understand and b) document! This PR will make these (and possibly other) classes explicitly unpack their config options up-front so we can be really clear about what our API is. Hopefully as a follow up I'll be able to document what all these options actually do :-)

### DIFF
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -105,82 +105,96 @@ const mapNativeTypeToKeystoneType = (type, listKey, fieldPath) => {
 module.exports = class List {
   constructor(
     key,
-    config,
+    {
+      fields,
+      hooks = {},
+      mutations = [],
+      schemaDoc,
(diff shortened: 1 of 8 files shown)


packages/keystone/lib/List/index.js
```diff
@@ -105,82 +105,96 @@ const mapNativeTypeToKeystoneType = (type, listKey, fieldPath) => {
...
+      labelField,
+      access,
+      adminConfig = {},
+      itemQueryName,
+      listQueryName,
+      label,
+      singular,
+      plural,
+      path,
+      ...adapterConfig
```
timleslie: There's a lot which gets bundled up into the `adapterConfig`. In a separate PR I'll try to tease out what all the various possibilities are.

PR #1056, opened 2019-04-23
Make the field adapter config parameters explicit

timleslie:


packages/fields/src/Implementation.js
```diff
@@ -4,12 +4,23 @@ import { parseFieldAccess } from '@keystone-alpha/access-control';
...
+      hooks = {},
+      isRequired,
+      defaultValue,
+      access,
+      label,
+      schemaDoc,
+      isUnique,
+      many,
+      ref,
+      options,
```
jesstelford: I'm not sure we should be explicitly calling out `many`/`ref`/`options` here; `many` & `ref` only have meaning to the `Relationship` field, and `options` only has meaning to the `Select` field.

Perhaps this really is a good use-case for splatting through _all the other options_?

```javascript
class Field {
  constructor(
    path,
    {
      hooks = {},
      isRequired,
      defaultValue,
      access,
      label,
      schemaDoc,
      isUnique,
      mongooseOptions,
      ...adapterConfig,
    },
    { getListByKey, listKey, listAdapter, fieldAdapterClass, defaultAccess }
  ) {
    this.adapter = listAdapter.newFieldAdapter(
      fieldAdapterClass,
      this.constructor.name,
      path,
      getListByKey,
      {
        isRequired,
        isUnique,
        many,
        ref,
        options,
        mongooseOptions,
      },
      adapterConfig,
    );
```

It means we're being explicit with all the common config options that can be passed for a field, but not being opinionated on what a field can accept.

As this PR stands, it's impossible to add a new config option for a field adapter, which greatly limits the ability of 3rd parties making new ones.

It's also tricky because the `adapterConfig` I mentioned above will now include all the config options for a field (such as `blocks` for the `Content` field).

JedWatson:
@timleslie can you share the problem you're solving with this?

I read through the changes and agree with @jesstelford's concern, in particular:

> As this PR stands, it's impossible to add a new config option for a field adapter, which greatly limits the ability of 3rd parties making new ones.

MadeByMike:
@timleslie I think this is now explicit with the recent API changes. I'm closing this PR but re-open it if I am wrong.

PR #1057, opened 2019-04-23
Fields provide (de)serialize methods for pre-save and post-read

jesstelford:
The `deserialize` method in particular will be used by the `Content` type to handle reconstructing the Slate.js editor state after reading from the API.

For each field Controller:
- Renames `.getInitialData()` -> `.getDefaultValue()`
- Renames `.getValue()` -> `.serialize()`
- Adds `.deserialize()` (where necessary)


jesstelford:
🤔  What am I missing here? Why is this test failing with the order of items reversed? I didn't touch any backend code in this PR :/

https://circleci.com/gh/keystonejs/keystone-5/16858

PR #1067, opened 2019-04-29
Make Blocks have similar API to fields

jesstelford:
Taking a page from the Field API; Blocks can now optionally have a `.implementation` which defines the server-side logic to run.

### DIFF
--- a/packages/fields/src/types/CloudinaryImage/ImageBlock.js
+++ b/packages/fields/src/types/CloudinaryImage/ImageBlock.js
@@ -2,28 +2,17 @@ import pluralize from 'pluralize';
 import { Block } from '../../Block';
 import RelationshipFieldType from '../Relationship';
 import CloudinaryImage from './';
-import Text from '../Text';
-import { importView } from '@keystone-alpha/build-field-types';
 
-export class CloudinaryBlock extends Block {
-  static get type() {
-    return 'cloudinaryImage';
-  }
-
-  static get isComplexDataType() {
-    return true;
-  }
-
-  static get viewPath() {
-    return importView('../Content/views/editor/blocks/image-container');
-  }
-
-  constructor({ adapter }, { fromList, createAuxList, getListByKey, listConfig }) {
+export class ImageBlock extends Block {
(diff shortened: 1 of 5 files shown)


packages/fields/src/types/CloudinaryImage/ImageBlock.js
```diff
@@ -2,37 +2,24 @@ import pluralize from 'pluralize';
...
+    // Ensure the list is only instantiated once per server instance.
     let auxList = getListByKey(auxListKey);
 
     if (!auxList) {
       auxList = createAuxList(auxListKey, {
         fields: {
           // We perform the requires here to avoid circular dependencies
           image: { type: CloudinaryImage, isRequired: true, adapter },
-          from: { type: RelationshipFieldType, isRequired: true, ref: fromList },
-          field: { type: Text, isRequired: true },
```
jesstelford: A touch of cleanup here too - these fields came in an earlier PR and make no sense yet (they may come back in a future PR)

JedWatson:
Nice 👌

packages/fields/src/types/CloudinaryImage/index.js
```diff
@@ -19,5 +20,11 @@ export default {
     mongoose: MongoCloudinaryImageInterface,
     knex: KnexCloudinaryImageInterface,
   },
-  block: CloudinaryBlock,
+  blocks: {
+    image: {
+      type: 'cloudinaryImage',
+      viewPath: imageContainer.viewPath,
```
mitchellhamilton: Maybe instead of viewPath, we should have a views key for consistency with field types and in case we want to add more views in the future?

jesstelford: Yes! It's a great idea. I went back and forth on this  one, especially since a block really can have multiple views: The Toolbar, the block itself, the pop-up thingy on the left, etc.

But then when it came down to it; they're all exported by the one view. And I can't really think of a case where they'd need to be separately loaded (with the `viewPath`, it can already `import` other files if it needs).

Given `viewPath` is how all the blocks currently do it ([for example](packages/fields/src/types/Content/blocks/blockquote.js)), I think it best if we revisit this in the future if we need to change the API at all.

PR #1068, opened 2019-04-29
Move Content Editor (de)serialization out of view

jesstelford:


packages/fields/src/types/Content/views/Field.js
```diff
@@ -45,7 +21,7 @@ let ContentField = ({ field, value: serverValue, onChange, autoFocus }) => {
             <Editor
               blocks={field.getBlocks()}
               value={value}
-              onChange={setValue}
+              onChange={onChange}
```
jesstelford: This has a perf impact because we're no longer wait for `onBlur` before triggering the whole state update cycle.

But, it also fixes a bug where clicking the "+" would blur the content editor, then the user would inject an image (for example), which changes the value of the field, but doesn't trigger another blur event, so now the local `value` is stale.

packages/fields/src/types/Content/views/Controller.js
```diff
@@ -60,10 +62,20 @@ export default class ContentController extends TextController {
...
-    return { document: data[path].document };
+    return { document: JSON.stringify(data[path].document) };
   };
 
-  deserialize = data => (data[this.path] ? data[this.path] : { document: {} });
+  deserialize = data => {
+    let value;
+    if (data[this.path] && data[this.path].document) {
+      value = {
+        document: JSON.parse(data[this.path].document),
```
mitchellhamilton: Are we not using graphql-type-json yet?

jesstelford: Not yet - it's on my TODO list ;)

PR #1069, opened 2019-04-29
Serialization of CloudinaryImage block from Content Editor type

jesstelford:
> Implementing the _"Mutations"_ portion of [the Rich Text Editor](https://github.com/keystonejs/keystone-5/issues/616) spec.

## tl;dr

Can add an Image to the Content Field which is saved to Cloudinary and the data is stored in KS5 in a separate, queryable list.

## How to understand this PR

There are 2 main halves to this:

- *Client Side*
    - This is where the Slate.js editor lives, and where Content _Blocks_ get rendered.
        - _For example: [the Cloudinary `single-image` block](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-30ea8b35ba5240e6003db87f1077a1c5)_
    - Serialisation of Slate's JSON tree happens on the client side before sending the result as a GraphQL mutation
        - The Slate.js tree is walked in [`Content` field's `Controller#serialize()`](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-09c7ea0554dca56eceeb08febd20cc9d)
        - Each block's `serialize` function is called (eg; the Cloudinary `single-image` block's [`serialize()` function](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-30ea8b35ba5240e6003db87f1077a1c5R182))
- *Server Side*
    - Serialised data is parsed into 2 halves in the [`processSerialised()` function](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-bc0f2c4a4eaf5befe9d49fca009523b5R39):
        1. The Slate.js JSON structure
        2. Any mutations for nested blocks (see below)
    - The [mutations are run and resolved to their IDs](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-bc0f2c4a4eaf5befe9d49fca009523b5R44)
    - The [JSON structure is updated to contain those IDs](https://github.com/keystonejs/keystone-5/pull/1069/files#diff-bc0f2c4a4eaf5befe9d49fca009523b5R74)
    - The JSON structure is saved (as the `.document` field)

## Why this is important

Here are two real-world cases which highlight the need for this:

### Use Case 1: Full-width image vs inline image resolution

A blog with posts powered by a `Content` Field. Each post can contain images in one of three formats:

1. _Full Width_: An image is the full width of the display, often breaking out of the column of the text content
    full-width-image
2. _Inline_: An image has the same width as the surrounding content.
    inline-image
3. _Aside_: An image is pulled to the side and text flows around it.
    aside

When adding these via the editor, they should be the same "block" with a setting for `['fullwidth', 'inline', 'aside']`.

When saving the content field, the image itself should be uploaded to Cloudinary, but the raw meta info (see gif below) should be stored with the `Content` Field's data.

When _reading_ the content field, the _type_ of image should determine the options sent to Cloudinary for resolution / cropping / etc.

For example, for a `fullwidth` image, we may want to leverage our existing `Cloudinary` image querying capabilities (`publicUrlFormatted`) to create something like this query:

```graphql
query Post(where: { id: "..." }) {
  content {
    cloudinaryImagesFullwidth: cloudinaryImages(where: { display: 'fullwidth' }) {
      publicUrl: publicUrlFormatted(transformation: { width: '1000' })
    }
    document # the slate.js JSON document
  }
}
```

Then for `inline` and `aside` images we might add:

```graphql
query Post(where: { id: "..." }) {
  content {
    cloudinaryImagesFullwidth: cloudinaryImages(where: { display: 'fullwidth' }) {
      publicUrl: publicUrlFormatted(transformation: { width: '1000' })
    }
    cloudinaryImagesInline: cloudinaryImages(where: { display: 'inline' }) {
      publicUrl: publicUrlFormatted(transformation: { width: '700' })
    }
    cloudinaryImagesAside: cloudinaryImages(where: { display: 'aside' }) {
      publicUrl: publicUrlFormatted(transformation: { width: '300' })
    }
    document # the slate.js JSON document
  }
}
```

Before this PR, Cloudinary image data would be embedded within the Slate.js JSON `document`, and there'd be no way to get the above queries to work (it's not possible to query within arbitrary JSON data from GraphQL).

This PR introduces an information extraction mechanism to allow `Content` Blocks to specify what data should be moved from the arbitrary slate.js JSON format out into well-typed GraphQL types (in this case a `Cloudinary` type). These types can be queried independently like any other type in Keystone 5. So, the above example query will work correctly.

#### On _Write-time_ vs _Query-time_

Ok, but really the above could all be calculated at _write-time_ when saving the slate.js JSON. But, what about mobile? You probably don't want 1000px images there. So, now do you save a different version of the post for mobile too? What about tablet? What about for post-previews? What about iframed embeds of posts? Each of them has a unique requirement on how what resolution/cropping/etc of images it should display.

By leaving it up to _query-time_, we don't have to duplicate any data, the client can dictate how it retrieves data (yay GraphQL!), and we can support any use-case going into the future.

### Use Case 2: Mentioning / Tagging a User

Within a wiki, each page is powered by a `Content` field. It is possible to `@`-mention a user (eg; @jesstelford).

Today, the design says it should just show in the final rendered page as `@jesstelford` (text-only). A query for that might look like:

```graphql
query Page(where: { id: "..." }) {
  content {
    userMention {
      username
    }
    document
  }
}
```

But, tomorrow the design changes to include an avatar preview `👩‍🎤 @jesstelford`. So now all we need to do is update the query like so:

```graphql
query Page(where: { id: "..." }) {
  content {
    userMention {
      username
      avatar {
        publicUrl: publicUrlFormatted(transformation: { width: '60' })
      }
    }
    document
  }
}
```

No need to update anything in the database.

🤯

---

NOTE: The _reading/deserialisation_ part is not implemented so you'll have to manually look at the DB to see the data. There's a future PR coming for the deserialisation which stitches the queries back together before attempting to render everything.

See the below screen recording for a run through of adding a Cloudinary Image to a Content Field. Also includes a look at what's stored in the DB. (The very end is me showing that the _reading_ part is not yet functional).

content-editor-save-cloudinary-image


packages/fields/src/types/CloudinaryImage/ImageBlock.js
```diff
@@ -19,15 +22,24 @@ export class ImageBlock extends Block {
       auxList = createAuxList(auxListKey, {
         fields: {
           // We perform the requires here to avoid circular dependencies
-          image: { type: CloudinaryImage, isRequired: true, adapter },
+          image: { type: requireDefault('./'), isRequired: true, adapter },
+          align: {
+            type: requireDefault('../Select'),
+            defaultValue: 'center',
+            options: ['left', 'center', 'right'],
+          },
```
jesstelford: ☝️  The `align` field will probably change, and is mostly there right now to show that it all works as expected.

packages/fields/src/types/CloudinaryImage/views/blocks/single-image.js
```diff
@@ -0,0 +1,213 @@
+/** @jsx jsx */
```
jesstelford: This file is mostly a copy+paste of the `image-container` block from the `Content` field with exceptions noted below.

packages/fields/src/types/CloudinaryImage/views/blocks/single-image.js
```diff
@@ -0,0 +1,213 @@
...
+import { useMemo } from 'react';
+import * as image from '../../../Content/views/editor/blocks/image';
+import * as caption from '../../../Content/views/editor/blocks/caption';
+import insertImages from 'slate-drop-or-paste-images';
+import imageExtensions from 'image-extensions';
+import { findNode } from 'slate-react';
+import { Block } from 'slate';
+import pluralize from 'pluralize';
+
+export const type = 'cloudinaryImage';
```
jesstelford: The type is specific to the `cloudinaryImage`

packages/fields/src/types/CloudinaryImage/views/blocks/single-image.js
```diff
@@ -0,0 +1,213 @@
...
+  reader.readAsDataURL(file);
+};
+
+const insertImageBlock = (editor, file, src) => {
+  editor.insertBlock({
+    type,
+    nodes: [
+      Block.create({
+        type: image.type,
+        data: { file, src },
```
jesstelford: Include the `file` in the data stored in Slate.js so we can later upload it.

packages/fields/src/types/CloudinaryImage/views/blocks/single-image.js
```diff
@@ -0,0 +1,213 @@
...
+          editor.moveToRangeOfNode(imgContainer);
+        }
+      }
+
+      next();
+    },
+  },
+];
+
+export function serialize({ node }) {
```
jesstelford: Each _Block_ handles its own serialisation (executed before sending the slate.js JSON to the server).

In this block, it looks for a nested `image` block which contains the `file` that needs uploading, and removes that `file` field from the `data` object so it doesn't get accidentally serialised and stored as JSON. Instead, it adds it as part of a `CloudinaryImage` mutation.

packages/fields/src/types/Content/Implementation.js
```diff
@@ -13,6 +15,92 @@ function flattenBlockViews(block) {
...
+ *       { data: { image: <FileObject>, align: 'center' } }
+ *     ]
+ *   },
+ *   relationshipUsers: {
+ *     create: [{ data: { id: 'abc123' } }],
+ *     connect: [{ id: 'xyz789' }],
+ *   },
+ * }
+ */
+async function processSerialised({ document, ...nestedMutations }, blocks, graphQlArgs) {
```
jesstelford: We're on the server, receiving some already-serialized data that we need to process for potential mutations (connects/creates).

packages/fields/src/types/CloudinaryImage/ImageBlock.js
```diff
@@ -1,7 +1,10 @@
 import pluralize from 'pluralize';
 import { Block } from '../../Block';
-import RelationshipFieldType from '../Relationship';
-import CloudinaryImage from './';
+
+const requireDefault = path => {
```
mitchellhamilton: Why is this necessary? It'll break when bundled with build-field-types

packages/fields/src/types/Content/views/Controller.js
```diff
@@ -62,7 +68,39 @@ export default class ContentController extends TextController {
...
+    // error thrown, or throw a new error indicating an unexpected Promise was
+    // thrown.
+    try {
+      blocks = this.getBlocks();
+    } catch (loadingPromiseOrError) {
+      if (isPromise(loadingPromiseOrError)) {
+        // `.getBlocks()` thinks it's in React Suspense mode, which we can't
+        // handle here, so we throw a new error.
+        throw new Error(
+          '`Content#getBlocks()` threw a Promise. This may occur when calling `Content#serialize()` before blocks have had a chance to fully load.'
```
mitchellhamilton: Is this a case that can actually happen right now or is it to provide a nice error for someone in the future if they change something which makes this be called at a different time?

packages/fields/types/Content/views/slate-walker/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "main": "dist/fields.cjs.js",
+  "module": "dist/fields.esm.js",
+  "preconstruct": {
+    "source": "../src/slate-walker"
+  }
+}
```
mitchellhamilton: What's this for?

packages/fields/src/types/Content/slate-walker.js
```diff
@@ -0,0 +1,64 @@
...
+    case 'inline': {
+      visitedNode = visitors.visitInline(node, recurse);
+      break;
+    }
+    case 'text': {
+      visitedNode = visitors.visitText(node, recurse);
+      break;
+    }
+    default: {
+      throw new Error(`Encountered unknown type '${node.object}' in Slate document`);
```
mitchellhamilton: I think `mark` is another type that should be here as well.

packages/fields/package.json
```diff
@@ -49,12 +49,14 @@
...
     "intersection-observer": "^0.5.1",
     "is-hotkey": "^0.1.4",
+    "lodash.get": "^4.4.2",
     "lodash.groupby": "^4.6.0",
     "lodash.isequal": "^4.5.0",
     "lodash.omitby": "^4.6.0",
     "luxon": "^1.11.4",
     "memoize-one": "^4.0.2",
     "mongoose": "^5.5.2",
+    "p-is-promise": "^2.1.0",
```
timleslie: If we're adding to package.json we should update our yarn.lock as well.

packages/fields/src/types/CloudinaryImage/ImageBlock.js
jesstelford: Because we do a dynamic `require` call in the constructor to avoid circular dependencies. And since all the modules are now ESModules, it needs `.default` interop.

Why would it break? This is a server-side file doing a synchronous require.

packages/fields/src/types/Content/views/Controller.js
jesstelford: For future safety. While I was hacking on this code I could easily see a future where refactoring accidentally called this at some point before blocks were loaded, so wanted to make that possible failure case explicit.

packages/fields/types/Content/views/slate-walker/package.json
jesstelford: Whoops - that's an artifact of a bad rebase. Will remove.

packages/fields/package.json
jesstelford: It's already there (I think it was brought in via a transitive dependency) :+1:

packages/fields/src/types/Content/slate-walker.js
jesstelford: `mark`s only show up as children of `leaves`. This algorithm only walks `nodes`.

Here's an example for the text

> H*el*lo

```json
{
  "object": "document",
  "data": {},
  "nodes": [
    {
      "object": "block",
      "type": "paragraph",
      "data": {},
      "nodes": [
        {
          "object": "text",
          "leaves": [
            {
              "object": "leaf",
              "text": "H",
              "marks": []
            },
            {
              "object": "leaf",
              "text": "el",
              "marks": [
                {
                  "object": "mark",
                  "type": "bold",
                  "data": {}
                }
              ]
            },
            {
              "object": "leaf",
              "text": "lo",
              "marks": []
            }
          ]
        }
      ]
    }
  ]
}
```

packages/fields/src/types/CloudinaryImage/ImageBlock.js
mitchellhamilton: ESM should mean the circular dependency just work no? and when it's bundled, the circular dependency won't exist at all.

It would break because everything is compiled into a single bundle(except for things that are in their own chunks like views)

jesstelford: Blow me down, you're right! ESModules have circular dependency resolution baked in. TIL 🎉

Now that I think about it; this was a problem before your refactor to use the new build system landed, and the `require()` calls were executed blindly by Node.

Just tested it with regular `import`s and it works 👍 

Have updated this branch.

PR #1070, opened 2019-04-29
Fix failing CI tests

timleslie:


.eslintrc.js
```diff
@@ -11,7 +11,6 @@ module.exports = {
   rules: {
     curly: ['error', 'multi-line'],
     'jsx-quotes': 'error',
-    'no-shadow': 'warn',
```
timleslie: This wasn't causing CI to fail, but it was treating these warnings as "important" in CI, which would mask the real failures that we care about and generally make things red. I'm going to simply turn it off, and if one day we decide that we care about this we can make it an "error" and fix all the existing problems.

PR #1077, opened 2019-05-02
Load Content Blocks in parallel to data fetching on Field view

jesstelford:
_NOTE: PR best viewed [with whitespace changes disabled](https://github.com/keystonejs/keystone-5/pull/1077/files?w=1)_

## Network requests before this PR

Screen Shot 2019-05-02 at 10 32 55 am

Note the `api` call is blocking the loading of the views.

## Network requests after this PR

Screen Shot 2019-05-02 at 10 33 53 am

The `api` call and the `views` calls are done in parallel

_(NOTE: The `api` call is starting a bit late due to Chrome's concurrent request limit (6). It will start even earlier on a prod deployment under either HTTP/2, or if the Admin UI is served from a different domain to the API)_

## More to come

You'll notice this PR only addresses the `Content` field's blocks and nothing else. I purposely wanted to keep this PR small to guage reception. If it all looks :+1:, then I can use the same technique with all the other field types, and for `Cell` / `Filters` too.

## Motivations

There are two main motivations for this change:

1. Performance as noted above
2. The Content Editor needs to be able to deserialise the Slate.js JSON blob returned by the API request. Until it's deserialised, it can't be rendered. But the way the code is structured before this PR; we don't load the views until we have the complete (deserialised) data. 🐓&🥚. This change makes it possible to ensure the deserialisation logic (which the views provide) is preloaded before trying to move on with deserialsation & display.

## Inspiration

This solution was partly inspired by Facebook's recent F8 video on how they make their dynamic view loading as performant as possible: https://developers.facebook.com/videos/2019/building-the-new-facebookcom-with-react-graphql-and-relay/

### DIFF
--- a/packages/admin-ui/client/pages/Item/index.js
+++ b/packages/admin-ui/client/pages/Item/index.js
@@ -1,11 +1,12 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { Component, Fragment, useMemo, useCallback } from 'react';
+import { Component, Fragment, Suspense, useMemo, useCallback } from 'react';
 import styled from '@emotion/styled';
 import { Mutation, Query } from 'react-apollo';
 import { withRouter } from 'react-router-dom';
 import { withToastManager } from 'react-toast-notifications';
 import memoizeOne from 'memoize-one';
+import isPromise from 'p-is-promise';
 
 import { Container } from '@arch-ui/layout';
 import { Button } from '@arch-ui/button';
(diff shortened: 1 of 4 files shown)


packages/admin-ui/client/pages/Item/index.js
```diff
@@ -274,83 +275,119 @@ const ItemNotFound = ({ adminPath, errorMessage, list }) => (
...
+          // Trigger the suspense boundary and wait for all fields in
+          // parallel.
+          throw Promise.all(initialisationPromises);
+        }
+
+        return children();
+      }}
+    </Render>
+  </Suspense>
+);
```
jesstelford: In a future PR, I will move this component somewhere re-usable as the logic is identical for `Cell`/`Filter`.

PR #1079, opened 2019-05-02
Create modal loading state

jesstelford:
## Before this PR

The modal wouldn't open until the views were loaded

create-modal-waiting

## This PR

Opens the modal immediately and shows a loading state so the user isn't left wondering what's going on

create-modal-loading

(_note: I cut out some of the loading animation in both because I had my network throttling settings too high_)


jesstelford:
Am force-merging ☝️ That CI failure is just Cypress being silly again.

PR #1080, opened 2019-05-02
Optimistically open Nested Create Item Modal and show loading spinner

jesstelford:
Same treatment as #1079, but for the nested create modal on Relationship types.

### DIFF
--- a/test-projects/basic/cypress/integration/relationship-field-re-hydration_spec.js
+++ b/test-projects/basic/cypress/integration/relationship-field-re-hydration_spec.js
@@ -2,10 +2,23 @@ describe('Testing re-hydration', () => {
   before(() => cy.visit('/reset-db'));
 
   it('Our new category should appear after we add it', () => {
+    function clickCategoriesSelect() {
+      // Wait for the select box to be rendered
+      cy.get('#react-select-ks-input-categories div[aria-hidden="true"]');
+      // Wait for any re-renders to happen which accidentally destroy the select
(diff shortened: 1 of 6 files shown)


test-projects/basic/cypress/integration/relationship-field-re-hydration_spec.js
```diff
@@ -5,7 +5,9 @@ describe('Testing re-hydration', () => {
     cy.visit('/admin/posts');
     cy.get('#list-page-create-button').click();
     cy.wait(150);
-    cy.get('#react-select-ks-input-categories div[aria-hidden="true"]').click();
+    cy.get('#react-select-ks-input-categories div[aria-hidden="true"]')
+      .first()
```
jesstelford: Fixes the test failures on CI. I can't replicate locally and it only appears to happen _sometimes_ on CI.

It's particularly annoying because the failure indicates that selector matches 2 elements, which when run locally I can only see a single element 😠

jesstelford: It's [still failing](https://circleci.com/gh/keystonejs/keystone-5/17394#artifacts/containers/0) :sob:

Trying a different method which will hopefully work.

jesstelford: Errmmmaagghheerrrd, it works 🎉

PR #1083, opened 2019-05-03
Loading update many

jesstelford:
_NOTE: Built on top of #1080_

## Before this PR

loading-update-many-old

Note that the entire Update modal would disappear.

## With this PR

loading-update-many

Show an inline loading spinner as each field view is loaded. No more disappearing acts!

### DIFF
--- a/.changeset/636920b6/changes.json
+++ b/.changeset/636920b6/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/admin-ui", "type": "patch" }], "dependents": [] }
--- a/.changeset/636920b6/changes.md
+++ b/.changeset/636920b6/changes.md
@@ -1,0 +1,1 @@
+- Update Many modal shows spinner while loading fields
--- a/packages/admin-ui/client/components/UpdateManyItemsModal.js
+++ b/packages/admin-ui/client/components/UpdateManyItemsModal.js
@@ -1,10 +1,13 @@
-import React, { Component, Fragment, useMemo, useCallback, Suspense } from 'react';
+/** @jsx jsx */
+import { jsx } from '@emotion/core';
+import { Component, Fragment, useMemo, useCallback, Suspense } from 'react';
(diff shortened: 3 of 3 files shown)


jesstelford:
Yeah, that's definitely possible too, it just means the user pays the loading cost for _all_ fields even if they only update one or two.

I figured the main use-case for the Update Many modal is to only update one or two fields at a time, so it's more optimal to load the views on demand.

mitchellhamilton:
IMO, it’s worth it to pre-load the views since the views aren’t really that big and it’s a much nicer experience to not see a loading indicator.

jesstelford:
The Content field view is quite large (pulls in all of Slate + blocks + whatever they require like Cloudinary Images, etc).

Here's what it looks like on a *Good 3G* connection to load a `Text` vs `Content` field:

update-many-slow-content

If they're updating the `Content` field when the user opens the Update Many modal, they'll pay the worst case loading time. But if they're only ever editing other fields (even a couple); they're hit with less of a loading time. Whereas if we load all fields always, they'll always pay the worst case loading time no matter what.

mitchellhamilton:
Why would it be slower if we preload all of them? I’m not saying we should require that they’re all loaded before rendering any, just that they’re preload them all. (I’m ignoring parse and evaluation time because I’m assuming it won’t be overly significant but it would probably be a good thing to test) Also, preloading them would mean we’re starting the request earlier so by the time that they’ve selected an option, the view will probably be loaded.

jesstelford:
I think I misunderstood what you meant by _preload_, sorry.

What would happen if the views aren't loaded by the time the user selects an item from the drop-down? They'd still need to see a loader there so they know things are happening and it's not broken.

mitchellhamilton:
Yeah, I totally agree we need to have a loader there. We can still preload the views though so, in most cases, people don’t see the loader.

jesstelford:
Awesome. Sounds like that would be an addition _on top of_ this PR? Can we merge this one and create an issue for that enhancement?

PR #1085, opened 2019-05-03
Show loading spinner while loading views in List Table

jesstelford:
_NOTE: PR best viewed [with whitespace changes disabled](https://github.com/keystonejs/keystone-5/pull/1085/files?w=1)_

This moves the loading spinner _inside_ the table which allows the user to interact with the sorting / columns display even while data is loading.

It also implements the things in #1077 so the data and the views can load in parallel.

posts-load-views-in-table

### DIFF
--- a/packages/admin-ui/client/components/ListTable.js
+++ b/packages/admin-ui/client/components/ListTable.js
@@ -1,9 +1,10 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import React, { Component } from 'react';
+import React, { Component, Suspense, Fragment } from 'react';
 import styled from '@emotion/styled';
 import { Link } from 'react-router-dom';
 
+import { captureSuspensePromises, noop } from '@keystone-alpha/utils';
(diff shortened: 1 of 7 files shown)


jesstelford:
@mitchellhamilton I broke the integration tests with this change so had to rework things slightly (just moved the `<Suspense>` component up the tree a little bit and added a `data-` attribute so Cypress can wait for things to be loaded before trying to interact with the table.

Can I get a re-review?

packages/admin-ui/client/components/ListTable.js
```diff
@@ -304,21 +309,27 @@ export default function ListTable(props) {
...
   const cypressId = 'ks-list-table';
 
-  return (
-    <Card css={{ marginBottom: '3em' }}>
-      <Table id={cypressId} style={{ tableLayout: isFullWidth ? null : 'fixed' }}>
+  // +2 because of check-boxes on left, and overflow menu on right
+  const columns = fields.length + 2;
+
+  const TableContents = ({ isLoading, children }) =>
+    console.log({ isLoading }) || (
```
mitchellhamilton:
```suggestion
    (
```

PR #1088, opened 2019-05-03
Admin UI static building

mitchellhamilton:


mitchellhamilton:
The this is mostly ready but I have some questions about the CLI @jesstelford:

- Why are we using the CLI to execute custom servers? From what I understand, a large part of the reason behind using a custom server is being able to run keystone programmatically and using a CLI to execute it defeats that. I'm especially confused because the CLI was modeled after Next's and Next doesn't execute custom servers, people just run node on the server file.
- Why is the `--out` flag a thing? I much prefer [what Next does where you specify the output folder in the Next config](https://github.com/zeit/next.js#setting-a-custom-build-directory) so for Keystone, I would think it would be a `distDir` export from `index.js`. It's especially nice because it removes the problem of having to specify the dist directory in two commands and in a custom server file if you have one.

jesstelford:
> Why are we using the CLI to execute custom servers?

No particular reason. It's basically a very thin wrapper over just executing the file with `node` directly.

> Why is the `--out` flag a thing over an export in `index.js`?

Note that `next` allows specifying that directory on the command line too (`-o`):

```
next export --help

    Description
      Exports the application for production deployment

    Usage
      $ next export [options] <dir>

    <dir> represents where the compiled dist folder should go.
    If no directory is provided, the 'out' folder will be created in the current directory.

    Options
      -h - list this help
      -o - set the output dir (defaults to 'out')
      -s - do not print any messages to console
```

I don't mind there being a config option too. Personally, I go back and forth between preferring the CLI option and the the config file option for various utils. I like the atomicity of the CLI option - I can run the CLI with different options and not worry about a config file getting in the way. But on the other hand, when there are lots of options, a config file is much easier to read and edit ¯\_(ツ)_/¯

mitchellhamilton:
> No particular reason. It's basically a very thin wrapper over just executing the file with node directly.

How would you feel about removing it because I feel like it's confusing because you can't use the CLI to pass in options because you have to pass it into `keystone.prepare` so it misleads people into thinking the CLI is doing something?

Procfile
```diff
@@ -1 +1 @@
-web: yarn start:test
+web: cd demo-projects/blog && node node_modules/@keystone-alpha/keystone/bin/cli
```
mitchellhamilton: I know this is strange, Heroku was being weird and this fixed it. I think it's only because of bolt monorepo reasons so it won't cause an issue for consumers

jesstelford:
Yeah, take it out if it's getting in the way 👍

jesstelford:
Looks like #1096 caused some conflicts with this branch - once they're fixed, let me know and I'll give this a review 👍

mitchellhamilton:
Fixed.

packages/keystone/bin/commands/build.js
```diff
@@ -1,20 +1,42 @@
...
+      let adminOutputPath = path.join(resolvedDistDir, 'admin');
+      await fs.remove(adminOutputPath);
+      console.log('Building Admin UI!');
+      await admin.staticBuild({
+        apiPath: '/admin/api',
+        outputPath: adminOutputPath,
+        graphiqlPath: '/admin/graphiql',
+      });
+      console.log('Built Admin UI!');
     }
```
jesstelford: Probably needs some form of output when there's nothing to do, with a hint of what to add maybe?

```suggestion
    } else {
      console.log('Nothing to build.');
      console.log(`To create an Admin UI build, make sure you export 'admin' from ${path.relative(_cwd, entryFile)}`);
    }
```

packages/keystone/bin/commands/start.js
```diff
@@ -0,0 +1,28 @@
...
+    '-o':       '--out',
+  },
+  help: ({ exeName }) => `
+    Usage
+      $ ${exeName} start --port=3000
+
+    Options
+      --port, -p  Port to start on [${keystone.DEFAULT_PORT}]
+      --entry     Entry file exporting keystone instance [${keystone.DEFAULT_ENTRY}]
+      --out, -o   Directory where build was saved [dist]
```
jesstelford: 🤔 Is `--out` potentially confusing here? Perhaps it should be `--build` or `--dist`?

Or maybe not an option at all, like next's CLI:

```
next start --help

    Description
      Starts the application in production mode.
      The application should be compiled with `next build` first.

    Usage
      $ next start <dir> -p <port>

    <dir> is the directory that contains the compiled dist folder
    created by running `next build`.
    If no directory is provided, the current directory will be assumed.
    You can set a custom dist folder in config https://github.com/zeit/next.js#custom-configuration

    Options
      --port, -p      A port number on which to start the application
      --hostname, -H  Hostname on which to start the application
      --help, -h      Displays this message
```

I like their `<dir>` option which doesn't need a command line switch.

jesstelford: Looks like `--out` isn't being used. It'll need to be picked up within `executeDefaultServer()`, and passed into `keystone.prepare()` inside `packages/keystone/bin/utils.js`.

packages/keystone/bin/utils.js
```diff
@@ -0,0 +1,37 @@
...
+      `)
+    );
+  }
+}
+
+function executeDefaultServer(args, entryFile) {
+  const port = args['--port'] ? args['--port'] : keystone.DEFAULT_PORT;
+
+  return keystone
+    .prepare({ entryFile, port })
```
jesstelford: `keystone.prepare()` needs to accept a new parameter `adminDistDir` so it knows where to load the `adminUI` build files from.

Then later within `packages/admin-ui/server/AdminUI.js`, `'dist'` can be set to the value of `adminDistDir` (see comment below).

packages/admin-ui/server/AdminUI.js
```diff
@@ -60,6 +62,102 @@ module.exports = class AdminUI {
...
+      ...this.keystone.getAdminMeta(),
+    };
+  }
+
+  createProdMiddleware() {
+    const app = express();
+
+    app.use(compression());
+
+    const builtAdminRoot = path.join(process.cwd(), 'dist', 'admin');
```
jesstelford: `'dist'` here shouldn't be hard coded (see above comment).

packages/admin-ui/server/AdminUI.js
```diff
@@ -60,6 +62,102 @@ module.exports = class AdminUI {
...
+      adminPath,
+      apiPath,
+      graphiqlPath,
+      ...this.getAdminMeta(),
+      ...this.keystone.getAdminMeta(),
+    };
+  }
+
+  createProdMiddleware() {
+    const app = express();
```
jesstelford: Should probably be an express `Router`:

```suggestion
    const app = express.Router();
```

mitchellhamilton:
> Could probably use some kind of sanity check test like checking that doing a build actually outputs files, and starting the server from the build returns a 200?

What do think about making one of the cypress test projects do `build` and `start` instead of running `dev` so it essentially does exactly that?

jesstelford:
> What do think about making one of the cypress test projects do build and start instead of running dev so it essentially does exactly that?

Yeah, great idea!

Copy one of the existing tests to a new test file so it's isolated, and we can build on it in the future :+1:

mitchellhamilton:
> Copy one of the existing tests to a new test file so it's isolated, and we can build on it in the future 👍

Do you mean one of the jest tests? I'd rather not do that right now because those particular tests do so much mocking and the internals will change soon because of the servers stuff so the tests would have to change really soon because of that. After the servers stuff is done, I'd like to change those tests to not do very much mocking at all and test everything nicely.

---

I've made it so the login integration tests run keystone build and start the server with NODE_ENV=production on CI so we have a smoke test for the build and prod start.

I've also fixed the output directory stuff and updated the docs in the root README so it's accurate with what's currently implemented.

packages/core/lib/core.js
```diff
@@ -73,6 +74,7 @@ module.exports = {
         // Force the admin & port
         ...(appEntry.admin ? { adminUI: appEntry.admin } : {}),
         port,
+        distDir: distDir || appEntry.distDir || 'dist',
```
jesstelford: `'dist'` here should probably be a default like the `entryFile`/`port` above:

```suggestion
        distDir: distDir || appEntry.distDir || DEFAULT_DISTDIR,
```

That can then be exported above, and imported into the `build.js` command to be used there:

https://github.com/keystonejs/keystone-5/pull/1088/files#diff-e3b5d2c74bc1774830696f662545611cR24

packages/keystone/bin/commands/dev.js
```diff
@@ -1,46 +1,16 @@
...
-    '-p':       '--port',
-    '--entry':  String,
+    '--port':  Number,
+    '-p':      '--port',
+    '--entry': String,
   },
   help: ({ exeName }) => `
     Usage
-      $ ${exeName} dev --port=3000
+      $ ${exeName} dev <dist> --port=3000
```
jesstelford: It doesn't look like `<dist>` here does anything (but it does in `start.js`)

```suggestion
      $ ${exeName} dev --port=3000
```

jesstelford:
Looks like that test you added is catching a bug already! 🎉

PR #1094, opened 2019-05-07
Simplify Content Type setup to extend Relationship

jesstelford:
I was struck with inspiration 🛰💥, and realised that the `Content` type should extend the `Relationship` type (instead of `Text`), which means all the tricky stuff I was trying to do with manually triggering nested mutations can be handled by KS5 like normal.

This also greatly simplifies the next bit of work which is querying for the data, plus enables the `from` back-link `Relationship` to work out of the box.

### DIFF
--- a/packages/fields/src/types/CloudinaryImage/ImageBlock.js
+++ b/packages/fields/src/types/CloudinaryImage/ImageBlock.js
@@ -4,12 +4,25 @@ import CloudinaryImage from './';
 import SelectType from '../Select';
 import RelationshipType from '../Relationship';
 
+const RelationshipWrapper = {
+  ...RelationshipType,
+  implementation: class extends RelationshipType.implementation {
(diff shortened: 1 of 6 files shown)


packages/fields/src/types/CloudinaryImage/ImageBlock.js
```diff
@@ -4,12 +4,25 @@ import CloudinaryImage from './';
 import SelectType from '../Select';
 import RelationshipType from '../Relationship';
 
+const RelationshipWrapper = {
+  ...RelationshipType,
+  implementation: class extends RelationshipType.implementation {
+    async resolveNestedOperations(operations, item, context, ...args) {
+      const result = await super.resolveNestedOperations(operations, item, context, ...args);
```
jesstelford: Taking advantage of work from https://github.com/keystonejs/keystone-5/pull/1033

PR #1096, opened 2019-05-07
Remove custom server execution from the cli

mitchellhamilton:
As discussed in #1088, this removes custom server execution from the cli as it's confusing because it's the same as directly running the server file with node.


mitchellhamilton:
Updated all the docs I could find and the template in create-keystone-app.

Issue #1039, opened 2019-04-16
Separate AdminUI presentational props from field config props


jesstelford:
From https://github.com/keystonejs/keystone-5/pull/1038#issuecomment-483854532

@jossmac says:

> Would a `Textarea` (or other) field type make more sense? I'm not sure that we should let presentational component props inform the config for our types.

I like the idea of separating config for the field and config for the admin UI.

A new field type of `Textarea` would be identical to `Text` in everything except the one `isMultiline` prop, so I'm not a huge fan of making that a new field. It could also lead to many many field types all because they display slightly differently in the admin UI.

Perhaps an alternative is to accept an `admin` prop on the field config level (like we do on the list level config already) which can pass these kinds of options through? That would make it clear this setting is purely for the Admin UI, and opens it up for custom field types to have a way of accepting these presentational-only  props instead of as a generic config.

JedWatson:
So the catch here (for me) is that not every option is absolutely either admin UI or db / graphql related.

If you have two options, `a` and `b`, which both affect the Admin UI, but `a` also affects the graphQL API or db adapter, then you'd probably end up with this:

`path: { type: Field, a: 'value', admin: { b: 'value' } }`

... which I have historically found super hard to make consistent and predictable.

IMO the cleanest approach is to say that "this is the field config API" - document all the options, what they do, and what they relate to. Group options that make sense, but not with artificial and inconsistent boundaries between the Admin UI and the back-end.

jesstelford:
Unfortunately that wont work for things like API keys.

A great example is the CloudinaryImage type - we don't want to accidentally pass the cloudinary api key to the Admin UI.

This has recently come up with Blocks which have this separation of some things executed server side and some things executed client side. See #1097

jesstelford:
The good thing is; there's no reason the server can't have access to the `adminConfig` options.

On the server side, it can access things directly on the `config` object, and/or on `config.adminConfig` :+1:

jesstelford:
Bringing in some more suggestions from https://github.com/keystonejs/keystone-5/pull/1286#issuecomment-499749679

[Some features start] to add a lot of functionality to schema definitions which is only applicable to the _optional_ Admin UI.

Can I suggest we move this to the configuration _of the Admin UI_?

For example (I'm not wedded to this API at all, it's only to demo my thoughts):

```javascript
keystone.createList('Task', {
  fields: {
    title: { type: Text },
    // I want this field hidden in the Admin UI
[...]

PR #1097, opened 2019-05-08
Merge default block config and only pass admin config to admin UI

jesstelford:
## Update

Based on the below discussion, this PR has a competing implementation:

- #1121: Implement an `extendAdminMeta()` API for blocks.
## Original PR

Ref #1039

Note: I added a `adminConfig` key to block configs so we can separate the server-side config options (Cloudinary keys), and the client side options (an upcoming PR will add a `query` config for generating the correct graphQL query client side).

### DIFF
--- a/.changeset/91cec723/changes.json
+++ b/.changeset/91cec723/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/fields", "type": "minor" }], "dependents": [] }
--- a/.changeset/91cec723/changes.md
+++ b/.changeset/91cec723/changes.md
@@ -1,0 +1,1 @@
+Allow Content Blocks to specify default config which is deeply merged into user-supplied config
--- a/.changeset/b865199c/changes.json
+++ b/.changeset/b865199c/changes.json
@@ -1,0 +1,90 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/cypress-project-basic", "type": "patch" },
+    { "name": "@keystone-alpha/fields", "type": "major" }
+  ],
+  "dependents": [
+    {
+      "name": "@keystone-alpha/admin-ui",
+      "type": "patch",
+      "dependencies": ["@keystone-alpha/fields"]
+    },
+    {
+      "name": "@keystone-alpha/fields-wysiwyg-tinymce",
+      "type": "patch",
+      "dependencies": ["@keystone-alpha/fields"]
+    },
+    {
+      "name": "@keystone-alpha/keystone",
+      "type": "patch",
+      "dependencies": ["@keystone-alpha/fields"]
+    },
+    {
+      "name": "@keystone-alpha/passport-auth",
+      "type": "patch",
+      "dependencies": ["@keystone-alpha/fields"]
+    },
+    {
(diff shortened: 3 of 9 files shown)


jesstelford:
To see this in action, check out the `deserialize-blocks` branch:

https://github.com/keystonejs/keystone-5/blob/237dc28b0613d8de2dc8fac3a1375b1322576461/packages/fields/src/types/CloudinaryImage/index.js#L30-L42

mitchellhamilton:
I don't totally get the motivation behind this, why is this better than having an `extendAdminMeta` method on block types and by default not passing the options to the admin UI like field types?

jesstelford:
To see this in action, check out the `deserialize-blocks` branch:

https://github.com/keystonejs/keystone-5/blob/237dc28b0613d8de2dc8fac3a1375b1322576461/packages/fields/src/types/CloudinaryImage/index.js#L30-L42

Which then enables overwriting that config here:

https://github.com/keystonejs/keystone-5/blob/237dc28b0613d8de2dc8fac3a1375b1322576461/test-projects/basic/index.js#L125

For example, in the Admin UI, I might only want to load smaller images:

```javascript
      blocks: [
        [CloudinaryImage.blocks.image, {
          adapter: cloudinaryAdapter,
          adminConfig: {
	          defaultQuery: `
	            cloudinaryImages {
	              id
	              image {
	                publicUrl: publicUrlTransformed(transform: { width: '60px' height: '60px' })
	              }
	              align
	            }
	          `,
        }],
```

jesstelford:
The `adminConfig` key prevents accidentally leaking credentials by trying to send `adapter: cloudinaryAdapter` to the client.

mitchellhamilton:
Yeah, I totally get what this enables but having something like `extendAdminMeta` enables the same thing and doesn't create cognitive overhead for consumers where with this, they need to think "is this an option for the field/block or for the adminConfig?" 

So, i.e. it's a question of control, what should control what is sent to the admin UI? The user or the field/block type? I'm trying to say, the field/block type should control it since it has the information and it means users don't have to care.

jesstelford:
One important point I forgot to mention with the above example: the user might also have a website powered by the GraphQL API, in that case the embedly API key won't make its way to the website.

Ie;

- the `embed` block needs an embedly API key _client side_. When one is supplied via the Keystone block config;
    - The Admin UI will work as expected
    - The application website will _not_ with without also adding the key elsewhere
- the `cloudinaryImage` block needs an API key _server side_. When one is supplied via the Keystone block config;
    - The Admin UI will work as expected
    - The application website will also work as expected

This was the confusion I was referring to.

jesstelford:
I'm going to extract out of this PR the `default config` logic (which is not controversial and should be able to merged quickly).

I'll also open a new PR with the `extendAdminMeta()` method added to blocks.

dcousens:
`extendAdminMeta` and `extendAdminMetaPrivate`, to discern in the latter, it will *never* be sent to client?

jesstelford:
Both those methods will be implemented on the blocks, and probably never seen by the app developers, so it wouldn't make much difference. Also, there's no such thing as a "private" admin config. It all gets bundled up into the JS, which is all public.

jesstelford:
Closing in favour of #1121

PR #1100, opened 2019-05-08
Meetup demo event details

SeanDoyleGit:
Implemented event details page with dynamic routing for the meetup demo.


demo-projects/meetup/server.js
```diff
@@ -19,6 +19,11 @@ Promise.all([keystone.prepare({ port }), nextApp.prepare()])
     // Attach the auth routes
     server.app.use('/', createAuthRoutes(keystone));
 
+    server.app.get('/event/:id', (req, res) => {
```
JedWatson: Irrelevant after next-routes has been added, but this should really be colocated with the next request handler

PR #1105, opened 2019-05-09
Move Admin UI session middleware registering to Admin UI internals

mitchellhamilton:
I'm working on the "servers" thing and noticed that the Admin UI session middleware registering happened outside of admin UI internals for no reason(i think) and that won't work with the encapsulation that servers will have.

I've added a changeset with major bumps since using the current version of `@keystone-alpha/server` and a version of `@keystone-alpha/admin-ui` with this change or vice versa would break things.


mitchellhamilton:
Closing because the session stuff will change and this is intentionally out of place for now.

PR #1109, opened 2019-05-10
Custom apps & BYO `express`

mitchellhamilton:
Closes #960 
Implements #1150

See PR description from @jesstelford below 👇 https://github.com/keystonejs/keystone-5/pull/1109#issuecomment-493277984


packages/keystone/bin/commands/build.js
```diff
@@ -21,21 +21,25 @@ module.exports = {
...
-      await admin.staticBuild({
-        apiPath: '/admin/api',
-        distDir: resolvedDistDir,
-        graphiqlPath: '/admin/graphiql',
-      });
+    if (servers) {
+      await Promise.all(
+        servers.map((server, index) => {
+          return server.build({
+            apiPath: '/admin/api',
```
gautamsi: should we use the config value for api and graphiql path?

mitchellhamilton: Yes, this is WIP for a reason.

gautamsi: just making sure as previous code also used same thing :)

jesstelford:
Heads up I've taken over this branch and am pushing my WIP work :+1:

jesstelford:
## Changes

- The `index.js` export for `admin` must now be exported in the `servers`
  array:
  ```diff
   module.exports = {
     keystone,
  -  admin,
  +  servers: [admin],
   }
  ```
- The `keystone.prepare()` method (often used within a _Custom Server_
  `server.js`) no longer returns a `server`, it now returns a `middlewares`
  array:
  ```diff
  +const express = require('express');
   const port = 3000;
   keystone.prepare({ port })
  -  .then(async ({ server, keystone: keystoneApp }) => {
  +  .then(async ({ middlewares, keystone: keystoneApp }) => {
       await keystoneApp.connect();
  -    await server.start();
  +    const app = express();
  +    app.use(middlewares);
  +    app.listen(port)
     });
  ```
- Removed the `@keystone-alpha/core` package; it has been merged into `@keystone-alpha/keystone`.
  - `core` had a very limited scope, and added confusion about what to import within a custom server.
  - We were unnecessarily exporting a bunch of constants from `core` for use within the `keystone` package.
  - Merging them together reduces the confusion (it's just `keystone` everywhere), cleaned up the exports, and also had the side-effect of making the tests a little better.
- `@keystone-alpha/server` was renamed to `@keystone-alpha/server-graphql` to be consistent with other 'servers' (`@keystone-alpha/server-next` & `@keystone-alpha/server-static`).
- The default route handler (`/` which shows the "Welcome to Keystone [Open the Admin UI]" screen) has been moved from the `@keystone-alpha/server-graphql` package to the `@keystone-alpha/admin-ui` package.
  - A new option to disable the default route handler has been added to the AdminUI's constructor:
  ```javascript
  const admin = new AdminUI(keystone, { disableDefaultRoute: true });
  ```

## Open Questions

- Should we rename `@keystone-alpha/admin-ui` -> `@keystone-alpha/server-admin-ui` to be consistent with the others (`@keystone-alpha/server-graphql`, `@keystone-alpha/server-next` & `@keystone-alpha/server-static`)?
- Should the `disableDefaultRoute` option be inverted so the default route is _not_ added by default, but we can keep it in for the demos? So far it appears we almost always want to disable it. Ie;
  ```javascript
  const admin = new AdminUI(keystone, { enableDefaultRoute: true /* default: false */ });
  ```

## Up Next

- Documentation
  - The expected API for `servers`.
  - Different ways to specify `servers`:
    - Export the `servers` array from `index.js`
    - Create a custom `server.js` _and_ export the `servers` array from `index.js`
    - Create a custom `server.js` which instantiates all of its own middlewares
  - How to customise the GraphQL middleware (it's just another server 😉) 
- Add more tests for the different edge cases of the `prepare` method + CLI (currently written as `test.todo()` in the code as placeholders)
- There is a `port` parameter being passed to `prepareMiddleware()` which is only used for some `console.log`s. I want to move all that into the default server and stop passing the `port` parameter.
- This new structure makes it easier to start listening to requests immediately rather than waiting for all of KS to be ready + the AdminUI, etc. We can avoid the bad DX of having to hit F5 constantly until it boots up.
- Most of the custom servers now only exist to insert default data. Can we provide a better API for that which is setup in `index.js`, or (better yet) is a separate script that is run only once? This would make most of our custom servers go away leaving us with a nice clean `index.js` everywhere.
- Get the `build()` step working for static assets.

jesstelford:
Did an in-person review with @JedWatson, came up with these 3 things todo for this PR:

- [ ] Don't automagically insert the `GraphQLApp` instance if it's not present. Now that we've got the concept of `apps`, we should make instantiating the `GraphQLApp` explicit.
- [ ] `keystone.prepare()` shouldn't be calling `require('./index.js')`, instead the app developer's code in `server.js` should do `const { keystone, apps } = require('./index'); keystone.prepare({ apps }).then(...)`
- [ ] Make sure the GraphQL API can live at any arbitrary endpoint, and that the AdminUI can access it

And these ones todo after this PR lands (added to #1182):

1. Rename all the `path`s to `route`s (to be consistent with the `StaticApp` API of accepting a `route` (the URL) and a `path` (local file system).

JedWatson:
@jesstelford thinking about it further... I really do like talking about the route of an app as a "path". It's pretty consistent terminology across other parts of keystone... like the "Admin path" or the "API path" that has emerged naturally, as opposed to an "Admin route" or "API route" which to me sounds like the actual route handler (or, really, middleware 🤷‍♂)

Even in the Admin UI, we use `path` for custom pages to mean the mount point and `component` to mean the path to the file that exports a react component to render when the user navigates to that path.

Maybe instead of switching all our "paths" to "routes", we can change the Static app's config instead. I could totally get my head around having a "path" and "src", where `path` is the bit of the url it is mounted under, and `src` is the directory to serve static assets from.

I think this is cognitively consistent with [express.static](https://expressjs.com/en/starter/static-files.html) as well, from those docs they call the argument `root` and describe "setting a mount path" in the same context as I define "path" above.

> The function signature is:

```js
express.static(root, [options])
```

> To create a virtual path prefix (where the path does not actually exist in the file system) for files that are served by the express.static function, specify a mount path for the static directory, as shown below:

```js
app.use('/static', express.static('public'))
```

What do you think?

```js
const keystone = new Keystone();
const admin = new AdminUI(keystone, { path: '/admin'});
const static = new StaticApp({ path: '/', src: './assets'});
```

(if you're inclined to match express more closely, substitute `src` for `root` above, I'm not sure if it's clearer to say that the directory is the source being served or avoid implying it may be modified or processed in some way that `src` folders often are these days. I personally don't find `root` very clear, so another substitute for `src` could be `from` like this)

```js
const static = new StaticApp({ path: '/', from: './assets'});
```

jesstelford:
That's a pretty sound argument - let's go with `path` for the URL portion, and `src` for the file-system portion 👍

jesstelford:
In-person review with @timleslie:

- [ ] The apps should have named exports, not defaults (to be consistent with the rest of our exports)
- [ ] Rename `@keystone-alpha/app-admin` to `@keystone-alpha/app-admin-ui`
- [ ] Rename `disableDefaultRoute` option to `enableDefaultRoute` and default to `false` (see [_Open Questions_ above](https://github.com/keystonejs/keystone-5/pull/1109#issuecomment-493277984))

PR #1114, opened 2019-05-12
Explicit field config

timleslie:


packages/fields/src/Implementation.js
```diff
@@ -4,23 +4,24 @@ import { parseFieldAccess } from '@keystone-alpha/access-control';
 class Field {
   constructor(
     path,
-    config,
+    { hooks = {}, isRequired, defaultValue, access, label, schemaDoc, ...config },
```
timleslie: This is the crux of this PR. In a future PR I intend to look at how the remaining config options are used in relation to the field adapters.

packages/fields/src/types/Content/Implementation.js
```diff
@@ -99,7 +99,7 @@ async function processSerialised(document, blocks, graphQlArgs) {
 }
 
 export class Content extends Relationship {
-  constructor(path, fieldConfig, listConfig) {
+  constructor(path, { blocks, ...fieldConfig }, listConfig) {
```
jesstelford: Can minimise the diff a bit here by renaming the input parameter:

```suggestion
  constructor(path, { blocks: inputBlocks, ...fieldConfig }, listConfig) {
```

PR #1115, opened 2019-05-13
Some notes on the "universal" field config

molomby:
Some notes from the discussion Friday, in addition to those in the [notion doc](https://www.notion.so/thinkmill/Hairy-Session-May-10-cbc7920f1ef14ff5b8e79da8f4921765). Probably needs editing/cleaning up.. 

Not sure if notes should be added to the repo or stored elsewhere?


MadeByMike:
@molomby I've captured these in Notion.

PR #1118, opened 2019-05-13
[docs] Add discussions on Schemas and Relationships

jesstelford:
See a rendered version of the docs here: https://github.com/keystonejs/keystone-5/blob/relationship-discussion/docs/discussions/schema.md


JedWatson:
👌 this looks great @jesstelford

jesstelford:
That test failure seems like an error in our linting config - the URL does work:

links

PR #1099, opened 2019-05-08
Display previously saved/serialised CloudinaryImage block

jesstelford:
_NOTE: Includes the changes from #1097_

Continuing the work on #161. For previous work, see PRs #684 #1013 #1014 #1017 #1018 #1030 #1033 #1057 #1067 #1068 #1069 #1077 #1094 

deserialise

PR #1120, opened 2019-05-13
Blocks have default config merged into user config

jesstelford:
Adds some functionality which #1099 requires; setting a default config that the user can overwrite for querying cloudinary image blocks.

NOTE: This change was originally part of #1097, but has been extracted here to avoid blocking on that discussion.


jesstelford:
Nope, this shouldn't be a separate PR... See #1097 for the implementation of this.

PR #1121, opened 2019-05-13
Allow blocks to pick data sent to the adminUI via extendAdminMeta()

jesstelford:
Note: This is a competing implementation to #1097 based on the discussion there.

### DIFF
--- a/packages/fields/src/types/Content/Implementation.js
+++ b/packages/fields/src/types/Content/Implementation.js
@@ -203,6 +203,7 @@ export class Content extends Relationship {
     this.auxList = auxList;
     this.listConfig = listConfig;
     this.blocks = blocks;
+    this.complexBlocks = complexBlocks;
   }
 
   /*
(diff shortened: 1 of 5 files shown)


packages/fields/src/types/Content/Implementation.js
```diff
@@ -227,6 +227,21 @@ export class Content extends Relationship {
...
+        .filter(([, blockConfig]) => blockConfig && Object.keys(blockConfig).length)
+        .reduce(
+          (options, block) => ({
+            ...options,
+            [block[0].type]: block[1],
+          }),
+          {}
+        ),
+
       blockOptions: this.blocks
```
mitchellhamilton: This should be removed right? Otherwise the blockOptions above would have no effect?

jesstelford: 🤦‍♂ Copy pasta error. Thanks!

packages/fields/src/types/Content/Implementation.js
```diff
@@ -226,9 +227,22 @@ export class Content extends Relationship {
...
       blockTypes: this.blocks.map(block => (Array.isArray(block) ? block[0] : block).type),
 
-      // Key the block options by type to be serialised and passed to the client
       blockOptions: this.blocks
-        .filter(block => Array.isArray(block) && !!block[1])
+        // Normalise the input config so everything is an array
+        .map(block => (Array.isArray(block) ? [block[0], block[1]] : [block, {}]))
+        // Give complex blocks the opportunity to set default values
+        .map(([block, blockConfig]) => {
+          const complexBlockInstance = this.complexBlocks.find(({ type }) => type === block.type);
```
mitchellhamilton: Couldn't you do `this.complexBlocks.find((block) => block.constructor === block.implementation)`? and then you wouldn't need to set `this.type`?

jesstelford: Yeah, that's one way to do it. I realised I wanted to move `this.type` to the super class anyway as it's important information.

I think both implementations would work.

PR #1128, opened 2019-05-14
Meetup demo rsvp validation

SeanDoyleGit:
updated rsvp component to handle rsvp validation. Also updated rsvp schema with a validation hook


demo-projects/meetup/schema.js
```diff
@@ -20,7 +20,7 @@ const cloudinaryAdapter = new CloudinaryAdapter({
 });
 
 const access = {
-  userIsAdmin: ({ authentication: { item: user } }) => user && user.isAdmin,
+  userIsAdmin: ({ authentication: { item: user } }) => user && !!user.isAdmin,
```
jesstelford: Was this intentional?

demo-projects/meetup/schema.js
```diff
@@ -109,6 +109,38 @@ exports.Rsvp = {
...
+          id
+        }
+      }`);
+
+      const { event, allRsvps } = data;
+
+      if (
+        !event ||
+        !event.startTime ||
+        new Date() > event.startTime ||
```
jesstelford: This comparison would be a timestamp vs a string, should the right hand side be wrapped in a `new Date`?

demo-projects/meetup/schema.js
timleslie: Yep, it turns out if this function returns `undefined` then the access control doesn't behave correctly. This is probably a bug. `!!` forces the value to `false` which does behave correctly.

PR #1145, opened 2019-05-14
Fix Heroku again

mitchellhamilton:
Hopefully this is the last time I say this but this should fix the Heroku deployment.


JedWatson:
🤞

PR #1146, opened 2019-05-14
Admin UI Tweaks

JedWatson:
Just a few tweaks to the Admin UI, some minor styling fixes I've been playing with and a couple of bugs squashed:

* Generally increased font sizes slightly on the list screen
* Dim out the navigation items until you move your cursor over the nav (emphasises the selected item a bit more)
* Fix the checkbox for "select all" when no items are returned (would previously be checked)
* Actually displays the selected item value when you're filtering by relationship fields

@jossmac over to you to make sure you're happy with this


jesstelford:
Do you mind adding some before/after screenshots or links?

packages/fields/src/types/Relationship/views/RelationshipSelect.js
```diff
@@ -65,17 +65,17 @@ const Relationship = forwardRef(
...
-          label: val._label_,
-          value: val,
-        }));
+        currentValue = (Array.isArray(value) ? value : []).map(getOption);
       } else if (value) {
-        currentValue = {
-          label: value._label_,
-          value: value,
-        };
+        currentValue = getOption(value);
```
jesstelford: 😍😍😍😍

packages/arch/packages/navbar/src/PrimaryNav.js
```diff
@@ -72,7 +71,7 @@ export const PrimaryNavItem = styled(ItemElement)(({ depth, isSelected }) => {
   return {
     border: 0,
     borderRight: '1px solid transparent',
-    color: colors.N90,
+    color: isSelected ? colors.N90 : mouseIsOverNav ? colors.N70 : colors.N40,
```
jesstelford: What's the contrast like on this? Does it hit AA accessibility?

JedWatson: Almost certainly not, but I haven't changed the experience at all when your cursor is over the nav (only when it's inactive and shouldn't distract from the content)

Also I made the text bigger so that's a win 🙂

JedWatson:
@jesstelford 

New colours and sizes (mouse not over nav)

image

The nav goes back to its current scheme when you put the cursor over it:

image

Previous font sizes and spacing:

image

It's pretty subtle; the tweaks just make it feel slightly less cramped imo

PR #1149, opened 2019-05-14
Improved `EventItems` component

simonswiss:
This PR brings better styling to the `EventItems` component, and also pulls it into its own component directory.

Instead of needing to wrap the event items inside an `ul` tag, we call the `EventItems` component and pass it the list of events as a prop - it will take care of displaying them in a grid of 3 / 2 / 1 columns depending on the screen width.


demo-projects/meetup/site/components/EventItems/EventItem.js
```diff
@@ -0,0 +1,78 @@
...
+      <h2>Talks</h2>
+      {talks.map(talk => (
+        <div key={talk.id}>
+        <h3>{talk.name}</h3>
+        <h3>Speakers</h3>
+        {talk.speakers.map(speaker => (
+          <p key={speaker.id}>{speaker.name}</p>
+          ))}
+          </div>
+        ))}*/}
```
jesstelford: Can remove this.

demo-projects/meetup/site/components/EventItems/EventItem.js
```diff
@@ -0,0 +1,78 @@
...
+import { colors, gridSize, shadows } from '../../theme';
+import { isInFuture, formatPastDate, formatFutureDate } from '../../helpers';
+
+const EventItem = ({ id, name, startTime, description, talks, themeColor, ...props }) => {
+  const prettyDate = isInFuture(startTime)
+    ? formatFutureDate(startTime)
+    : formatPastDate(startTime);
+
+  // Breakpoints for the events grid
+  const breakpoints = [670, 890];
```
jesstelford: Should these numbers live in `demo-projects/meetup/site/theme.js` instead?

simonswiss: Probably! It was a quick implementation and nothing yet has media queries - we are in a very rapid _throw things at the wall_ mode and clean up as we go, but this is definitely something that should live in the theme file.

PR #1157, opened 2019-05-15
Use babel preset react app in admin ui

mitchellhamilton:
We discussed doing this a while ago. This means that anything that'll work in CRA will work in the Admin UI.


packages/admin-ui/server/getWebpackConfig.js
```diff
@@ -27,7 +25,6 @@ module.exports = function({ adminMeta, entry, outputPath }) {
         // since it doesn't have to compile these things with babel
         // note that the preconstruct aliases are also disabled on heroku to enable this
         // when we have static builds for the admin ui, we can remove this
-        ...(isHerokuEnv ? [/@keystone-alpha\/field/, /@arch-ui/] : []),
```
mitchellhamilton: This isn't necessary now because we have static building and we don't care as much about memory usage during building since it won't happen when the server is running

jesstelford: Can also remove the comment above it.

PR #1160, opened 2019-05-15
Refactoring out adapter config option for `dbName`

molomby:
The `dbName` option is redundant and confusing. This info should be supplied in the connection URI. 

Default mongoose connection URI updated to include default db name based on Keystone `name`; if you want to specify a different DB name in dev, just supply the whole URI, eg: 'mongodb://localhost:27017/my-other-db'.

### DIFF
--- a/packages/adapter-mongoose/README.md
+++ b/packages/adapter-mongoose/README.md
@@ -27,15 +27,11 @@ keystone.connect(mongoDbUri, mongooseOptions);
 
 ### `mongoDbUri`
 
-_**Default:**_ `'mongodb://localhost:27017/'`
+_**Default:**_ `'mongodb://localhost:27017/${SAFE_KEYSTONE_NAME}'`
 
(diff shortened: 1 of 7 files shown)


packages/adapter-mongoose/README.md
```diff
@@ -27,16 +27,12 @@ keystone.connect(mongoDbUri, mongooseOptions);
 
 ### `mongoDbUri`
 
-_**Default:**_ `'mongodb://localhost:27017/'`
+_**Default:**_ `'mongodb://localhost:27017/${SAFE_KEYSTONE_NAME}'`
```
dcousens: Maybe `${PUT_A_UNIQUE_DBNAME_HERE}` instead?

molomby: I was trying to convey that it gets the value you give for `name` in your Keystone config and makes it safe to be used as a DB name. The behaviour before was the same, it just used 2 vars.

PR #1192, opened 2019-05-20
Fix Admin UI building on Windows

mitchellhamilton:


packages/field-views-loader/index.js
```diff
@@ -1,10 +1,9 @@
 const loaderUtils = require('loader-utils');
-const endent = require('endent');
 
 function serialize(value, allPaths) {
   if (typeof value === 'string') {
     allPaths.add(value);
-    return `loaders['${value}']`;
+    return `loaders[${JSON.stringify(value)}]`;
```
mitchellhamilton: The paths need to be JSON.stringified since windows paths have `\` which need to be escaped

packages/field-views-loader/index.js
```diff
@@ -100,11 +99,11 @@ module.exports = function() {
 
   let loaders = `{\n${[...allPaths]
     .map(path => {
-      return `'${path}': () => import('${path}').then(interopDefault)`;
+      return `${JSON.stringify(path)}: () => import(${JSON.stringify(path)}).then(interopDefault)`;
     })
     .join(',\n')}\n}`;
 
-  const source = endent`
+  const source = `
```
mitchellhamilton: endent was breaking things and removing it fixed it, I know having it is nice but debugging things on Windows is annoyyyyyyyyying and I didn't want to do any more.

Issue #802, opened 2019-03-10
Admin UI dashboard not rendering


nathsimpson:
After porting a personal project of mine from the voussoir to keystone-alpha packages, I signed in to the admin ui successfully. However, as I was redirected to the dashboard of the admin ui, my screen went white and I received this error in the console...
```
tiny-invariant.esm.js:11 Uncaught Error: Invariant failed: You should not use <Route> outside a <Router>
```

WORKAROUND:
@mitchellhamilton recommended that I add the following snippit to my package.json, once I did so and ran yarn again, everything performed as expected. I'm happy to show anyone at Thinkmill my repo.
```
  "resolutions": {
    "**/react-router-dom/react-router": "4.4.0-beta.6"
  },
```

MichaelZaporozhets:
@nathsimpson @JedWatson This issue appears to be back as of react-router/react-router-dom's bump to v5

PR #1194, opened 2019-05-20
Fix react-router-dom error when installed through npm

mitchellhamilton:
Related to #802


JedWatson:
Thanks @mitchellhamilton 👌

PR #1197, opened 2019-05-20
Remove unnecessary keystone param for AdminUIApp

jesstelford:
Follow on task from #1109

### DIFF
--- a/.changeset/e523ee4b/changes.json
+++ b/.changeset/e523ee4b/changes.json
@@ -1,0 +1,14 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/demo-project-blog", "type": "patch" },
+    { "name": "@keystone-alpha/demo-project-meetup", "type": "patch" },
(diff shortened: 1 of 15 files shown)


JedWatson:
@jesstelford on this one, I was planning to add some validation to the `pages` option that would require the `keystone` instance to be provided in the constructor (as it previously was).

e.g validate list keys are correct, throw if you specify one that doesn't exist.

We could do this validation in the `prepare` step, or even later I guess, but this sort of validation kind of seems like the sort of thing that should be done in the constructor.

What do you think? worth adding it back for this (and other possible things like it) or should we defer things like config validation to `prepare`?

jesstelford:
I'm happy with both :+1:

I think the important difference is; we should be passing it as part of the options object, not a stand alone param (if we decide to pass it in).

If we defer the validation to the prepare step, it means we'd have to also add it to the build step for when we do static builds.

PR #1198, opened 2019-05-20
Meetup demo signup

SeanDoyleGit:
Implemented sign up page mutation,

fixed bug on signout

fixed bug on event page on sign in


demo-projects/meetup/site/components/Talks.js
```diff
@@ -28,7 +27,9 @@ const Talks = ({ talks }) => {
               <H5 as="h3">{talk.name}</H5>
             </div>
             <Content>
-              {talk.description ? <Html markup={talk.description} /> : null}
+              {talk.description ? (
+                <Html markup={talk.description.replace(/style=".*?"/g, '')} />
```
jossmac: Is this something that the `Html` component should handle?

demo-projects/meetup/site/pages/signup.js
```diff
@@ -12,19 +13,25 @@ import { Button, Field, Label, Input } from '../primitives/forms';
...
   const { isAuthenticated, signin } = useAuth();
   const { meetup } = publicRuntimeConfig;
 
-  const handleSubmit = async event => {
+  const handleSubmit = createUser => {
     event.preventDefault();
+    createUser({ variables: { name, email, password } });
+  };
+
+  const handleSignin = async () => {
```
jossmac: Should probably rename these artifacts from "sign in"

demo-projects/meetup/site/components/Talks.js
JedWatson: This regexp is too greedy, it'll cause issues.

A quick and dirty fix would at least be this: `/style="[^"]*"/g`

What's the problem being guarded against here?

demo-projects/meetup/site/pages/event.js
```diff
@@ -62,7 +62,7 @@ export default class Event extends Component {
                 title={name}
               >
                 <p css={{ fontWeight: 100 }}>{locationAddress}</p>
-                <Html markup={description} />
+                {description ? <Html markup={description.replace(/style=".*?"/g, '')} /> : null}
```
JedWatson: Same as comment above, if this logic is required it should be encapsulated somewhere (like a `Description` component if not the `Html` component itself) rather than being repeated through the codebase

demo-projects/meetup/site/pages/signup.js
```diff
@@ -12,19 +13,25 @@ import { Button, Field, Label, Input } from '../primitives/forms';
...
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
   const [isLoading, setIsLoading] = useState(false);
   const [errorState, setErrorState] = useState(false);
   const { isAuthenticated, signin } = useAuth();
   const { meetup } = publicRuntimeConfig;
 
-  const handleSubmit = async event => {
+  const handleSubmit = createUser => {
     event.preventDefault();
```
JedWatson: This event is no longer being passed to the function, and refers to the global event. I'm pretty sure that is open to problems.

demo-projects/meetup/site/pages/signup.js
```diff
@@ -48,48 +55,76 @@ export default () => {
...
-            <Button disabled>Creating account...</Button>
-          ) : (
-            <Button type="submit">Sign up</Button>
-          )}
-        </form>
-      </Container>
+              <form
+                css={{ marginTop: gridSize * 3 }}
+                noValidate
+                onSubmit={() => handleSubmit(createUser)}
```
JedWatson: Really should be passing the event through from here, so we can cancel default correctly (unless I'm mistaken how this api works)

demo-projects/meetup/site/components/Talks.js
SeanDoyleGit: the problem is that many of the descriptions for talks and events have custom styling that clash with the designs.

demo-projects/meetup/site/pages/signup.js
jossmac: Pretty sure this will grab `window.event`, but better to be explicit:

```jsx
  const handleSubmit = createUser => event => {...}
  
  <form
    onSubmit={handleSubmit(createUser)}
  />
```

PR #1203, opened 2019-05-21
Consolidate CLI logging and remove subsequently unnecessary `port` arg for prepareMiddleware

jesstelford:
cli-output

💯% better console output.

Note the above is shown when using the `keystone` CLI (which most projects should anyway). If the app developer opts out and creates a custom server, then they can implement their own pretty printing logic 👍


jesstelford:
I went down a rabbit hole for both of those things!

Gladly, I pulled myself out of it and stuck with this basic implementation first.

> the Admin UI app and GraphQL app are [treated] special[ly]

I thought about a more generalised solution where `.prepareMiddleware` could return more than just the middleware, or even "register" some meta info or... something. But it ended up being hacky and felt wrong.

This version which I landed on feels better because the hackyness is contained, and since we use named exports and we're in the same monorepo, I don't mind reaching in like this (but it's still not ideal).

> The loading thing pauses and it's annoyyyyyyying(I'm assuming because a bunch of synchronous work happens?

Yep! I tried using the `threads` library, then the `tiny-worker` library, but both introduced more complexity than I was comfortable with. It's purely because the main thread is locked up while `require`ing the file, so the CLI output doesn't update.

PR #1204, opened 2019-05-21
Remove unnecessary custom servers from blog demo

jesstelford:
Follow on from #1200's `onConnect()` implementation - the blog demo no longer needs a custom server.


packages/keystone/lib/Keystone/index.js
```diff
@@ -111,7 +111,9 @@ module.exports = class Keystone {
         })
       )
     ).then(() => {
-      if (this.eventHandlers.onConnect) this.eventHandlers.onConnect(this);
+      if (this.eventHandlers.onConnect) {
+        return this.eventHandlers.onConnect.call(this, this);
```
jesstelford: Making sure to return any promises so the `.connect()` call will wait.

test-projects/basic/server.js
```diff
@@ -7,7 +7,7 @@ const initialData = require('./data');
 keystone
   .prepare({ apps, port, dev: process.env.NODE_ENV !== 'production' })
   .then(async ({ middlewares }) => {
-    await keystone.connect(process.env.MONGODB_URI);
+    await keystone.connect();
```
jesstelford: Turns out this wasn't actually being set anywhere.

dcousens: @jesstelford it is informative to have as part of demo though.

packages/keystone/lib/Keystone/index.js
mitchellhamilton: I agree with returning it for that reason but why call it and set the `this` value? I feel like that just adds confusion because then there's another way to get the keystone instance.

test-projects/basic/server.js
jesstelford: It's since been replaced by `CONNECT_TO` in #1201

packages/keystone/lib/Keystone/index.js
jesstelford: Whoops, that was me trying something and I forgot to take it out. Thanks for the catch!

PR #1205, opened 2019-05-21
Capture early requests to the keystone server while still booting

jesstelford:
👇 gif

first-responder

### DIFF
--- a/packages/keystone/bin/loading.html
+++ b/packages/keystone/bin/loading.html
@@ -1,0 +1,162 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="Content-type" content="text/html; charset=utf-8" />
+    <title>KeystoneJS</title>
+    <style type="text/css">
+      body {
+        background-color: white;
+        color: #172b4d;
+        font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial,
+          sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+        font-size: 14px;
+        letter-spacing: -0.005em;
+        margin: 0;
+        text-decoration-skip: ink;
+        text-rendering: optimizeLegibility;
+        -ms-overflow-style: -ms-autohiding-scrollbar;
+        -moz-font-feature-settings: 'liga' on;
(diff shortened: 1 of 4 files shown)


jesstelford:
The prod-mode opt out can be done by using a custom server which I think it's an ok compromise.

JedWatson:
This was so good I couldn't help merging it immediate 🎉

packages/keystone/bin/utils.js
```diff
@@ -59,44 +59,71 @@ function extractAppMeta(apps) {
...
   spinner.succeed('Initialised Keystone instance');
-  spinner.start('Preparing Keystone server');
+  spinner.start('Connecting to database');
+
+  status = 'db-connect';
 
-  const app = express();
   const dev = process.env.NODE_ENV !== 'production';
 
   const { middlewares } = await keystone.prepare({ apps, port, distDir, dev });
```
JedWatson: This could be re-written so that we connect to the database (first) but also start preparing apps in parallel.

Remembering that we now have an `onConnect` method, which can do things like create initial data from imports, it would be good to kick this off as early as possible.

By waiting for both promises to resolve asynchronously, we would then likely (in most cases) connect first (this is pretty instant) then go into "preparing to accept requests" mode while the apps' `prepare` promises are being resolved - which is _slightly_ more correct than what's displayed in the current setup, which will make "connecting to database" look like it's taking forever when what's actually happening is we're waiting on a build process. At least in dev.

This is mainly important because people might start wondering why connecting to the database is so slow, and try to fix it 🤷‍♂

JedWatson:
@jesstelford left a comment for followup if you agree

packages/keystone/bin/loading.html
```diff
@@ -0,0 +1,162 @@
...
+
+      let cancelLastCall = checkAndSetStatus();
+      const interval = setInterval(() => {
+        cancelLastCall();
+        if (ready) {
+          clearInterval(interval);
+          return;
+        }
+        cancelLastCall = checkAndSetStatus();
+      }, 500);
```
JedWatson: imo we could make this faster; like every 120ms. the load is trivial and the current setup could lead to up to 1/2s wait _longer_ than necessary before bouncing into the initialised app, making startup feel slower than it is.

PR #1209, opened 2019-05-21
Correct docs for manual getting started & demos

jesstelford:
I tried to run through the getting started guides in the README, and... couldn't.

I've updated the docs on how to do it.

NOTE: no changeset because it's only a `README.md` change.


README.md
```diff
@@ -35,33 +35,34 @@ installed](https://docs.mongodb.com/manual/installation/#mongodb-community-editi
...
+cd keystone-5/demo-projects/todo
 ```
 
-There are currently two projects available: `todo` and `blog`. You can specify the project you want to start:
+Now install and run the project:
 
 ```bash
-yarn start blog
+yarn
+yarn start
```
mitchellhamilton: This will break if we make changes to demo projects that require things in keystone packages that haven't been released yet.

jesstelford: D'oh, you're right!

I want to make sure we're encouraging did to use the `npm` versions and not the local versions.

Perhaps the `git clone` link can point to the most recent release tag?

Ideally, this while section would be replaced with something like:

```bash
yarn create keystone-app --demo blog
```

mitchellhamilton: > Perhaps the git clone link can point to the most recent release tag?

If there's an easy way to do this that doesn't require manually updating after every release, do it.

> Ideally, this while section would be replaced with something like:
>
> ```
>yarn create keystone-app --demo blog
> ```

Strongly agree. That would be great.

README.md
```diff
@@ -35,33 +35,34 @@ installed](https://docs.mongodb.com/manual/installation/#mongodb-community-editi
...
-```bash
-yarn global add bolt
-```
+- [Node.js](https://nodejs.org/) >= 10.x
+- [MongoDB](https://v5.keystonejs.com/quick-start/mongodb) >= 4.x
 
-You'll also need MongoDB installed. If you need help check out our [MongoDB Guide](https://v5.keystonejs.com/quick-start/mongodb)
-
-Then clone this repo and use Bolt to install the dependencies:
+Download a copy of the Keystone 5 repo:
```
jesstelford:
```suggestion
Download a copy of the Keystone 5 repo, and check out the latest release:
```

README.md
```diff
@@ -35,33 +35,34 @@ installed](https://docs.mongodb.com/manual/installation/#mongodb-community-editi
...
 
 ```bash
-git clone https://github.com/keystonejs/keystone-5.git
-cd keystone-5
-bolt
+git clone --depth=1 https://github.com/keystonejs/keystone-5.git
```
jesstelford:
```suggestion
git clone https://github.com/keystonejs/keystone-5.git
cd keystone-5
git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
```

README.md
```diff
@@ -35,33 +35,34 @@ installed](https://docs.mongodb.com/manual/installation/#mongodb-community-editi
...
+- `todo`: A Todo app showcasing the AdminUI and how to create a minimal List
+- `blog`: A starting point for a blog including a WYSIWYG editor
+- `meetup`: A local community event website with speakers and sponsors
+
+Then move into the directory for that demo, for example:
 
 ```bash
-yarn build
-yarn start
+cd keystone-5/demo-projects/todo
```
jesstelford:
```suggestion
cd demo-projects/todo
```

README.md
jesstelford: > If there's an easy way to do this that doesn't require manually updating after every release, do it.

Figured it out (43c96e2eb55ff97f7eb7a021e912f770d82e6175), check the instructions now - it will always get you to the latest release 🎉

README.md
```diff
@@ -35,33 +35,36 @@ installed](https://docs.mongodb.com/manual/installation/#mongodb-community-editi
...
-You'll also need MongoDB installed. If you need help check out our [MongoDB Guide](https://v5.keystonejs.com/quick-start/mongodb)
-
-Then clone this repo and use Bolt to install the dependencies:
+Download a copy of the Keystone 5 repo, and check out the latest release:
 
 ```bash
 git clone https://github.com/keystonejs/keystone-5.git
 cd keystone-5
-bolt
+git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
```
jesstelford: This magic is from https://github.com/tj/git-extras/pull/372/files

PR #1210, opened 2019-05-21
attempt to fix  env variables with windows

SeanDoyleGit:
Attempt to fix setting environment variables on windows


packages/create-keystone-app/templates/todo/package.json.ejs
```diff
@@ -2,7 +2,7 @@
   "name": "<%= appName%>",
   "version": "0.0.0",
   "scripts": {
-    "start": "NODE_ENV=development DISABLE_LOGGING=true keystone"
+    "start": "<%= envVariables%> keystone"
```
jesstelford: Perhaps we can use [`cross-env`](https://www.npmjs.com/package/cross-env), which has the upshot of making the code more portable.

```suggestion
    "start": "cross-env NODE_ENV=development DISABLE_LOGGING=true keystone"
```

SeanDoyleGit: will change now :)

packages/create-keystone-app/lib/generator.js
```diff
@@ -61,6 +61,7 @@ function generate(name, noDeps) {
         copyTemplate(`${templateDir}/todo`, projectDir, {
           name,
           appName,
+          envVariables,
```
mitchellhamilton: Is this meant to be here?

packages/create-keystone-app/templates/todo/package.json.ejs
```diff
@@ -2,7 +2,7 @@
   "name": "<%= appName%>",
   "version": "0.0.0",
   "scripts": {
-    "start": "NODE_ENV=development DISABLE_LOGGING=true keystone"
+    "start": "cross-env NODE_ENV=development DISABLE_LOGGING=true keystone"
   },
   "private": true,
   "dependencies": {
```
mitchellhamilton: `cross-env` needs to be in the dependencies here, not in create-keystone-app's dependencies.

packages/create-keystone-app/lib/generator.js
SeanDoyleGit: no it is not

jesstelford:
@SeanDoyleGit Once you've added the `cross-env` dep to the `packages/create-keystone-app/templates/todo/package.json.ejs` project, it'd be great to see it also used in each of the projects in `demo-projects` & `test-projects` 🎉

packages/create-keystone-app/package.json
```diff
@@ -17,6 +17,7 @@
     "@samverschueren/stream-to-observable": "^0.3.0",
     "arg": "^4.1.0",
     "chalk": "^2.4.2",
+    "cross-env": "^5.2.0",
```
jesstelford:
```suggestion
```

This isn't needed here (it's meant to be in `packages/create-keystone-app/templates/todo/package.json.ejs`)

PR #1212, opened 2019-05-21
restore process.env.MONGODB_URI for test-project

dcousens:
@jesstelford as discussed,  #1201 did __not__ introduce `CONNECT_TO` as a Keystone-wide environment variable,  but only for the `keystone` command.  Thereby this had nothing set.

This resolves that,  possibly.

### DIFF
--- a/.changeset/cb278d0e/changes.json
+++ b/.changeset/cb278d0e/changes.json
@@ -1,0 +1,4 @@
+{
+  "releases": [{ "name": "@keystone-alpha/cypress-project-basic", "type": "patch" }],
+  "dependents": []
+}
--- a/.changeset/cb278d0e/changes.md
+++ b/.changeset/cb278d0e/changes.md
@@ -1,0 +1,1 @@
+Restore process.env.MONGODB_URI to test project example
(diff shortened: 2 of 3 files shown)


test-projects/basic/.env.example
```diff
@@ -2,3 +2,4 @@
 # CLOUDINARY_KEY=abc123
 # CLOUDINARY_SECRET=abc123
 PORT=1
+MONGODB_URI=mongodb://127.0.0.1:27017/tp_basic
```
jesstelford: No need to add the value here, we don't want to force its usage, instead we can rely on the fallback.

dcousens: I think providing this here provides a good example for users to follow?

jesstelford: `dotenv-safe` will read `.env.example` for which env vars are _required_. With this change, if the environment doesn't provide a `MONGODB_URI`, `dotenv-safe` will throw an error, which is not what we want (we want it to default to the localhost fallback when not set), especially as this is a test project and we will be using an in-memory mongo server which is passed in manually (ie; not via an env var)

dcousens: Rebased and removed.

PR #1223, opened 2019-05-23
Compile Controller base class to ES5 so that non-native classes can extend it 

mitchellhamilton:


babel.config.js
```diff
@@ -14,12 +14,18 @@ module.exports = {
...
     '@babel/preset-flow',
   ],
   plugins: [
     '@babel/plugin-proposal-class-properties',
     '@babel/proposal-object-rest-spread',
     '@babel/plugin-syntax-dynamic-import',
   ],
+  overrides: [
+    {
+      include: 'packages/fields/src/Controller.js',
```
jesstelford: Our browserlist target means every browser supports [ES6 classes](https://caniuse.com/#feat=es6-class), can we skip compiling _every_ class?

mitchellhamilton: I'm slightly confused, do you mean you want to skip compiling for every class? That's what currently happens before this PR but that creates a problem that if a consumer uses a compiled class to extend this class for whatever reason, this could happen if a consumer compiles a class that extends this class with Babel themselves and it also currently happens because the Admin UI uses babel-preset-react-app and we could remove babel-preset-react-app from the Admin UI but there's still the problem that a consumer could compile a class that extends this class so this fix would still be necessary.

jesstelford: Ah, ok, makes sense. Let's stick with this change for now, and I'll open an issue to move to pure classes (if possible) in the future 👍

jesstelford:
[Odd errors after this change](https://circleci.com/gh/keystonejs/keystone-5/19512?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link):

```
ReferenceError: regeneratorRuntime is not defined
```

PR #1232, opened 2019-05-24
hasChanged is a synchronous method, and was inversed

jesstelford:


packages/fields/src/Controller.js
```diff
@@ -40,8 +40,8 @@ export default class FieldController {
    * passed to this.serialize() for you
    * @return boolean
    */
-  hasChanged = async (initialData, currentData) =>
```
jesstelford: This function should _not_ be `async`, it is consume _synchronously_ here:

https://github.com/keystonejs/keystone-5/blob/fix-has-changed/packages/app-admin-ui/client/pages/Item/index.js#L134-L137

packages/fields/src/Controller.js
```diff
@@ -40,8 +40,8 @@ export default class FieldController {
...
-    isEqual(initialData[this.path], currentData[this.path]);
+  hasChanged = (initialData, currentData) =>
+    !isEqual(initialData[this.path], currentData[this.path]);
```
jesstelford: Whoops, the logic was inverted! So I've fixed that... with a bang! 💥

PR #1238, opened 2019-05-24
Reveal homepage video on hover/touch rather than click on button

simonswiss:
Large GIF (1198x800)

### DIFF
--- a/.changeset/083e01ed/changes.json
+++ b/.changeset/083e01ed/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/website", "type": "patch" }], "dependents": [] }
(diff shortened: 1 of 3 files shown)


jesstelford:
On a simulated 3G connection, if I hover as the page is still loading, I first see a grey box - I'm guessing this is Egghead loading, and we can't do anything about it?

4g-video

It doesn't happen if I want about 20s before trying to hover.

simonswiss:
We could take care of it by adding an `onLoad` listener to the `iframe` and delay the reveal of the video until the video is actually ready.

Will merge this for now - but something to keep in mind to improve it! ✨

jesstelford:
@all-contributors please add @simonswiss for example & tutorial & video

PR #1241, opened 2019-05-25
[website] Fix mobile menu on homepage and 404 

jordanoverbye:
On mobile, if you click the hamburger menu button on a page which does not have the sidebar menu, 
the menu does not toggle as expected. This can be seen on the [homepage](http://v5.keystonejs.com) and the [404 page](http://v5.keystonejs.com/1234). This is happening because the sidebar component isn't actually part of any non-documentation pages.

What I've done for this PR:

1. Take out the Sidebar component from `templates/docs.js` and move into its own component named `Sidebar`.
2. `Sidebar` now renders the `Footer`, `SidebarNav` and `Search` (rather than passing them as children) as it's now used in more than 1 file.
3. Added a new prop `mobileOnly` to `Sidebar` which will set display block or none 
4. Rename the current `Sidebar` component to `SidebarNav`.
5. Add the new `Sidebar` component to all pages

### DIFF
--- a/website/src/components/Sidebar.js
+++ b/website/src/components/Sidebar.js
@@ -1,123 +1,93 @@
 /** @jsx jsx */
 
-import React from 'react'; // eslint-disable-line no-unused-vars
-import { StaticQuery, graphql, Link } from 'gatsby';
-import { colors, gridSize } from '@arch-ui/theme';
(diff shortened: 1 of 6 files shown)


JedWatson:
Looks great @jordanoverbye!

PR #1247, opened 2019-05-27
Make previous and next buttons consistent with order in sidebar

mitchellhamilton:
This has been annoying me for ages so I finally fixed it.

### DIFF
--- a/website/src/utils/hooks.js
+++ b/website/src/utils/hooks.js
@@ -32,7 +32,6 @@ export function useNavData() {
       }
     }
   `);
-  console.log(data.allSitePage.edges);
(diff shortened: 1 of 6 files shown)


website/src/utils/hooks.js
```diff
@@ -10,3 +11,42 @@ export function useDimensions() {
...
+              navGroup
+              isPackageIndex
+              pageTitle
+            }
+          }
+        }
+      }
+    }
+  `);
+  console.log(data.allSitePage.edges);
```
jossmac: Intentional log?

```suggestion
```

PR #1260, opened 2019-05-30
Slight tinker with loading view

jossmac:
- shade the background
- slightly larger font
- smoother spinner aesthetic via opacity


jesstelford:
Screen Shot 2019-05-30 at 3 11 46 pm

PR #1263, opened 2019-05-30
windows compatible npm scripts

### LINKED ISSUE #1262: windows compatibility: project does not install
* windows 10.
* Node v10.13.0
* Yarn 1.16.0

I get an error initialising working copy.

```
> bolt
⚡️   bolt v0.22.6 (node v10.13.0)
🔎   [1/4] Validating project...
📦   [2/4] Installing project dependencies...
...
yarn install v1.9.4
[1/4] Resolving packages...
success Already up-to-date.
$ yarn build:build-field-types && preconstruct dev && node ./packages/build-field-types/bin.js dev  && yarn run --silent contributing-guide
yarn run v1.12.3
$ rimraf packages/build-field-types/lib && babel packages/build-field-types/src --out-dir packages/build-field-types/dist --ignore '**/__tests__/**/*','**/__mocks__/**/*'
Successfully compiled 42 files with Babel.
Done in 3.73s.
🎁 info project is valid!
🎁 success created links!
🎁 info project is valid!
🎁 success created links!
Input required
error Command failed with exit code 1.
```

Seems that the npm scripts in the root `package.json` is not windows friendly.

Please let me know if any further info is required for this bug.

I will raise a corresponding PR shortly.

bholloway:
Fixes #1262 

## Code Changes

* Remove `\n` which is somehow resulted in the `Input Required` error (root cause unknown)
* Use `\"` everywhere instead of `'`
* Use `cross-env` harder
* Convert `/bin/sh -c` to `yarn --cwd` (see below)
* Convert `rm -rf` to `rimraf`
* Fix path manipulation in `wysiwyg-tinymce`

## Conceptual Change

I'm proposing a change to how scripts are invoked in sub-directories.

IMO the `yarn --cwd` is far more elegant than  `cd`.

A example is the `website`.

| before | after |
|--------|--------|
|`"website:start": "cd website && yarn start"`| "website": "yarn --cwd website"|
|`> yarn website:start` | `> yarn website start` |

In comparison a more complex example is `demo-projects`.

We have more than one directory and it is a deep path.  Any scripting is problematic cross platform, but without scripting we cannot easily use substitution to construct a deep directory path.

We could...

* Enumerate all the nested packages. 👎 

* Somehow delegate to the `./demo-projects` directory without any knowledge of deep directory structure. 👍 

We can do the latter but the cost is an additional `package.json` placed there.

| before | after |
|--------|--------|
|  `"start:demo": "/bin/sh -c 'cd demo-projects/${1:-$0} && yarn start' todo"` | `"demo-projects": "yarn --cwd ./demo-projects"` |
| `> yarn start:demo` | `> yarn demo-projects start` |

In `demo-projects/package.json` we have.

```json
"start": "yarn todo start"
"todo": "yarn --cwd todo"
```

At each step we simply defer to the scripts in the local `package.json`. IMO this is the more elegant and scalable approach so long as we can live with the additional `package.json`.

## testing

So I have run the basic scripts and successfully launched the `todo` demo on Windows.

We would want to exhaustively test all the scripts on Mac and Windows platforms.


package.json
```diff
@@ -6,32 +6,35 @@
   "author": "The KeystoneJS Development Team",
   "license": "MIT",
   "scripts": {
-    "coverage": "DISABLE_LOGGING=true NODE_ENV=test jest --maxWorkers=1 --coverage",
+    "demo-projects": "yarn --cwd ./demo-projects",
+    "test-projects": "yarn --cwd ./test-projects",
```
jesstelford: If I run `yarn test-projects`, what happens? Does it do the equivalent of `cd test-projects && yarn install`?

package.json
```diff
@@ -6,32 +6,35 @@
   "author": "The KeystoneJS Development Team",
   "license": "MIT",
   "scripts": {
-    "coverage": "DISABLE_LOGGING=true NODE_ENV=test jest --maxWorkers=1 --coverage",
+    "demo-projects": "yarn --cwd ./demo-projects",
```
jesstelford: I think I'd prefer `demo` here, so it's `yarn demo blog start`, which is consistent with the `website`, but wont work for the `test-projects` folder (because there's already a `test` script).

package.json
```diff
@@ -6,32 +6,35 @@
...
     "lint": "yarn lint:prettier && yarn lint:eslint && yarn lint:flow && yarn lint:markdown",
     "start": "yarn start:demo",
-    "start:demo": "/bin/sh -c 'cd demo-projects/${1:-$0} && yarn start' todo",
-    "start:test": "/bin/sh -c 'cd test-projects/${1:-$0} && yarn start' basic",
-    "website:start": "cd website && yarn start",
-    "website:build": "cd website && yarn build",
+    "start:demo": "yarn demo-projects start",
+    "start:test": "yarn test-projects start",
+    "website:build": "yarn website build",
+    "website:start": "yarn website start",
```
jesstelford: We can ditch all these as long as `yarn website start`/`yarn demo blog start` works from the root directory.

package.json
bholloway: It is the equivalent of `cd test-projects && yarn` so I believe that is the same result

packages/fields-wysiwyg-tinymce/src/index.js
```diff
@@ -1,10 +1,10 @@
+import { dirname } from 'path';
 import express from 'express';
 import { Text } from '@keystone-alpha/fields';
 import { importView } from '@keystone-alpha/build-field-types';
 
 function prepareMiddleware() {
-  const tinymce = require.resolve('tinymce');
-  const tinymcePath = tinymce.substr(0, tinymce.lastIndexOf('/'));
+  const tinymcePath = dirname(require.resolve('tinymce'));
```
jesstelford: 👍

package.json
bholloway: Yes I agree. But what to do about `test` I guess. Personally I prefer consistency with `test` over terseness.

package.json
jesstelford: And does that actually do anything (given that directory's `package.json` doesn't have any `dependencies`)? (sorry, I'd test myself, but I'm neck deep in another branch right now)

package.json
bholloway: I think will need to update (serveral) `readme.md` too

package.json
bholloway: No. It is true that it is a little annoying for new users.

We can fix _allthethings_ but doing
```json
"website:": "yarn --cwd website",
"demo:": "yarn --cwd demo-projects",
"demo:": "yarn --cwd test-projects",
```

Then we have `yarn test: start`  and `yarn test: basic: start` so it is more obvious that `yarn test:` is a oops.

However I can see that in docs the space may be considered a type-o. Its a case of chose-the-lesser-confusion. 🤔

package.json
jesstelford: Yeah, a quick grep for `(website|start):.*` should find all the uses

package.json
jesstelford: Thinking about it more; I don't ever expect a user to _run_ the test projects. They're designed to be run by CI only really.

So perhaps the naming inconsistency is ok:

```
$ yarn website start
$ yarn demo blog start
$ yarn test-project basic start # only run by CI
```

jesstelford: Or... we don't even expose the test projects to be run like that, which makes even more sense to me:

```
$ yarn website start
$ yarn demo blog start
$ cd test-projects/basic && yarn start # a user is discouraged from doing this
```

bholloway:
There is also a problem with the paths written by `dev.js` in both `preconstruct` and in `packages/build-field-types`. The `\` is not escaped in the windows paths when writing `dist/*.cjs.js`.

I'll make the suggest changes to the scripts and open an issue with `preconstruct` before adding any other commits.

jesstelford:
Work has been moved to #1299

PR #1268, opened 2019-06-01
WIP: Fields markdown

bholloway:
This is mostly copied from `fields-wysiwyg-tinymce` (as requested).

After trying a couple of other solutions I got this one to work with least effort - The engine is `draft-js`, but the version that allows plugins [`draft-js-plugins`](https://github.com/draft-js-plugins/draft-js-plugins). It should be easy to add buttons and other features for which there are plugins.

Essentially this is a "preview" type view that converts markdown to rich text as you type it. The rich text is then converted back to markdown on change.

We could easily make this toggle between "raw" and "preview". I think having side-by-side views is not such a good idea - scrolling is non-trivial and in some cases we have limited width.

image


bholloway:
I locally have this based on my `fix-windows-compatibility` branch but I've rebased to `master` for this PR.

This would need to be tested against `master` before merging.

timleslie:
@bholloway Is this PR still relevant? If not, can we close it out?

bholloway:
I don't know what happened with this feature. I am not attached to the code so feel free.

timleslie:
Closing as we now have a `fields-markdown` package.

PR #1276, opened 2019-06-05
Verify Keystone Session id on complete

ticidesign:
If we don't have a keystoneSessionId on complete it will fail and send back to the loginMiddleware


MadeByMike:
Confirmed solved

PR #1284, opened 2019-06-07
Adding keystone.addLists method

JedWatson:
In another (not in this repo) project I found it getting really repetitive repeating list keys in the schema exports _and_ the index file (twice - for import, and for list creation).

This PR addresses that. Usage optional, so I only used it in the demo project (the meetup schema is quite complex and I find it a bit more readable in that project to see which lists are actually being created explicitly)


JedWatson:
Good points @timleslie - I've been thinking about this.

> Note: I'm pretty 🤷‍♂ about this PR getting in, I'd still like it to, but the comment below is mainly because I think it's an interesting conversation. TL;DR this falls under my category of "tone setting" where you see "pure helper, unnecessary" and I think there's value digging into that!

For me, especially for minor API like this (the implementation is trivial) the issue is not so much surface area, as _what are we telling people to do with our API_.

Adding a single list seems pretty useful (unless we decide to make list creation an atomic action, where you pass lists into the Keystone constructor, which I think has merit for debate but consider outside the scope of this conversation)

However, I don't think encouraging people to repeat themselves _when it doesn't add value to the implementation_ is a good thing to do too much of. Explicitness is 💯 when there's a point to what you're being explicit about; but here (e.g in the Meetup demo) I just see unnecessary boilerplate & duplication.

For sure we could add that loop / function into the demo. But then those few lines of code will be repeated across hundreds and thousands of projects, adding no unique value each time. This seems like the sort of thing Keystone is good at _removing_ from projects.

In most cases, I expect developers to define several lists across one or multiple files; import them; and register them. This API supports that with as little repetition and with minimal unnecessary expression in the project codebase - so from a "what is our API encouraging people to do" I think that's worth it.

> because then where does it end?

Really good question, especially in the context of Keystone's history. I think it'll always be a trade-off, if we can get good at codifying how and why we make those trade-offs it would be awesome!

Keystone sits somewhere in the middle of the slider between "minimal framework API" and "minimal project code". Might be a good topic for a workshop to dig into this more and see what principles we can distill. 

Real-world case study: our work on the CLI has come pretty directly out of trying to manage that balance - my original API was really quite verbose in the examples, which led to @jesstelford removing a lot of the non-differentiating code (read: boilerplate) through the CLI which in turn shrank the code required to set up a keystone project dramatically. The cost was more than one way to do things, and some confusion because the API was no longer as explicit. We've iterated through that to a really good place now, where there's minimal code required in each project but everyone feels it's a lot clearer what's going on (less "magic").

But even with the CLI we haven't really addressed the branching ways we've introduced to set up a keystone project, and probably should. Every quality of life feature we add introduces more complexity and we should probably get clearer on how we think about that and plan to manage it long term.

MadeByMike:
@JedWatson I also think that bing explicit is helpful for smaller projects. Writing `createList()` 4-5 times is better than hiding the names of the lists you're creating in different files.

For larger projects I imagine writing a for loop is a pretty quick and obvious decision, that is not much of a convenience method when you consider the scale of the project. 

If you're ok I'd like to add this suggestion to the documentation and close this?

PR #1285, opened 2019-06-07
Fix demo scripts

JedWatson:
Fixing a paper cut: our demo projects aren't consistent with their scripts, and the main demo start script in the root package doesn't actually "work out of the box" because it runs `start` instead of `dev` and you need to go into the blog demo project directory first to build it.

Also the meetup project was only doing a next build, etc.

I made the slightly opinionated decision that `yarn start :project?` in the root script would run the project's `dev` script, because that's _probably_ the intention if you're running keystone in the monorepo. The individual packages themselves have `build` and `start` scripts so that from the project directory they behave as keystone projects should.

@jesstelford you might want to eyeball this and make sure I've gotten it right.

### DIFF
--- a/demo-projects/blog/package.json
+++ b/demo-projects/blog/package.json
@@ -9,8 +9,8 @@
     "node": ">=8.4.0"
   },
   "scripts": {
-    "dev": "cross-env NODE_ENV=development DISABLE_LOGGING=true keystone dev",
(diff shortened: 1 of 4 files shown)


JedWatson:
Note: this intentionally includes no changeset

JedWatson:
Closing in favour of #1299

PR #1286, opened 2019-06-07
[WIP] Adding support for `isVisible` and `isEditable` options on fields

JedWatson:
WIP

Pushing this early for API Feedback

Not sure how I feel about the "default true" options, we've avoided them so far (these were `hidden` and `noedit` in keystone 4)

To allow dynamic evaluation client-side, we'll need to be able to specify a custom client-side controller for fields (which I'll open an issue for); so that needs a bit more thought, just going to implement them as boolean options in the first pass.


jesstelford:
Agreed we should have an option for toggling visibility & editability of fields for the Admin UI.

This implementation starts to add a lot of functionality to schema definitions which is only applicable to the _optional_ Admin UI.

Can I suggest we move this to the configuration _of the Admin UI_?

For example (I'm not wedded to this API at all, it's to demo my thoughts):

```javascript
keystone.createList('Task', {
  fields: {
    title: { type: Text },
    // I want this field hidden in the Admin UI
    sortOrder: { type: Integer },
  }
});

const admin = new AdminUIApp({
  lists: {
    Task: {
      fields: {
        sortOrder: { visible: false } // default: true
      }
    }
  }
});

module.exports = {
  keystone,
  apps: [new GraphQLApp(), admin],
};
```

And we can move more of the existing AdminUI only stuff there such as `defaultColumns`, etc:

```javascript
const admin = new AdminUI({
  lists: {
    Task: {
      fields: {
        sortOrder: { visible: false } // default: true
      },
      // Only display the '_label', not the 'name' by default
      defaultColumns: ['_label'],
    }
  }
});
```

jesstelford:
My suggestion introduces a separation of where things are configured. Personally I like this as it means I am configuring discrete separate things.

However, if it makes more sense to you to configure both schema and admin ui at the same time, my suggestion would also allow doing something like:

```javascript
const _ = require('lodash');

const lists = {
  Task: {
    fields: {
      title: { type: Text },
      // I want this field hidden in the Admin UI
      sortOrder: {
        type: Integer,
        adminUI: { isVisible: false },
      },
    }
  }
}

Object.entries(lists).forEach(([key, list]) => {
  // The adminUI key on fields is silently ignored
  keystone.createList('Task', list);
});

const admin = new AdminUIApp({
  lists: _.mapValues(lists, ({ fields }) => ({
    // Pick out only the adminUI key
    fields: _.mapValues(fields, field => _.pick(field, 'adminUI'))
  })),
});

module.exports = {
  keystone,
  apps: [new GraphQLApp(), admin],
};
```

Whereas the inverse cannot be true; this PR does not allow separating the configuration.

molomby:
> Not sure how I feel about the "default true" options [..]

Why don't we invert them then?

* !`isVisible` => `isHidden`
* !`isEditable` => `isDisabled` I guess? This is actually what it ends up as anyway in [many](https://github.com/keystonejs/keystone-5/pull/1286/commits/412f58ac8ffd8da892a0d0919aaa4244a35cb990#diff-aaec383ac3900831448782db3600da5fR27) [cases](https://github.com/keystonejs/keystone-5/pull/1286/commits/c5d789ad8e34a333a58b81120f0476275878e0e7)

MadeByMike:
@JedWatson Where is this branch at? I kinda like the ideas in here but want to close stale PRs :)

PR #1288, opened 2019-06-08
Added required @keystone-alpha/app-static to manual setup instructions

Vultraz:


jesstelford:
Thanks for the PR!

The example for manual setup (https://github.com/Vultraz/keystone-5/blob/bf740b8a4c10a94e99301269b19605a8eb72e9b9/README.md#manual-setup) doesn't use the `app-static` import:

```javascript
const { Keystone }        = require('@keystone-alpha/keystone');
const { Text }            = require('@keystone-alpha/fields');
const { GraphQLApp }      = require('@keystone-alpha/app-graphql');
const { AdminUIApp }      = require('@keystone-alpha/app-admin-ui');
```

Were you using another example which requires it?

Vultraz:
Oh, derp 🤦‍♂ I had tried the `yarn` quickstart, but it failed setting up packages (see my comment on #1263), so I did a manual install in the same directory using the package list in the readme. On running the project (just `keystone`), startup failed with a message saying `app-static` was missing. Which makes sense, since I'm using the `index.js` file from the yarn setup.

jesstelford:
Ah, that makes sense! At least you were able to get a version running 👍

Issue #719, opened 2019-02-14
Improve support for isRequired on Relationships


timleslie:
Due to the use of back-references, the `isRequired` flag is tricky to enforce for Relationship fields. At the moment I don't believe it's adequately implemented. This tasks needs someone to go through the Relationship field and ensure that the `isRequired` semantics are correctly implemented across all use cases.

jesstelford:
Just got stung by this:

```javascript
keystone.createList('User', {
  fields: {
    name: { type: Text },
  },
});

keystone.createList('Post', {
  fields: {
    author: {
      type: Relationship,
      ref: 'User',
      isRequired: true,
    }
  }
});
```

The `isRequired` check appears to do nothing:

```graphql
mutation createPost {
  createPost(data:{}) {
    id
    author {
      id
    }
  }
}
```

Successfully creates a new item.

---

I think the most basic implementation should at least:

1. Set the field as required in the graphQL schema. (currently it's just `input PostCreateInput { author: UserRelateToOneInput }`)
2. Check for at least one of the possible operations (`create` / `connect` / `disconnect` / `disconnectAll`) having any value.

MadeByMike:
We should update the docs until this is resolved.

christopherliu:
Hello! It looks like isRequired is not supported on relationships in [Keystone v6](https://keystonejs.com/docs/apis/fields#relationship) as well as v5. Is this the proper channel to hear updates on this, given that the issue is closed, or should I open a new feature request issue?

dcousens:
Hi @christopherliu,  there is an open feature request / question around this https://github.com/keystonejs/keystone/discussions/7310,  we'll endeavour to try and answer with our reasoning for why `isRequired` is missing soon

PR #1290, opened 2019-06-09
Consistent support for isRequired across all field types.

### LINKED ISSUE #1245: isRequired has no effect
``` js
  fields: {
   description: {
     type: Text,
     isRequired: true
   }
  }
```

Expectation is that it fails if `null` (or `''`, aka empty string) in the administration UI on creation.

Actuality: It creates an item without an error.

**edit:** Unclear if this is meant be enforced by GraphQL,  but,  I can verify that you can `create*` `isRequired` fields as `null` without error too.  `isUnique` **is** enforced by GraphQL.

timleslie:
Fixes #1245

Add a check during for the `_validateInput` phase for non-relationship fields which have `isRequired` set but have not supplied/supplied a null value.

### DIFF
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -1098,10 +1098,10 @@ module.exports = class List {
     // Check for isRequired
     const fieldValidationErrors = this.fields
       .filter(
-        f =>
-          f.isRequired &&
-          !f.isRelationship &&
-          (resolvedData[f.path] === undefined || resolvedData[f.path] === null)
+        field =>
+          field.isRequired &&
+          !field.isRelationship &&
+          (resolvedData[field.path] === undefined || resolvedData[field.path] === null)
       )
       .map(f => ({
         msg: `Required field "${f.path}" is null or undefined.`,
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -1094,6 +1094,24 @@ module.exports = class List {
       originalInput,
       actions: mapKeys(this.hooksActions, hook => hook(context)),
     };
+
+    // Check for isRequired
+    const fieldValidationErrors = this.fields
+      .filter(
+        f =>
+          f.isRequired &&
+          !f.isRelationship &&
+          (resolvedData[f.path] === undefined || resolvedData[f.path] === null)
+      )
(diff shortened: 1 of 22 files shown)


packages/keystone/lib/List/index.js
```diff
@@ -1094,6 +1094,24 @@ module.exports = class List {
       originalInput,
       actions: mapKeys(this.hooksActions, hook => hook(context)),
     };
+
+    // Check for isRequired
+    const fieldValidationErrors = this.fields
```
jesstelford: This happens _after_ relationships are resolved: https://github.com/keystonejs/keystone-5/blob/16386a2d96e0378e2e2a433c8de3bbce21745515/packages/keystone/lib/List/index.js#L1280

This could result in a related item being created or deleted before the input is validated (and then fails validation)

packages/keystone/lib/List/index.js
```diff
@@ -1094,6 +1094,24 @@ module.exports = class List {
...
+      .filter(
+        f =>
+          f.isRequired &&
+          !f.isRelationship &&
+          (resolvedData[f.path] === undefined || resolvedData[f.path] === null)
```
jesstelford: Can we rename `f` -> `field`?

packages/keystone/lib/List/index.js
```diff
@@ -1094,6 +1094,24 @@ module.exports = class List {
       originalInput,
       actions: mapKeys(this.hooksActions, hook => hook(context)),
     };
+
+    // Check for isRequired
+    const fieldValidationErrors = this.fields
+      .filter(
+        f =>
+          f.isRequired &&
+          !f.isRelationship &&
```
jesstelford: When is an `isRequired` relationship validated?

timleslie: Never :-) Need to create a separate issue for that, because as you noted above, we attempt to resolve relationships before the validation step.

jesstelford: Got it 👍

timleslie:
Note: This PR does not address `isRequired` on `Relationship` fields. As noted in the review comments above, the relationship value is used before the validation step in the lifecycle, and so needs to be handled separately. This is captured in issue #719.

Issue #1133, opened 2019-05-14
Implement a "read-only" option for lists


molomby:
Leading on from [thoughts](https://github.com/keystonejs/keystone-5/issues/1117#issuecomment-491725665) in #1117.

It would be very useful (and probably not especially difficult?) to support a flag on lists that made them read-only (`isReadOnly` I guess?). When set to `true` this option would remove all mutations related to the list from the GraphQL schema as well as edit/create/delete functionality from the Admin UI. All read operations/interface would operate in the same manner.

I can think of two scenarios in which this would be useful:

* Supporting database views -- Both core database adapters can be used with DB platforms that support views. DB view let app devs easily incorporate business and reporting logic into their DB then surface it through their Keystone. This also gives app developers the opportunity to leverage tools like materialised views to improve performance of common, time-intensive operations. This was [previously discussed in #1777](https://github.com/keystonejs/keystone-5/issues/1117#issuecomment-491725665).
* Supporting tables/collections managed outside the system -- Databases often become de facto APIs to other systems. Imagine for example, a piece of legacy software that can sync business data to a relational DB. We can add the data to our DB but the Keystone instance itself _does not manage it_. By configuring a read-only list, users can view and link to these sync'ed records through the admin UI and surface them through GraphQL without allowing operations that will cause errors.

An example of the "view" scenario -- here we have a "high scores" list with a field that calculates each score's current ranking. This kind of operation is impractical to perform in JS, even with the planned "virtual" fields (#1117).

Lists:

```js
keystone.createList('HighScore', {
  fields: {
    player: { type: Text },
    points: { type: Number, isIndexed: true },
    ranking: { type: Relationship, ref: 'HighScoreRanking', many: false }
  }
});
keystone.createList('HighScoreRanking', {
  options: { isReadOnly: true },
  fields: {
    ordinal: { type: Integer },
    score: { type: Relationship, ref: 'HighScore', many: false }
  }
});
```

Underlying view (pgSQL):

```sql
create view "highScoreRankings" as (
	select "id", row_number() over (order by "points" desc) as "ordinal"
	from "highScores"
);
```

This can then be used in GraphQL queries:

```graphql
query {
  # Get a player's scores and the rank of those scores:
  myScores: allHighScores (where: { player: 'molomby' }) {
    player
    points
    ranking: { ordinal }
  }

  # Get the current top 10 scores:
[...]

Issue #776, opened 2019-03-06
Access Control Policies


gautamsi:
I found that when using Auth Strategy, it is still possible to create an admin user using graphql anonymously. To prevent this I had to set default access on Keystone instance or individual list.

I propose to have Access Control Policies
1. Default Access Control policy should kick in when you use any authentication strategy
2. Default policy should be Opt out, by default all list will be tagged with this policy and you have to exclude lists. User list  should be locked in or You have to specifically say you want to remove user list from this policy.
3. It should prevent write access in above case by default
4. It should prevent update to relational fields for user (like post.author -> User) in case a list is excluded from policy

I am also thinking the authentication scenario should be a short hand for imperative/static access

```
type StaticAccess = boolean | 'auth' | 'own';
```
question is how do we find owner user, it would be easy for user list but for other, it may not be much difficult as we can have createdBy meta field.  another way to make list configuration for own item assignment. for example
```
keystone.createList('Post', {
  fields: {
    author: {
      type: Relationship,
      ref: 'User', // we can infer here but it is highly likely that ref field may not be where you want to add restriction.
      access: ({ authentication: { item, listKey }, existingItem }) => {
        return true;
      },
    },
  },
ownerResolver: item => item.author, // here we can use the resolver for owner
});
```

We can also have delegated access, like `createdBy`  field can be meta, and user can change this field in future, but there can be assigned field. like I post the blog for another user, I become createBy but I can add that user as author, then that user should also be able to change the  blog.  then I should change the config like this

`ownerResolver: item => [item.createdBy, item.author],`

PR #1291, opened 2019-06-09
implement tracking createdAt, updatedAt, createdBy and updatedBy

### LINKED ISSUE #184: Add special fields `updatedAt` and `createdAt` of type DateTime

gautamsi:
closes #184 

Now using plugin feature from #1491.

See [README for List Plugin](https://github.com/keystonejs/keystone-5/blob/b6c3d1ee8c479f68dbf3a9fde6b9fde5d879b61a/packages/list-plugins/README.md) this PR creates.

=====================
Initial API (deprecated) (Now using List Plugins as in #1491)

have used name in the lib like atTracking and byTracking. probably needs more suited name. 

to use you have to wrap the schema of field with 

- `atTracking` - for createdAt and modifiedAt
and/or 
- `byTracking` - for createdBy and modifiedBy

options for both atTracking and byTracking can be provided as 

```js
const atSchema = atTracking(<fieldSchema>, { chainBeforeChangeHook: true /*default true*/})

const bySchema = byTracking(<fieldSchema>, { chainBeforeChangeHook: true /*default true*/, userList: 'User' /*default 'User'*/})
```

both can be chained like `byTracking(atTracking(<fieldSchema>))`


gautamsi:
@jesstelford @timleslie any feedback on naming? should this be renamed like this:
`atTracking` -> `timeTracking`
`byTracking` -> `authorTracking`

gautamsi:
`resolveInput` caveat:
had to turn off access control and allow all action, later in code delete or overwrite the field with correct value.

it breaks the graphQL as admin ui is sending each and every field with default value rather than changed fields only.

this can be fixed when we have read only field in admin ui. #1133

MadeByMike:
@gautamsi everyone I've spoken to really loves this idea. We're struggling to determine how it should be implemented. Essentially there are 3 options: 

1. It's something we add as a core feature
2. It's something we add via a plugin architecture
3. It's something we add via higher order functions that wrap a list - similar to what you have done

My feeling is higher order functions are perfect for individual examples, but if something is likely to be used more widely (and I feel think is) it might be better implemented as some kind of plugin architecture. I've created an issue to discuss how this might work: #1489 

I like to see many features added as "plugins" rather than as core features. The idea being that anything that is more opinionated about the structure of an application should be a plugin. This includes assumptions about tracking, user list, and similar. On top of this we can build additional layers of abstraction that accommodate for more specific need of many applications. Does that make sense?

Apart from tracking, some other examples that might be 'plugins' include:

- Revision history,
- Roles,
- Migration\Seeding

Building plugins might help find limitations in the core product. An example of this is orderable\sortable fields. To support this I think we need to make some changes to the core platform and if other plugin ideas surface requirements that would be great!

I'm going to discuss plugins more with @timleslie today and add to the issue I've opened. 

This is 100% a good idea and sorry for the slow response. It's difficult as it is the first thing of this type.

gautamsi:
@MadeByMike I have proposed small change in core for this see https://github.com/keystonejs/keystone-5/issues/184#issuecomment-478442994

gautamsi:
> I like to see many features added as "plugins" rather than as core features. The idea being that anything that is more opinionated about the structure of an application should be a plugin. This includes assumptions about tracking, user list, and similar. On top of this we can build additional layers of abstraction that accommodate for more specific need of many applications. Does that make sense?

I have advocated about the plugin system for long, I will write more in #1489 . some example from https://github.com/keystonejs/keystone-5/pull/980#issuecomment-480836291

MadeByMike:
Thanks @gautamsi! @timleslie and I had a discussion and think we're going to move forward with an idea around a very simple implementation of 'list-plugins' that are basically a way to defined higher order function that modify config.

This will allow us to move forward with simple 'plugins' like this without shaving a yak. I suspect that we will need to re-visit plugins again when we look at use-cases that might need to do things like register a new list or something. That's a harder one to get right so we're going to move slow. :)  Looking forward to hearing your ideas.

gautamsi:
First thing I would propose to identify things we do eagerly (like `list.initFields` at creation of list) and try to delay them as much as possible to be point of `keystone.connect()`. This will also enable many scenarios which probably called hacking the system, but paves the way for adding more **extension points** to the system based on what others made us of. 

I should have taken #776 forward with RFC PR for authentication extensions.

dcousens:
Love the idea,  but not as a part of the core implementation; 100% prefer a plugin.

I think typically immutable rows are the ideal in any database system,  for testing,  race conditions,  verifiability,  and rollbacks.
However,  they have many problems of their own - each significantly impacting a systems design.

Typically immutable data means that you need to order your data by some kind of sequence (timestamp or otherwise) to determine what is old and what is new.
This can mean additional indexes that can blow out in database size quickly if not done carefully.
That said,  they are exceedingly appropriate and performant in scenarios where your only index may have been over only 1 field - and each group over that field doesn't update too frequently.

If that isn't a match for the requirements,  then you could use a **separate** table that tracks an appropriate delta or changeset (not necessarily complete),  timestamp,  and a created/updated `by` field;  thus keeping the source data agnostic.

I think both systems have a place,  and both are preferable to an inline set of `createdBy` or `updatedBy` fields on the source data.

dcousens:
For example,  a comparison of the systems

### Pull request
#### posts
id | text | createdBy | createdAt | updatedBy | updatedAt
-|-|-|-|-|-
0 | "I am going on a holiday" | Alice | 2019-07-19 | Alice | 2019-07-22
1 | "My holiday was a hoot" | Alice | 2019-08-13 | null


### Immutable system
#### posts
id | sequence | text | author | timestamp
-|-|-|-|-
0 | 0 | "I am thinking of going on a holiday" | Alice | 2019-07-19
0 | 1 | "I am going on holiday" | Alice | 2019-07-20
0 | 2 | "I am going on a holiday" | Alice | 2019-07-22
1 | 0 | "My holiday was a hoot" | Alice | 2019-08-13 | null

### Mutable system with detached table
#### posts
id | text | author | createdAt
-|-|-|-
0 | "I am going on a holiday" | Alice | 2019-07-19
1 | "My holiday was a hoot" | Alice | 2019-08-13

#### posts_history
id | sequence | text | author | changedAt
-|-|-|-|-
0 | 0 | "I am thinking of going on a holiday" | Alice | 2019-07-19
0 | 1 | "I am going on holiday" | Alice | 2019-07-20

gautamsi:
now that #1491 is in progress, the question is- should this be part of core? or should I create a package (userland) for this?

timleslie:
> now that #1491 is in progress, the question is- should this be part of core? or should I create a package (userland) for this?

I would like to see these available in core as

```js
import { updatedAt, createdAt, updatedBy, createdBy } from `@keystone-alpha/list-plugins`
```

gautamsi:
Thoughts on `@keystone-alpha/plugins/list`?

MadeByMike:
This is great! Now that we've merged the list plugin feature we have a better way of calling this. Let's nail down the scope and update the PR.

Can I suggest the following:

### No imputable in first release

Immutable tracking should be out of scope for now. Simple first is always a good idea. - but we can keep in mind as a config option we expose in the future.

### I like `/plugins/list` but will defer to @timleslie 

Anything is fine

### Importing `atTracking` is better

 Importing `atTracking` is better. The list plugins API was meant to save repetition and we're re-introducing it. I had a friendly argument with @timleslie yesterday and I wanted just a single `trackingFields` plugin where I think the default should be everything on and users can config `updatedAt`, `createdAt`, `updatedBy` and `createdBy` off.

@timleslie would win a real fight so I gave in but can I ask for:

```
import { atTracking, byTracking } from `@keystone-alpha/plugins/list`
```
It's actually not hard to do both `atTracking` could be a pre-composed function that combines `updatedAt`, `createdAt` and we let the user choose.

### Plugin options

For **options**  basically it is just going to be a [datetime field](https://v5.keystonejs.com/keystone-alpha/fields/src/types/date-time/) that gets added so we should pass through all the options for date field including access control options.

We should set some default access control to read only if not specified.

The only additional option I imagine we need are the field names. Making these options: `updatedAtField`, `createdAtField` and similar for byTracking...

```
atTracking({
 updatedAtField: 'updatedAt'
 createdAtField: 'updatedAt'
})
```

### Questions

Use `Anonymous User` for un-authed users? 
We can derive the User list and field from Auth settings?
Throw is there is Auth and User not found? Prevent update?

timleslie:
My motivation for going with `list-plugins` over `plugins/list` is that the later implies we have other plugins, which we don't. Even if we did, it's not clear to me that they should live together, as I suspect other future "plugins" (whatever that means) will be totally different beasts, which behave in different ways and are used for different things.

All that said, I'm not going to block either option, I'm more concerned that potential options to each individual plugin are 👌

MadeByMike:
I had an attempt with `atTracking` and here is what I came up with:

```
const { DateTime } = require('@keystone-alpha/fields');

const composeResolveInput = (originalHook, newHook) => async params => {
  const resolvedData = await originalHook(params);
  return newHook({ ...params, resolvedData });
};

const _atTracking = ({ created = true, updated = true }) => ({
  updatedAtField = 'updatedAt',
  createdAtField = 'createdAt',
  ...atFieldOptions
}) => ({ fields = {}, hooks = {}, ...rest }) => {
  const dateTimeOptions = {
    type: DateTime,
    format: 'MM/DD/YYYY h:mm A',
    access: {
      read: true,
      create: false,
      update: false,
    },
    ...atFieldOptions,
  };

  if (updated) {
    fields[updatedAtField] = {
      ...dateTimeOptions,
    };
  }

  if (created) {
    fields[createdAtField] = {
      ...dateTimeOptions,
    };
  }

  const { resolveInput: originalResolveInput } = hooks;
  const newResolveInput = ({ resolvedData, existingItem, originalInput }) => {
    if (Object.keys(originalInput).length === 0) {
      return;
    }
    const dateNow = new Date().toISOString();
    if (existingItem === undefined) {
      // create mode
      if (created) {
        resolvedData[createdAtField] = dateNow;
      }
      if (updated) {
        resolvedData[updatedAtField] = dateNow;
      }
    } else {
      // update mode
      if (created) {
        delete resolvedData[createdAtField]; // This is a bug?, data for createdAtField should not be sent
      }
      if (updated) {
        resolvedData[updatedAtField] = dateNow;
      }
    }
    return resolvedData;
  };
  hooks.resolveInput = composeResolveInput(originalResolveInput, newResolveInput);
  console.log({ fields, hooks, ...rest });
  return { fields, hooks, ...rest };
};

const createdAt = options => _atTracking({ created: true, updated: false })(options);
const updatedAt = options => _atTracking({ created: false, updated: true })(options);
const atTracking = options => _atTracking({ created: true, updated: true })(options);

module.exports = { createdAt, updatedAt, atTracking };
```

gautamsi:
@MadeByMike I have used your version, tweaked and fixed it, still need to have field access to allow all due to how admin-ui tries to send all field on creation and updation. Admin-UI does not recognize the field with readonly value, not sure if I missed anything.

packages/list-plugins/lib/tracking/atTracking.js
```diff
@@ -0,0 +1,69 @@
...
+  updatedAtField = 'updatedAt',
+  createdAtField = 'createdAt',
+  ...atFieldOptions
+} = {}) => ({ fields = {}, hooks = {}, ...rest }) => {
+  const dateTimeOptions = {
+    type: DateTime,
+    format: 'MM/DD/YYYY h:mm A',
+    access: {
+      read: true,
+      create: true, // TODO: revert to false when read only fields are available
```
gautamsi: someone can inadvertently choose the field and try update that in admin-ui as this is available even if we say `false` here. This breaks admin ui creation of item with error 400. admin-ui sends graphql mutation with the field and that is not added to schema.

MadeByMike: Rather than shipping this with todo comments let's fix the issue that is holding us back.
I need to understand it better. Is there a current issue open with a simple test case?

packages/list-plugins/lib/tracking/atTracking.js
```diff
@@ -0,0 +1,69 @@
...
+    };
+  }
+
+  if (created) {
+    fields[createdAtField] = {
+      ...dateTimeOptions,
+    };
+  }
+
+  const originalResolveInput = hooks.resolveInput || (({ resolvedData }) => resolvedData); // possible to have no hooks provided
```
MadeByMike:
```suggestion
  const originalResolveInput = hooks.resolveInput
```
Can we deal with the no hooks situation in `composeHooks` rather than creating a noop function? 

Let's also move `originalResolveInput` under `newResolveInput` as it get's lost up here.


I also think we need to make a new issue so that hooks can accept an array or we expose a generic `composeHooks` function. It should not be the responsibility of plugin authors to do this.

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,70 @@
+const { Relationship } = require('@keystone-alpha/fields');
+
+const composeResolveInput = (originalHook, newHook) => async params => {
```
MadeByMike: Let's make composeResolveInput a utility function and import it into both of these rather than write the same function twice.

packages/list-plugins/index.js
```diff
@@ -0,0 +1,12 @@
+const { atTracking, createdAt, updatedAt } = require('./lib/tracking/atTracking');
+
```
MadeByMike:
```suggestion
```

test-projects/login/index.js
```diff
@@ -36,6 +45,23 @@ keystone.createList('Post', {
   },
 });
 
+keystone.createList('ListWithPlugin', {
```
MadeByMike: Why are we adding this to the login test? And I'm not sure we need a cypress test. It should probably be part of API testing and ideally its own section?

I actually wouldn't mind if this (and other plugins) are published as their own npm modules with self contained tests and everything... does that mean outside mono repo @jesstelford, @timleslie let's discuss.

timleslie: I don't mind having one of our cypress tests utilise these plugins purely from a smoke testing point of view. I would put it onto the `basic` project rather than the `login` project, as that's where we have the kitchen sink.

The plugins should definitely have their own unit tests, but there's no reason for them to live outside the mono repo. The `@keystone-alpha/list-plugin` package will contains all of our sanctioned plugins, which we should be testing as much as possible.

MadeByMike: I'm fine with all that :)

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,70 @@
...
+
+const _byTracking = ({ created = true, updated = true }) => ({
+  updatedByField = 'updatedBy',
+  createdByField = 'createdBy',
+  trackingListName = 'User',
+  ...byFieldOptions
+} = {}) => ({ fields = {}, hooks = {}, ...rest }) => {
+  const relationshipOptions = {
+    type: Relationship,
+    ref: trackingListName,
```
MadeByMike: We don't need this. We spread `byFieldOptions` so they can just pass `ref`

MadeByMike:
```suggestion
```

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,70 @@
...
+
+const composeResolveInput = (originalHook, newHook) => async params => {
+  const resolvedData = await originalHook(params);
+  return newHook({ ...params, resolvedData });
+};
+
+const _byTracking = ({ created = true, updated = true }) => ({
+  updatedByField = 'updatedBy',
+  createdByField = 'createdBy',
+  trackingListName = 'User',
```
MadeByMike:
```suggestion
```

packages/list-plugins/lib/tracking/atTracking.js
gautamsi: there is no specific issue for this, but there is on open for simple feature https://github.com/keystonejs/keystone-5/pull/1286

test-projects/login/index.js
gautamsi: initially used `basic` project but soon realised that due to no authentication the `byTracking` features would not work.

packages/list-plugins/lib/tracking/byTracking.js
gautamsi: moved to util.js

packages/list-plugins/lib/tracking/byTracking.js
gautamsi: done

packages/list-plugins/lib/tracking/atTracking.js
MadeByMike: This is now resolved

packages/list-plugins/lib/tracking/atTracking.js
```diff
@@ -0,0 +1,65 @@
...
+  updatedAtField = 'updatedAt',
+  createdAtField = 'createdAt',
+  ...atFieldOptions
+} = {}) => ({ fields = {}, hooks = {}, ...rest }) => {
+  const dateTimeOptions = {
+    type: DateTime,
+    format: 'MM/DD/YYYY h:mm A',
+    access: {
+      read: true,
+      create: true, // TODO: revert to false when read only fields are available
```
MadeByMike:
```suggestion
      create: false,
```

packages/list-plugins/lib/tracking/atTracking.js
```diff
@@ -0,0 +1,65 @@
...
+      update: true, // TODO: revert to false when read only fields are available
```
MadeByMike:
```suggestion
      update: false,
```

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,65 @@
...
+  updatedByField = 'updatedBy',
+  createdByField = 'createdBy',
+  ...byFieldOptions
+} = {}) => ({ fields = {}, hooks = {}, ...rest }) => {
+  const relationshipOptions = {
+    type: Relationship,
+    ref: 'User',
+    access: {
+      read: true,
+      create: true, // TODO: revert to false when read only fields are available
```
MadeByMike:
```suggestion
      create: false,
```

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,65 @@
...
+      update: true, // TODO: revert to false when read only fields are available
```
MadeByMike:
```suggestion
      update: false,
```

MadeByMike:
@gautamsi this is ready to go once you commit those suggestions. I would have merged it last week if they permissions were correct. I think it's better that this is correct and we surface the bug elsewhere in the code rather than add todo like this in a PR. 

Either way that issue is fixed now so this should work as expected. 

This is an excellent contribution! Thanks!

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -0,0 +1,65 @@
...
+      if (created) {
+        resolvedData[createdByField] = id;
+      }
+      if (updated) {
+        resolvedData[updatedByField] = id;
+      }
+    } else {
+      // update mode
+      if (created) {
+        delete resolvedData[createdByField]; // createdByField No longer sent by api/admin, but access control can be skipped!
```
gautamsi: @MadeByMike this is no longer sent from admin/graphql as addressed in #1521 
but may still be sent by skipping access control (not tested)

MadeByMike: It can't hurt to leave it here just incase it is ever sent. But if you're confident you can remove it.

gautamsi: will leave it for now unless tested later.

Issue #1275, opened 2019-06-05
Expose programatic API for executing graphql queries/mutations via the `keystone` instance


jesstelford:
A suggested API is:

```javascript
    keystone.createItem('Session', {
      outputFragment: `
        id
        posts {
          id
          title
        }
      `,
      variables: {}
    });
```

Where `createItem` matches our GraphQL API naming for `create<List>` (so we'd have ones like `keystone.update<List>`/ `keystone.delete<List>`), and `'Session'` is the list we're operating on.

Internally, this should follow all the same rules as a HTTP request that triggers a query which adds an extra complication: How can we ensure that the Apollo Server injects the correct `context` and runs the query through all the same error formatting, etc? It doesn't appear that `apollo-graphql` exposes a programatic API. Perhaps we have to fake it and do a localhost http request to the listening port internally? Is there an overhead with doing that?

timleslie:
If you create an `ApolloServer` object then you can use

```
server = new ApolloServer(...)
const { schema } = server;
```

to obtain a `GraphQLSchema` object. This is can be used as follows:

```
const { graphql } = require('graphql');
graphql(schema, query, null, context, variables)
```

see https://graphql.org/graphql-js/graphql/#graphql for details of the `query` API.

At the moment, every time we call `createApolloServer` we have this call `keystone.registerSchema(schemaName, server.schema);`, which keeps track of all these schema objects in the `_graphQLQuery` property.

These are then used for example in the the `test-utils` packages as such:

```
function graphqlRequest({ keystone, query }) {
  return keystone._graphQLQuery[SCHEMA_NAME](query, keystone.getAccessContext(SCHEMA_NAME, {}));
}
```

In this case the auth details on the context are not specified.

If we want to create generic `keystone.createItem` methods, we'll need to decide a) which schema to use (maybe just default to the "admin" schema?) and what context to use. Do we want skip access control altogether, or do we want to be able to specify a particular authed user to masquerade as?

jesstelford:
Having to setup the `context` each time is the tricky part.

Perhaps we could extend the `keystone.registerSchema` method to set that up for us correctly, then expose that via a `keystone.query()` or `keystone.graphQL()` method?

jesstelford:
Another use-case is when given a `List` instance, being able to execute a query against it directly without having to specify the list name, etc. For example:

```javascript
    const list = this.getList();
    list.createItem({
      outputFragment: `
        id
        posts {
          id
          title
        }
      `,
[...]

PR #1294, opened 2019-06-12
Social auth fixes

ticidesign:
Refactor social auth to include:
 - GraphQL to fetch data
 - Change social scope to collect emails


packages/passport-auth/lib/Passport.js
```diff
@@ -44,7 +49,7 @@ class PassportAuthStrategy {
       this.keystone.createList(this.config.sessionListKey, {
         fields: {
           [FIELD_SERVICE_NAME]: { type: Text },
-          [FIELD_USER_ID]: { type: Text },
+          [FIELD_USER_ID]: { type: Text, mongooseOptions: { index: true } },
```
jesstelford: This will only add an index for Mongoose, but not for any other database adapter (eg; Postgres).

packages/passport-auth/lib/Passport.js
```diff
@@ -76,7 +81,7 @@ class PassportAuthStrategy {
         }
         resolve({
           id: profile.id,
-          username: profile.username || null,
+          username: profile.displayName || null,
```
jesstelford: Should this be:

```
username: profile.username || profile.displayName || null
```
?

I think different services return the profile with different sets of data.

But really, is the `displayName` a _`username`_? They're different things I think 🤔

packages/passport-auth/lib/Passport.js
```diff
@@ -89,59 +94,84 @@ class PassportAuthStrategy {
...
+              # eg; facebookUserId / googleUserId
+              ${this.serviceIdField}: $serviceId
+            }
+          ) {
+            id
+          }
+        }
+      `,
+        { serviceId: validatedInfo.id }
+      );
```
jesstelford: @timleslie Is there a better way to do this graphql query? We have access to `this.keystone` (the Keystone instance) here, but couldn't figure out how to correctly do a query. c.f. #1275

packages/passport-auth/lib/Passport.js
```diff
@@ -174,18 +206,50 @@ class PassportAuthStrategy {
...
-      });
+      const passportSessionMutationName = this.getSessionList().gqlNames.updateMutationName;
+      const passportSessionMutationInputName = this.getSessionList().gqlNames.updateInputName;
+      const serviceItem = await request(
+        this.config.endpoint,
+        `
+          mutation($id: ID!, $data: ${passportSessionMutationInputName}) {
+            ${passportSessionMutationName}(id: $id , data: $data) {
+              id
+              serviceUserId
```
jesstelford:
```suggestion
              ${FIELD_USER_ID}
```

packages/passport-auth/lib/Passport.js
```diff
@@ -174,18 +206,50 @@ class PassportAuthStrategy {
...
+              serviceUsername
```
jesstelford:
```suggestion
              ${FIELD_USERNAME}
```

packages/passport-auth/lib/Passport.js
```diff
@@ -174,18 +206,50 @@ class PassportAuthStrategy {
...
+      const newServiceItemFields = {
+        [this.serviceIdField]: serviceItem[passportSessionMutationName][FIELD_USER_ID],
+        [this.serviceUsernameField]: serviceItem[passportSessionMutationName][FIELD_USERNAME],
+      };
+
+      await request(
+        this.config.endpoint,
+        `
+          mutation($id: ID!, $newServiceItemFields: ${userMutationInputName}) {
+            ${userMutationName}(id: $id ,data: $newServiceItemFields) {
```
jesstelford:
```suggestion
            ${userMutationName}(id: $id, data: $newServiceItemFields) {
```

packages/passport-auth/lib/Passport.js
ticidesign: You're right, Facebook returns a `name` object. Twitter returns `username` and Google `displayName`

gautamsi: I believe all of them gets normalized to username inside passportjs strategy

packages/passport-auth/lib/Passport.js
gautamsi: we should have universal index option and have documentation if the adapter supports it or not

essentially `index: true` on field rather than mongooseOptions or knexOptions

jesstelford: Brilliant idea, I've captured that in this issue: https://github.com/keystonejs/keystone-5/issues/1297

jesstelford:
@ticidesign This PR needs the `graphql-request` package added correctly;

```
cd packages/passport-auth
bolt add graphql-request
```

That will fix all the failing tests.

packages/passport-auth/lib/Passport.js
```diff
@@ -76,7 +81,7 @@ class PassportAuthStrategy {
         }
         resolve({
           id: profile.id,
-          username: profile.username || null,
+          username: profile.username || profile.displayName || `${profile.name.givenName} ${profile.name.familyName}`,
```
gautamsi: assuming `profile.name` is never null!

packages/passport-auth/lib/Google.js
```diff
@@ -10,7 +10,7 @@ class GoogleAuthStrategy extends PassportAuthStrategy {
       {
         sessionIdField: 'googleSession',
         keystoneSessionIdField: 'keystoneGoogleSessionId',
-        scope: ['profile'],
+        scope: ['openid profile', 'openid email'],
```
gautamsi: I don't remember but is adding more scopes gets you a user consent screen on google side?
users should be able to add more scope in config if needed.

packages/passport-auth/lib/Passport.js
```diff
@@ -89,59 +94,86 @@ class PassportAuthStrategy {
...
     };
 
-    // Only add a reference to the parent list when we know the link exists
-    if (pastSessionItem) {
-      newSessionData[this.config.itemField] = fieldItemPopulated.id;
+    let itemId;
+    // Find an existing user that matches the given validated service id
+    try {
+      const queryName = this.getList().gqlNames.listQueryName;
+      const data = await request(
```
jesstelford: I've added an issue to come back and fix this later: https://github.com/keystonejs/keystone-5/issues/1300

packages/passport-auth/lib/Passport.js
ticidesign: Every service returns a different set of data for the name, username and displayName. So we decide to return the whole profile object and use just the service ID in this case.

packages/passport-auth/lib/Google.js
jesstelford: Yeah, it probably will. I think its ok to have this as a default, but allow the app developer to overwrite it if they want less scope. In my experience, most apps want email anyway, so this makes the default work as expected for them.

PR #1301, opened 2019-06-13
Fix behaviour of isRequired within update operations

timleslie:


packages/keystone/lib/List/index.js
```diff
@@ -1094,14 +1094,15 @@ module.exports = class List {
...
     // Check for isRequired
     const fieldValidationErrors = this.fields
       .filter(
         field =>
           field.isRequired &&
           !field.isRelationship &&
-          (resolvedData[field.path] === undefined || resolvedData[field.path] === null)
+          ((operation === 'create' &&
+            (resolvedData[field.path] === undefined || resolvedData[field.path] === null)) ||
+            (operation === 'update' && resolvedData[field.path] === null))
```
jesstelford: Should this check also for `undefined`?

```javascript
(operation === 'update' && Object.prototype.hasOwnProperty.call(resolvedData, field.path) && (resolvedData[field.path] === null || resolvedData[field.path] === undefined)))
```

timleslie: Good call, fixed.

PR #1306, opened 2019-06-14
Refactoring block loading logic

jesstelford:
To support the upcoming OEmbed block, I needed to be able to dynamically specify a depended upon view file whilst still having it dynamically loaded with `import()`. After I pulled that card out from the bottom of the house, this is what it looked like after rebuilding...

See the `oembed-block` branch for the upcoming OEmbed block work.

Changes:

## `@keystone-alpha/fields`

- **Blocks**
  - Blocks must now be classes, which implement the `Block` interface.
  - `type` is no longer passed to a Block's constructor (it is expected to know it already).
  - Remove the `embed` block
  - `Block#dependencies` is removed; any depended upon views should be returned from `Block#getAdminViews()`.
  - `Block#extendAdminMeta()` is renamed to `Block#getViewOptions()`.
  - Add new `Block#getAdminViews()` method.
- **Field Types**
  - `Field#extendViews()` is renamed to `Field#extendAdminViews()`.

## `@keystone-alpha/keystone`

- Gather views from fields via the renamed method `#extendAdminViews()` (was `#extendViews()`)

## `@arch-ui/input`

- When rendered as a <textarea>, the <Input> component is no longer limited to 100px height. `height` is replaced with `min-height`.

## `test-projects/basic`

- Remove the deleted `embed` block from the test project


packages/app-admin-ui/server/env.js
```diff
@@ -1,4 +1,4 @@
 exports.mode = process.env.NODE_ENV === 'production' ? 'production' : 'development';
 
 exports.enableDevFeatures =
-  process.env.ENABLE_DEV_FEATURES === 'true' ? true : !(exports.mode === 'production');
+  process.env.ENABLE_DEV_FEATURES === 'true' ? true : exports.mode !== 'production';
```
jesstelford: DeepScan complained about this for some reason... Did their config change? If so, how do we stop that from blocking PRs in the future?

https://deepscan.io/dashboard/#view=project&tid=3252&pid=4799&bid=38273&subview=pull-request&prid=114374

packages/fields/src/types/Content/Implementation.js
```diff
@@ -110,45 +101,35 @@ export class Content extends Relationship {
...
+      Array.isArray(block) ? block : [block, {}]
+    );
+
+    blocks.push(...DEFAULT_BLOCKS);
 
-    blocks.push(
-      ...DEFAULT_BLOCKS.filter(
-        defaultBlock =>
-          !blocks.find(
-            block => (Array.isArray(block) ? block[0] : block).type === defaultBlock.type
```
jesstelford: Removed this check as there is a duplicate check following. I know the semantics are slightly different, but the end result is the same; there should only ever be a single `paragraph` block, and that's added by default.

packages/fields/src/types/Content/serialiser.js
```diff
@@ -65,6 +65,14 @@ export function serialiseSlateValue(value, blocks) {
           );
         }
 
+        if (!block.path) {
+          throw new Error(
+            `No mutation path set for block view type '${
+              block.type
+            }'. Ensure the block's view exports a 'path' key corresponding to the mutation path for saving block data`
+          );
+        }
```
jesstelford: There's a check for `.type`, but the other required `.path` didn't have a check so I added it.

Note: This `block` could be a _client-side_ block, so would not trigger the getter which throws an error if `get path()` isn't implemented (see `Block.js`), hence we need the check here.

test-projects/basic/index.js
```diff
@@ -132,7 +132,6 @@ keystone.createList('Post', {
         Content.blocks.blockquote,
         Content.blocks.orderedList,
         Content.blocks.unorderedList,
-        [Content.blocks.embed, { apiKey: process.env.EMBEDLY_API_KEY }],
```
jesstelford: Coming Soon™️

jesstelford:
The test failure is the "Timed Out expecting Our Post" one. Yet the screenshot of the failure shows that it _is_ there: https://20734-128193054-gh.circle-artifacts.com/0/tmp/screenshots/add-edit-delete-data.js/Deleting%20data%20--%20Deleting%20data%20to%20Posts%20%28failed%29.png

It's happening on other PRs too - perhaps we can ignore it for this PR? 🙏

packages/app-admin-ui/server/env.js
mitchellhamilton: A while ago, I was talking to @timleslie and we thought we should remove DeepScan because there are so many false positives and we aren't really getting any value out of it now.

packages/fields/src/Block.js
```diff
@@ -32,15 +38,22 @@ export class Block {
...
+  /**
+   * @param graphQlArgs {Object}
+   */
+  getMutationOperationResults() {
+    return;
   }
 
-  extendAdminMeta(meta) {
-    return meta;
+  getAdminViews() {
```
mitchellhamilton: Not sure if there's a better name for this but when I was reading through this, it wasn't totally clear to me how the views that `getAdminViews` return are used on the client. Eventually I looked at the implementation of the Content field type Controller and understood that they're all interpreted as blocks on the client but that seemed strange because it's a very different meaning to `extendViews`/now `extendAdminViews` on field Controllers where it's an object and each view on the object is used by its own thing but the names are really similar.

jesstelford: Yeah, I got stuck on naming this one for that exact reason. There's another wrinkle: the paths are loaded and interpreted as Blocks _if they have an export of `type`_, otherwise they're just taking advantage of the `import()` mechanism.

Open to suggestions (with minimal bike shedding) 👍

mitchellhamilton: > if they have an export of type, otherwise they're just taking advantage of the import() mechanism.

What's an example of something that would be here and wouldn't be a block?

jesstelford: This one: https://github.com/keystonejs/keystone-5/pull/1307/files#diff-c850d4389f69a80bcd0223eee6a2cf48

Used here: https://github.com/keystonejs/keystone-5/pull/1307/files#diff-eef8d04a752abe36a598e4cdca7cbdb4R66

And loaded here: https://github.com/keystonejs/keystone-5/pull/1307/files#diff-f72926d16a4c17fb27bd1aa14ec46339R24

mitchellhamilton: Hmmmmmmm, I don't really like that the path has to be specified in the options and getAdminViews but I can't think of a better way to do all this right now so I guess this is all fine then.

PR #1307, opened 2019-06-14
Oembed block

jesstelford:
Depends on #1306 

Adds an `oembed` block type which can generically render whatever embed code is required by the configured adapter.

Screen Shot 2019-06-14 at 3 19 28 pm


gautamsi:
do we have any sample for how to render this on next app or on pug view?

jesstelford:
Not yet. The same goes for all the other Blocks: the only "renderer" so far is the AdminUI, but I will be building some examples soon.

The core of what's required it's too make a query like so:

```graphql
{
  Post(where: {id: "5ce222f898c9bf47598db9c2"}) {
    value {
      document
      cloudinaryImages {
        id
        image {
          publicUrl
          __typename
        }
        align
        __typename
      }
      oEmbeds {
        id
        embed {
          type
          originalUrl
          version
          title
          cacheAge
          provider {
            name
            url
            __typename
          }
          author {
            name
            url
            __typename
          }
          thumbnail {
            url
            width
            height
            __typename
          }
          ... on OEmbedPhoto {
            url
            width
            height
            __typename
          }
          ... on OEmbedVideo {
            html
            width
            height
            __typename
          }
          ... on OEmbedRich {
            html
            width
            height
            __typename
          }
          __typename
        }
        __typename
      }
      __typename
    }
    __typename
  }
}
```

Then there'll be a keystone-supplied deserialisation lib which will convert the result into a format ready for rendering by slate.js 👍

Issue #1335, opened 2019-06-26
Test framework expects all field types to support all adapters


molomby:
I get that we want all core field types to support all core database adapters but, until we get there, can we make it so the test suite doesn't assume this is the case?

jesstelford:
This _is_ the case today; otherwise the tests wouldn't be passing.

Is there a specific use-case you have in mind for a core field type not supporting a core database adapter?

molomby:
Ha, quite right. I want to add a type for UUIDs.

Knex/pg support is easy; they're stored in a binary format and implicitly cast between it and Strings on the way in and out of the DB.

Without a fair bit of effort, Mongoose/Mongo will treat them as Strings. This bloats the DB and breaks equality comparisons due to case sensitivity. Supporting them _properly_ for Mongo requires explicitly converting them to and from a binary type whenever they move in or out of the DB.

It's totally doable -- I'd love UUIDs (and all core types) to be fully supported -- I just don't want to do it today :)

This is part of a larger plan to make UUIDs the default id type for lists using the Knex adapter (or at least a core option). That what's stopping me from pulling this out into a non-core package, etc.

molomby:
Uuids added in #1372 with basic Mongoose support

PR #1310, opened 2019-06-17
[WIP] JSON & Memory Database Adapters

jesstelford:
Fixes #1177

Introduces 2 new database adapters:
- `@keystone-alpha/adapter-json`: A local `.json` file as the database storage engine
- `@keystone-alpha/adapter-memory`: An in-memory storage engine

## `@keystone-alpha/adapter-json`

```javascript
const { JSONAdapter } = require('@keystonejs/adapter-json');
const keystone = new Keystone({
  name: 'My Awesome Project',
  adapter: new JSONAdapter(),
});
// ...
keystone.connect('./database.json');
```

## `@keystone-alpha/adapter-memory`

```javascript
const { MemoryAdapter } = require('@keystonejs/adapter-memory');
const keystone = new Keystone({
  name: 'My Awesome Project',
  adapter: new MemoryAdapter(),
});
// ...
keystone.connect();
```

## TODO

Things left to do:

- [x] handling of `where` clauses
  - [x] Simple field clauses
  - [x] `AND` clauses
  - [x] `OR` clauses
  - [x] Nested relationship where clauses
  - [x] Support all field types
    - [x] `CalendarDay`
    - [x] `Checkbox`
    - [x] `CloudinaryImage`
    - [x] `Color`
    - [x] `DateTime`
    - [x] `Decimal`
    - [x] `File`
    - [x] `Float`
    - [x] `Integer`
    - [x] `OEmbed`
    - [x] `Password`
    - [x] `Relationship`
    - [x] `Select`
    - [x] `Slug`
    - [x] `Text`
    - [x] `Unsplash`
    - [x] `Url`
    - [x] `Uuid`
- [x] All of ☝️ but for the `Memory` adapter
- [ ] Tests
  - [ ] api-tests/uniqueness/unique.test.js
  - [ ] api-tests/queries/search.test.js
  - [ ] api-tests/queries/limits.test.js
  - [ ] packages/email/tests/index.test.js // Not sure if I need to fix this snap or if a local issue
- [ ] Relationship tests
  - [ ] api-tests/relationships/nested-mutations/two-way-backreference/to-many.test.js
  - [ ] api-tests/relationships/nested-mutations/two-way-backreference/to-one-required.test.js
  - [ ] api-tests/relationships/nested-mutations/create-and-connect-many.test.js
  - [ ] api-tests/relationships/nested-mutations/create-and-connect-singular.test.js
  - [ ] api-tests/relationships/nested-mutations/disconnect-many.test.js
[...]

PR #1336, opened 2019-06-26
Allow core field types that don't support all core DB adapters

molomby:
See #1335

### DIFF
--- a/api-tests/fields.test.js
+++ b/api-tests/fields.test.js
@@ -9,15 +9,23 @@ jest.setTimeout(60000);
 
 describe('Test CRUD for all fields', () => {
   const typesLoc = path.resolve('packages/fields/src/types');
-  const testModules = fs
+  const testModulePaths = fs
     .readdirSync(typesLoc)
     .map(name => `${typesLoc}/${name}/filterTests.js`)
     .filter(filename => fs.existsSync(filename));
-  testModules.push(path.resolve('packages/fields/tests/idFilterTests.js'));
+  testModulePaths.push(path.resolve('packages/fields/tests/idFilterTests.js'));
 
-  multiAdapterRunners().map(({ runner, adapterName }) =>
+  multiAdapterRunners().map(({ runner, adapterName }) => {
+    // Skip tests for field adapters that don't exist
+    const testModulesForAdapter = testModulePaths.map(require).filter(mod => {
+      if (mod.name === 'ID') return true;
+      if (mod.type && mod.type.adapters.hasOwnProperty(adapterName)) return true;
+      console.log(`Skipping CRUD tests for ${mod.name}; no ${adapterName} field adapter supplied`);
+      return false;
+    });
+
     describe(`Adapter: ${adapterName}`, () => {
-      testModules.map(require).forEach(mod => {
+      testModulesForAdapter.forEach(mod => {
         describe(`All the CRUD tests for module: ${mod.name}`, () => {
           const listName = 'test';
           const keystoneTestWrapper = (testFn = () => {}) =>
@@ -43,6 +51,6 @@ describe('Test CRUD for all fields', () => {
           mod.filterTests(keystoneTestWrapper);
         });
       });
-    })
-  );
+    });
+  });
 });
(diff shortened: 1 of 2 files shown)


molomby:
> I'm concerned this change would open a path to where we break our promise of "just pick an adapter and KS5 will work!" - some subsets of types would only work in some subsets of database adapters.

This is a tricky one... On one hand I share you concern -- people should be able to pick up KS, pick an adapter and have everything work as they'd expect. On the other hand I think we need to be careful about oversimplifying things.

Different storage mechanisms have different abilities and different strengths. Not all adapters (even those in core) will or should support the same functionality.

Eg -- I think there should be a Mongo ObjectID field type. This would be useful if you wanted to reference an item in a different DB (so not a relationship field, just an ID). Could we support this on Knex? Absolutely. Should we bother? Well.. maybe, but I wouldn't want to block its release on that basis. 95% of the people who will want to use this type will be on the Mongoose adapter.

> This problem will only be exacerbated as we add more database adapters (eg; #1310)

Completely.. but do you really want to support many-to-many relationships on a JSON DB adapter? Or full text indexing with configurable languages on an in memory DB? It's doable, sure, but you risk building a stack of functionality no one actually wants. I'm 100% ok saying that, if you want branching full text search over a set of items, you need to pick a storage adapter that supports that functionality and, more to the point, _it's ok to build a new adapter that doesn't necessarily support that functionality_.

I'm worried that if we say "every core adapter must support every core field type/list config" is a hard rule we'll end up either:

* Creating a huge amount of work and holding back useful functionality until it supports all the everything, or..
* Be forced to reduce our core functionality to the lowest common denominator that all core adapter can support

> The linked issue mentions you want to create a new type that only supports Knex for now; can I suggest you do that as a separate package instead of as a core KS5 type? That way you don't run into any failing tests here, and can still use the type ASAP.

Yeah, I'm working on other branches that need this to be core. Specifically..

* I want ID columns to use existing field types and be added (automatically) to the field array, rather than being exceptional
* Devs should be able to specify the type used for the list ID

Happy to discuss all of this as well ^^

molomby:
Interested to here @timleslie and @JedWatsons thoughts on this too ^^

jesstelford:
> people should be able to pick up KS, pick an adapter and have everything work as they'd expect.

Yes, agreed. All core field types and core database adapters should _just work as expected_.

> Eg -- I think there should be a Mongo ObjectID field type. [..] Could we support this on Knex? Absolutely. Should we bother?

Yes.

If not, then it shouldn't be a core field type.

> do you really want to support many-to-many relationships on a JSON DB adapter?

Yes.

> full text indexing with configurable languages on an in memory DB?

No. This is not a feature of Keystone.

> it's ok to build a new adapter that doesn't necessarily support that functionality.

Agreed. That adapter can be a seperate package, and is not a core adapter.

> I'm worried that [..] holding back useful functionality until it supports all the everything

Until it classifies as a core field type, it can still be a perfectly usable third party field type. Publishing npm modules is cheap, and if it fully qualifies for inclusion in core, then we can bring it in 👍

> - I want ID columns to use existing field types and be added (automatically) to the field array, rather than being exceptional
> - Devs should be able to specify the type used for the list ID

Can you open an issue for this functionality? There is ongoing work that will make this much easier (specifically; de-special-casing the `label` field in a similar way).

molomby:
Aaah ok, so we're actually talking about the definition of "core". I'd aways though the core Keystone field types and adapters should be designed and tested to work well together but wouldn't have said "all core field types work with all core adapters" was a _hard_ rule. 

Spoke to @JedWatson about this all last night and he's a lot closer to your interpretation. He reckons the `Uuid` field should be core but also should support Mongoose as a priority (even if it does so in a crappy way). I still tend to think a small number of well justified, short-term exceptions is better than a half baked implementation (but not strongly).

There's also talk of building out `Increment` (ie. an auto-incrementing serial value) and `MongoID` field types that would exist in the monorepo but outside the core fields package.

> Can you open an issue for this functionality? There is ongoing work that will make this much easier (specifically; de-special-casing the label field in a similar way).

Nice one. Yeah, I'll be back on this tomorrow (2019-06-28); let's catch up then? Would be great to hash some of these other aspects/field types out.

PR #1344, opened 2019-06-27
Upgrade react-select

mitchellhamilton:


timleslie:
Last time we tried to upgrade we ran into this issue, which is why this is version pinned to `2.3.0` at the moment. It would be good to get it back into a `^` range if possible.

https://github.com/keystonejs/keystone-5/issues/827

PR #1346, opened 2019-07-01
Docs/new project

vlad-elagin:
Hello guys!

I've written few pages of Guides, while trying out Keystone 5. There would be more soon probably. Also english isn't my first language, so pardon any typos. It would be nice if you read them out.

I'm open to comments.
Thanks.


jesstelford:
Thank you @vlad-elagin 🙏 !

I'll give it a more detailed look over soon, but after a cursory glance; this all looks great!

PR #1352, opened 2019-07-02
(arch-ui/tooltip) Allow modifiers to be passed in

jordanoverbye:
This allows you to pass in `popper-js` [modifiers](https://popper.js.org/popper-documentation.html#modifiers) to the `arch-ui/tooltip` component. This is especially useful when you have components with `overflow` set which cause overlapping UI issues. @jossmac can you please review.


packages/arch/packages/tooltip/src/index.js
```diff
@@ -172,7 +177,7 @@ export default class Tooltip extends Component<Props, State> {
   };
 
   render() {
-    const { children, content, onHide, onShow, placement, className } = this.props;
+    const { children, content, onHide, onShow, placement, modifiers, className } = this.props;
```
jossmac: I'd prefer not to extend the tooltip's API. I think, in this case, we're better off just applying the applicable modifiers:

```jsx
<Popper
  modifiers={{ preventOverflow: { boundariesElement: 'scrollParent' } }}
/>
```

The user shouldn't have to know about Popper's API to have the tooltip work as expected in an overflow container.

PR #1357, opened 2019-07-04
Markdown field type

### LINKED ISSUE #316: Markdown field type
A basic WYSIWYG editor, supporting Markdown formatting and output (bold, italic, lists, headings, links, images, tables).

- Admin UI editor functionality
- Sanitisation concerns (as per [KS4](https://github.com/keystonejs/keystone/pull/4478/commits/5cb6405dfc0b6d59003c996f8a4aa35baa6b2bac))
- Field customisation options (eg, for sanitisation, ...)

Relates to #10

mitchellhamilton:
Closes #316

markdown editor

There's also a little admin ui performance thing in here that I fixed when debugging why editing was slow.


packages/fields-markdown/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@keystone-alpha/fields-markdown",
+  "description": "KeystoneJS Markdown Field Type",
+  "version": "3.0.2",
```
timleslie: Should we reset this to `0.0.0` and let `changesets take care of things?

mitchellhamilton: Done (the version was only at 3.0.2 because I copied this package from somewhere else)

packages/fields-markdown/src/views/Field.js
```diff
@@ -0,0 +1,133 @@
...
+        >
+          {icon}
+          <A11yText>{label}</A11yText>
+        </ToolbarButton>
+      )}
+    </Tooltip>
+  );
+};
+
+export default function WysiwygField({ field, errors, value, onChange }) {
```
jesstelford:
```suggestion
export default function MarkdownField({ field, errors, value, onChange }) {
```

PR #1358, opened 2019-07-04
Extract `Content` field into its own package: `@keystone-alpha/field-content`

jesstelford:
This is to start isolating some of the functionality of "Blocks", but also give a place where we can export the (de)serialisation libraries.


packages/field-content/src/Implementation.js
```diff
@@ -224,7 +218,7 @@ export class Content extends Relationship {
   extendAdminViews(views) {
     return {
       ...views,
-      blocks: unique(flatMap(this.blocks, block => block.getAdminViews())),
+      blocks: unique(flatMap(this.blocks, block => block.getAdminViews({ availableBlocks }))),
```
jesstelford: Dependency inject so that other blocks can depend-upon the defaults.

packages/field-content/src/blocks/heading.js
```diff
@@ -0,0 +1,14 @@
...
+import Block from '../Block';
+
+export default class HeadingBlock extends Block {
+  get type() {
+    return 'heading';
+  }
+  getAdminViews({ availableBlocks }) {
+    return [
+      importView('../views/editor/blocks/heading'),
+      ...new availableBlocks.paragraph().getAdminViews({ availableBlocks }),
```
jesstelford: Here we consume the dependency injected blocks, and further pass it on to the block we're depending upon. Dependency-ception!

packages/field-content/src/slate-serializer.js
```diff
@@ -42,19 +82,19 @@ function shallowNodeToJson(node) {
...
+export function serialize(value, blocks) {
   const allMutations = {};
 
   const serializedDocument = walkSlateNode(value.document, {
     visitBlock(node) {
       const block = blocks[node.type];
 
       // No matching block that we're in charge of
-      if (!block) {
+      if (!block || typeof block.serialize !== 'function') {
```
jesstelford: Not everything has a serialization process (such as a paragraph)

packages/field-content/src/views/Controller.js
```diff
@@ -71,6 +67,59 @@ export default class ContentController extends TextController {
     });
   }
 
+  getFilterGraphQL = ({ type, value }) => {
```
jesstelford: Instead of extending the `Text` controller, we bring the implementation in directly and modify it slightly so it's more appropriate for the Content type.

packages/field-content/src/views/editor/index.js
```diff
@@ -25,8 +25,8 @@ function getSchema(blocks) {
     blocks: {},
   };
   Object.keys(blocks).forEach(type => {
-    if (blocks[type].schema !== undefined) {
-      schema.blocks[type] = blocks[type].schema;
+    if (typeof blocks[type].getSchema === 'function') {
+      schema.blocks[type] = blocks[type].getSchema({ blocks });
```
jesstelford: Dependency inject the blocks into a `getSchema` function.

packages/field-content/src/views/editor/index.js
```diff
@@ -37,28 +37,30 @@ function Stories({ value: editorState, onChange, blocks, className }) {
...
-            return <block.Node {...props} />;
+  let plugins = useMemo(
+    () =>
+      Object.values(blocks).reduce(
+        (combinedPlugins, block) => {
+          if (typeof block.getPlugins !== 'function') {
+            return combinedPlugins;
           }
-          return null;
+          return [...combinedPlugins, block.getPlugins({ blocks })];
```
jesstelford: Dependency inject the blocks into a `getPlugins()` function.

test-projects/basic/index.js
```diff
@@ -12,11 +12,11 @@ const {
   DateTime,
   Color,
   Url,
-  Content,
   Decimal,
   OEmbed,
   Unsplash,
 } = require('@keystone-alpha/fields');
+const Content = require('@keystone-alpha/field-content').default;
```
jesstelford: The outward API of the field type hasn't changed at all 👍

packages/field-content/README.md
```diff
@@ -0,0 +1,9 @@
+<!--[meta]
+section: field-types
+title: Content Field
+[meta]-->
+
+# Conent Field
```
mitchellhamilton:
```suggestion
# Content Field
```

packages/field-content/src/Block.js
```diff
@@ -1,4 +1,4 @@
-export class Block {
+export default class Block {
```
mitchellhamilton: I believe we made a decision a bit ago to not use default exports in package entrypoints. I noticed default exports in some other places too, why don't we have a single entrypoint with named exports?

packages/field-content/src/Implementation.js
mitchellhamilton: Why dependency injection over exporting the available blocks?

packages/fields/src/types/Unsplash/UnsplashBlock.js
```diff
@@ -1,7 +1,7 @@
 import pluralize from 'pluralize';
 import { importView } from '@keystone-alpha/build-field-types';
 
-import { Block } from '../../Block';
+import Block from '@keystone-alpha/field-content/Block';
```
mitchellhamilton: It feels strange to be depending on `@keystone-alpha/field-content` from `@keystone-alpha/fields`. Why move it out of fields if it's still depended on by fields. (My ideal solution would be move the block types like this out of fields)

packages/field-content/src/Block.js
timleslie: Oh yeah, strongly agree on this ☝️

packages/field-content/src/Implementation.js
jesstelford: Because they're instances which aren't instantiated until the `Content` constructor is executed.

Then later on, those instances can dynamically define their view requirements (via `.getAdminViews()`).

Why are they instances? Because blocks such as the OEmbed block take an adapter ("iFramely" or "embedly"), and it's up tot he adapter to determine what views to include. That adapter isn't known until execution time.

packages/field-content/src/Block.js
jesstelford: Oh yeah, I'll fix that up. Good catch! 👍

packages/fields/src/types/Unsplash/UnsplashBlock.js
jesstelford: Yeah, I agree - extracting the blocks out would make all of this nicer.

I think we can leave it as a half-way solution like this ☝️ for now, then later extract the Blocks out into their own packages to break that almost-circular-dependency.

packages/field-content/src/Implementation.js
mitchellhamilton: I'm a little confused, `availableBlocks` is importing the blocks, right?
https://github.com/keystonejs/keystone-5/blob/3082dce73ea272f5d7bc8b6fe008d7bbd6e145fe/packages/field-content/src/Implementation.js#L4

And those aren't instances, are they?

packages/field-content/src/Block.js
jesstelford: Fixed.

packages/field-content/src/Implementation.js
jesstelford: 🤦‍♂ You're right! And the blocks are being instantiated on demand.

I was thinking of the blocks injected into the `Node`/`Sidebar` components exported from the `Block` views.

I'll remove this injection and just export the blocks like you suggested 👍

jesstelford: Done 👍

packages/fields/src/types/Unsplash/UnsplashBlock.js
gautamsi: I think we should remove all service dependent fields form `@keystone-alpha/fields` and keep this one simplistic. Unsplash should move out to its own package like the tinyMCE based field. 

however markdown field seems basic and should remain in there.

Issue #1353, opened 2019-07-03
one relation appears many times in beforeChange hook


Olya-Yer:
in before change hook one object is represented several times 

in resolved Data 

![image](https://user-images.githubusercontent.com/33322677/60573777-690c3980-9d89-11e9-81d2-acd0c3c1a97d.png)

jesstelford:
Are you able to share your schema & the hook code so we can reproduce the issue?

Olya-Yer:
keystone.createList('Object', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    // This list is setup with a single field, 'name'.
    // The type of Text influences the types and queries available via GraphQL.
    // See the Text field docs:
    // https://v5.keystonejs.com/keystone-alpha/fields/src/types/text/
    name: { type: Text, schemaDoc: 'This is the thing you need to do', isUnique: true },
    field: { type: Relationship, ref: 'Field', many: true}
  },
  hooks: {
    beforeChange: async ({
      resolvedData,
      existingItem,
      originalInput,
      context,
      list,
    }) => {
      beforeChange({
        resolvedData,
        existingItem,
        originalInput,
        context,
        list,
      })
    }
  }
});
const beforeChange = ({
  resolvedData,
  existingItem,
  originalInput,
  context,
  list,
}) => {
  console.log('resolvedData',resolvedData)
  console.log('existingItem',existingItem)
  console.log('originalInput',originalInput)
  // console.log('context',context)
  console.log('list',list)
};

Olya-Yer:
keystone.createList('Field', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    name: { type: Text, schemaDoc: 'this is field of ypur object'},
  },
});

keystone.createList('Object', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    name: { type: Text, schemaDoc: 'This is the object/document/table', isUnique: true },
    field: { type: Relationship, ref: 'Field', many: true}
  },
  hooks: {
    beforeChange: async ({
      resolvedData,
      existingItem,
      originalInput,
      context,
      list,
    }) => {
      beforeChange({
        resolvedData,
        existingItem,
        originalInput,
        context,
        list,
      })
    }
  }
});
const beforeChange = ({
  resolvedData,
  existingItem,
  originalInput,
  context,
  list,
}) => {
  console.log('resolvedData',resolvedData)
  console.log('existingItem',existingItem)
[...]

PR #1359, opened 2019-07-04
Fix removing relationships from the Admin UI

mitchellhamilton:


packages/fields/src/types/Relationship/views/Controller.js
```diff
@@ -28,26 +28,43 @@ export default class RelationshipController extends FieldController {
...
       }
-      return this.buildRelateToManyInput(data[path], path);
+      return {
+        connect: ids.map(id => ({ id })),
+        disconnect: this.lastSavedState
+          .filter(id => {
+            return !ids.includes(id);
+          })
+          .map(id => ({ id })),
+      };
```
jesstelford: The other option here is:

```javascript
return {
  disconnectAll: true,
  connect: ids.map(id => ({ id })),
};
```

This is [the equivalent of a `set`](https://v5.keystonejs.com/keystone-alpha/fields/src/types/relationship/#overriding-a-to-many-relationship).

jesstelford: Also removes the need for storing the `lastSavedState`.

mitchellhamilton: Just tried that and it changes the order of the items which would probably be really confusing for users

jesstelford: 🤔 Hmm, it should set them in the order that they are in the array passed to the mutation. @timleslie we spoke about this the other day, and I think we landed on "it's up to the database to say what the order should be" IIRC, but Mongo should respect the order.

packages/fields/src/types/Relationship/views/Controller.js
```diff
@@ -28,26 +28,43 @@ export default class RelationshipController extends FieldController {
...
     if (many) {
-      if (!Array.isArray(data[path]) || !data[path].length) {
-        return { disconnectAll: true };
+      let ids = [];
+      if (Array.isArray(value)) {
+        ids = value.map(x => x.id);
       }
-      return this.buildRelateToManyInput(data[path], path);
+      return {
+        connect: ids.map(id => ({ id })),
```
JedWatson: Looks like we're instructing the resolver to `connect` each id present in the field in the Admin UI _each time the item is saved_, while only removing ids that are no longer present. I've realised that how this API works is a bit unclear; I expect this would create _duplicate connections_ which could explain #1353

If that's the case, we could ensure we only connect IDs that aren't present in `lastSavedState`...

We should also figure out what to do with multiple instances of the same value in a relationship field. AFAIK that's actually potentially a valid use case, and would (should?) be supported by the underlying database adapters, but at the moment react-select wouldn't allow you to select the same value more than once without some tweaking.

In the future this might be something we want to introduce an option for, like `allowDuplicates` or something... just to make the behaviour really explicit at a schema level; and then if that's not enabled, we could ensure duplicate values are stripped at the resolver level.

packages/fields/src/types/Relationship/views/Controller.js
JedWatson: Just opened an issue for this: #1362

PR #1370, opened 2019-07-08
Some FieldAdapter helpers

molomby:
Adding some helper functions to the BaseFieldAdapter (for getting the list and field references); Updating the Password type implementation

### DIFF
--- a/packages/keystone/lib/adapters/index.js
+++ b/packages/keystone/lib/adapters/index.js
@@ -159,6 +159,13 @@ class BaseFieldAdapter {
     this.isUnique = isUnique;
   }
 
+  get list() {
+    return this.getListByKey(this.listAdapter.key);
+  }
+  get field() {
+    return this.list.fieldsByPath[this.path];
+  }
+
   setupHooks() {}
 }
 
--- a/packages/fields/src/types/Password/Implementation.js
+++ b/packages/fields/src/types/Password/Implementation.js
@@ -94,18 +94,16 @@ const CommonPasswordInterface = superclass =>
     setupHooks({ addPreSaveHook }) {
       // Updates the relevant value in the item provided (by referrence)
       addPreSaveHook(async item => {
-        const list = this.getListByKey(this.listAdapter.key);
-        const field = list.fieldsByPath[this.path];
-        const plaintext = item[field.path];
+        const plaintext = item[this.path];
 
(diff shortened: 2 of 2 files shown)


packages/keystone/lib/adapters/index.js
```diff
@@ -159,6 +159,13 @@ class BaseFieldAdapter {
     this.isUnique = isUnique;
   }
 
+  get list() {
+    return this.getListByKey(this.listAdapter.key);
+  }
+  get field() {
+    return this.list.fieldsByPath[this.path];
+  }
```
jesstelford: Can we make these `getList()` and `getField()` methods instead of getters? I'm not a fan of getters in general, but `list` and `field` in particular are almost impossible to grep in this codebase without a lot of noise.

molomby: Hmm.. but this is exactly what getters are good at though and they _meant_ to look like fields from the outside. `this.field.blah` is a lot more natural than `this.getField.blah` don't you think? The latter would confuse me -- looks like a typo, should be `this.getField().blah`.

It's always going to be hard to grep for things with common names, I don't think that makes the names wrong in this case.

Can we get a third opinion? @JedWatson @timleslie

jesstelford: > should be `this.getField().blah`

Yes, that's what I meant by _"`getFields()` methods_".

> It's always going to be hard to grep for things with common names, I don't think that makes the names wrong in this case.

It's getting worse and worse in the Keystone codebase. It's pain I'm feeling daily, so want to avoid it at every possible turn. This is one of them.

If I `rg` the codebase for `field` today, there are 3,328 results. `getField()` however; 61 results. Similarly for `list`; 2,742 vs 88.

gautamsi:
I will keep pushing for TypeScript, it will be more manageable with grep, rename, jumping to source and other actions. 

Now that we do "build" step for many things, it would be more natural. 

We do have typescript as first class citizen in React, Next and WebPack. Running Dev instance without the build is no longer a problem with Typescript.

jesstelford:
To avoid a TypeScript discussion can of worms going 💥 in this PR, I've created a new issue to track discussion: #1373

molomby:
Closing in favour of #1385 🦑

PR #1371, opened 2019-07-08
Update Slate for Content Editor

jesstelford:
I ran into bugs when trying to use Slate in a project, and realised we're quite out of date in terms of Slate's API, so updated here and fixed a couple issues on top.


packages/arch/packages/tooltip/src/index.js
```diff
@@ -40,7 +40,7 @@ let TooltipPositioner = (props: PositionerProps) => {
     <Popper
       referenceElement={props.targetNode}
       placement={props.placement}
-      modifiers={{ preventOverflow: { enabled: false } }}
+      modifiers={{ hide: { enabled: false }, preventOverflow: { enabled: false } }}
```
jesstelford: Without this, I get lots of `'preventOverflow' modifier is required by 'hide' modifier in order to work, be sure to include it before 'hide'!` warnings from `popper.js`.

packages/field-content/src/views/editor/constants.js
```diff
@@ -7,11 +7,7 @@ export const initialValue = {
         nodes: [
           {
             object: 'text',
-            leaves: [
-              {
-                text: '',
-              },
-            ],
+            text: '',
```
jesstelford: Latest slate doesn't have `leaves` on `text` nodes.

packages/field-content/src/views/editor/index.js
```diff
@@ -44,12 +44,12 @@ function Stories({ value: editorState, onChange, blocks, className }) {
           if (typeof block.getPlugins !== 'function') {
             return combinedPlugins;
           }
-          return [...combinedPlugins, block.getPlugins({ blocks })];
+          return [...combinedPlugins, ...block.getPlugins({ blocks })];
```
jesstelford: Fixing a bug introduced in #1358 where we weren't correctly flattening the plugins returned.

packages/field-content/src/views/editor/index.js
```diff
@@ -44,12 +44,12 @@ function Stories({ value: editorState, onChange, blocks, className }) {
...
         },
         [
           ...markPlugins,
           {
-            renderNode(props) {
+            renderBlock(props) {
```
jesstelford: API change in `slate-react`.

packages/fields/src/types/CloudinaryImage/views/blocks/single-image.js
```diff
@@ -188,19 +188,24 @@ export function serialize({ node, blocks }) {
...
-      create: {
-        image: file,
-      },
+  const mutations = joinIds && joinIds.length ? ({
+    connect: { id: joinIds[0] },
+  }) : ({
+    create: {
+      image: file,
     },
+  });
```
jesstelford: Correctly `connect` to existing items rather than creating new on every save.

packages/field-content/src/views/editor/index.js
mitchellhamilton: I'm pretty sure this will break with links since they're not Slate blocks, they're inlines.

jesstelford: Excellent catch! I've had to add a `renderInline` as well as a `renderBlock` to get the original behaviour back. All tested and working now 👍

PR #1372, opened 2019-07-08
Add `Uuid` field type

molomby:
Adding `Uuid` type to the core fields package. Mongoose support isn't terrific; we use Strings instead of a binary type. Ideally we can fix this later.

Pros: Field type has tests _and_ docs 🥳


packages/fields/src/types/Uuid/Implementation.js
```diff
@@ -0,0 +1,105 @@
...
+  /^[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}$/.test(a);
+
+// TODO: UUIDs _should_ be stored in Mongo using binary subtype 0x04 but strings are easier; see README.md
+export class MongoUuidInterface extends MongooseFieldAdapter {
+  addToMongooseSchema(schema, mongoose) {
+    const schemaOptions = {
+      type: mongoose.Schema.Types.String,
+      validate: {
+        validator: this.buildValidator(validator),
+        message: '{VALUE} is not a valid UUID',
```
jesstelford:
```suggestion
        message: '{VALUE} is not a valid UUID. Must be 8-4-4-4-12 hex format',
```

PR #1383, opened 2019-07-09
Adding db level config for knex adapter

molomby:
Adding `knexOptions` to the KnexFieldAdapter to support DB-level config for nullability (`isNotNullable`) and defaults (`defaultTo`). These are defaulted based on their KS-level equivalents (`isRequired` and `defaultValue`).

Also updated existing Knex field adapters to use new db-level config options. This switches `isRequired` flag for `isNotNullable` and adds db-level defaulting to all fields (for which it makes sense).

I'm expecting some of the naming to be a little spicy for some 🌶

### DIFF
--- a/.changeset/early-tools-help/changes.json
+++ b/.changeset/early-tools-help/changes.json
@@ -1,0 +1,8 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/adapter-knex", "type": "minor" },
+    { "name": "@keystone-alpha/fields", "type": "minor" },
+    { "name": "@keystone-alpha/keystone", "type": "patch" }
+  ],
+  "dependents": []
+}
--- a/.changeset/early-tools-help/changes.md
+++ b/.changeset/early-tools-help/changes.md
@@ -1,0 +1,1 @@
+Adding `knexOptions` to the KnexFieldAdapter to support DB-level config for nullability (`isNotNullable`) and defaults (`defaultTo`)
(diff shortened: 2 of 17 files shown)


molomby:
This is now dependant on #1385; it should be merged first

molomby:
**Usage**

This..

```js
keystone.createList('User', {
  fields: {
    name: { type: Text },

    // Here we're setting a default in KS but preventing the adapter from also using
    // the value as column default
    email: {
      type: Text,
      defaultValue: 'noone@example.com',
      knexOptions: { defaultTo: null }
    },

    // Both isRequired and the defaultValue will carry though to the DB as `not null` 
    // and `default false`
    isEnabled: {
      type: Checkbox,
      isRequired: true,
      defaultValue: true,
    },

    // This field is required according to KS but we'll keep the field nullable 
    // (eg. say there's a legacy integration that adds users to the DB directly)
    password: {
      type: Password,
      isRequired: true,
      knexOptions: { isNotNullable: false }
    },

    // This field isn't required by KS but the knexOptions mean it will always 
    // be populated in the DB
    registeredAt: {
      type: DateTime,
      isRequired: false,
      knexOptions: {
        defaultTo: knex => knex.raw('localtimestamp'),
        isNotNullable: true
      }
    },

    // This field is being defaulted using a DB-side function
    // defaultTo needs to be a function in this case so we can get the knex referrence
    externalId: {
      type: Uuid,
      knexOptions: { defaultTo: knex => knex.raw('gen_random_uuid()') }
    },
  },
});
```

Produces a schema like this...

```sql
CREATE TABLE keystone."User" (
    id integer DEFAULT nextval('keystone."User_id_seq"'::regclass) PRIMARY KEY,
    name text,
    email text,
    "isEnabled" boolean NOT NULL DEFAULT true,
    password character varying(60),
    "registeredAt_utc" timestamp without time zone NOT NULL DEFAULT LOCALTIMESTAMP,
    "registeredAt_offset" text NOT NULL,
    "externalId" uuid DEFAULT gen_random_uuid()
);
```

PR #1385, opened 2019-07-10
Adding a `field` reference to field adapter instances

molomby:
This tries to solve the same problem as #1370 but, rather than getters that dereference the field at run time, we just pass the `field` instance in to the `fieldAdapter` constructor.

Easiest to see what's going on if you look at the commits in series.


molomby:
No changeset; this isn't a functional update

molomby:
Ah nice catch, thanks @jesstelford

Issue #1384, opened 2019-07-09
Lists don't leverage field types for their primary keys


molomby:
The `id` column for lists will always be special in a number of ways but they still have a tonne in common with normal field types; eg. all the GraphQL filtering, etc. should be reused.

The blocker at the moment is the lack of field types that are actually appropriate for use as primary keys. See #317 for current work to resolve this.

Issue #317, opened 2018-09-14
ID field type(s) for storing Mongo ObjectIds, UUID, etc.


molomby:
Sometime you need to store an ID but relationship fields (as currently implemented) are inappropriate. I've seen this come up with polymorphic relationships (eg. `relatedId`, `relatedType`) and when versioning records (eg. `lineageId`). Also if you're using more than one DB (eg. @dominikwilkowski wants to use Mongo and pgSQL for Jerrick so there'll be some cross-linking).

It's possible these use cases could be addressed with updates to the relationship field type and the versioning functionality. Alternatively (or in the meantime) a generic ID type might be useful. I imagine it could be configured as either an Mongo ObjectIds or UUID (or alternatively maybe we should have separate MongoObjectId and UUID types).

Prior art: KS Next has [Uuid](https://github.com/Thinkmill/groki-platform/tree/master/common/keystoneNext/fieldTypes/Uuid) and [UuidArray](https://github.com/Thinkmill/groki-platform/tree/master/common/keystoneNext/fieldTypes/UuidArray) field types.

molomby:
Uuid added in #1372. Plans to also add a non-core `MongoID` field type so will leave this open.

PR #1390, opened 2019-07-10
Adding `MongoId` field type for standalone (ie. non-primay key) IDs

molomby:
Adding the `MongoId` field type -- a non-core, standalone, Mongo ObjectID type. Supports the Mongoose and Knex adapters.

This type can be used to store arbitrary Mongo ObjectIds in a list (in addition to your primary key, if you're on Mongo).

Relates to #317, follows on from #1372 and lays more of the groundwork for fixing #1384.

### DIFF
--- a/package.json
+++ b/package.json
@@ -306,10 +306,11 @@
   },
   "field-types": {
     "packages": [
-      "packages/fields-wysiwyg-tinymce",
+      "packages/field-content",
       "packages/fields",
       "packages/fields-markdown",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -306,10 +306,11 @@
   },
   "field-types": {
     "packages": [
-      "packages/fields-wysiwyg-tinymce",
```
molomby: Correcting to alpha order

PR #1392, opened 2019-07-11
Adding `AutoIncrement` field type

molomby:
This is a pretty weird field type in a lot of ways. The README has details.

Usage looks like...

```js
keystone.createList('Todo', {
  fields: {
    name: { type: Text },
    todoInc: { type: AutoIncrement, knexOptions: { isPrimaryKey: false } },
  },
});
```

It only supports Knex and, at the moment, really only suports Postgres for non-primary keys.

This completes #317 and paves the way for #1384.


molomby:
Oh there's also a v. minor fix to the `KnexFieldAdapter` constructor so it stores additional `knexOptions`

PR #1396, opened 2019-07-15
Switching lists to use standard field types for primary keys (instead of weird pseudo-field)

molomby:
Until now list adapters automatically added a semi-magical pseudo-field that surfaced in Keystone as `id`. Fields and filters for this pseudo-field were added to the GraphQL types and filters as exceptional cases. This PR replaces that approach by leveraging the existent functionality in of generic field types.

Database adapters can now (optionally) provide a `getDefaultPrimaryKeyConfig()` function that returns a their suggested config for the `id` field. (The Knex and Mongoose adapters defer this responsibility decision to the types themselves.)

The intention is to give solution builders sensible defaults but with multiple opportunities to intervene:

```js
const { Keystone } = require('@keystone-alpha/keystone');
const { MongoId } = require('@keystone-alpha/fields-mongoId');
const { Text } = require('@keystone-alpha/fields');

const adapter = new MongooseAdapter();
const keystone = new Keystone({ name: 'Keystone To-Do List', adapter });

// Let Keystone add a default id field automatically
keystone.createList('Todo', {
  fields: { name: { type: Text } },
});

// Ask the adapter for it's default primary key config
keystone.createList('Todo', {
  fields: {
    id: adapter.getDefaultPrimaryKeyConfig(),
    name: { type: Text },
  },
});

// Pick the field type yourself and use the types recommended primary key config
keystone.createList('Todo', {
  fields: {
    id: MongoId.primaryKeyDefaults[adapter.name].getConfig(),
    name: { type: Text },
  },
});

// Build the field config manually using whatever field type and config you want
keystone.createList('Todo', {
  fields: {
    id: { type: MongoId },
    name: { type: Text },
  },
});
```

These ^^^ configs result in identical lists.

### DIFF
--- a/.changeset/six-pets-dress/changes.json
+++ b/.changeset/six-pets-dress/changes.json
@@ -1,0 +1,130 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/adapter-knex", "type": "major" },
+    { "name": "@keystone-alpha/adapter-mongoose", "type": "major" },
+    { "name": "@keystone-alpha/fields-auto-increment", "type": "major" },
+    { "name": "@keystone-alpha/fields-mongoId", "type": "major" },
+    { "name": "@keystone-alpha/fields", "type": "major" },
+    { "name": "@keystone-alpha/app-admin-ui", "type": "minor" },
+    { "name": "@keystone-alpha/keystone", "type": "minor" },
+    { "name": "@keystone-alpha/test-utils", "type": "minor" }
+  ],
+  "dependents": [
+    {
+      "name": "@keystone-alpha/api-tests",
+      "type": "patch",
+      "dependencies": [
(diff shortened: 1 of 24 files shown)


molomby:
I believe we need fixes for these issues before this can be merged:

- #1397: Access control not respected by Admin UI
- #1398: Create item form submits values for blank/unedited fields
- #1399: Field values aren't correctly reloaded from the DB after create

jesstelford:
The `id` field can be excluded from the AdminUI just as it's excluded from the `Create` and `Update` input types in the GraphQL layer, here:

https://github.com/keystonejs/keystone-5/blob/master/packages/app-admin-ui/client/classes/List.js#L17-L20

molomby:
I believe the issues blocking this have been fixed or (in the case of #1397 and #1398) worked around.

PR #1411, opened 2019-07-17
Upgrade babel to 7.5.4

timleslie:
https://github.com/babel/babel/releases


timleslie:
Hmm, can't immediately reproduce the failure locally. Will have to dig deeper.

jesstelford:
Looks like the fix might be mentioned in the error:

```
you can use
the @babel/plugin-syntax-dynamic-import plugin and let your
bundler handle dynamic imports.
```

timleslie:
Tried this, didn't work, will try again some other time.

PR #1414, opened 2019-07-17
Adding non-core `DateTimeUtc` field type

molomby:
Like the core type but simpler and, most of the time, what we actually want. 

Insists on being given offsets anytime it ingests a value but stores only UTC so contributes a single field to the schema:

* A `timestamp` on Knex (`timestamptz` on Postgres)
   - Precision limited to match JS (milliseconds); avoids silent rounding in JS world
* A `Date` on Mongoose

Supports:

* Field config `isIndexed`, `isUnique` and `defaultValue`
* Knex options `defaultTo` and `isNotNullable`
* The same filtering as the core `DateTime` type (in, equality, "ordering")
* Ordering

Values are exposed as `Date` objects in JS and plain `String` scalars in GraphQL. 

I'd argue this should actually be the default/core implementation (but not today!).

### DIFF
--- a/.changeset/two-pets-doubt/changes.json
+++ b/.changeset/two-pets-doubt/changes.json
@@ -1,0 +1,4 @@
+{
+  "releases": [{ "name": "@keystone-alpha/fields-datetime-utc", "type": "major" }],
+  "dependents": []
+}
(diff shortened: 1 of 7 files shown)


gautamsi:
was this necessary to be a separate field? 
For me this could easily be an option on the `DateTime` field rather than brand new field.

PR #1421, opened 2019-07-18
Ensure resolveInput for list receives result from fields

jesstelford:


packages/keystone/lib/List/index.js
```diff
@@ -447,7 +449,9 @@ module.exports = class List {
         type ${this.gqlNames.authenticateOutputName} {
           """ Used to make subsequent authenticated requests by setting this token in a header: 'Authorization: Bearer <token>'. """
           token: String
-          """ Retreive information on the newly authenticated ${this.gqlNames.outputTypeName} here. """
+          """ Retreive information on the newly authenticated ${
+            this.gqlNames.outputTypeName
+          } here. """
```
jesstelford: No idea why prettier decided this needed to change :/

jesstelford: Heh, it didn't... I had an older version of `prettier` installed.

PR #1422, opened 2019-07-18
Fixing field config application for knex field adapters

molomby:
Fixing a few problems around field and field adapter config:

* Falsy defaults weren't being applied to the DB (ie. giving a `Checkbox` a DB-level default of `false`)
* Explicitly setting `isNotNullable` on a singular relationship field was temporarily disabled; forgot to re-enable it after fixing logic around the how the setting was defaulted

### DIFF
--- a/packages/fields/src/types/Checkbox/Implementation.js
+++ b/packages/fields/src/types/Checkbox/Implementation.js
@@ -25,20 +25,16 @@ export class Checkbox extends Implementation {
   }
 }
 
-const CommonCheckboxInterface = superclass =>
-  class extends superclass {
-    getQueryConditions(dbPath) {
-      return this.equalityConditions(dbPath);
-    }
-  };
-
-export class MongoCheckboxInterface extends CommonCheckboxInterface(MongooseFieldAdapter) {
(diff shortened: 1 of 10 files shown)


packages/fields/src/types/Checkbox/Implementation.js
```diff
@@ -25,20 +25,16 @@ export class Checkbox extends Implementation {
...
 
-const CommonCheckboxInterface = superclass =>
-  class extends superclass {
-    getQueryConditions(dbPath) {
-      return this.equalityConditions(dbPath);
-    }
-  };
-
-export class MongoCheckboxInterface extends CommonCheckboxInterface(MongooseFieldAdapter) {
+export class MongoCheckboxInterface extends MongooseFieldAdapter {
```
jesstelford: Why the class refactor here?

molomby: Didn't seem worth the complexity to avoid repeating 3 lines. The new version achieves the same thing, is easier to understand, creates less variables/classes and is shorter/less code.

I'm not that fussed if you want to revert it though.

PR #1428, opened 2019-07-19
Update Contributing guide & Changeset template

jesstelford:


CONTRIBUTING.md
```diff
@@ -61,15 +61,16 @@ The `bolt version-packages` command will generate a release commit, which will b
...
-bolt version-packages
-bolt format
-git add .
-git commit -m "Run version-packages"
+git checkout master && \
+git pull && \
+git branch -D temp-release-branch && \
+git checkout -b temp-release-branch && \
+bolt fresh && \
+bolt build && \
```
jesstelford: NOTE: I added a `bolt build` command here to catch any issues that might otherwise only be caught in the publish phase (like in #1427)

PR #1438, opened 2019-07-23
Ensure searchUnsplash mutation returns results with a non-null id

jesstelford:
Previously, this was returning `id: null`, which would confuse Apollo's cache who expected `id`s to always have a value if included in the results. Ultimately, apollo would key the cache by the `id`, but that `id` is `null`, so it would only store a single item 5 times (for example).

This fix ensures that the `searchUnsplash` query always injects an id.

### DIFF
--- a/packages/fields/src/types/Unsplash/Implementation.js
+++ b/packages/fields/src/types/Unsplash/Implementation.js
@@ -14,8 +14,9 @@ const defaultTransforms = {
   fit: 'max',
 };
 
-function transformUserFromApiToKs5(user) {
+function transformUserFromApiToKs5(user, { includeId = false } = {}) {
   return {
+    ...(includeId && { id: user.id }),
     unsplpashId: user.id,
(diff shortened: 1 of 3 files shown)


packages/fields/src/types/Unsplash/Implementation.js
```diff
@@ -14,8 +14,9 @@ const defaultTransforms = {
   fit: 'max',
 };
 
-function transformUserFromApiToKs5(user) {
+function transformUserFromApiToKs5(user, { includeId = false } = {}) {
   return {
+    ...(includeId && { id: user.id }),
     unsplpashId: user.id,
```
mitchellhamilton: I know this isn’t changed in this PR but is `unsplpashId` a typo?

jesstelford: Hah! It is! I'll fix it in this PR, thanks for catching that 👍

PR #1439, opened 2019-07-23
Update knex readme

MadeByMike:
The warning is no longer relevant. `./build-test-db.sh` is not available when installed from node modules.


packages/adapter-knex/README.md
```diff
@@ -5,21 +5,12 @@ title: Database Adapter - Knex
...
 
-Before running Keystone, you must set up a database, a schema, and a user.
-Assuming you're on MacOS and have Postgres installed the `build-test-db.sh` does this for you:
-
-```sh
-./packages/adapter-knex/build-test-db.sh
-```
-
-Otherwise, you can run the steps manually:
+Before running Keystone, you must set up a database, a schema, and a user. You can run the following commands:
```
timleslie: Can we keep a reference to the `build-test-db.sh` script and include a note that it's only for development within the mono-repo itself?

MadeByMike: That's not an awesome developer experience. We could link to the script in the repo for people to download and run?

MadeByMike:
Out of date with: https://github.com/keystonejs/keystone-5/pull/1468

Issue #1377, opened 2019-07-09
Running `yarn install` when using `create-keystone-app ` causes `Error: spawn keystone ENOENT`


petronbot:
## What I did

Followed the "Quick Start" instructions (https://v5.keystonejs.com/quick-start/) to get up and running with a fresh Keystone instance:

```
yarn create keystone-app my-app
cd my-app
yarn start
```

Woohoo! Keystone is running and I'm happy, until I find myself clearing my node modules folder:

```
rm -rf ./node_modules
yarn install
yarn start
```

## What I expected to see

A happy instance of Keystone starting up all its services:

```
$ cross-env NODE_ENV=development DISABLE_LOGGING=true keystone dev
ℹ Command: keystone dev
✔ Validated project entry file ./index.js
✔ Keystone server listening on port 3000
✔ Initialised Keystone instance
✔ Connected to database
✔ Keystone instance is ready 🚀
🔗 Keystone Admin UI:	http://localhost:3000/admin
🔗 GraphQL Playground:	http://localhost:3000/admin/graphiql
🔗 GraphQL API:		http://localhost:3000/admin/api
```

## What I saw

I get Node’s `ENOENT` error and can’t start the app. There’s no `keystone` in `./node_modules/bin`, even after running `yarn install`.

```
$ cross-env NODE_ENV=development DISABLE_LOGGING=true keystone dev
events.js:174
      throw er; // Unhandled 'error' event
      ^
Error: spawn keystone ENOENT`
```

During this `yarn install` step I see Yarn throwing some warnings, most of which I ignore, except this one:

```
warning Pattern ["@keystone-alpha/keystone@latest"] is trying to unpack in the same destination "/Users/thinkmill/Library/Caches/Yarn/v4/npm-@keystone-alpha-keystone-7.0.2-cc60ffc0cc65735f413f65131679c1679f4ae5d2/node_modules/@keystone-alpha/keystone" as pattern ["@keystone-alpha/keystone@^7.0.0","@keystone-alpha/keystone@^7.0.0"]. This could result in non-deterministic behavior, skipping.
```

If I run `npm install` I don't see any such warning and the app fires up quite happily.

Also: in `package.json` if I change `"@keystone-alpha/keystone": "latest,` to `"@keystone-alpha/keystone": "7.0.2",` the `yarn install` works just fine.

## System

* Yarn: `v1.15.2`
* Node: `v10.15.3`
* `@keystone-alpha/keystone`: `v7.0.2`
[...]

PR #1442, opened 2019-07-24
Expand create keystone app CLI

MadeByMike:
Updates the create-keystone-app cli.

1. Add the ability to create an app in an existing empty directory
1. Add the ability to select from available templates
1. Added the blank template

ks-create-prompt


gautamsi:
I think the plan is to get away from template directory and use `create-next-app` like `--example` switch

MadeByMike:
We can still do that. Is there an issue that describes how it would work? 

I don't think any future changes to create-keystone-app would need to block this. It can evolve :)

mitchellhamilton:
IMO, we shouldn't add more templates like this, we should rip out the templates and make create-keystone-app essentially clone the projects from `demo-projects` since it will solve #1377 and mean we won't have to maintain even more projects.

MadeByMike:
Generally I like the idea. Currently I don't think the demo projects are 100% what you'd want from a CLI but I think they could be aligned without too much work. I can't see someone wanting scaffold out the meetup demo as a starting point. It's really to specific. It works as a learning exercise but not so much a scaffolding project.

Having said that, I've talked with @jesstelford about the need for more a few more granular demos, for example maybe one with simple a detached React front-end that you'd deploy separately to your Keystone application. And maybe a Gatsby one as well as a few others.... what's most important here is good examples of how to structure these projects rather than the content. The current todo example is the opposite of this. To be fair warns people of this in the code and is a good into to lists etc.... Maybe it should be a pure JS example and not React.

On this note, it would also be nice if the CLI templates\demo guided not towards the desired application, ie, blog, todo, meetup, but towards the desired tech. Eg. Help me choose my stack, Mongo or Postgres + React, Next, Gatsby etc... often these are the decisions I have in mind when I start and don't mind if the example is bare bones (or virtually blank). I need just enough to show me how to work with it. I care more that webpack, babel and npm scripts etc... are configured for me.

Perhaps the end goal of the CLI does need more thought.

mitchellhamilton:
Makes sense 👍 

The thing that I care about isn’t what projects that exist right now that we pull from, the thing I care about is not doing the templating stuff because it’s causing issues like that package manager thing and we can’t test the templates locally.

So I think the answer might be create a new directory for examples with tech stack things like you mentioned.

MadeByMike:
Sounds good @mitchellhamilton. We could also selectively include some demos as options in the CLI. That would add a manual step to include it somehow but it's a tiny bit of work and never needs updating. 

The only challenge here is we're creating a dependency the demos repo and the CLI package. Is there any way to manage this so that editing the demos can't break the CLI. One solution I guess it to have a CI test for each of them?

mitchellhamilton:
I think having a quick smoke test that starts the app should be fine but also, I don’t think next has tests like that and they work fine with all their examples.

MadeByMike:
@mitchellhamilton I took your feedback onboard and once we land #1468 I've got some plans for a new method with the CLI I think you might like. Going to close this for now.

PR #1443, opened 2019-07-25
Add @keystone-alpha to package.json 

MadeByMike:
Updates the `package.json` in the template to add `@keystone-alpha` and hopefully fixes: #1377

### DIFF
--- a/packages/create-keystone-app/templates/todo/package.json.ejs
+++ b/packages/create-keystone-app/templates/todo/package.json.ejs
@@ -6,6 +6,7 @@
   },
   "private": true,
   "dependencies": {
+    "@keystone-alpha": "latest",
     "@keystone-alpha/adapter-mongoose": "latest",
     "@keystone-alpha/app-admin-ui": "latest",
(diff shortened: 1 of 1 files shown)


MadeByMike:
> I don't think this will work, `@keystone-alpha` isn't a package that exists, if someone did an install, it should fail saying that `@keystone-alpha` doesn't exist.

Interesting, so how does the KS package end up in `node_modules`?

mitchellhamilton:
There's isn't such thing as "the KS package", there are lots of packages, like `@keystone-alpha/keystone` and others which are what's installed.

MadeByMike:
Gotcha! So it's `@keystone-alpha/keystone` that's missing?

MadeByMike:
Nope, it's there. Any thoughts Mitchell?

mitchellhamilton:
There isn’t any missing packages, the problem is that they all have their version range set to latest which package managers seem to complain about so hence why I think we should do the thing mentioned in the other PR https://github.com/keystonejs/keystone-5/pull/1442#issuecomment-514863096

PR #1446, opened 2019-07-27
Slug field

jesstelford:
So, err... this one turned out to be _huge_.

Who'd have thought that a "simple slug" would be so complicated?

But anyway, it's pretty comprehensive, and covers most of the cases I could think of with one caveat which I've thoroughly documented in the `README.md`.

So, for all the deets on this new field, [check out the `README.md`](https://github.com/keystonejs/keystone-5/blob/slug-field/packages/fields/src/types/Slug/README.md).

I've added 29 unit tests to cover the various scenarios of creating / updating slugs.

Please let me know if I missed any!

I also had to make a tweak to how the `@keystone-alpha/keystone` package's `List` class handles `resolveInput()`: It used to only execute `resolveInput()` on fields where data was passed in to the mutation (or, for `create` mutations, had a default value). It now runs `resolveInput()` over _all_ fields in the list regardless of if they have received data or a default value. This enables setting a value for a field during an update when no data was received, which was previously impossible.

Finally, I updated the `blog` demo so it uses next.js 9's dynamic routes, and those routes are based on the new `Slug` field ❤️


jesstelford:
Whoops, toys test failures are because our mocks don't have a `resolveInput` method which is now being executed. I'll update them when I'm at my laptop again.

packages/fields/src/types/Slug/Implementation.js
```diff
@@ -0,0 +1,258 @@
...
+
+    if (from && generate) {
+      throw new Error(
+        `Only one of 'from' or 'generate' can be supplied as an option to the Slug field on ${listAndFieldPath}.`
+      );
+    }
+
+    if (from) {
+      if (typeof from !== 'string') {
+        if (typeof from === 'function') {
```
MadeByMike: I like this use of nested if over `&&` :)

jesstelford: Honestly, without it, I was going around in loops in my head trying to keep a handle on all the different branches!

jesstelford:
The failing test is due to a database not being dropped correctly between test runs. I'm not sure how to fix it - it seems there's a difference between the knex adapter and the mongo adapter where the mongo adapter will create a new database on every connection, but the knex one will re-use a single _Schema_ (database?) and only drop it _sometimes_.

jesstelford:
Ok, I think I've fixed it by forcing unique, random list names for that set of tests. This is not ideal, but it works, and tests things how I expect them to be tested (in isolation).

PR #1450, opened 2019-07-29
Remove incorrect Cell view config from Slug field

jesstelford:


packages/fields/src/types/Slug/index.js
```diff
@@ -8,7 +8,6 @@ const Slug = {
     Controller: importView('../Text/views/Controller'),
     Field: importView('../Text/views/Field'),
     Filter: importView('../Text/views/Filter'),
-    Cell: importView('../Text/views/Cell'),
```
jesstelford: This file doesn't actually exist, it just relies on the default view which is to render the value as text (which is exactly what we want)

Issue #1373, opened 2019-07-08
Should we migrate to TypeScript?


jesstelford:
**Pros**

- TypeScript's strict types can catch a small percentage of type-related bugs
- Code navigation tooling is very mature, especially in VSCode

**Cons**

- Requires a separate compile step (but can integrate nicely with Babel)
- There is [a cost to adding TypeScript](https://medium.com/javascript-scene/the-typescript-tax-132ff4cb175b).
- All core developers have to learn a new language (TypeScript)
- There are a lot of [issues](https://mobile.twitter.com/phaser_/status/1126547233162387456) [with](https://mobile.twitter.com/mpjme/status/1135660290048217094) [TypeScript](https://mobile.twitter.com/sebmck/status/1145919537071525889).
- Opting out of TypeScript can be very hard

**Let's vote!**

Click 👍 or 👎 on this issue to cast your vote for migrating to TypeScript

gautamsi:
[Why TypeScript is the best way to write Front-end in 2019](https://medium.com/@jtomaszewski/why-typescript-is-the-best-way-to-write-front-end-in-2019-feb855f9b164)

Some more encouragement for using TypeScript (especially when you have trouble):

* you can ignore certain errors with `// @ts-ignore`  [in `.ts` files](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#suppress-errors-in-ts-files-using--ts-ignore-comments)
* You do not have to invest in _**all or nothing**_ approach when moving to TypeScript, [you can roll gradually](https://gist.github.com/OliverJAsh/13b5410843d6bee692927897f0c84197#how-we-gradually-migrated-to-typescript-at-unsplash). 
* You dont have to convert the code to use TypeScript, it can statically check `.js` files and infer types. with jsdoc in place (properly) it will be even more useful. (VS Code features)

It does integrate well with [Babel 7
](https://babeljs.io/docs/en/babel-preset-typescript)

a two part blog from MS on how to smoothly migrate to Typescript: 
[How to Upgrade to TypeScript Without Anybody Noticing](https://devblogs.microsoft.com/typescript/how-to-upgrade-to-typescript-without-anybody-noticing-part-1/)


On Opting out:  You can be careful when you start and use it in least invasive manner, when you have more confidence you can abandon the idea of opting out :) 

[...]

PR #1454, opened 2019-07-30
[POC] Migrate (Front End React) to TypeScript

gautamsi:
#### Goal is to convert to TypeScript and adapt to this without having major issues.

_my plan is to convert as much front end code to TypeScript first, as most packages are either built as part of npm post install hook, or they are loaded using babel flow preset/config. this way there will be no change in developer flow. once it is acceptable, natural next step would be to touch back end code._

#### 1st commit: https://github.com/keystonejs/keystone-5/commit/99f2376dca7953ee46d6016c7a43454a94e00ee9 : 

- Renames `admin-ui/client/**/*.js` files to `.ts` and `.tsx` (if there is react/jsx code, `.tsx` otherwise `.ts`)
- convert `Flow` specific type information to `TypeScript` with help of `babel-plugin-flow-to-typescript` (there were not many changes so could be done by hand)
- updates `webpack` to load these files using existing `babel-loader` config.
- checks if run works fine for demo projects (all work fine with `admin-ui`)
>could not find any test file to convert and run test for this

#### Next Steps

- [x] fix initial typing errors 
~- [ ] Add type information to `admin-ui` to make is really using typescript~ - Planning to go by incremental approach and add Type information (return type and preferred types when change of code needed. 
- [ ] more packages to be converted - planning to have this as separate isue to list which all package needed conversion and converted.


gautamsi:
this is all in approach where we convert all fiels to typescript and use `any` type and `// @ts-ignore` for suppressing errors. When this file changes or someone has time, they can improve the typing.

using `type $TSFixMe = any;` type alias for easy search and fixup with strict checking. 

in commit https://github.com/keystonejs/keystone-5/pull/1454/commits/c8bf12ba1fa2b10c4dfbccc3871e9d7ccdc5ecfd I had to fix the snapshot, I am not sure if this one is critical to have double quote form typescript output.

gautamsi:
@JedWatson ☝️

MadeByMike:
@gautamsi I think we're still in the process of discussing the merits of moving to TypeScript or Not. That's not to say we won't at some point but I don't think we're ready to accept PRs on this yet. It needs more unpicking.

gautamsi:
This PR is just for review and feedback, this also showcases how little I had to change the admin-ui to make is typescript ready.

MadeByMike:
Sadly closing this for now since we are not ready. The discussion will continue in issue #1373

PR #1458, opened 2019-08-02
Update mongo adapter docs

MadeByMike:
Updates the mongo adapter docs to correct old example


MadeByMike:
This will be out of date with: https://github.com/keystonejs/keystone-5/pull/1468

PR #1464, opened 2019-08-06
Tweaks to make tests run cleanly

jesstelford:
Notes inline.

### DIFF
--- a/packages/arch/packages/drawer/src/index.js
+++ b/packages/arch/packages/drawer/src/index.js
@@ -126,6 +126,7 @@ type Props = {
   attachTo: HTMLElement,
   children: Node,
   closeOnBlanketClick: boolean,
+  isOpen: boolean,
   component: ComponentType<*> | string,
   footer?: Element<*>,
   heading?: string,
@@ -167,6 +168,7 @@ let ModalDialog = memo<Props>(function ModalDialog({
   width,
   onKeyDown,
   transitionState,
+  isOpen,
 }) {
   let stackIndex = useStackIndex(
     transitionState === 'entered' || transitionState === 'entering',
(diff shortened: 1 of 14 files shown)


packages/arch/packages/drawer/src/index.js
```diff
@@ -181,11 +183,13 @@ let ModalDialog = memo<Props>(function ModalDialog({
...
-        onClick={closeOnBlanketClick ? onClose : undefined}
-        isTinted
-      />
+      {isOpen ? (
+        <Blanket
+          style={fade(transitionState)}
+          onClick={closeOnBlanketClick ? onClose : undefined}
+          isTinted
+        />
+      ) : null}
```
jesstelford: Without this check, Cypress would _sometimes_ complain about a button not being clickable because this element was still on the page.

The `<Blanket>` element is used to capture clicks outside the modal to then close that modal. When `isClosed === true`, that functionality is no longer needed, so this `<Blanket>` should be removed.

packages/fields/src/types/CloudinaryImage/ImageBlock.js
```diff
@@ -92,7 +92,10 @@ export class ImageBlock extends Block {
 
   getMutationOperationResults({ context }) {
     return {
-      [this.path]: context._blockMeta[this.joinList][this.path],
+      [this.path]:
+        context._blockMeta &&
+        context._blockMeta[this.joinList] &&
+        context._blockMeta[this.joinList][this.path],
```
jesstelford: In an upcoming PR, it's possible that this data is not hydrated (eg; creating a content value which has no cloudinary images), so this check is more thorough.

packages/fields/src/types/OEmbed/views/Field.js
```diff
@@ -41,15 +41,15 @@ const PlaceholderPreview = ({ originalUrl, fieldPath }) => (
...
+export default class OEmbedField extends Component {
   onChange = event => {
     this.props.onChange({
       originalUrl: event.target.value,
     });
   };
 
   render() {
-    const { autoFocus, field, value, savedValue = {}, errors } = this.props;
+    const { autoFocus, field, value = {}, savedValue = {}, errors } = this.props;
```
jesstelford: In an upcoming PR, the `value` passed through here could be `undefined`, so we default it to an empty object (the same as we did for `savedValue`).

PR #1465, opened 2019-08-06
changed password auth strategy type to correct one

prvit:


jesstelford:
@all-contributors add @prvit for docs

PR #1468, opened 2019-08-07
Update adapter config API

MadeByMike:
Some reasonable changes to the adapter APIs to allow a better getting started experience and more consistent API.

Adapters are now responsible for their own configuration and don't rely on the user setting anything on the keystone object environmental vars. The `name` from the keystone configuration is still silently passed through to the Adapter for use as a default and if no configuration options are set the adapters look for a set of environmental variable (this may differ between adapters).

For consistency and to avoid confusion with multiple ways of doing thing `keystone.connect()` no longer accepts a parameter. There was ambiguity over whether this or the adapter config should take precedence and future adapters may not have a connect string or require other connection options so it makes sense for all this to be on the adapter.

I tried to involve everyone working on this but if you have any questions or concerns come see me.

More info in the committed readme files.


packages/keystone/lib/adapters/index.js
```diff
@@ -1,15 +1,14 @@
...
+  constructor(config = {}) {
     this.config = { ...config };
-    this.name = this.config.name;
     this.listAdapters = {};
   }
 
   newListAdapter(key, { listAdapterClass, ...adapterConfig }) {
     const _listAdapterClass =
-      listAdapterClass || this.config.listAdapterClass || this.constructor.defaultListAdapterClass;
+      listAdapterClass || this.listAdapterClass || this.constructor.defaultListAdapterClass;
```
jesstelford: Why did this change?

packages/keystone/lib/adapters/index.js
```diff
@@ -18,9 +17,9 @@ class BaseKeystoneAdapter {
     return this.listAdapters[key];
   }
 
-  async connect(to, config = {}) {
+  async connect({ name }) {
     // Connect to the database
-    await this._connect(to, config);
+    await this._connect({ name }, { ...this.config });
```
jesstelford:
```suggestion
    await this._connect({ name }, this.config);
```

packages/test-utils/lib/test-utils.js
```diff
@@ -8,20 +8,29 @@ const { MongooseAdapter } = require('@keystone-alpha/adapter-mongoose');
...
 
-function setupServer({ name, adapterName, createLists = () => {} }) {
+function setupServer({ name, adapterName, createLists = () => {}, createApps, keystoneOptions }) {
   const Adapter = { mongoose: MongooseAdapter, knex: KnexAdapter }[adapterName];
-  const args = { mongoose: {}, knex: { dropDatabase: true } }[adapterName];
+  const args = {
+    mongoose: {},
+    knex: {
+      dropDatabase: true,
+      knexOptions: { connection: process.env.KNEX_URI || 'postgres://localhost/keystone' },
```
jesstelford: These are the defaults anyway, so we don't need to set them here.

```suggestion
      knexOptions: {},
```

packages/adapter-knex/README.md
```diff
@@ -35,47 +35,43 @@ const { KnexAdapter } = require('@keystonejs/adapter-knex');
...
-
-const knexOptions = { ... };
-
-keystone.connect(uri, {
-  client,
-  schemaName,
-  ...knexOptions,
+  adapter: new KnexAdapter({
+    schemaName,
+    knexOptions,
```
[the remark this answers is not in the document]
MadeByMike: Yeah, if you think that makes more sense I'm down with it.

packages/adapter-mongoose/README.md
```diff
@@ -12,29 +12,36 @@ const { MongooseAdapter } = require('@keystonejs/adapter-mongoose');
 
 const keystone = new Keystone({
   name: 'My Awesome Project',
-  adapter: new MongooseAdapter(),
+  adapter: new MongooseAdapter({
+    mongoUri,
+    mongooseOptions,
```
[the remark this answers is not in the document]
MadeByMike: Hmmm how do we say the API is for the options then?

packages/test-utils/lib/test-utils.js
MadeByMike: Nope, we need this makes the database for all tests `keystone` rather than the test name. We could make it `test` or something else, but this would require everyone to make a postgres db named `test` for testing.

packages/keystone/lib/adapters/index.js
MadeByMike: Might be a mistake

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -16,28 +17,43 @@ const {
...
-  async _connect(to, config = {}) {
-    const {
-      connection = to ||
-        process.env.KNEX_URI ||
-        'postgres://keystone5:k3yst0n3@127.0.0.1:5432/ks5_dev',
-      schemaName = 'keystone',
-      ...rest
-    } = config;
+  async _connect({ name }) {
+    const { schemaName = 'public', knexOptions = {} } = this.config;
```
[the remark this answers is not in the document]
MadeByMike: No, we need this. It's added to `this.schemaName` further down.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -16,28 +17,43 @@ const {
...
-      ...rest,
       client: this.client,
-      connection,
+      connection: uri,
+      ...knexOptions,
     });
+
+    // Knex will not error until a connection is made
+    // To check that the connection we run a test query
+    this.knex.raw('select 1+1 as result').catch(result => {
```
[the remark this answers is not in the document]
MadeByMike: Do we need to? I guess if we want the warning to come first.... but it will slow it down.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -16,28 +17,43 @@ const {
...
+      logger.error(`Could not connect to database: '${uri}'`);
+      logger.warn(`Check your database configuration.`);
+      logger.error(result.error);
+      throw result.error;
+    });
+
     this.schemaName = schemaName;
```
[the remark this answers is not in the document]
MadeByMike: This is used elsewhere

MadeByMike:
Merged all suggestions... just the issue of what to do with the docs whether we need to show the options in use or the simplest default... nothing ;)

packages/adapter-knex/README.md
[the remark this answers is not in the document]
jesstelford: In general, we want to make sure all code examples can be copy+pasted and _mostly_ work. Things like this make it one little step closer to working for the user who lands on this page 👍

packages/adapter-knex/lib/adapter-knex.js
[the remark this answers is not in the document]
jesstelford: If we don't `await`, then line 54 (`throw result.error`) will not be caught, and so the `_connect()` function will not fail, and then Keystone will continue on with the connection, potentially throwing a different error later on.

packages/adapter-mongoose/README.md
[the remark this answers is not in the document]
jesstelford: Actually write it out in the docs:

```markdown
# API

## `new MongooseAdapter(options)`

### `options`

These options are passed directly to `mongoose.connect()`, with the exceptions noted below:

#### `options.mongoUri` (optional)

... etc etc
```

packages/test-utils/lib/test-utils.js
jesstelford: Ahh, gotchya 👍

packages/adapter-mongoose/README.md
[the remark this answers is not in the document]
MadeByMike: Happy with that!

packages/adapter-knex/README.md
```diff
@@ -37,46 +15,61 @@ const keystone = new Keystone({
...
-Any extra options provided will be passed through to the Knex configuration function. See the [Knex docs](https://knexjs.org/#Installation-client) for possible values.
+or `'postgres://localhost/<DATABASE_NAME>'`where `DATABASE_NAME` is be derived from the Keystone project name.
 
 ## Debugging
 
 To log all Knex queries, run the server with the environment variable `DEBUG=knex:query`.
+
+## Setting Up a Database
+
+Before running Keystone with the Knex adapter you should configure a database. By default Knex will look for Postgres database on the default port, matching the project name, as the current user.
```
jesstelford:
```suggestion
Before running Keystone with the Knex adapter you should configure a database. By default Knex will look for a Postgres database on the default port, matching the project name, as the current user.
```

packages/adapter-knex/README.md
```diff
@@ -37,46 +15,61 @@ const keystone = new Keystone({
...
+
+In most cases this database will not exist and you will want configure a user and database for your application.
```
jesstelford:
```suggestion
In most cases this database will not exist and you will want to configure a user and database for your application.
```

packages/adapter-knex/README.md
```diff
@@ -37,46 +15,61 @@ const keystone = new Keystone({
...
+
+Otherwise, you can run these steps manually:
+
+```shell
+createdb -U postgres keystone
+psql ks5_dev -U postgres -c "CREATE USER keystone5 PASSWORD 'k3yst0n3'"
+psql ks5_dev -U postgres -c "GRANT ALL ON DATABASE keystone TO keystone5;"
+```
+
+If using the above, you will want to set a connection string of: `postgres://keystone5:k3yst0n3@localhost:5432/keystone`
```
jesstelford: This whole section feels like it'll be ultimately confusing for users; By following these instructions we're taking them away from the defaults we mentioned earlier. Now they have to know two different things just to get started?

Can I suggest we rename this section to something like `## Preparing for Production`:

```markdown
## Preparing for Production

When developing locally, using the defaults is fine. When it's time to deploy to production, your database may have a different name, and you will most likely require a username & password to access it.

Let's say your production database was setup with the following information:

- **Database host**: `awesome-pg-host.example.com`
- **Schema name**: `my-company`
- **Database name**: `keystone`
- **User**: `keystone5`
- **Pass**: `k3yst0n3` _(note: this is **not** a secure password!)_

Then to access your database, you would set the `schemaName` & `knexOptions`:

```javascript
const { KnexAdapter } = require('@keystonejs/adapter-knex');
  name: 'My Awesome Project',
  adapter: new KnexAdapter({
    schemaName: 'my-company',
    knexOptions: { connection: 'postgres://keystone5:k3yst0n3@awesome-pg-host.example.com/keystone' },
  }),
});
```

Then again, we can delete the `.sh` file, and remove any confusing language around running SQL commands, etc.

packages/adapter-knex/README.md
```diff
@@ -37,46 +15,61 @@ const keystone = new Keystone({
...
-  schemaName,
-  ...knexOptions,
-});
 ```
 
-## API
+## API `new KnexAdapter(options)`
 
-### `uri`
+### `schemaName`
```
jesstelford: Where did `schemaName` come from? It's not clear where to set this variable. I find it helps to tie it back to the api like so:

```suggestion
### `options.schemaName`
```

packages/adapter-knex/README.md
```diff
@@ -37,46 +15,61 @@ const keystone = new Keystone({
...
-_**Default:**_ `'postgres://keystone5:k3yst0n3@127.0.0.1:5432/ks5_dev'`
+_**Default:**_ `'public'`
 
-Either a connection string, or a connection object, as accepted by Knex.
-See [knex docs](https://knexjs.org/#Installation-client) for more details.
-If the environment variable `KNEX_URI` is set, its value will be used as the default.
+All postgres tables are grouped within a schema and `public` is the default schema.
 
-### `client`
+### `knexOptions`
```
jesstelford:
```suggestion
### `options.knexOptions`
```

packages/adapter-knex/README.md
MadeByMike: Most user will want to setup up a database and I think having this with the knex adapter make some sense. The defaults may not work for everyone depending how they installed postgress and access on their machine.

I moved this to the bottom and I'd only expect people to be reading this if they need help.

jesstelford:
@all-contributors add @MadeByMike for code

PR #1472, opened 2019-08-08
Refactor Knex.js query building

sarneaud:
We had some bugs in generated SQL having ambiguous column references
for GraphQL queries containing "many" relationships.  We also had a
problem with "many" relationship queries triggering intermediate SQL
queries that were blowing out in size (specifically the auto-generated
IN clause).

It wasn't easy to fix both without reworking the SQL generator to make
subqueries easier to build, and aliases easier to track.

The implementation of "many" relationships still has room for
improvement, but it's working well enough for our site now.

While I was at it, I pushed the work of populateMany() into the main
query by using array_agg subqueries.  This reduces the number of
database calls a lot.


sarneaud:
By the way, the only changes to the generated SQL are supposed to be the extra subqueries for populateMany(), the aliases, and the implementation of where clauses on "many" relationships.

MadeByMike:
Wow @sarneaud was this as painful to write as it was to read? Good job!

MadeByMike:
If this fixes existing bug, what tests should we have had to capture the bug and can we add them now?

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -277,35 +292,30 @@ class KnexListAdapter extends BaseListAdapter {
...
-        )
-      )),
-    };
+  // Add subqueries that return arrays of IDs for each "many" relationship
+  _addManySubqueries(query, baseTableAlias) {
+    const manyAdapters = this.fieldAdapters.filter(a => a.isRelationship && a.config.many);
+    query.column(manyAdapters.map(a => {
+      const manyTable = this._manyTable(a.path);
+      const manyTableAlias = `s${baseTableAlias}__${a.path}`;
+      return this._query().select(this._knexRaw('array_agg(??)', `${manyTableAlias}.${a.refListId}`))
```
jesstelford: Adding `array_agg` will tie our adapter directly to postgres. We want to retain as much support for different DBs via knex as possible (hence why we chose knex, and not postgres directly). Is there another way to do this without dropping to raw SQL?

sarneaud: No, array operations aren't supported by Knex.

Unfortunately, array handling isn't very portable.  (I didn't realise KS is meant to support DBs other than PG.)  But doing something like this is a lot better for performance than looping over the results and running extra queries.  Maybe it's worth special-casing for PG and MySQL, with a fallback to looping.

sarneaud:
The current test failures don't seem to be related to this PR.  Are they known to be problematic?

sarneaud:
I've reverted the change to `populateMany()` because it isn't required for the bug fix I need.  Heads up, though: the adapter is using `returning()` on mutation queries, which isn't supported by MySQL or Sqlite3.  Maybe it's a good idea to add an in-memory Sqlite3 DB to the test run to catch some of these DB-specific quirks.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -105,6 +104,14 @@ class KnexAdapter extends BaseKeystoneAdapter {
     const { AutoIncrement } = require('@keystone-alpha/fields-auto-increment');
     return AutoIncrement.primaryKeyDefaults[this.name].getConfig(this.client);
   }
+
```
timleslie: It looks like we're not using these methods. Can we remove them?

api-tests/relationships/nested-mutations/two-way-backreference/to-many.test.js
```diff
@@ -9,7 +9,7 @@ const toStr = items => items.map(item => item.toString());
 
 // `mongodb-memory-server` downloads a binary on first run in CI, which can take
 // a while, so we bump up the timeout here.
-jest.setTimeout(60000);
+//jest.setTimeout(60000);
```
timleslie: Do we want to make this change?

timleslie:
After a chat with Simon the main take aways for this PR are:
 * Fixes an existing bug with many to many queries by being smarter about table-alias generation
 * Uses subqueries to perform `every/some/none` operations, rather than separate queries.
 * Does not introduce any new postgres-specific operations.

Based on this I'm happy to give this a 👍 once the cosmetic issues are addressed and CI is green.

PR #1474, opened 2019-08-10
changed syntax to match

kennedybaird:


jesstelford:
Thanks @kennedybaird!

@all-contributors can you add @kennedybaird for docs?

PR #1480, opened 2019-08-12
Default function

jesstelford:
Add support for setting default values at mutation time via a function.

This PR adds support for default-as-a-function to both:
- The Admin UI `Controller`s
- The GraphQL API `Implementation`s

The Admin UI `Controller`s isn't actually used in the Admin UI because there's no way to serialize a function and pass it through the Webpack build process (weeeeellll, there _is_, but we're not doing it yet, and it needs to be discussed if we add that support. See [how Puppeteer does it](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageevaluatepagefunction-args)).

The GraphQL API `Implementation` looks like this:

```javascript
keystone.createList('Post', {
  fields: {
    createdAt: {
      type: DateTime,
      defaultValue: () => Date.now().toISOString(),
    },
  }
});
```

The `defaultValue()` function receives an object as its only parameter, very similar to the rest of the hooks:

```javascript
keystone.createList('Post', {
  fields: {
    createdAt: {
      type: DateTime,
      defaultValue: ({
        existingItem,
        context,
        originalInput,
        actions: { query }
      }) => Date.now().toISOString(),
    },
  }
});
```

This PR retains the ability to set a static default value (`defaultValue: 'foobar'`), and the default `defaultValue` is still `undefined`.

I'm using this new functionality as above, but also to set the author of a post like so:

```javascript
keystone.createList('Post', {
  fields: {
    createdAt: {
      type: Relationship,
      ref: 'User',
      access: {
        // Only admins can explicitly set the author of a post.
        // Otherwise, the author is set in the defaultValue hook.
        create: ({ authentication: { item } }) => item && item.isAdmin,
        read: true,
        update: ({ authentication: { item } }) => item && item.isAdmin,
      },
      defaultValue: ({ context: { authedItem } }) => {
        if (!authedItem) {
          return null;
        }
        return { connect: authedItem.id };
      },

    },
  }
});
```


api-tests/default-value/defaults.test.js
```diff
@@ -0,0 +1,190 @@
...
+    adapterName,
+    createLists: keystone => {
+      keystone.createList('User', { fields });
+    },
+  });
+
+describe('defaultValue field config', () => {
+  multiAdapterRunners().map(({ runner, adapterName }) =>
+    describe(`Adapter: ${adapterName}`, () => {
+      it('Has no default by default', () => {
```
MadeByMike: Amazing test name!

docs/discussions/mutation-lifecycle.md
```diff
@@ -140,6 +140,8 @@ Any fields which a) are not set on the provided item and b) have a configured de
 
 The default value of a field can be configured at `List` definition time with the config attribute `defaultValue`.
 
+The `defaultValue` may be a static value, or a function which returns the value / a Promise for the value.
```
MadeByMike:
```suggestion
The `defaultValue` may be a static value, or a function which returns either the value or a Promise.
```

Maybe this is a little clearer?

test-projects/basic/cypress/integration/search_spec.js
```diff
@@ -12,7 +12,7 @@ describe('Search', () => {
       url: '/admin/posts',
       searchTerm: 'hello planet',
       found: [],
-      notFound: ['Hello World'],
+      notFound: ['Hello Things'],
```
MadeByMike: Vital change :)

Issue #1489, opened 2019-08-13
Do we need a plugin architecture and what does it look like?


MadeByMike:
It seems there is a need for some kind of plugin architecture. Sometimes this can be achieved via hooks and HOC however if there are many popular 'plugins' wrapping many HOCs is not a fantastic API.

I've created this issue to discuss.

Some examples that might be 'plugins' include:

- Revision history, #321 
- Roles, #1273 
- Migration\Seeding, #299, #311, #144, #139, #1091
- AdminUI customisation, #1037 (this one might be a slightly different issue)

I'm very pro plugins as I think they help find limitations in the core product. An example of this is orderable\sortable fields. To support this I think we need to make some changes to the core platform and if other plugin ideas surface requirements that would be great!

I'd love to see 'plugins' stay really simple. essentially a better API for HOCs something like:

```javascript
keystone.createList('User', Users, 
{ 
plugins: [
  withTimestamps(timestampPluginOptions), 
  withRoles(rolePluginOptions), 
  ]
}
);
```

I think plugins should be able to add fields, add hooks and register their own hooks.

timleslie:
# List Plugin API Proposal

In this comment I'll outline a proposal for adding `list plugins` as a mechanism for adding functionality to a `List`. 

A new field will be added to the `config` object consumed by `createList` called `plugins`, which takes a list of functions with the signature `config => config`. Each of these functions will have an opportunity to take the config object and return either a new config or a modified config.

The plugin function can make changes such as adding new fields or hooks, adding new options to existing fields (e.g. adding access control), or modifying existing fields.

A trivial example would be
```js
keystone.createList('User', {
  fields: { ... },
  hooks: { ... },
  plugins: [config => ({...config, fields: { ...config.fields, newField: { type: Text })],
}
```

`createList` would apply each of the plugins in the order specified, allowing for chaining of different plugins.

The following two pieces of code would be equivalent.

```js
keystone.createList('User', {..., plugins: [plugin1, plugin2] })
```

```js
keystone.createList('User', plugin2(plugin1({...})))
```

In this sense, the `plugins` option is simply a convenience wrapper around what is already possible. The advantage is that it gives us a well defined pattern which people can follow. It also allows the plugin list to be defined programatically.

gautamsi:
>The plugin function can make changes such as adding new fields or hooks

[...]

PR #1491, opened 2019-08-13
List plugins

timleslie:
#1489

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -74,7 +74,9 @@ module.exports = class Keystone {
   createList(key, config, { isAuxList = false } = {}) {
     const { getListByKey, adapters } = this;
     const adapterName = config.adapterName || this.defaultAdapter;
-    const list = new List(key, config, {
+    const compose = fns => o => fns.reduce((acc, fn) => fn(acc), o);
+
+    const list = new List(key, compose(config.plugins || [])(config), {
       getListByKey,
       getGraphQLQuery: schemaName => this._graphQLQuery[schemaName],
       adapter: adapters[adapterName],
--- a/.changeset/rich-hairs-fix/changes.json
+++ b/.changeset/rich-hairs-fix/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/keystone", "type": "minor" }], "dependents": [] }
--- a/.changeset/rich-hairs-fix/changes.md
+++ b/.changeset/rich-hairs-fix/changes.md
@@ -1,0 +1,1 @@
+Added `plugins` option to the config of `createList`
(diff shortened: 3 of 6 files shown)


packages/keystone/lib/Keystone/index.js
```diff
@@ -74,7 +74,9 @@ module.exports = class Keystone {
   createList(key, config, { isAuxList = false } = {}) {
     const { getListByKey, adapters } = this;
     const adapterName = config.adapterName || this.defaultAdapter;
-    const list = new List(key, config, {
+    const compose = fns => o => fns.reduce((acc, fn) => fn(acc), o);
+
+    const list = new List(key, compose(config.plugins || []), config, {
```
gautamsi: We should also have option to run this for Keystone level plugins

gautamsi:
can we remove `list.initFields();` from `keystone.createList` and delay this until `keystone.connect`, this can also enable some more scenarios, like modifying list config or option dynamically before the `connect` call.

packages/keystone/lib/Keystone/index.js
gautamsi: What about running plugins over the instance of `List`/`Field` as well. We can separate the hooks of plugins, one that run on config and one that run on instance of `List/Field`

jesstelford: @gautamsi that's an interesting idea, and it might come in the future. This PR is purely for setting up a "blessed" API to enhance field config.

Anything more than what's here already is out of scope for now.

jesstelford: Should this have executed the composed function?

```suggestion
     const list = new List(key, compose(config.plugins || [])(config), {
```

jesstelford: See my other comment: https://github.com/keystonejs/keystone-5/pull/1491#discussion_r313246556

timleslie: Hah, yep, 💯

MadeByMike:
@gautamsi this is not the last PR we will have for plugins - just a simple first step. We're gathering more feedback regarding "Keystone level plugins" and I think we will keep #1489 open for now.

Are you happy with this as a very basic fist step?

MadeByMike:
I want to get into the habit of adding documentation at the time we add features @timleslie are you ok if I add doco to this PR?

MadeByMike:
@timleslie I updated the API docs. I still have to do the fields documentation but I don't want to hold up this PR any longer :)

gautamsi:
@timleslie I can create a PR for following (either to be merged with this or after), thoughts

> can we remove `list.initFields();` from `keystone.createList` and delay this until `keystone.connect`, this can also enable some more scenarios, like modifying list config or option dynamically before the `connect` call.

PR #1496, opened 2019-08-16
Speed up Knex query condition lookup

sarneaud:
For some some of our more complex GraphQL queries, the Knex adapter was
spending ~100ms generating Knex queries in total.  This patch drops
that down to a few ms.

We have some complex-to-build-but-quick-on-the-DB API calls that are
about twice as fast with this patch.  It also improves performance when
multiple requests are in flight because tasks aren't blocking one other
trying to build queries so much.


jesstelford:
Tests are all green (ignore the Netlify ones) 👍 

@timleslie you know more about the knex adapter; any reason not to merge this?

PR #1500, opened 2019-08-18
Expose 'originalInput' to access control functions for lists & fields

jesstelford:
Let's say I've got a "publishing workflow" where a `Post`, once published, can't be edited except by admins, or if the author first marks it as draft.

I've got two options for setting access control, but neither seem ideal:

**Option 1: access per field**

```javascript
keystone.createList('Post', {
  fields: {
    content: {
      type: Text,
      access: {
        read: true,
        create: true,
        update: ({ authentication: { item }, existingItem }) => {
          if (!item) {
            return false;
          }

          if (item.isAdmin) {
            return true;
          }

          // Allow updating when in draft
          // NOTE: This check is now required on _every_ field
          return existingItem.publishStatus !== 'published';
        },
      }
    },
    publishStatus: {
      type: Select,
      options: ['draft', 'published'],
      defaultValue: 'draft',
      access: {
        read: true,
        create: true,
        update: ({ authentication: { item }, existingItem }) => {
          if (!item) {
            return false;
          }

          // Admins or authors can adjust the published status
          if (item.isAdmin || existingItem.author === item.id) {
            return true;
          }
        },
      }
    },
  },
});
```

This isn't great because it would be easy to accidentally forget to add that access control on every field.

**Option 2: List-level access**

```javascript
keystone.createList('Post', {
  fields: {
    content: { type: Text },
    author: { type: Relationship, ref: 'User' },
    publishStatus: {
      type: Select,
      options: ['draft', 'published'],
      defaultValue: 'draft',
    },
  },
  access: {
    create: true,
    read: true,
    update: ({ authentication: { item } }) => {
      if (!item) {
        return false;
      }

      if (item.isAdmin) {
        return true;
      }

      // How do I allow the author to set a 'published' post back to 'draft'?
      return {
        publishStatus_not: 'published'
        author: { id: item.id },
      }
    }
  },
});
```

This seems like the best bet, but I can't access the incoming changes so I can first check to see what the author is trying to change, and if it involves changing the publishedStatus to draft, then it's ok, otherwise, don't allow it.

---

So, this PR introduces the `originalInput`, which allows it to come out like this:

```javascript
keystone.createList('Post', {
  fields: {
    content: { type: Text },
    author: { type: Relationship, ref: 'User' },
    publishStatus: {
      type: Select,
      options: ['draft', 'published'],
      defaultValue: 'draft',
    },
  },
  access: {
    create: true,
    read: true,
    update: ({ authentication: { item }, originalInput }) => {
      if (!item) {
        return false;
      }

      if (item.isAdmin) {
        return true;
      }

      // The only valid edit an author can make to publishStatus is to move it
      // from 'published' -> 'draft'.
      if (originalInput.publishStatus && originalInput.publishStatus !== 'draft') {
        return false;
      }

      // Otherwise, the author must be trying to edit a post that's already got
      // publishStatus in draft
      return {
        author: { id: item.id },
        publishStatus_not: 'published',
      };
    }
  },
});
```

### DIFF
--- a/.changeset/rare-flies-care/changes.json
+++ b/.changeset/rare-flies-care/changes.json
@@ -1,0 +1,7 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/access-control", "type": "minor" },
+    { "name": "@keystone-alpha/keystone", "type": "minor" }
+  ],
+  "dependents": []
+}
--- a/.changeset/rare-flies-care/changes.md
+++ b/.changeset/rare-flies-care/changes.md
@@ -1,0 +1,1 @@
+Expose 'originalInput' to access control functions for lists & fields
--- a/docs/api/access-control.md
+++ b/docs/api/access-control.md
@@ -113,6 +113,7 @@ type AccessInput = {
     item?: {},
     listKey?: string,
   },
+  originalInput?: {},
 };
 
 type StaticAccess = boolean;
@@ -141,6 +142,7 @@ ie; for a list `User`, it would match the input type `UserWhereInput`.
 - `authentication` describes the currently authenticated user.
   - `.item` is the details of the current user. Will be `undefined` for anonymous users.
   - `.listKey` is the list key of the currently authenticated user. Will be `undefined` for anonymous users.
+- `originalInput` for `create` & `update` mutations, this is the data as passed in the mutation.
 
 When resolving `StaticAccess`;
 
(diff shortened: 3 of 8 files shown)


jesstelford:
100% agreed! I didn't want to clutter up this PR too much, but am happy to come back to it in the future 👍

jesstelford:
In aggregate, the tests have all passed. Each run produces a different failure, all related to the known issues with our Cypress suite.

PR #1503, opened 2019-08-19
Implemented custom types for mutations

SeanDoyleGit:
This is a quick solution to adding custom types for mutations. 
Let me know what you think. 

This is how I am thinking we could add types to the custom mutations
```
  fields: {},
  mutations: [
    {
      types: [
        `type ExampleType {
          name: String
	  amount: Int
        }`,
      ],
      schema: 'getExampleType(id: ID!): ExampleType',
      resolver: async (obj, args, context, info, { query }) => {}
    }
  ]
})
```

### DIFF
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -690,7 +690,7 @@ module.exports = class List {
   }
 
   getGqlMutationTypes() {
-    return [...flatten(this.mutations.filter(x => x.types).map(x => x.types);
+    return flatten(this.mutations.filter(x => x.types).map(x => x.types));
   }
 
   checkFieldAccess(operation, itemsToUpdate, context, { gqlName, extraData = {} }) {
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -689,6 +689,13 @@ module.exports = class List {
     return mutations;
   }
 
+  getGqlMutationTypes() {
+    const types = [];
+    types.push(...flatten(this.mutations.filter(({ types }) => types).map(({ types }) => types)));
+
+    return types;
+  }
+
   checkFieldAccess(operation, itemsToUpdate, context, { gqlName, extraData = {} }) {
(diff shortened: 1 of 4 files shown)


packages/keystone/lib/List/index.js
```diff
@@ -689,6 +689,13 @@ module.exports = class List {
     return mutations;
   }
 
+  getGqlMutationTypes() {
+    const types = [];
+    types.push(...flatten(this.mutations.filter(({ types }) => types).map(({ types }) => types)));
```
dcousens:
``` js
return [...flatten(this.mutations.map(({ types }) => types).filter(x => x)]
```

dcousens: or
``` js
return [...flatten(this.mutations.filter(x => x.types).map(x => x.types)
```

If the complexity mattered

packages/keystone/lib/List/index.js
```diff
@@ -689,6 +689,13 @@ module.exports = class List {
     return mutations;
   }
 
+  getGqlMutationTypes() {
+    const types = [];
```
jesstelford:
```suggestion
```

packages/keystone/lib/List/index.js
jesstelford:
```suggestion
    return [...flatten(this.mutations.filter(x => x.types).map(x => x.types);
```

packages/keystone/lib/List/index.js
```diff
@@ -689,6 +689,13 @@ module.exports = class List {
...
+    types.push(...flatten(this.mutations.filter(({ types }) => types).map(({ types }) => types)));
+
```
jesstelford:
```suggestion
```

packages/keystone/lib/List/index.js
```diff
@@ -689,6 +689,13 @@ module.exports = class List {
...
+    return types;
```
jesstelford:
```suggestion
```

PR #1506, opened 2019-08-20
Create new CLI and example projects

MadeByMike:
- Adds a new CLI with Adapter choices
- Add example projects as starter templates

I intend to add a bunch of new starter templates in a new PR.

Note: I would have loved if the example projects could have lived in the root directory but not sure this is possible.

Future Todo: I think I'd like to move the demo projects somewhere else to avoid possible confusion.


gautamsi:
Earlier plan was to create more examples and use them like `create-next-app --example`. I suggested to use similar `examples` directory. demo-projects would continue to house specific demo, we should create another folder and map --examples from there.

MadeByMike:
> Earlier plan was to create more examples and use them like `create-next-app --example`. I suggested to use similar `examples` directory. demo-projects would continue to house specific demo, we should create another folder and map --examples from there.

I had a flag but decided there was little need for it with a detailed prompt. I can see that flags might be a useful way to preselect options that would be useful for either power users or testing. But hopefully people are not scaffolding projects too often so I think the 1sec it takes to select a project is ok :)

Would love an `example-projects` to be at the root - but I can't do this and have it published as a standalone package.

MadeByMike:
@mitchellhamilton I've been around the block with this a few times :( I've pitched your approach to Jed and and Jess who both had input and we eventually arrived here. This is perhaps a place where everyone is only slightly unhappy?? If so, I think maybe that's a win.

I've tried to take on board all suggestions but to do everything would be conflicting. I think we need to take a step back and look at the goals. For me the goals were: 

- A better CLI experience especially for first time users
- Ability to select adapters
- Remove templating engines from the CLI code (and work without building from CLI)
- Examples that were closer to "deployable" 
- Examples that provided good starting points for different types of applications

Of the above I think I achieved the first 4 or maybe 4.5... and the ease with which I can add examples now means I can quickly add PRs for `next`, `react` and any other apps we build. That puts me on the way to hitting all the goals. This is not the last improvement.

I should say too, I am fully on board with these examples being deployed to their own repos. I'd love people to be able to clone them if that is their preferred way of working and that's why I've gone to the lengths I have to avoid templating and make sure examples can run like normal projects without compiling.

To that end - if in future you'd like to disconnect the `example-projects` into its own repo or multiple repos you have my full support.


For now I think we should push this as a step in the right direction as I'd like to talk and write about the example projects.

mitchellhamilton:
Yep, that sounds good, sorry, please ignore my previous comment, I was thinking too much about little things(the npm thing I said is overly complex and there are easier ways to solve it, ignore that)

> Ability to select adapters
> Remove templating engines from the CLI code (and work without building from CLI)

I think these two things are sort of in conflict if we are doing what's in this PR(the adapter replacement stuff) and by using any form of templating, we stop ourselves from being able to easy test examples that have code replaced because we have to use the cli to write the example rather than run the example that's in the repo but maybe it's not really an issue since at least we're not doing templating on the package.json files so changesets can update the package.json files when we do releases. ¯\\\_(ツ)_/¯

So I think the right answer is to merge this and in another PR, change this to clone the examples from github(and to resolve the "examples that depends on unpublished things" problem, we can set a git tag and clone the examples from that tag) so we can not ship the examples in the create-keystone-app package and have community examples from other repos.

----

unrelated code thing: The example projects should be treated as packages in the repo so that the package.json files are updated with new versions of packages when we do releases. (Add them to `bolt.workspaces` in the root package.json)

package.json
```diff
@@ -49,6 +49,7 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/runtime": "^7.4.3",
+    "@babel/standalone": "^7.5.5",
```
mitchellhamilton: Where is this used? I can't find it anywhere

MadeByMike: Hmm you might be right...

MadeByMike:
@mitchellhamilton +1 on the workspace set-up thing.

@mitchellhamilton I get that replacing a 'magic comment' with the adapter stuff is kinda templating. But user experience of selecting an adapter is much better than them needing to understand which projects have what adapters. We could probably hide this from them with some clever code where they make the same selections and it clones the right project.... 

The only thing that worries me is that as we get more adapters we quickly end up with many project to make each time we want to support a specific adapter with just a few more adapters it gets huge and we are small.

Ability to do smoke tests on the projects would be nice.... but we can have an open issue for solving that in future.

PR #1508, opened 2019-08-20
Update auth packages

MadeByMike:
- `PasswordAuthStrategy` is now own package `auth-password`
- Renamed `passport-auth` to `auth-passport`


MadeByMike:
Split this into 2 PRs

PR #1509, opened 2019-08-20
Implemented custom queries

SeanDoyleGit:
Implemented custom queries in the same format as custom mutations. 

Let me know what you think. 

usage would look like: 

```
  fields: {},
  queries: [
    {
      types: [
        `type ExampleType {
          name: String
	  amount: Int
        }`,
      ],
      schema: 'getExampleType(id: ID!): ExampleType',
      resolver: async (obj, args, context, info, { query }) => {}
    }
  ]
})
```


gautamsi:
This should also be added to `Keystone` instance directly. there are scenarios where there is no need to attach the query/mutation with the list, it will be easier to have them attached to keystone. 

```js
/*  .... other options   */
queries: {},
mutations: {},
});
```
or better have to pass this as parameter to `app-graphql` additionally.

I would assume I can pull in multiple queries and mutations and have this merged like regular graphql thing. 

if not possible, we should have a way to create custom graphql based on apollo and merge definitions from keystone to that,

timleslie:
> This should also be added to `Keystone` instance directly. there are scenarios where there is no need to attach the query/mutation with the list, it will be easier to have them attached to keystone.
> 

This PR is a placeholder implementation while we finalise discussion around what the correct API is for adding custom queries/mutations. We know that the current API is not what we want to have long term. This quick fix is an important stop-gap measure until that design is complete.

MadeByMike:
Great work!

jesstelford:
[There was no changeset](https://github.com/keystonejs/keystone-5/pull/1509#issuecomment-522864928)!

I've added one: https://github.com/keystonejs/keystone-5/pull/1516

PR #1510, opened 2019-08-20
Add `Slug.alwaysMakeUnique` option & fix uniquess bug

jesstelford:
The `Slug.alwaysMakeUnique` option forces calling `makeUnique` even when
initially generated slug may already be unique to prevent accidental
data leak. For example; a deactivated user with slug `sam-smith` will cause a collision for new accounts with a `name` of `Sam Smith`, revealing the existence of previously created accounts of that name by checking if the returned `slug` has a unique string appended.

In the process, this fixes a bug where items restricted via access
control weren't considered when testing a `Slug` for uniqueness.


packages/fields/src/types/Slug/Implementation.js
```diff
@@ -223,11 +233,16 @@ export class SlugImplementation extends Text {
...
+            [this.path]: uniqueSlug,
             // Ensure we ignore the current item when doing an update
             ...(existingItem && existingItem.id && { id_not_in: [existingItem.id] }),
           },
         },
+        // Access Control may filter out some results, so we wouldn't be
+        // retreiving an accurate list of all existing items. Because we add the
+        // unique constraint to the field, the database will throw an error if
+        // we miss a match and try to insert anyway.
+        skipAccessControl: true,
```
jesstelford: This was the bug... Which led to adding `alwaysMakeUnique`.

MadeByMike:
I understand the need now. But I think you should just run `alwaysMakeUnique` whenever `makeUnique` is set. It doesn't need to be a user option.

BTW: Great pick-up! Took me a bit to even understand.

jesstelford:
> you should just run alwaysMakeUnique whenever makeUnique is set.

The desired default here is to generate a slug directly from the input value. So, for a `url` generated based on a user's `name`, it would take `Jess Telford` and turn it into `jess-telford`. But if we _always_ ran `makeUnique` by default, then it would be `jess-telford-2j54lkdf` which is not desirable. However, we want to ensure that in the case where there _is_ an existing `jess-telford` url, we can supply a `makeUnique` method which is run for us at the right time.

PR #1514, opened 2019-08-21
Rename passport-auth to auth-passport

MadeByMike:
Rename `passport-auth` to `auth-passport`.


.changeset/cold-numbers-push/changes.json
```diff
@@ -0,0 +1,69 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/auth-passport", "type": "major" },
+    { "name": "@keystone-alpha/keystone", "type": "major" },
```
timleslie: Is this really a major bump to `@keystone-alpha/keystone`?

.changeset/cold-numbers-push/changes.md
```diff
@@ -0,0 +1 @@
+Renamed `passport-auth` to `auth-passport`
```
timleslie: Can we add a bit more details here, as this is actually a pretty significant change, and we don't want people to miss it. I'm thinking something like:

```
This change renames the package `@keystone-alpha/passport-auth` to `@keystone-alpha/auth-passport`. Anyone using `passport-auth` should switch over to the new package name, as the old one will no longer be receiving any updates.
```

.changeset/cold-numbers-push/changes.json
MadeByMike: Actually no, only the other one is because you can't import `PasswordAuthStrategy` from keystone anymore. But now that I think about it, I don't think this has any breaking change for Keystone itself.

.changeset/cold-numbers-push/changes.json
```diff
@@ -0,0 +1,69 @@
...
+    { "name": "@keystone-alpha/cypress-project-login", "type": "major" },
```
timleslie: This package hasn't had a real change, so shouldn't get bumped at all.

.changeset/cold-numbers-push/changes.json
```diff
@@ -0,0 +1,69 @@
...
+    { "name": "@keystone-alpha/cypress-project-social-login", "type": "major" }
```
timleslie: This is just a dependency change, so should only be a patch bump.

.changeset/cold-numbers-push/changes.json
```diff
@@ -0,0 +1,69 @@
+{
+  "releases": [
+    { "name": "@keystone-alpha/auth-passport", "type": "major" },
+    { "name": "@keystone-alpha/keystone", "type": "minor" },
+    { "name": "@keystone-alpha/cypress-project-login", "type": "major" },
```
MadeByMike:
```suggestion
    { "name": "@keystone-alpha/cypress-project-login", "type": "minor" },
```

MadeByMike: Has a dep change

.changeset/cold-numbers-push/changes.json
```diff
@@ -0,0 +1,69 @@
...
+    { "name": "@keystone-alpha/cypress-project-social-login", "type": "major" }
```
MadeByMike:
```suggestion
    { "name": "@keystone-alpha/cypress-project-social-login", "type": "minor" }
```

PR #1515, opened 2019-08-21
Make `auth-password` a package

MadeByMike:
PasswordAuthStrategy is now imported from the `auth-password` package for a more consistent top-level API:

E.g.:

```
const { PasswordAuthStrategy } = require('@keystone-alpha/auth-password');
```


.changeset/plenty-laws-attend/changes.md
```diff
@@ -0,0 +1 @@
+Makes the password auth strategy its own package
```
mitchellhamilton: Could you add a short explanation/code example of how consumers should migrate?

MadeByMike:
```suggestion
Makes the password auth strategy its own package.
Previously: `const { Keystone, PasswordAuthStrategy } = require('@keystone-alpha/keystone');`
After change: `const { PasswordAuthStrategy } = require('@keystone-alpha/auth-password');`
```

MadeByMike: Sure, done :)

PR #1518, opened 2019-08-22
Remove node_modules check in CLI

MadeByMike:
Fixes a bug found when the cli was running outside the monorepo

### DIFF
--- a/.changeset/rich-zebras-rule/changes.json
+++ b/.changeset/rich-zebras-rule/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "create-keystone-app", "type": "minor" }], "dependents": [] }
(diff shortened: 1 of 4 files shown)


.changeset/rich-zebras-rule/changes.json
```diff
@@ -0,0 +1 @@
+{ "releases": [{ "name": "create-keystone-app", "type": "minor" }], "dependents": [] }
```
jesstelford:
```suggestion
{ "releases": [{ "name": "create-keystone-app", "type": "patch" }], "dependents": [] }
 ```

File: packages/create-keystone-app/example-projects/todo/public/styles.css [hunk 2]
```diff
@@ -40,7 +39,7 @@ body {
   max-width: 600px;
   background: #fff;
   border-radius: 18px;
-  border: solid 1px hsla(220, 84%, 85%, 1);
```
dcousens: Sneaking in changes I see

MadeByMike: ;)

PR #1519, opened 2019-08-22
Allow display and filtering of 'Id' column in Admin UI

jesstelford:
Also adds a beautiful new button next to any relationship fields so you can quickly and easily click through to the related item (for to-single), or a list view filtered to the related items (for to-many):

link-to-relations

### DIFF
--- a/packages/fields-mongoid/src/views/Controller.js
+++ b/packages/fields-mongoid/src/views/Controller.js
@@ -1,6 +1,6 @@
-import FieldController from '../../../Controller';
+import FieldController from '@keystone-alpha/fields/Controller';
 
-export default class UuidController extends FieldController {
+export default class TextController extends FieldController {
   getFilterGraphQL = ({ type, value }) => {
     switch (type) {
       case 'is': {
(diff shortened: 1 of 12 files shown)


packages/fields-mongoid/src/views/Controller.js
```diff
@@ -0,0 +1,68 @@
...
+          .split(',')
+          .map(value => `"${value.trim()}"`)
+          .join(',')}]`;
+      }
+      case 'not_in': {
+        return `${this.path}_not_in: [${value
+          .split(',')
+          .map(value => `"${value.trim()}"`)
+          .join(',')}]`;
+      }
```
jesstelford: The magic source which enables filtering by IDs!

packages/fields/src/Controller.js
```diff
@@ -6,6 +6,7 @@ export default class FieldController {
     this.label = config.label;
     this.path = config.path;
     this.type = config.type;
+    this.isPrimaryKey = config.isPrimaryKey;
```
jesstelford: We don't want primary key fields to be editable within the Admin UI, so we make sure that setting is easily accessible to allow filtering out later.

PR #1521, opened 2019-08-22
Respect static field-level access control in the Admin UI

jesstelford:
For example;

```javascript
keystone.createList('Post', {
  fields: {
    name:  { type: Text },
    slug: { type: Slug, from: 'name' },
    createdAt: {
      type: DateTime,
      defaultValue: () => new Date().toISOString(),
      access: {
        create: false,
        read: true,
        update: false,
      }
    }
  }
});
```

Will show the `createdAt` field in the list view as a column, but will not show the field in either the create modals or the item details form.

NOTE: As a future enhancement, we should show the field in the item details form, but with a "read only" view.

Screen Shot 2019-08-23 at 9 11 29 am
Screen Shot 2019-08-23 at 9 10 49 am
Screen Shot 2019-08-23 at 9 16 49 am

### DIFF
--- a/packages/app-admin-ui/client/components/CreateItemModal.js
+++ b/packages/app-admin-ui/client/components/CreateItemModal.js
@@ -48,12 +48,14 @@ class CreateItemModal extends Component {
       return;
     }
 
-    const nonPrimaryKeyFields = fields.filter(({ isPrimaryKey }) => !isPrimaryKey);
+    const creatable = fields
+      .filter(({ isPrimaryKey }) => !isPrimaryKey)
+      .filter(({ maybeAccess }) => !!maybeAccess.create);
 
-    const data = arrayToObject(nonPrimaryKeyFields, 'path', field => field.serialize(item));
+    const data = arrayToObject(creatable, 'path', field => field.serialize(item));
 
     if (!countArrays(validationWarnings)) {
-      const { errors, warnings } = await validateFields(nonPrimaryKeyFields, item, data);
+      const { errors, warnings } = await validateFields(creatable, item, data);
 
       if (countArrays(errors) + countArrays(warnings) > 0) {
         this.setState(() => ({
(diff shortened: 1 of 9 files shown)


packages/fields/src/types/Relationship/views/CreateItemModal.js
```diff
@@ -60,7 +60,11 @@ class CreateItemModal extends Component {
       return;
     }
 
-    const data = arrayToObject(fields, 'path', field => field.serialize(item));
+    const creatable = fields
+      .filter(({ isPrimaryKey }) => !isPrimaryKey)
```
jesstelford: This inadvertently fixes a bug introduced in #1519 where the Id field would still be visible when doing a nested create from an item details page. 🎉

test-projects/access-control/cypress/integration/field/admin-ui.js
```diff
@@ -152,42 +203,55 @@ describe('Access Control Fields > Admin UI', () => {
...
+        fieldAccessVariations
+          .filter(({ update, read }) => !update && read)
+          .forEach(access => {
+            const field = getFieldName(access);
+            cy.get(`label[for="ks-input-${field}"]`)
+              .should('not.exist')
+              .then(() => cy.get(`#ks-input-${field}`).should('not.exist'));
+          });
+      });
+
```
jesstelford: I unrolled these tests as I needed to `.skip` specific ones instead of whole categories of tests.

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -48,12 +48,14 @@ class CreateItemModal extends Component {
       return;
     }
 
-    const nonPrimaryKeyFields = fields.filter(({ isPrimaryKey }) => !isPrimaryKey);
+    const creatable = fields
+      .filter(({ isPrimaryKey }) => !isPrimaryKey)
```
thiagodebastos: ~Is `isPrimaryField` an optional `boolean` that is always `true` if it exists on an object?
If so then this part of the function tells me that we are creating an array of `field`s who's~
This is hard for me to follow, what is `createTable` doing?

thiagodebastos: I'm going to guess that the first filter returns items without a `isPrimaryKey` prop, which when exists is always true.

jesstelford: Yep, that's right; it checks to see if `isPrimaryKey` exists and/or is not truthy. (ie; we want to filter out anything which might be flagged as `isPrimaryKey`)

PR #1526, opened 2019-08-23
List Plugin refactor for tracking for #1291

gautamsi:
blocked by: #1291 

This PR refactors #1291. `atTracking` and `byTracking` is now uses `trackingBase`. `trackingbase` is higher order function, which returns a plugin for tracking a field on create or update operation based on config. 

### Config Options for `trackingBase({...})`:

`fieldConfig`: [Required] Field config, similar to field as in list schema
`createField`: [Required*] Name of the field for create hook
`createFieldFunc`: [Required*] function to return value for create operation, this function accepts parameters same as resolveInput hook
`updateField`: [Required*] Name of the field for update hook
`updateFieldFunc`: [Required*] function to return value for update operation, this function accepts parameters same as resolveInput hook

> only one of `createField, createFieldFunc` or `updateField, updateFieldFunc` combination is required, both combination can be provided.

### Example
this is now possible:
```js
const { uptime } = require('os');
const { trackingBase } = require('@keystone-alpha/list-plugins');
const sysUptimeTracker = trackingBase({ // track system uptime when item is created or updated
  fieldConfig: {
    type: Text,
  },
  createdField: 'uptimeWhenCreated',
  createFieldFunc: () => uptime(),
  updatedField: 'uptimeWhenUpdated',
  updateFieldFunc: () => uptime(),
});
keystone.createList('ListWithPlugin', {
  fields: {
    // ... fields
  },
  plugins: [sysUptimeTracker],
});
```


MadeByMike:
@gautamsi I don't understand, why do we want uptime?

gautamsi:
nope, it is just an example to showcase power of `trackingBase` it could be anything for the devs to use. like: hostname at the time of creation and updation if there are multiple containers connecting to same DB. depends on the need of anyone.

MadeByMike:
@gautamsi I discussed this with you in back channels and have thought about it some more and I'm going to close this issue. I'll add my reasoning here incase anyone else comes across it:

Plugins are the place to provide more "opinionated" starting points for applications. I.e. "We assume you want our standard tracking features and have a User list".

These kind of assumptions are not made in the core infrastructure and therefore we try to make this more configurable and flexible. List plugins are one of the mechanisms we provide for this. 

For these reasons, I don't think the kind of configurability suggested here is what we want in 'core' plugins and we definitely think tracking should be one of those.

My preference would be to have many simple plugins over one complex plugin that does it all. I'd strongly recommend that to other plugin developers too but of course, in contrib land that's up to them. :)

PR #1527, opened 2019-08-23
CloudinaryImage & File fields no longer lose their value during update mutations

jesstelford:
A recent change to ensure `resolveInput()` is run on every field every mutation surfaced a known bug where the `File` (and hence the `CloudinaryImage`) type would incorrectly return `null` instead of `undefined`.

ie; Save a cloudinary image, then updating any other fields would cause the cloudinary image to be reset to `null`.

This PR fixes that bug.

**Before**
image-upload-bug

**After**
image-upload-fixed


jesstelford:
I'm going to merge this as it's a critical bug fix, and @gautamsi gave me a 👍 🎉

PR #1532, opened 2019-08-26
Add a keystone.extendGraphQLSchema() method for custom types, queries and mutations.

timleslie:
This PR adds a new method for adding custom types, queries and mutations to the graphQL schema.

```javascript
keystone.extendGraphQLSchema({
  types: ['type Foo { foo: Int }', ...],
  queries: [{ schema: '...', resolver: () => {} }, ...],
  mutations: [{ schema: '...', resolver: () => {} }, ...],
});
```

It also removes the `queries` and `mutations` options from the `createList()` API.

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -45,6 +45,9 @@ module.exports = class Keystone {
     this.lists = {};
     this.listsArray = [];
     this.getListByKey = key => this.lists[key];
+    this._extendedTypes = [];
+    this._extendedQueries = [];
+    this._extendedMutations = [];
     this._graphQLQuery = {};
     this.registeredTypes = new Set();
     this.eventHandlers = { onConnect };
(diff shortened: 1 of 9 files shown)


packages/keystone/lib/Keystone/index.js
```diff
@@ -155,8 +164,7 @@ module.exports = class Keystone {
     // Deduping here avoids that problem.
     return [
       ...unique(flatten(this.listsArray.map(list => list.getGqlTypes({ skipAccessControl })))),
-      ...unique(flatten(this.listsArray.map(list => list.getGqlQueryTypes()))),
-      ...unique(flatten(this.listsArray.map(list => list.getGqlMutationTypes()))),
+      ...this._extendedTypes,
```
jesstelford: We should `unique` these still.
```suggestion
      ...unique(this._extendedTypes),
```

MadeByMike:
I like this a lot 👍

packages/keystone/lib/Keystone/index.js
```diff
@@ -316,11 +330,77 @@ module.exports = class Keystone {
...
+                      return Promise.reject(
+                        new Error(
+                          'No executable schema is available. Have you setup `@keystone-alpha/app-graphql`?'
+                        )
+                      );
+                    }
+
+                    return graphQLQuery(queryString, passThroughContext, variables);
+                  },
+                }),
```
jesstelford: Some dupe code here ☝️ can we factor that out?

timleslie: Yep, this is also duplicated in `List/index.js`, so I just need to work out the right place to factor this out to 👍 Wanted to get the API 👍 from everyone before I spent too long on polish :-)

packages/keystone/lib/Keystone/index.js
timleslie: Good point 👍

thiagodebastos:
This is looking great @timleslie !

PR #1536, opened 2019-08-27
Bind context to query function provided to resolvers

timleslie:


MadeByMike:
Why is this WIP @timleslie Looks good - Do you need to test more?

timleslie:
Needs testing and changeset and testing :-)

PR #1556, opened 2019-08-29
Updates to most API docs

MadeByMike:
Progressive updates to most API docs.

Still to do in a future PR are updates to field types and guides.

The PR contains only docs.


docs/api/access-control.md
```diff
@@ -1,68 +1,37 @@
...
-# Access Control API
+# Access Control
 
 Control who can do what with your GraphQL API.
 
-_Note: This is the API documentation for Access Control. For getting started,
-see the [Access Control Guide](../guides/access-control.md) or the
-[Authentication Guide](../guides/authentication.md)._
+_Note_: This is the API documentation for Access Control. For getting started,
+see the [Access Control Guide](/guides/access-control.md) or the
```
jesstelford: These link changes will mean they no longer work on GitHub which we were trying to maintain for folks browsing the source. We have a gatsby plugin which rewrites them to work for the website on build.

docs/api/access-control.md
```diff
@@ -472,357 +441,3 @@ _NOTE:_ Even when returning `false`, this field _will_ be included in GraphQL
...
-const publicSchema = keystone.createGraphQLMiddleware();
-
-app.use('/graphql', (req, res, next) => {
-  // Assuming req.user is set by some earlier authentication middleware
-  if (req.user.role === 'ADMIN') {
-    return adminSchema(req, res, next);
-  }
-  publicSchema(req, res, next);
-});
-```
```
jesstelford: Happy to remove this 👍 

Did you capture it somewhere else?

docs/api/authentication.md
```diff
@@ -0,0 +1,47 @@
...
+
+Authentication strategies allow users to identify themselves to Keystone. This can be used to restrict access to the AdminUI, and to configure [access controls](/guides/access-control/).
+
+- For password logins see: [`auth-password`](/keystone-alpha/auth-password/)
+- For social logins using [Passport.js](http://www.passportjs.org/) see: [`auth-passport`](/keystone-alpha/auth-passport/)
+
+## Usage
+
+```javascript
+const authStrategy = keystone.createAuthStrategy({
```
jesstelford: I like to include the imports of things that are the focus of code snippets to ensure no one gets confused when copy+pasta-ing.

```suggestion
const { PasswordAuthStrategy } = require('@keystone-alpha/auth-password');

const authStrategy = keystone.createAuthStrategy({
```

docs/api/authentication.md
```diff
@@ -0,0 +1,47 @@
...
+  apps: [new AdminUIApp({ authStrategy })],
+};
+```
+
+## Config
+
+| Option | Type           | Default    | Description                                                        |
+| ------ | -------------- | ---------- | ------------------------------------------------------------------ |
+| `type` | `AuthStrategy` | (required) | A valid authentication strategy.                                   |
+| `list` | `String`       | (required) | The list that contains and authenticated item, for example a user. |
```
jesstelford:
```suggestion
| `list` | `String`       | (required) | The list that contains and authenticated item, for example a user. |
| `config` | `Object`       | undefined | Different authentication strategies may have additional config options. See the documentation for individual authentication strategies for more details. |
```

docs/api/authentication.md
```diff
@@ -0,0 +1,47 @@
...
+
+_Note_: Different authentication strategies may have additional config options. See the documentation for individual authentication strategies for more details.
```
jesstelford:
```suggestion
```

docs/api/create-list.md
```diff
@@ -1,66 +1,73 @@
 <!--[meta]
 section: api
-title: createList
+title: Create List
+order: 2
 [meta]-->
 
-# keystone.createList(name, options)
+# createList(name, options)
```
jesstelford: I'd like to keep the `keystone.` part on this heading so it's obvious what method we're talking about (as it's the main page heading).

```suggestion
# `keystone.createList(name, options)`
```

docs/api/create-list.md
```diff
@@ -1,66 +1,73 @@
...
+| `adapterConfig` | `Object`                            |                               | Override the adapter config options for a specific list.       |
+| `itemQueryName` | `String`                            |                               | Changes the _item_ name in GraphQL queries and mutations.      |
+| `listQueryName` | `String`                            |                               | Changes the _list_ name in GraphQL queries and mutations.      |
+| `singular`      | `String`                            |                               | Specify a singular noun for Keystone to use for the list.      |
+| `plural`        | `String`                            |                               | Specify a plural for Keystone to use for the list.             |
+| `path`          | `String`                            |                               | Changes the path in the Admin UI.                              |
+| `plugins`       | `Array`                             | `[]`                          | An array of `plugins` that can modify the list config.         |
```
jesstelford: A lot of these are missing their `default` value which makes it look like they're required.

docs/api/create-list.md
```diff
@@ -1,66 +1,73 @@
...
-where it's possible a user is never able to create new items because they don't
-have access to create a certain field, which attempts to use the `defaultValue`,
-but then throws an error because no `defaultValue` has been set.
+See: [Fields](/keystone-alpha/fields/) for more information on configuring field options.
 
-### schemaDoc
+### `schemaDoc`
 
-A description for the list. Used in the Admin UI.
+A description for the list used in the AdminUI.
```
jesstelford:
```suggestion
A description for the list used in the GraphQL schema.
```

This isn't actually used in the Admin UI, whoops!

docs/api/create-list.md
MadeByMike: This is why I've chose to follow a format where usage always follows the main heading

docs/api/create-list.md
MadeByMike: It should be 🤣

docs/api/create-list.md
MadeByMike: I put (required) when required. What would you suggest for no default? I originally thought `null` but... technically the default is not `null` I think nothing is best

docs/api/authentication.md
MadeByMike: Config is not a property of the password password auth. Only `list` and `type` are common properties.

docs/api/create-list.md
1337cookie: If the default is `undefined` would it make sense to specify how it will be resolved? 

schemaDoc for example: 
 `undefined` Will display in the docs as `A keystone list` 
adapterConfig:
`undefined` The `defaultAdapter`  specified when creating the Keystone instance will be used.
singular
`undefined` pluralize npm package will generate a singular label.

Or something to that note?

MadeByMike: Yeah I think that makes sense in a lot of cases

docs/api/access-control.md
MadeByMike: Yep in notion

docs/api/create-list.md
MadeByMike: I'm going to come back and revisit this later.

PR #1091, opened 2019-05-05
[WIP] Remove createItems

timleslie:

PR #1557, opened 2019-08-30
 First pass at a Programatic keystone.executeQuery method

jesstelford:
Synthesising a bunch of discussions from #1275 & #1091, I've attempted to implement a `keystone.executeQuery` method which takes into account as much of the possible use-cases I can think of.

What do you think?

From the docs:

## `keystone.executeQuery(queryString, options)`

Use this method to execute queries or mutations directly against a Keystone
instance.

> ℹ️ When querying or mutating via `keystone.query`, there are differences to keep
in mind:
> - No access control checks are run (everything is set to `() => true`)
> - The `context.req` object is set to `{}` (you can override this if necessary,
  see options below)
> - Attempting to authenticate will throw errors (due to `req` being mocked)

Returns a Promise representing the result of the given query or mutation.

### `queryString`

A graphQL query string. For example:

```graphql
query {
  allTodos {
    id
    name
  }
}
```

Can also be a mutation:

```graphql
mutation newTodo($name: String) {
  createTodo(name: $name) {
    id
  }
}
```

### `options`

| Option    | Type     | Default | Description                                                            |
| --------- | -------- | ------- | ---------------------------------------------------------------------- |
| `variables` | `Object` | `{}`  | The variables passed to the graphql query for the given queryString. |
| `context` | `Object` | `{}`  | Override the default `context` object passed to the GraphQL engine. Useful for adding a `req` or setting the `schemaName` |

---

**Todo**:
- [x] ~Get tests passing~
- [x] ~Split into logical PR chunks~
- [x] ~Add tests for new method(s)~
- [x] ~Add a changeset~
- [x] ~Move PR out of draft~

### DIFF
--- a/packages/keystone/README.md
+++ b/packages/keystone/README.md
@@ -188,8 +188,8 @@ Disconnect all adapters.
 Use this method to execute queries or mutations directly against a Keystone
 instance.
 
-> ℹ️ When querying or mutating via `keystone.query`, there are differences to keep
-> in mind:
+> ℹ️ When querying or mutating via `keystone.executeQuery`, there are differences
+> to keep in mind:
 >
 > - No access control checks are run (everything is set to `() => true`)
 > - The `context.req` object is set to `{}` (you can override this if necessary,
--- a/packages/keystone/README.md
+++ b/packages/keystone/README.md
@@ -16,6 +16,7 @@ order: 1
 | `prepare`             | Manually peapare Keystone middlewares.                                       |
 | `createItems`         | Add items to a Keystone list.                                                |
 | `disconnect`          | Disconnect from all adapters.                                                |
+| `executeQuery`        | Run GraphQL queries and mutations directly against a Keystone instance.      |
 
 <!--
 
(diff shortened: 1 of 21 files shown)


packages/test-utils/lib/test-utils.js
```diff
@@ -175,5 +202,6 @@ module.exports = {
   setupServer,
   multiAdapterRunners,
   graphqlRequest,
+  networkedGraphqlRequest,
```
timleslie: Ooh, nice addition 👍

packages/keystone/README.md
```diff
@@ -181,3 +182,48 @@ See: [Custom Server](https://v5.keystonejs.com/guides/custom-server).
 ## disconnect()
 
 Disconnect all adapters.
+
+## `keystone.executeQuery(queryString, options)`
+
+Use this method to execute queries or mutations directly against a Keystone
+instance.
+
+> ℹ️ When querying or mutating via `keystone.query`, there are differences to keep
```
timleslie: `keystone.query` -> `keystone. executeQuery`

jesstelford: Good catch! Fixed.

PR #1560, opened 2019-09-02
Move Session handling logic from graphQL app to keystone package

jesstelford:
This change is for 2 reasons:

1. Sessions are a Keystone-wide thing, and should be handled as such. We're already treating all the other sesssion/auth stuff as a Keystone-wide thing, so no the session middlewares are too.
2. It's needed for https://github.com/keystonejs/keystone-5/pull/1557

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -17,6 +17,11 @@ const {
   validateFieldAccessControl,
   validateListAccessControl,
 } = require('@keystone-alpha/access-control');
+const {
+  startAuthedSession,
+  endAuthedSession,
+  commonSessionMiddleware,
+} = require('@keystone-alpha/session');
 
 const {
   unmergeRelationships,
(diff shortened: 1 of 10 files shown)


packages/keystone/lib/Keystone/index.js
```diff
@@ -63,6 +72,64 @@ module.exports = class Keystone {
     }
   }
 
+  getCookieSecret() {
+    if (!this._cookieSecret) {
+      throw new Error('No cookieSecret set in Keystone constructor');
```
MadeByMike: This is good! Can we log an example of how to set it too?

packages/keystone/lib/Keystone/index.js
```diff
@@ -63,6 +72,64 @@ module.exports = class Keystone {
...
+    }
+    return this._cookieSecret;
+  }
+
+  // The GraphQL App uses this method to build up the context required for each
+  // incoming query.
+  // It is also used for generating the `keystone.query` method
+  getGraphQlContext({ schemaName = 'admin', req = {}, skipAccessControl = false } = {}) {
```
timleslie: Should mention this changed API in the changeset.

PR #1564, opened 2019-09-02
Prepare release

jesstelford:
### Bug fixes

- [12668191](https://github.com/keystonejs/keystone-5/commit/12668191): Fix assertion logic of cookieSecret deprecation message


jesstelford:
NOTE: I have unpublished `@keystone-alpha/auth-passport@4.0.0` as it's completely broken.

PR #1566, opened 2019-09-02
Remove `hostURL` and `apiPath` options from `PassportAuthStrategy` constructor

jesstelford:
Thanks to usage of the new `keystone.executeQuery()` internally.

I've tested this against the `test-projects/social-auth`, and it all works great 👍 

You'll notice I've added access control checks in the test project. These checks would previously cause the social sign in flow to error out because it made a networked graphQL request as an anonymous user 😱  Now it hits `keystone.executeQuery` directly, so doesn't get blocked by Access Control 👍

### DIFF
--- a/packages/auth-passport/README.md
+++ b/packages/auth-passport/README.md
@@ -100,6 +100,7 @@ const cookieSecret = '<Something super secret>';
 const keystone = new Keystone({
   name: 'Login With Google Example',
   adapter: new MongooseAdapter(),
+  cookieSecret,
 });
 
 keystone.createList('User', {
@@ -120,9 +121,6 @@ const googleStrategy = keystone.createAuthStrategy({
     idField: 'googleId',
     appId: '<Your Google App Id>',
     appSecret: '<Your Google App Secret>',
-    cookieSecret,
-    hostURL: 'http://localhost:3000',
-    apiPath: '/admin/api',
     loginPath: '/auth/google',
     callbackPath: '/auth/google/callback',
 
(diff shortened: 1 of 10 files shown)


packages/auth-passport/README.md
```diff
@@ -147,7 +145,7 @@ const googleStrategy = keystone.createAuthStrategy({
 
 module.exports = {
   keystone,
-  apps: [new GraphQLApp({ cookieSecret }), new AdminUIApp()],
```
jesstelford: These examples are out of date, I missed them in #1560

packages/auth-passport/lib/Passport.js
```diff
@@ -64,6 +61,20 @@ class PassportAuthStrategy {
...
+        'The `hostURL` config option for `PassportAuthStrategy` has been removed and is no longer necessary.'
+      );
+    }
+
+    // NOTE: Remove after March 1st, 2020 (ie; 6mo since deprecation)
+    if (typeof config.apiPath !== 'undefined') {
+      console.warn(
+        'The `apiPath` config option for `PassportAuthStrategy` has been removed and is no longer necessary.'
+      );
+    }
```
jesstelford: I think patch release + `warn` is a good combo here. Any objections?

packages/auth-passport/lib/Passport.js
```diff
@@ -453,8 +479,7 @@ class PassportAuthStrategy {
 
     // Here we create both the Passport Session Item and the User Item
     // in KS5 as a single, nested mutation.
-    const sessionItem = await request(
```
gautamsi: nice, this is something I did not like in the first place, glad to see it now removed

PR #1568, opened 2019-09-02
[RFC] Password Auth identity filtering

MadeByMike:
This is a pretty simple change and it adds the ability to filter a list to exclude users from authentication.

For example I wanted to restrict users that can login to the AdminUI so I made this `identityFilter`  only try to match admin users:

```javascript
  type: PasswordAuthStrategy,
  list: 'User',
 config: {
   identityFilter: ({ identityField, identity }) => ({
      AND: [{ [identityField]: identity }, { isAdmin: true }],
    })
  }
});
```

Any thoughts

### DIFF
--- a/packages/auth-password/index.js
+++ b/packages/auth-password/index.js
@@ -12,6 +12,9 @@ class PasswordAuthStrategy {
       identityField: 'email',
       secretField: 'password',
       protectIdentities: false,
+      identityFilter: ({ identityField, identity }) => ({
+        [identityField]: identity,
+      }),
       ...config,
     };
   }
@@ -52,8 +55,13 @@ class PasswordAuthStrategy {
       );
     }
 
+    const identityFilter = this.config.identityFilter({
+      identityField,
+      identity,
+    });
+
     // Match by identity
-    const results = await list.adapter.find({ [identityField]: identity });
+    const results = await list.adapter.find(identityFilter);
 
     // If we failed to match an identity and we're protecting existing identities then combat timing
     // attacks by creating an arbitrary hash (should take about as long has comparing an existing one)
--- a/.changeset/giant-readers-argue/changes.json
+++ b/.changeset/giant-readers-argue/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/auth-password", "type": "minor" }], "dependents": [] }
--- a/.changeset/giant-readers-argue/changes.md
+++ b/.changeset/giant-readers-argue/changes.md
@@ -6,8 +6,10 @@ The `identityFilter` has been added to configurations options:
 const authStrategy = keystone.createAuthStrategy({
   type: PasswordAuthStrategy,
   list: 'User',
-  identityFilter: ({ identityField, identity }) => ({
-    AND: [{ [identityField]: identity }, { isAdmin: true }],
-  }),
+  config: {
+    identityFilter: ({ identityField, identity }) => ({
+      AND: [{ [identityField]: identity }, { isAdmin: true }],
+    }),
+  },
 });
 ```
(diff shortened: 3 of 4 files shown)


gautamsi:
this affects only the login to `admin-ui`, correct? good to have this configurable, I am hoping to have admin customization option in future (including branding) to have normal user login and change few things based on access control.

MadeByMike:
@gautamsi not just the adminUI but anywhere that auth strat is used.

gautamsi:
usually you want to keep normal users away from `admin-ui`, in rare cases you would not want a normal user to login. there is no point having a user who can not login if they are not admin. am I missing anything?

MadeByMike:
In a lot of applications you have users with different levels of access. You may want to keep some of those out of the admin UI. 

In other case is you might want to keep archived or inactive users accounts around so that they show up in lists (author of a post etc...) but you don't wan these users logging in anymore.

MadeByMike:
After more testing I found that you can only have one auth strat. Doing the following results options from the last call to `createAuthStrategy` overriding `adminAuthStrategy`. 

```

const adminAuthStrategy = keystone.createAuthStrategy({
  type: PasswordAuthStrategy,
  list: "User",
  config: {
    identityFilter: ({ identityField, identity }) => ({
      AND: [{ [identityField]: identity }, { isAdmin: true }]
    })
  }
});

// User Auth Strat
keystone.createAuthStrategy({
  type: PasswordAuthStrategy,
  list: "User"
});
```

Having only one auth strat is possibly not a bad idea? Although some apps might want both social and password? 

I think that could be a separate issue and this PR could be resolved by adding some kind of access control to the entire adminUI app.

gautamsi:
something I envisioned earlier, did not find time yet,
see https://github.com/keystonejs/keystone-5/issues/878

some inspiration from this (for ui):
https://github.com/estilles/keystone-social-login

basically, as soon as you add more auth strategy, it should load a view (similar to views for field, how we load them from custom field types) and add the login method to `admin-ui` login page.

gautamsi:
> In other case is you might want to keep archived or inactive users accounts around so that they show up in lists (author of a post etc...) but you don't wan these users logging in anymore.

I don't remember seeing this, but do we have a login hook with authstrategy or keystone in general? like I would want to throw if user is stale or disabled or pending verification?

jesstelford:
> you can only have one auth strat

This is slightly incorrect. You can have only one _of each type_ of auth strategy.

> do we have a login hook with authstrategy

This would be a much nicer solution for this particular problem - it provides a place to do other actions at auth time if the app developer wants to.

packages/auth-password/index.js
```diff
@@ -12,6 +12,9 @@ class PasswordAuthStrategy {
       identityField: 'email',
       secretField: 'password',
       protectIdentities: false,
+      identityFilter: ({ identityField, identity }) => ({
+        [identityField]: identity,
```
jesstelford: This would be very easy to accidentally overwrite and completely break sign in.

Also, what about `auth-passport`? Does it need this kind of thing too?

gautamsi: if we have login hook, (before issuing token/cookies) all auth strategy would be covered not just password. then admin-ui may have another hook for preventing users from login to actual admin-ui pages not just login. 

in ks4, if you are not allowed to login to admin-ui, it would just sit on login page saying, you are already logged in, do you want to sign out (something like that), similar solution would be much appreciated. I earlier suggested to have `redirectUrl` for admin-ui options in case user does not have admin-ui access or they are trying to access admin-ui only for authentication.

MadeByMike:
Some really good points here I still think this might be worth pursuing and perhaps a hook is the best way to go! I'm going to close this and capture comments in an issue.

Regarding access filtering access to the to the admin UI which seem like a different but parallel issue, I've split this off into a new PR. #1570 1570

PR #1570, opened 2019-09-02
[RFC] Filter admin access

MadeByMike:
Added a filter option to control access to the adminUI. I still think a hook is needed as per #1568 but this is a separate issue. 

I'm not the best with Express routes and it always does my head in a little so if someone has the time to suggest a neater approach I'd be much appreciated.

Still to do is docs, changeset and everything...

Looking for feedback on general idea.

```
module.exports = {
  keystone,
  apps: [
    new GraphQLApp(),
    new AdminUIApp({
      authStrategy: adminAuthStrategy,
      filterAdminAccess: user => {
        console.log({ user });
        return false;
      }
    })
  ]
};
```


MadeByMike:
This has been working really great for me. If I've got in principal approval I'll add docs and a changeset.

MadeByMike:
> This is probably a "let's fix it later" thing but from what I understand from this code, if a user doesn't have admin access, they'll be redirected to a completely empty page which I imagine would be pretty confusing, could we have a "you don't have permission to access the admin ui" page or something?

Totally agree!

packages/app-admin-ui/index.js
```diff
@@ -52,6 +54,15 @@ class AdminUIApp {
     };
   }
 
+  isAccessAllowed(req) {
+    return (
+      req.user &&
+      this._isAccessAllowed({ authentication: { item: req.user, listKey: req.authedListKey } }) &&
```
jesstelford: I made this API consistent with the access control APIs

packages/app-admin-ui/index.js
```diff
@@ -52,6 +54,15 @@ class AdminUIApp {
...
+      req.session.audiences &&
+      req.session.audiences.includes('admin')
```
jesstelford: This is potentially redundant now? @timleslie this was part of your initial experimental audiences work - is it useful to keep?

packages/app-admin-ui/index.js
```diff
@@ -149,22 +185,26 @@ class AdminUIApp {
     }
     const secureBuiltRoot = path.join(builtAdminRoot, 'secure');
     const secureStaticMiddleware = express.static(secureBuiltRoot);
-    const secureFallbackMiddleware = fallback('index.html', { root: secureBuiltRoot });
+    const secureFallbackMiddleware = fallback('index.html', {
+      root: secureBuiltRoot,
+    });
```
jesstelford: Thanks prettier ¯\\\_(ツ)\_/¯

packages/app-admin-ui/index.js
```diff
@@ -129,16 +138,43 @@ class AdminUIApp {
...
+      // All unauthenticated requests go to the signin page
+      res.format({
+        // For everything else, we have middleware to handle it later down the
+        // line
+        // From the docs: "If the header is not specified [or is */*], the first
+        // callback is invoked."
+        default: () => {
+          next();
+        },
+        // For everything else, we have middleware to handle it later down the
```
MadeByMike: Did you mean to copy this comment?

packages/app-admin-ui/index.js
```diff
@@ -176,33 +216,36 @@ class AdminUIApp {
...
 
     _app.use(this.adminPath, app);
 
     if (this.enableDefaultRoute) {
       // Attach this last onto the root so the `this.adminPath` can overwrite it
       // if necessary
-      _app.get('/', (req, res) => res.sendFile(path.resolve(__dirname, './server/default.html')));
+      _app.get('/', (req, res) => {
+        if (!this.isAccessAllowed(req)) {
+          res.sendFile(path.resolve(__dirname, './server/default.html'));
```
jesstelford: We should still send this even when the user is allowed access; if they deployed with the defaults of `/admin`, but have `/` bookmarked, they'll then be shown a blank screen, for example.

packages/app-admin-ui/index.js
```diff
@@ -244,13 +287,18 @@ class AdminUIApp {
       const publicHotMiddleware = webpackHotMiddleware(publicCompiler, webpackHotMiddlewareConfig);
 
       // app.use(adminMiddleware);
```
MadeByMike: Should this old comment go?

packages/app-admin-ui/index.js
timleslie: Um, I'd prefer to keep it for now and either get it all to work in the future or tear it all out in one big hit if we decide its redundant.

jesstelford:
Looks like this:

admin-filter

packages/app-admin-ui/index.js
```diff
@@ -176,33 +216,36 @@ class AdminUIApp {
...
     }
 
     _app.use(this.adminPath, app);
 
     if (this.enableDefaultRoute) {
       // Attach this last onto the root so the `this.adminPath` can overwrite it
       // if necessary
-      _app.get('/', (req, res) => res.sendFile(path.resolve(__dirname, './server/default.html')));
+      _app.get('/', (req, res) => {
+        if (!this.isAccessAllowed(req)) {
```
MadeByMike:
```suggestion
    
```

packages/app-admin-ui/index.js
```diff
@@ -176,33 +216,36 @@ class AdminUIApp {
...
+          res.sendFile(path.resolve(__dirname, './server/default.html'));
+        }
```
MadeByMike:
```suggestion
        
```

packages/app-admin-ui/index.js
jesstelford:
```suggestion
        res.sendFile(path.resolve(__dirname, './server/default.html'));
```

packages/app-admin-ui/index.js
jesstelford: Whoops, nope. Good catch!

packages/app-admin-ui/index.js
jesstelford: 💥

PR #1586, opened 2019-09-04
Docs/quick start

simonswiss:
Quick start guide suggested enhancements:

- consolidates the *requirements* in one place (some of it was in the next step)
- adds a couple of heading tags for readability
- adds a "what next" suggestion for the readers discovering Keystone


simonswiss:
Haven't fixed the broken image yet as this PR focuses on the copy. There is an open issue https://github.com/keystonejs/keystone-5/issues/1581 for broken images.

PR #1591, opened 2019-09-09
New NuxtJS app

simonswiss:
This PR adds a new `app-nuxt` package which is used to generate a new NuxtJS starter project via the `create-keystone-app` CLI.

Cannot actually test the CLI command before the package is released, but generating the app from inside `packages/create-keystone-app/example-projects/nuxt` works well!

## New folders/packages:

- `packages/app-nuxt`
- `packages/create-keystone-app/example-projects/nuxt`


packages/create-keystone-app/example-projects/nuxt/index.js
```diff
@@ -0,0 +1,29 @@
...
+const { GraphQLApp } = require('@keystone-alpha/app-graphql');
+const { AdminUIApp } = require('@keystone-alpha/app-admin-ui');
+const { NuxtApp } = require('@keystone-alpha/app-nuxt');
+
+const nuxtConfig = {
+  srcDir: 'src',
+  buildDir: 'dist',
+};
+
+const { MongooseAdapter: Adapter } = require('@keystone-alpha/adapter-mongoose');
```
mitchellhamilton: Doesn't this need a special comment like this? https://github.com/keystonejs/keystone-5/blob/eb84ddead30772fb86fff955e97697129a3848a7/packages/create-keystone-app/example-projects/starter/index.js#L6-L9

gautamsi:
should this (`app-nuxt`) be part of this repo or a separate repo?

jesstelford:
@gautamsi

> should this (`app-nuxt`) be part of this repo or a separate repo?

Good question! So far we've been keeping the apps in this repo (see `app-static` & `app-next`), so I think it's ok in this instance, but we might want to move them out into their own place in the future to keep the repo clean.

MadeByMike:
I'd be pro moving them out in the future

PR #1593, opened 2019-09-09
Create ks app nuxt

simonswiss:
This should be merged _after_ #1591 

This PR brings a new example project for NuxtJS in the `create-keystone-app` package.


packages/keystone/bin/utils.js
```diff
@@ -97,6 +97,7 @@ async function executeDefaultServer(args, entryFile, distDir, spinner) {
 
   spinner.succeed('Initialised Keystone instance');
   spinner.start('Connecting to database');
+  console.log('\n');
```
jesstelford: Why is this here? It may collide with other spinner output.

packages/create-keystone-app/example-projects/nuxt/index.js
```diff
@@ -0,0 +1,29 @@
...
+const { GraphQLApp } = require('@keystone-alpha/app-graphql');
+const { AdminUIApp } = require('@keystone-alpha/app-admin-ui');
+const { NuxtApp } = require('@keystone-alpha/app-nuxt');
+
+const nuxtConfig = {
+  srcDir: 'src',
+  buildDir: 'dist',
+};
+
+const { MongooseAdapter: Adapter } = require('@keystone-alpha/adapter-mongoose');
```
jesstelford: This needs some magic comments to ensure everything works correctly when generated via the CLI. See the `static` project for an example: https://github.com/keystonejs/keystone-5/blob/5fc2bd597ed8f6b0dfaf7761e1f20bc2ad14caf5/packages/create-keystone-app/example-projects/todo/index.js#L7-L10

packages/keystone/bin/utils.js
MadeByMike: Without it the logging from Nuxt ends up on the same line as 'Connecting to database'

jesstelford: Hmm 🤔 Does it interfere with the spinner from spinning correctly? Got a recording of what it looks like?

Perhaps we can insert the newline at the end of the `spinner.start`?

PR #1601, opened 2019-09-09
Add support for grouping access control by schema name

timleslie:
`parseListAccess` and `parseFieldAccess` now take `schemaNames` as an argument, and return a nested access object, with the `schemaNames` as keys.

For example,

```js
parseListAccess({ defaultAccess: false, access: { public: true }, schemaNames: ['public', 'internal'] }
```

will return

```js
{
  public: { create: true, read: true, update: true, delete: true },
  internal: { create: false, read: false, update: false, delete: false },
}
```

These changes are backwards compatible with regard to the `access` argument, so

```js
const access = { create: true, read: true, update: true, delete: true };
parseListAccess({ access, schemaNames: ['public', 'internal'] }
```

will return

```js
{
  public: { create: true, read: true, update: true, delete: true },
  internal: { create: true, read: true, update: true, delete: true },
}
```

### DIFF
--- a/packages/access-control/lib/access-control.js
+++ b/packages/access-control/lib/access-control.js
@@ -75,7 +75,9 @@ const parseAccess = ({
   // Check that none of the schemaNames match the accessTypes
   if (intersection(schemaNames, accessTypes).length > 0) {
     throw new Error(
-      `${JSON.stringify(intersection(schemaNames, accessTypes))} are reserved words and cannot be used as schema names.`
+      `${JSON.stringify(
+        intersection(schemaNames, accessTypes)
+      )} are reserved words and cannot be used as schema names.`
     );
   }
 
--- a/packages/access-control/lib/access-control.js
+++ b/packages/access-control/lib/access-control.js
@@ -75,13 +75,13 @@ const parseAccess = ({
   // Check that none of the schemaNames match the accessTypes
   if (intersection(schemaNames, accessTypes).length > 0) {
     throw new Error(
-      `Invalid schemaNames: ${JSON.stringify(intersection(schemaNames, accessTypes))}`
+      `${JSON.stringify(intersection(schemaNames, accessTypes))} are reserved words and cannot be used as schema names.`
     );
   }
 
-  const N = intersection(Object.keys(access), schemaNames).length;
+  const providedNameCount = intersection(Object.keys(access), schemaNames).length;
   const type = getType(access);
-  if (type === 'Object' && N === Object.keys(access).length) {
+  if (type === 'Object' && providedNameCount === Object.keys(access).length) {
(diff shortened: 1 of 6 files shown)


packages/access-control/lib/access-control.js
```diff
@@ -64,11 +64,64 @@ const parseAccess = ({
...
+  validateGranularType,
+}) => {
+  // Check that none of the schemaNames match the accessTypes
+  if (intersection(schemaNames, accessTypes).length > 0) {
+    throw new Error(
+      `Invalid schemaNames: ${JSON.stringify(intersection(schemaNames, accessTypes))}`
+    );
+  }
+
+  const N = intersection(Object.keys(access), schemaNames).length;
```
MadeByMike: `N`?

packages/access-control/lib/access-control.js
```diff
@@ -64,11 +64,64 @@ const parseAccess = ({
...
+  accessTypes,
+  access,
+  defaultAccess,
+  onGranularParseError,
+  validateGranularType,
+}) => {
+  // Check that none of the schemaNames match the accessTypes
+  if (intersection(schemaNames, accessTypes).length > 0) {
+    throw new Error(
+      `Invalid schemaNames: ${JSON.stringify(intersection(schemaNames, accessTypes))}`
```
jesstelford: Perhaps a more meaningful error message might be useful here?

```suggestion
      `${JSON.stringify(intersection(schemaNames, accessTypes))} are reserved words and cannot be used as schema names.`
```

packages/access-control/lib/access-control.js
jesstelford: What is `N`? Could we make this at least 3 characters long so searching for it is easier?
```suggestion
  const totalSchemasWithAccess = intersection(Object.keys(access), schemaNames).length;
```

packages/access-control/lib/access-control.js
```diff
@@ -64,11 +64,64 @@ const parseAccess = ({
...
+          access: access.hasOwnProperty(schemaName) ? access[schemaName] : defaultAccess,
+          defaultAccess,
+          onGranularParseError,
+          validateGranularType,
+        }),
+      }),
+      {}
+    );
+  } else if (type === 'Object' && N > 0) {
+    // If some are in, and some are out, throw an error!
```
jesstelford: Does this mean we can't do:

```
access: {
  public: { read: true },
  create: false
}
```
?

timleslie: Correct.

timleslie:
> LGTM! I'm on mobile so can't load the Circle tests; any false positives?

Will double check before merging 👍

PR #1604, opened 2019-09-10
Use schemaName based .access

timleslie:


packages/keystone/lib/List/index.js
```diff
@@ -219,7 +219,7 @@ module.exports = class List {
       listKey: key,
       access,
       defaultAccess: this.defaultAccess.list,
-    })['public'];
+    });
```
timleslie: This is the crux of the PR. We want to change `.access` to be something like `{ public: { read: true, ... } }`. All the other changes stem from this (and the corresponding change in the base field class).

For now we hard-code `schemaName = 'public'` in a bunch of places. These will be resolved in the next PR, which will make changes to the `Keystone` level APIs.

gautamsi: should public be defaults to read: true and rest false? just to be more secure by default

jesstelford: @gautamsi I believe that's coming in a future PR 👍

timleslie: Yep, the question of default access control is definitely on our minds at the moment and will be addressed once we have these mechanisms in place.

timleslie:
The `access-control` package handles all the cases of user configuration, and when this gets parsed it always gets returned in the shape `{ schemaName: { read: ... } }`. The previous PR which updated the parser checked all those conditions, and so the tests changed in this can assume the normalised shape for the actual `.access` property of lists and fields.

PR #1605, opened 2019-09-10
Add automatic releases

mitchellhamilton:
This only does the create a PR bit, not auto-publishing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Dependencies
         run: bolt
 
-      - name: Create Release Pull Request or Publish to npm
+      - name: Create Release Pull Request
         uses: changesets/action@master
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,18 +18,13 @@ jobs:
         with:
           version: 10.x
 
-      - name: Install Yarn
-        run: npm install --global yarn
+      - name: Install Bolt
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+          version: 10.x
+
+      - name: Install Bolt
+        run: npm install --global bolt
+
+      - name: Install Dependencies
+        run: bolt
+
+      - name: Create Release Pull Request
+        uses: changesets/action@master
```
timleslie: Is this a pre-canned action which already exists somewhere else?

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - master
```
timleslie: I wonder if we *always* want to do this on push to master? A lot of the time I'd prefer to group a whole bunch of PRs in a single combined release.

.github/workflows/release.yml
mitchellhamilton: Yes, I wrote it, https://github.com/changesets/action

timleslie: Aha, Excellent!

.github/workflows/release.yml
mitchellhamilton: It always runs on master but that doesn't mean it creates a release on every push. It creates a PR like the ones we have now so you can do exactly that, if there are new changesets on master, the PR will be updated.

timleslie: Ah, of course, very good! Let's ship it then :-)

Issue #1595, opened 2019-09-09
Make `keystone.sid` cookie secure in Prod mode


jesstelford:
We automatically set the `keystone.sid` cookie when performing an `authenticate<List>with<Strategy>`, but it's not set to `secure`. We should set it to `secure: true` when `NODE_ENV === 'production'`.

See: https://www.npmjs.com/package/express-session#cookiesecure

MadeByMike:
Fixed in #1612

PR #1612, opened 2019-09-11
Add cookieMaxAge and secureCookies options

### LINKED ISSUE #1596: Configurable cookie settings for `keystone.sid`
A core setting may be to set `maxAge` to force a re-login every day, etc.

This is similar to https://github.com/keystonejs/keystone-5/issues/1595 but a more generic "all the cookie settings", and perhaps this issue supersedes #1595.

MadeByMike:
Resolves #1596 and #1595

I followed the pattern of adding options to the keystone object. Not sure if I should have grouped these under cookies or something, we're getting quite a few config options here and I'm not sure these should be on the Keystone constructor?

We could probably save the above for a more general discussion in future.

For both options I've set defaults for `dev` and `production` - Can you check these are right?

Can I also get some help on how best to test this? Commits welcome.

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -45,7 +45,7 @@ module.exports = class Keystone {
     cookieSecret = 'qwerty',
     sessionStore,
     secureCookies = process.env.NODE_ENV === 'production', // Default to true in production
-    cookieMaxAge = process.env.NODE_ENV === 'production' ? 1000 * 60 * 60 * 24 : null,
+    cookieMaxAge = 1000 * 60 * 60 * 24 * 30, // 30 days
   }) {
     this.name = name;
     this.adapterConnectOptions = adapterConnectOptions;
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -63,6 +63,8 @@ module.exports = class Keystone {
     this._cookieMaxAge = cookieMaxAge;
     this._sessionStore = sessionStore;
     this.eventHandlers = { onConnect };
+    this._sessionStore = sessionStore;
+    this.registeredTypes = new Set();
 
     if (adapters) {
       this.adapters = adapters;
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -44,6 +44,8 @@ module.exports = class Keystone {
     onConnect,
     cookieSecret = 'qwerty',
     sessionStore,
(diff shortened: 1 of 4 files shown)


jesstelford:
Yeah, it's starting to get a bit crowded there.

I'm thinking perhaps we have a `session` config option which is an object that contains some things.

One thing I realised about #1596 is the `maxAge` doesn't actually invalidate the session. So it gives a false sense of security. Really we need to make sure the sessionStore is the one given that setting, but since we don't control that, we can't. Chicken & Egg.

packages/keystone/lib/Keystone/index.js
```diff
@@ -44,6 +44,8 @@ module.exports = class Keystone {
     onConnect,
     cookieSecret = 'qwerty',
     sessionStore,
+    secureCookies = process.env.NODE_ENV === 'production', // Default to true in production
+    cookieMaxAge = process.env.NODE_ENV === 'production' ? 1000 * 60 * 60 * 24 : null,
```
jesstelford: Why make `cookieMaxAge` different between environments? I'm happy if this one is defaulted to something like 30 days _always_, and folks can overwrite it for their app.

packages/keystone/lib/Keystone/index.js
```diff
@@ -544,7 +547,13 @@ module.exports = class Keystone {
       // Used by other middlewares such as authentication strategies. Important
       // to be first so the methods added to `req` are available further down
       // the request pipeline.
-      commonSessionMiddleware(this, this._cookieSecret, this._sessionStore),
+      commonSessionMiddleware({
```
jesstelford: This constitutes a `breaking` change to the `Session` package, so the changeset needs to be updated to reflect that.

packages/keystone/lib/Keystone/index.js
MadeByMike: I'm fine with that too. Sasa wanted one day and I thought it might be be annoying in dev but I'm happy to make it 30 days and we can change that for the app Sasa is working on?

packages/keystone/lib/Keystone/index.js
MadeByMike: Ahhh yes. A worthwhile change though?

jesstelford: Yeah, totally! Brings it more inline with the rest of our APIs 👍

timleslie:
> I'm thinking perhaps we have a `session` config option which is an object that contains some things.

Yeah, I'm a big fan of that kind of grouping. Doesn't have to be on this PR, but something to consider for the future.

PR #1613, opened 2019-09-11
Remove default schemaName from getGraphQlContext

timleslie:
This default is never used and will only lead bugs.


packages/app-graphql/README.md
```diff
@@ -39,7 +39,7 @@ module.exports = {
 | Option         | Type     | Default                               | Description                                                                                          |
 | -------------- | -------- | ------------------------------------- | ---------------------------------------------------------------------------------------------------- |
 | `cors`         | `Object` | `{ origin: true, credentials: true }` | Options passed directly to the express [cors](https://github.com/expressjs/cors) middleware package. |
-|  |
+|                |          |                                       |                                                                                                      |
```
MadeByMike: Delete the row? This leaves an empty row.

timleslie: *shrug* I'm just doing what prettier tells me to.

Issue #1616, opened 2019-09-11
`npm init keystone-app my-app` fails if you don’t have yarn installed


dzigg:
Followed the docs on creating a todo app ~ `https://v5.keystonejs.com/quick-start/`
Failed after "Copying project files"
![Screen Shot 2019-09-11 at 1 37 03 pm](https://user-images.githubusercontent.com/4436922/64666378-75072c80-d499-11e9-824a-121783d30615.png)

Ran `npm install` to install dependencies and the app worked fine.

MadeByMike:
Thanks @dzigg it's meant to catch an error if `yarn` is not installed and fallback to `npm`. Let me see if I can replicate this :)

MadeByMike:
![image](https://user-images.githubusercontent.com/1320567/64667813-76872380-d49e-11e9-9f79-dcf5060dcbba.png)

I found the cause and updated the messaging so that you know if dependencies have been installed with yarn or npm.

Thanks @dzigg PR incomming

dzigg:
Thanks!

dzigg:
Thanks!

masonmonu:
Hi all, npm init keystone-app my-app is failing.

PR #1617, opened 2019-09-11
Fix CLI error when yarn is not detected

MadeByMike:
The CLI is meant to catch an error if `yarn` is not installed and fallback to npm. It wasn't doing this correctly.

I took the chance to improve the error messaging so that people know how packages have been installed.

Fixes:  #1616


packages/create-keystone-app/lib/util.js
```diff
@@ -22,9 +22,15 @@ const getProjectDirectory = async () => {
...
-  return execa.commandSync(cmd, {
-    cwd: path.resolve(newProjectDir),
-  });
+  let result = false;
+  try {
+    result = execa.commandSync(cmd, {
+      cwd: path.resolve(newProjectDir),
+    });
+  } catch (error) {
+    result = error;
```
jesstelford: Does this `error` object have a `.failed` key? Is that something special which `execa` adds?

MadeByMike: Yes exactly.

PR #1619, opened 2019-09-11
Add a new GraphQLPlaygroundApp to decouple from the GraphQLApp

timleslie:
We want to be able to set up these apps independently of each other, particularly in a situation where there are multiple `GraphQLApp` instances on a single endpoint.


packages/app-graphql-playground/README.md
```diff
@@ -0,0 +1,44 @@
...
+draft: true
+[meta]-->
+
+# GraphQL Playground App
+
+A Keystone App that creates an Apollo GraphQL playground.
+
+## Usage
+
+FIXME
```
MadeByMike: I don't want todo's or fixme in docs. I just spent ages removing all of them. 

When published it makes the website look broken. It's better to just not include these sections. Or to take a moment to write them.

packages/app-graphql-playground/README.md
```diff
@@ -0,0 +1,44 @@
...
+        apollo : {},
+    }),
+    new AdminUIApp()
+  ],
+};
+```
+
+## Config
+
+FIXME
```
MadeByMike: Same here

packages/app-graphql-playground/README.md
timleslie: Ooh, good catch, I had fully intended to do these docs on this PR and completely forgot :-( Will fix.

timleslie:
> This one is a pretty significant breaking change... I wonder if it's worth maintaining some backwards compatibility for a while? Or at the very least we should say in the docs, this was the old way and the new way is...

This change is backwards compatible. The existing `GraphQLApp` will instantiate a `GraphQLPlaygroundApp` if a `graphiqlPath` is supplied (which it is, by default).

MadeByMike:
> > This one is a pretty significant breaking change... I wonder if it's worth maintaining some backwards compatibility for a while? Or at the very least we should say in the docs, this was the old way and the new way is...
> 
> This change is backwards compatible. The existing `GraphQLApp` will instantiate a `GraphQLPlaygroundApp` if a `graphiqlPath` is supplied (which it is, by default).

Brilliant!

PR #1635, opened 2019-09-15
Added a createLists helper function

Vultraz:
Figured this might be useful. Wasn't able to run a test that it works, but I'm pretty sure I got it right 😬


timleslie:
Thanks for the contribution @Vultraz. We're going to pass on this one, the reason being that we want to keep our core APIs as small as possible. While creating multiple lists is a common operation, this API addition doesn't represent a major saving for the user, and so the cost of additional API surface area outweighs the benefits to the user.

PR #1643, opened 2019-09-16
Parse custom mutation access

timleslie:


packages/keystone/lib/Keystone/index.js
```diff
@@ -53,7 +54,7 @@ module.exports = class Keystone {
   }) {
     this.name = name;
     this.adapterConnectOptions = adapterConnectOptions;
-    this.defaultAccess = { list: true, field: true, ...defaultAccess };
+    this.defaultAccess = { list: true, field: true, custom: true, ...defaultAccess };
```
MadeByMike: Does this mean all custom schemas share 1 default access control?

timleslie: Yep, this is consistent with the list and field defaults.

Issue #1182, opened 2019-05-17
Custom Servers + BYO `express` cleanup tasks


jesstelford:
A master-list of tasks to do after #1109 lands.

- [x] [Stop hard-coding the admin path on static build](https://github.com/keystonejs/keystone-5/pull/1109#discussion_r282741011)
- [x] Rename all the paths to routes (to be consistent with the StaticApp API of accepting a route (the URL) and a path (local file system).
  - [x] Change `FileAdapter` API from `{ route, directory }` -> `{ path, src }` to match everything else. #1644

---

- [x] Get the `build()` step working for static assets.
- [x] Rename options for `server-static` to `path` for the URL portion, and `src` for the file-system portion.
- [x] Add more tests for the different edge cases of the `prepare` method + CLI (currently written as `test.todo()` in the code as placeholders) #1109
- [x] Remove `keystone` from `@keystone-alpha/app-admin`'s constructor (it's passed into the `prepareMiddlewares()` function. #1197
- [x] There is a `port` parameter being passed to `prepareMiddleware()` which is only used for some `console.log`s. I want to move all that into the default server and stop passing the `port` parameter. #1203
- [x] Most of the custom servers now only exist to insert default data. Can we provide a better API for that which is setup in `index.js`, or (better yet) is a separate script that is run only once? This would make most of our custom servers go away leaving us with a nice clean `index.js` everywhere.
  - [x] Provide an alternate mechanism for running things after `connect` (#1200)
  - [x] Remove the custom servers (#1204)
- [x] This new structure makes it easier to start listening to requests immediately rather than waiting for all of KS to be ready + the AdminUI, etc. We can avoid the bad DX of having to hit F5 constantly until it boots up. (#1205)
- [ ] ~Move the default server into its own file `default-server.js` so it's easier to reference / discover for app developers.~ The default server is now heavily customised and might actually be confusing for devs, so I wont be doing this.
- [x] Documentation #1214 
  - The expected API for `servers`.
  - Different ways to specify `servers`:
    - Export the `servers` array from `index.js`
    - Create a custom `server.js` _and_ export the `servers` array from `index.js`
    - Create a custom `server.js` which instantiates all of its own middlewares
  - How to customise the GraphQL middleware (it's just another server 😉)

JedWatson:
> We can avoid the bad DX of having to hit F5 constantly until it boots up.

🎉

[...]

PR #1644, opened 2019-09-16
Change file adapter API 

MadeByMike:
Just knocking off a few tiny outstanding things so I can close some issues.

See:  #1182


packages/file-adapters/lib/local-file.js
```diff
@@ -3,19 +3,19 @@ const path = require('path');
 const mkdirp = require('mkdirp');
 
 module.exports = class LocalFileAdapter {
-  constructor({ directory, route }) {
-    this.directory = path.resolve(directory);
-    this.route = route;
+  constructor({ src, path }) {
```
jesstelford: `path` will be shadowed if you do it like this, so you need to either reassign it to a different variable or don't destructure:

```suggestion
  constructor({ src, path: inputPath }) {
```

packages/file-adapters/lib/local-file.js
```diff
@@ -3,19 +3,19 @@ const path = require('path');
...
+    this.src = path.resolve(src);
+    this.path = path;
```
jesstelford:
```suggestion
    this.path = inputPath;
```

Issue #1649, opened 2019-09-16
`originalInput` not correctly passed to `beforeChange()` and `afterChange()` hooks


molomby:
There's a bug in the `List` class where the (internal) `_beforeChange()` and `_afterChange()` functions have these signatures:

```js
async _beforeChange(resolvedData, existingItem, context, originalInput) { ... }
async _afterChange(updatedItem, existingItem, context, originalInput) { ... }
```

But when they're called, an `operation` argument is also provided. Eg..

```js
this._beforeChange(resolvedData, existingItem, context, operation, originalInput);
this._afterChange(newItem, existingItem, context, operation, originalInput);
```

As a result, when one of these hooks is executed, the `originalInput` comes though as a string (ie. "update" or "create") rather than an object [as documented](https://github.com/keystonejs/keystone-5/blob/master/docs/api/hooks.md#hooks-function-properties).

Vultraz:
Looks like this one can be closed @jesstelford

molomby:
Was merged

PR #1650, opened 2019-09-16
Molomby/fix 1649 hooks original input

molomby:
The bug is described in #1649.

Strictly speaking this should be a major version bump..


jesstelford:
@all-contributors add @molomby for bug fixes!

PR #1656, opened 2019-09-17
Add GraphQL query validators

sarneaud:
https://github.com/keystonejs/keystone-5/issues/1469


MadeByMike:
Love it! But I also want @timleslie or @jesstelford to look at it

PR #1664, opened 2019-09-17
Add support for alpha channel in color field

### LINKED ISSUE #602: RGB colour picker shouldn't include alpha
The `Color` field type uses a colour picker UI component which includes an 'A' value. At present we have no way of storing an alpha value. We should either store RGBA or not have alpha available in the UI. Discussion required.

jordanoverbye:
Closes #602

Currently Keystone stores the value from the `Color` field as a `HEX` string. However in the Admin UI, we have an option for the user to change the alpha value - but this does not get saved. I have updated the `Color` field package so the value will now be stored as an `RGBA` value. For example: `rgba(255, 255, 255, 0.5)`. For backwards compatibility, if a value is currently being stored as a hex value that will still work.

I have also removed the 'preset colors' from the color picker UI to make things less busy. I think this would be cool to have as an option when creating the field - but I think that is for another PR.

#### UI Updates

Screen Shot 2019-09-17 at 4 40 09 pm

Screen Shot 2019-09-17 at 4 40 30 pm


packages/fields/src/types/Color/views/Field.js
```diff
@@ -12,10 +10,23 @@ const ColorField = ({ field, value: serverValue, errors, onChange }) => {
   const value = serverValue || '';
   const htmlID = `ks-input-${field.path}`;
 
+  const colorPickerValue = useMemo(() => {
+    // keystone previously stored values as a hex string and this should still be supported
+    // it is now stored as an rgba string
+    if (value) {
+      if (value.indexOf('rgba', 0) === 0) {
+        const rgbaValues = value.replace(/^rgba?\(|\s+|\)$/g, '').split(',');
```
jesstelford:
```suggestion
        const rgbaValues = value.replace(/^rgba\(|\s+|\)$/g, '').split(',');
```
The `a` is required based on the `if` check above.

.changeset/nice-grapes-compare/changes.md
```diff
@@ -0,0 +1 @@
+Added alpha channel support to the color field
```
jesstelford:
```suggestion
Added alpha channel support to the `Color` field. NOTE: This introduces a backwards-compatible change in the format saved to the database. The old format was a 6-character hex string preceded by a `#`, eg; `#663399`. The new format is an `rgba` string, eg; `rgba(102, 51, 153, 1)`. Both formats are fully supported in CSS, and the `Color` field will correctly handle old values stored in the hex format.
```

jordanoverbye:
Thanks @jesstelford ! I've just committed your suggestions 🙏

MadeByMike:
Love it. Great work @jordanoverbye. I want to capture our other ideas here.

GraphQL color functions!

```
query {
  color {
    hue
    hex
    rgb
    rgba
    lightness
    hsl
    hsv
    etc...
    r
    g
    b
    a
    contrast(with: '#555')
  }
  color(mix: 'dodgerblue', .5) {
    rgba
  }
  color(lighten: .5) {
    rgba
  }
}
```

PR #1665, opened 2019-09-17
[RFC] Admin UI hooks initial framework

MadeByMike:
I got sucked down a rabbit hole today of adding hooks to the admin UI. It's a door I really wanted to open as I'm realising how many issues could be addressed through these hooks.

I took influence from the hooks and pages APIs, I don't see many other ways 🤔

The API looks something like this:

```javascript
new AdminUIApp({
  enableDefaultRoute: true,
  hooks: {
    hookName: [path.resolve('./hooks/function')],
  },
})
```
The hook name can be anything we set-up in the admin UI. I haven't set-up any in this PR yet but I did tested it and it works pretty well I think. Admin UI set-up looks like this:

```javascript
const hookData = { this: { canBe: 'anything' } }
const testHook = useHook('testHook', hookData);
const result = testHook(<Component>Example</Component>);
```

You can also wrap multiple components:

```js
const result = testHook([
  <Component>Example 1</Component>,
  <Component>Example 2</Component>
]);
```

The useHook function comes from `AdminMeta.js`. It will compose all of the functions in the hooks array, load the views and pass the `hookData`.

Hook functions can do a bunch of things and don't actually need to modify the component array

```js
// do something with data
(arrayofComponents, data) => {
  console.log(data);
  return arrayofComponents;
}

// Append the array
(arrayofComponents, data) => [<p>Hello</p>,...arrayofComponents]

// Nest content (This hook should probably be last in the array)
(arrayofComponents, data) => [<p>{arrayofComponents}</p>]

// Delete everything
(arrayofComponents, data) => []
```

Thoughts? Comments?

One thing I would LOVE and I am not sure it is possible... I'm sure @mitchellhamilton will know but is there anyway we could do some magic so that this could be an inline function:

```
hooks: {
  hookName: [path.resolve('./hooks/function')],
}
```

I assume we could render a function to a string which would not work for anything complicated... Would love if someone smarter than me could have a think about it.

My other concern is that this doesn't move us closer to an adminUI that can be run independently from Keystone. It ties it closer?


mitchellhamilton:
I have mixed feelings on the name hooks, it certainly makes sense since it goes with the other hooks in Keystone but given this is for the front end, I feel like there could be confusion with React hooks. `overrides` might be a good other name since from what I understand, most of this is about changing existing behavior in the Admin UI?


My thoughts on the API: It seems very specific to a small number of use cases. IMO, the ideal API would be that a hook can be anything and each specific hook will expect whatever it requires. A common convention would probably be that lots of the hooks are React components but making it generic would mean that we could allow overriding non-visual behavior if that's a thing we want to do. E.g. something like this:

```jsx
let SomeComponent = useHook('someHookName')


<SomeComponent {...someProps} />


let onDeleteItem = useHook('someHookName')

<Thing onDelete={() => onDeleteItem(item)} />
```

> One thing I would LOVE and I am not sure it is possible... I'm sure @mitchellhamilton will know but is there anyway we could do some magic so that this could be an inline function:
> 
> ```
> hooks: {
>   hookName: [path.resolve('./hooks/function')],
> }
> ```
> 
> I assume we could render a function to a string which would not work for anything complicated... Would love if someone smarter than me could have a think about it.

It's probably possible with some build hackery to compile the keystone entry point and alias the server side things to stubs of them but also it's probably a bad idea and would likely cause problems especially since our current opinion on compiling server-side stuff is that Keystone doesn't have an opinion.

> My other concern is that this doesn't move us closer to an adminUI that can be run independently from Keystone. It ties it closer?

This doesn't really change anything in this regard, the Admin UI is already built dynamically based on things that a user passes to Keystone. We could relatively easily deploy the Admin UI separately from Keystone before and after this PR.

mitchellhamilton:
Oh, I just understood that the idea is that you can apply multiple hooks so that's why the API is like that. I don't totally get why you would want that for visual things, do you have any examples of where this API would work well?

MadeByMike:
I think hooks probably is the right term. I know it's a loaded term in the React world, but it's also what WordPress, Drupal and other CMSs use for this exact kind of thing. I think it is probably right but let's see what other think.

I love your ideas about non-visual hooks, I can definitely see how that makes sense and will probably amend this tomorrow to allow for it.

The reason for allowing multiple hooks is for the same reason we're going to make the other hooks arrays. I'm working on the assumption that these might be either added by plugins in future or at least that a user might like to include hooks from a third-party library.

Let's say there is a logo hook and ... I don't know a weather widget. These might be imported separately:

```
import { logo, weather} from 'crazy-hooks-package'
// It actually kinda sucks that these have to be paths and we can't pass them config before including them :(
...
hooks: {
  sidebar: [ logo, weather],
}
```

Now let's say that the sidebar hook wraps 3 items, the title the menu and the footer links. The logo replaces the first item (title) and the weather widget inserts itself before the last item (footer).

This is probably a really bad example, but the general idea is that rather than having a hook for every gap there are some places where being able to modify the array of things that are rendered makes more sense... I think.

jesstelford:
`hooks` is already in use in React, so wouldn't be a good idea to use here.

Have you seen how Prisma do their _Custom UI Components_? https://www.prisma.io/docs/prisma-admin/custom-components-h4jp/

gautamsi:
what about calling them `extension`? more generic and can be anything (UI, non UI) , `hooks` seems like something which is triggered, like webhook

gautamsi:
> The logo replaces the first item (title) and the weather widget inserts itself before the last item (footer).

I would say we mark each hook/extension point so that you can say (pretty exhaustive but we can start simple)
```
hooks: {
  sidebar: {
   logo,
   pre-dashboard: custom-dashboard,
   dashboard: replace-dashboard,
   pre-list: links-before-lists,
   post-list: links-after-list,
   copyright,
   footer:replace-footer,
 },
 list-item: {
   nav-buttons: [],
   item-link: [], // one in dropdown
},
item-list:......,
// so on
}
```

I would also expect that these hooks will be provided some data when the action is being called. and some predefined actions like:

1. For example, when you render the link in dropdown button for item in the item table, when you call that link, current item is passed on. 
2. graphql context should also be passed when rendering instance of that hook.
2. you can have pre-defined actions for those hook/extension like `create new item` with pre-filled data and specific list.

VinayaSathyanarayana:
Please refer to https://marmelab.com/react-admin/ for ideas/thoughts

PR #1666, opened 2019-09-17
Configure stalebot

MadeByMike:
This is going to automatically mark old issues as stale.

I've configured it with pretty generous timings. 60 days without comment until stale. Then half a year before automatically closed. The idea is to periodically go through it and close stale issues rather than leaving it until the bot does it. I think we can manage that :)


jesstelford:
I'm concerned that while this may help us with clearing out old issues, it's a horrible experience from a contributor's perspective.

Perhaps we miscommunicate and leave an issue for another maintainer with a response, or someone initially responds, but fails to follow up (we all get busy), both perfectly reasonable things to happen to a maintainer. But then the OP doesn't have a response. And they wait patiently, maybe even periodically checking back in on the issue, until 60 days later, a _bot_ of all things, tells that contributor that their issue is "stale". That's not a great feeling. "Maybe it's a bad issue?" "Maybe they don't like me?" "Maybe I'm not good enough to be a developer after all" "Maybe it's a silly question" are all things that could go through their heads. Then, 4 months later, the issue is _closed_. The conversation has been shut down. No more humans have spoken to that contributor. No more humans have even considered that contributor's original message or what they may have needed. Just closed. By a bot.

> The idea is to periodically go through it and close stale issues rather than leaving it until the bot does it. I think we can manage that

If we can manage that, then we don't need the stale bot.

jesstelford:
Oh, and not to mention ever-green issues which may be things like future planned features, or interesting notes we made for our future selves to come back to. They would get closed and quickly forgotten about.

MadeByMike:
I agree about the potential negative experience which is why I configured it the way I have. I can increase the initial timing but I'd rather re-word the message.

"If we can manage that, then we don't need the stale bot." - We need this to find stale issues.

How about we frame it even more positively. Something like:

"It looks like you haven't had a response in over 3 months. Sorry about that! We've flagged this issue to be manually reviewed by maintainers. We'll do our best to look at it as soon as possible. If you can provide any new information please leave us a comment. This issue will not be automatically closed."

MadeByMike:
Updated text to be super positive. Removed the tag `stale` and replaced with `needs-review`. This puts all the onus on us rather than the contributor. I also extended the close time to over a year.

I think that by apologising that we have not reviewed it and marking it for special attention this is a positive experience for contributors and will help us see if we've slipped up on responding to something.

Any issue that extends over a year in a `needs-review` state should be closed. But it should not happen. Even "ever-green" issues are not helpful if there is nothing to add to the discussion in over 12 months. At the very least this should alert us to look at them every 3 months.

Finally if there are issues that we want this bot to skip we can tag them as `pinned`, `high-priority`, or `security`. This bot will ignore all these tags.

PR #1673, opened 2019-09-18
Fix regression when parent list of a Content field had a `where` clau…

timleslie:
…se as its access control.


packages/fields/src/types/OEmbed/OEmbedBlock.js
```diff
@@ -51,7 +51,16 @@ export class OEmbedBlock extends Block {
               'A reference back to the Slate.js Serialised Document this embed is contained within',
           },
         },
-        access: listConfig.listAccess,
+        access: Object.entries(listConfig.listAccess).reduce(
```
MadeByMike: Util function?

MadeByMike: It's now used 3 times?

MadeByMike: 4 times!

timleslie: Yeah, maybe, but also I don't have time right now :-)

timleslie: The real refactor should come one level higher with the way we create these Aux lists for blocks, but that's future Tim's problem.

PR #1676, opened 2019-09-18
Show an error toast if there was an error creating a new item in the admin-ui

### LINKED ISSUE #1074: Show validation/network errors in Create Item form
Eg; a `Password` field with a min length of 8 characters, but only enter a password of length 3, it show an error in the admin UI.

This appears to be a regression - errors used to be surfaced as a toast notification.

jordanoverbye:
Currently there is no error messaging shown to the user in the admin-ui app if you try to create an item and it fails. What I have done is displaying an error toast to the user if this does happen.

I've also created a custom `ToastContainer` for `react-toast-notifications` because of an issue with z-indexes. Currently, toasts are hidden behind the popover modal. But now, toasts should always be visible.

I think this can close #1074  

Screen Shot 2019-09-18 at 5 43 24 pm


MadeByMike:
👍 Looks good. I'm going to pull this down and manually test tomorrow.

PR #1678, opened 2019-09-19
Add issue templates HEAVILY inspired by Next.js

MadeByMike:
Please consider this a tribute. I found few ways to improve on how next organised their issue templates.


MadeByMike:
> What do you think about making each of the sections' text commented out so we don't end up with a bunch of boilerplate text in issues if folks forget to delete it?

🤔 Sure, I'm deep in something else now but I will do this later

PR #1682, opened 2019-09-19
Add docs for Docker-based KS deployments

sarneaud:
I've added a placeholder section for explaining the keystone build/start
commands, etc.

https://github.com/keystonejs/keystone-5/issues/1266
https://github.com/keystonejs/keystone-5/issues/1257

@MadeByMike I'm not sure about Keystone's deployment "style" (e.g., recommended `yarn` usage or whatever), so the docs might need some tweaks.

### DIFF
--- a/docs/guides/deployment.md
+++ b/docs/guides/deployment.md
@@ -1,0 +1,79 @@
+<!--[meta]
+section: guides
+title: Deployment Recommendations
+[meta]-->
+
+# Deployment Recommendations
+
+## General
+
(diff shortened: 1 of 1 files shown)


VinayaSathyanarayana:
Should we not have separate Dockerfile?

sarneaud:
> Should we not have separate Dockerfile?

Sorry, I don't know what you mean.  Are you suggesting another kind of Dockerfile?

PR #1683, opened 2019-09-19
Allow configuration of Express settings

Vultraz:
In KS4 it was possible to set your desired Express template engine (among other things) in `keystone.init`. I couldn't find any way to do that simply in KS5 without running a custom server for a single `app.set` line. Since StaticApp is the bare-bones view wrapper (the Next and Nuxt apps use React and Vue), I think it would be useful to allow setting that (and other things) here.

To avoid unnecessary specificity, I made it so you can configure any Express setting.

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -58,6 +58,7 @@ module.exports = class Keystone {
     secureCookies = process.env.NODE_ENV === 'production', // Default to true in production
     cookieMaxAge = 1000 * 60 * 60 * 24 * 30, // 30 days
     schemaNames = ['public'],
+    expressOptions = {},
   }) {
     this.name = name;
     this.adapterConnectOptions = adapterConnectOptions;
@@ -77,6 +78,7 @@ module.exports = class Keystone {
     this.eventHandlers = { onConnect };
     this.registeredTypes = new Set();
     this._schemaNames = schemaNames;
+    this._expressOptions = Object.entries(expressOptions);
 
     if (adapters) {
       this.adapters = adapters;
@@ -685,13 +687,18 @@ module.exports = class Keystone {
           ...apps,
         ]
           .filter(({ prepareMiddleware } = {}) => !!prepareMiddleware)
-          .map(app =>
-            app.prepareMiddleware({
+          .map(app => {
+            const expressInstance = app.prepareMiddleware({
               keystone: this,
               dev,
               distDir: distDir || DEFAULT_DIST_DIR,
-            })
-          )
+            });
+
+            // Apply Express settings
+            this._expressOptions.forEach(([setting, value]) => expressInstance.set(setting, value));
+
+            return expressInstance;
+          })
       )),
     ]).filter(middleware => !!middleware);
 
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -50,7 +50,6 @@ module.exports = class Keystone {
     adapter,
     defaultAdapter,
     name,
-    adapterConnectOptions = {},
     onConnect,
     cookieSecret = 'qwerty',
     sessionStore,
@@ -61,7 +60,6 @@ module.exports = class Keystone {
     expressOptions = {},
   }) {
     this.name = name;
-    this.adapterConnectOptions = adapterConnectOptions;
     this.defaultAccess = { list: true, field: true, custom: true, ...defaultAccess };
     this.auth = {};
     this.lists = {};
--- a/.changeset/curly-paws-beg/changes.json
+++ b/.changeset/curly-paws-beg/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/keystone", "type": "patch" }], "dependents": [] }
--- a/.changeset/curly-paws-beg/changes.md
+++ b/.changeset/curly-paws-beg/changes.md
@@ -1,0 +1,1 @@
+Removed adapterConnectOptions key (unused as of 144e6e86)
--- a/.changeset/smart-buses-taste/changes.json
+++ b/.changeset/smart-buses-taste/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/keystone", "type": "minor" }], "dependents": [] }
--- a/.changeset/smart-buses-taste/changes.md
+++ b/.changeset/smart-buses-taste/changes.md
@@ -1,0 +1,1 @@
+Added expressConfig ctor option to pass settings to middleware Express instances
--- a/packages/keystone/README.md
+++ b/packages/keystone/README.md
@@ -20,20 +20,21 @@ const keystone = new Keystone({
 
 ### Config
 
-| Option                  | Type       | Default    | Description                                                                                                                                       |
-| ----------------------- | ---------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `name`                  | `String`   | `null`     | The name of the project. Appears in the Admin UI.                                                                                                 |
-| `adapter`               | `Object`   | Required   | The database storage adapter. See the [Adapter Framework](https://v5.keystonejs.com/keystone-alpha/keystone/lib/adapters/) page for more details. |
-| `adapters`              | `Array`    | `[]`       |                                                                                                                                                   |
-| `defaultAdapter`        | `Object`   | `null`     |                                                                                                                                                   |
-| `adapterConnectOptions` | `Object`   | `{}`       |                                                                                                                                                   |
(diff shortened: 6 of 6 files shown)


Vultraz:
How might that work? I noticed both the Admin UI and GraphQL apps have their own express instance, but not all the options makes sense for all the apps. What would the Admin UI do with a template engine, for example? It's already using React.

gautamsi:
I have a contrib package and demo here which makes use of view engine and also full  (almost) compatibility with KS4 View.

see [@keystonejs-contrib/view](https://github.com/keystonejs-contrib/keystonejs-contrib/tree/master/packages/view) and [blog-view example](https://github.com/keystonejs-contrib/keystonejs-contrib/tree/master/examples/blog-view)

gautamsi:
> Maybe even when instantiating the Keystone insurance? 🤔

did not know Keystone offers insurance as well 😋

Vultraz:
(Also, should I name the key something like `expressConfig` just to be very clear as to its function?)

Vultraz:
@jesstelford any more thoughts on this?

jesstelford:
Apollogies!

> both the Admin UI and GraphQL apps have their own express instance

They instantiate their own sub-instance, which gets used later in a global express server which is nice.

> What would the Admin UI do with a template engine, for example? It's already using React.

Nothing, which is nice in that any custom settings for such a thing wont accidentally break the Admin UI 👍 

> Should I name the key something like expressConfig just to be very clear as to its function?

My thinking is something along the lines of:

```javascript
  name: "Awesome",
  adapter: mongooseAdapter,
  expressOptions: {
    'view engine': 'pug',
  }
});
```

Vultraz:
Rebased and renamed key to `expressOptions` just in the static app for now. I was looking, and I can't actually find anywhere where `adapterConnectOptions` is used. It's only in the `Keystone` ctor, and [a commit from August](https://github.com/keystonejs/keystone-5/commit/144e6e86#diff-ea3ebd50af516b44587d453c083e2879) says it was removed. What should I do?

Vultraz:
Moved the config key to the `Keystone` ctor.

Vultraz:
\o/

Vultraz:
@MadeByMike you have any thoughts?

Vultraz:
@jesstelford is there anything else I should do with this?

MadeByMike:
@Vultraz I've been on holiday I'll make sure I look at this today 👍

packages/keystone/lib/Keystone/index.js
```diff
@@ -75,6 +74,7 @@ module.exports = class Keystone {
     this.eventHandlers = { onConnect };
     this.registeredTypes = new Set();
     this._schemaNames = schemaNames;
+    this._expressOptions = Object.entries(expressOptions);
```
MadeByMike:
```suggestion
```

packages/keystone/lib/Keystone/index.js
```diff
@@ -658,13 +658,18 @@ module.exports = class Keystone {
...
+            const expressInstance = app.prepareMiddleware({
               keystone: this,
               dev,
               distDir: distDir || DEFAULT_DIST_DIR,
-            })
-          )
+            });
+
+            // Apply Express settings
+            this._expressOptions.forEach(([setting, value]) => expressInstance.set(setting, value));
```
MadeByMike:
```suggestion
            Object.entries(expressOptions).forEach(([setting, value]) => expressInstance.set(setting, value));
```

packages/keystone/lib/Keystone/index.js
MadeByMike: We're not accessing these options elsewhere so there is no need to set this property?

MadeByMike:
I've got one minor suggestion otherwise I'm happy with this.

At the risk of bike shedding, I'm not totally convinced this is necessary given how easy it should be to create a custom server or app. If it is necessary I'd probably prefer to see it as an option on the static app rather than applied to all middlewares. The downside is each app would have to implement options like this or extend a base app class that implements generic express options.

I'm sorry, I know this is in conflict with what @jesstelford suggested. I'll happily go with whatever the  consensus is. @jesstelford if you're happy with this approach go ahead and approve this PR 👍

MadeByMike:
Sorry for bikeshedding

packages/keystone/lib/Keystone/index.js
Vultraz: I can't do this since `expressOptions` is a ctor argument. It's used in `prepare` so I need to set that.

Vultraz:
Rebased.

dcousens:
At what point do you simply let users provide an express instance instead of trying to wrap their API?

Vultraz:
Presumably that's where the custom server comes in.

gautamsi:
@Vultraz and others: possibly you have missed my comment https://github.com/keystonejs/keystone-5/pull/1683#issuecomment-533079347

it is easy to use express settings with [custom server](https://v5.keystonejs.com/guides/custom-server#minimal-custom-server) options, even the minimal would work.

I have used this in the blog view demo which uses `bodyparser` `view engines` and other things. 

https://github.com/keystonejs-contrib/keystonejs-contrib/blob/9a6304e68ac580434132c3542a2283bb765cf0f1/examples/blog-view/server.js#L14-L18

Vultraz:
@gautamsi yes, I know it's pretty simple. But it still seems like overkill if you're setting *one* setting, and if the idea is that you shouldn't always need a custom server, I think this is a better solution. But of course it's up to the devs to decide.

Also IIRC the log output isn't as nice 😛

gautamsi:
I have proposed earlier to export the Kestone cli utils methods for formatting it properly (cant find the comment yet) once that is done custom server can use those methods for log output.

Vultraz:
That could still be done as a separate PR. But that still leaves the user with something else to configure for their custom server.

I think we just need to wait and see what @jesstelford says.

MadeByMike:
@Vultraz I'm happy with the approach here as I think the option for 'view engine' should trickle down to other middlewares.

That said, I don't want to allow general modification of `expressOptions`. 

I think a custom App or a custom server remains the correct ways to deal with the general I want to control the express app problem. I believe apps are allowed to be slightly opinionated should and we should provide only narrow configuration options. On top of that I can see people doing a lot of things with it that go against the direction we want to take with the overall API.

I'm really sorry as I know you had conflicting advice on this but if you we can start with the narrower option of configuring the view engine I'll approve this. 

You might be able to use the same approach as #1758 depending on how this trickles down or do you need to apply to each middleware as done in this PR? I'll leave that to you to figure out.

Vultraz:
@MadeByMike well, on the narrowest of scopes I could just add a view engine option to StaticApp. Does it make sense to keep `expressOptions` as a `keystone` config option if it only has the view engine? It made sense if any setting was allowed, like @jesstelford mentioned, but if it's restricted maybe it should stay in StaticApp. 🤔

gautamsi:
I think this is still not in right direction. 

`view engine` is only needed when you have express routes to serve. to add that route you would need custom server as this PR does not have anything for this. 

I see that all it does is to set options on express instance on keystone side. 

one better approach would be to restructure the code as such
* all the apps (admin-ui, graphql, static etc.) would need express instance
* keystone cli would provide such instance
* or custom server would provide this instance

advantage, cleaner approach and have more sharing of express instance. we can also use express `Router` to create a mini-app for keystone specific things isolated (partially) from other application portions if that make sense

Vultraz:
@gautamsi I don't entirely follow what you're proposing 🤔 

I just want to reach some agreement here on what to do.

gautamsi:
I don't see a reason to pass on `view engine` to express. When you need view engine, you would be adding some routes to render the view engine.

Keystone v5 does not have express routes to use view engine or rendering as such. You are forced to use custom server, when you do that you can simply use express instance to add `view engine` and any other option.

I do not see a reasoning to use this PR, may be you can provide some example use of `view engine`.

my proposal is straight forward, all the `app-*` package creates its own instance in `prepareMiddleware` call. This is only used by keystone cli or the custom server. This `keystone.prepareMiddleware` call should accept `app` parameter and pass around this to other apps.

let me create a PR to make things clear.

Vultraz:
Ok, so you're suggesting we allow passing a whole custom Express instance to the `keystone` ctor?

gautamsi:
For custom server yes, possibly not to ctor but to prepare call. This should be done after standardizing the use of express router.

Vultraz:
Hm.... @MadeByMike @jesstelford thoughts?

Vultraz:
Closing in favor of #1783. I'll split the adapterConnectOptions into a new PR.

MadeByMike:
@Vultraz thanks for helping us _eventually_ realise the best way to do this 👍

Vultraz:
No problem! Glad I could help 😃

PR #1686, opened 2019-09-23
Support cache control hints in schemas

sarneaud:
https://github.com/keystonejs/keystone-5/issues/1453


sarneaud:
This PR covers all the use cases in the original feature request.  The global default max age is already an Apollo feature:

https://github.com/apollographql/apollo-server/tree/master/packages/apollo-cache-control#setting-a-default-maxage

(I've added an example to the Keystone docs.)

PR #1689, opened 2019-09-23
Rename Keystone to KeystoneJS in docs where possible

### LINKED ISSUE #865: Docs reference of KeystoneJS vs Keystone
I noticed @JedWatson mentioned in https://github.com/keystonejs/keystone-5/pull/796#pullrequestreview-212601579 that `KeystoneJS` should be used instead of `Keystone`. I see the https://v5keystonejs.netlify.com site is using numerous instance of `Keystone` instead of `KeystoneJS`

should that be fixed?

jordanoverbye:
I've gone through every page on the website and tried to rename every instance of "Keystone" to "KeystoneJS". It's a little tricky as sometimes it makes more sense to use `Keystone`, but in most cases it makes sense to use "KeystoneJS"

This PR closes #865

### DIFF
--- a/.changeset/hot-laws-travel/changes.json
+++ b/.changeset/hot-laws-travel/changes.json
@@ -1,21 +1,21 @@
 {
   "releases": [
-    { "name": "@keystone-alpha/demo-project-blog", "type": "minor" },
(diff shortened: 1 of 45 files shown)


jordanoverbye:
Sure thing @timleslie . Can you please review again when you have a chance

PR #1692, opened 2019-09-24
Replace alpha symbol on website with alpha version number

### LINKED ISSUE #1006: [Docs] We should explicitly call out the alpha-ness of v5
I find the alpha symbol (`α`) a confusing way to convey the state of Keystone 5.

Looking at other projects such as [lodash](https://lodash.com/docs), [tailwind](https://tailwindcss.com/docs), [react](https://reactjs.org/docs/getting-started.html), and [`date-fns`](https://date-fns.org/v2.0.0-alpha.27/docs/Getting-Started) the version and (sometimes) the current state of the project is explicitly called out in the header to reduce ambiguity and increase developer's confidence they're on the right page for them.

I propose we replace the alpha symbol with the version number + the word `alpha`.

Currently looks like:

Screen Shot 2019-04-08 at 11 45 15 am

Proposed new look:

Screen Shot 2019-04-08 at 11 44 51 am

jordanoverbye:
This PR attempts to close #1006 .

1. In the website header, I've removed the alpha symbol `(α)` and replaced that with `v5.x alpha`.
1. On the home page, I've removed the alpha symbol `(α)` and have not replaced it with anything. I think it looks weird having the version number here since its already right there in the header. What does everyone think?

### Screenshots

**Home Page**
Screen Shot 2019-09-25 at 9 20 27 am

**Content page**
Screen Shot 2019-09-25 at 9 20 10 am

**Mobile**
Screen Shot 2019-09-25 at 9 19 18 am

### DIFF
--- a/website/src/components/Header.js
+++ b/website/src/components/Header.js
@@ -66,17 +66,17 @@ const Logo = () => (
         KeystoneJS{' '}
       </span>
     </Link>
-    <abbr
+    <span
       css={{
-        cursor: 'help',
-        fontSize: '1.2em',
-        marginLeft: '0.2em',
-        textDecoration: 'none',
+        display: 'inline-block',
+        color: colors.N40,
+        fontStyle: 'italic',
+        marginLeft: '1.5em',
+        textDecoration: 'underline',
       }}
-      title="Keystone 5 is currently in alpha"
     >
-      (α)
-    </abbr>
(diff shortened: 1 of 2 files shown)


jesstelford:
Hey @all-contributors can you add @jordanoverbye for docs?

PR #1697, opened 2019-09-27
Always cast IDs to strings before comparison

timleslie:
This fixes a bug which was causing the adapter to try to re-insert *all* items into the table at once, which was causing crashes when there were more than 65K items.

### DIFF
--- a/packages/adapter-knex/lib/adapter-knex.js
+++ b/packages/adapter-knex/lib/adapter-knex.js
@@ -345,11 +345,10 @@ class KnexListAdapter extends BaseListAdapter {
           .select(a.refListId)
           .from(tableName)
           .where(`${this.key}_id`, item.id);
+        const currentRefIds = currentValues.map(x => x[a.refListId].toString());
 
         // Delete what needs to be deleted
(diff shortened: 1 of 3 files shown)


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -345,11 +345,10 @@ class KnexListAdapter extends BaseListAdapter {
           .select(a.refListId)
           .from(tableName)
           .where(`${this.key}_id`, item.id);
+        const currentRefIds = currentValues.map(x => x[a.refListId].toString());
```
gautamsi: having short named variable is fine for one liner or single statement (like `x` in this case). It takes time to figure out what exactly `a` is in this case. should this be simplified like `fieldAdapter`

timleslie: Quite possible (I'd be more inclined to simply doing `const { refListId } = this.fieldAdaptersByPath[path];`) but that's out of scope for this PR. I'm actually about to rejig the rest of the code around here for performance reasons, so I might make the variable name change on a future PR.

PR #1543, opened 2019-08-27
upgrade react-apollo and replace react-apollo-hooks

gautamsi:
closes #1533

PR #1701, opened 2019-09-27
Use official apollo react hooks (2nd Attempt)

### LINKED ISSUE #1533: depricated `react-apollo-hooks`
`react-apollo-hooks` is now deprecated in favor of `@apollo/react-hooks`

gautamsi:
attempting #1543 again which closed #1533

last time this had to be revert due to 
Error: `Content#getBlocks()` threw a Promise. This may occur when calling `Content#(de)serialize()` before blocks have had a chance to fully load.

see https://github.com/keystonejs/keystone-5/pull/1543#issuecomment-527349884 for more details

It was failing due to trying to deserialise item outside of `<Suspense />` which I have fixed and had to run function inside of suspense to make it work.

no false positive in E2E this time :)


packages/app-admin-ui/client/pages/Item/index.js
```diff
@@ -388,96 +388,96 @@ const ItemNotFound = ({ adminPath, errorMessage, list }) => (
...
-        {({ loading, error, data, refetch }) => {
-          // Now that the network request for data has been triggered, we
-          // try to initialise the fields. They are Suspense capable, so may
-          // throw Promises which will be caught by the above <Suspense>
-          captureSuspensePromises(
-            list.fields
-              .filter(({ isPrimaryKey }) => !isPrimaryKey)
-              .filter(({ maybeAccess }) => !!maybeAccess.update)
-              .map(field => () => field.initFieldView())
+      {(() => {
```
gautamsi: This is needed as we can not run [these 2 linee](https://github.com/keystonejs/keystone-5/pull/1701/files#diff-18fd6ba6fe6ea61ebfb3250f37792127R449) outside of `<Suspense />`, it throws promise which would be either caught and waited inside hook or let suspense take care of that.

I am sure there can be improvement which I am unable to think of right now.

packages/app-admin-ui/client/providers/Session.js
```diff
@@ -1,4 +1,4 @@
-import { withApollo } from 'react-apollo';
+import { withApollo } from '@apollo/react-hoc';
```
gautamsi: this is the only place `withApollo` is being used inside admin-ui code. once session is converted into functional component we can get rid of this as well and use standard query and mutation from hooks.

packages/app-admin-ui/client/pages/Item/index.js
gautamsi: @jesstelford @timleslie please suggest improvement to this block of code.

MadeByMike:
44 changes... I just wanted to say I don't think I'll get to this one this week. Sorry.

gautamsi:
if this is big change, I can split it into 2
1. will have all the core packages modified to use official react hooks
2. all the demo and test projects to used official hooks.

MadeByMike:
> if this is big change, I can split it into 2
> 
> 1. will have all the core packages modified to use official react hooks
> 2. all the demo and test projects to used official hooks.

Yes please! Maybe even one package at a time.

PR #1717, opened 2019-10-01
DRY the Admin UI middleware a bit

sarneaud:
There are separate routines for prod and dev middleware.  The prod
mode path had some bugs in redirect handling.

I've sucked some of the code out of the separate paths so more prod code
is covered by tests.  The redirects work now.


jesstelford:
Looks like some integration tests are no longer able to run. Can you see what went wrong there?

PR #1726, opened 2019-10-02
Handle errors better when knex connection is an object

### LINKED ISSUE #1725: Knex-adapter: Support for knex connection object
# Bug report
Knex adapter does not handle knex connection as object when connection failed

## Describe the bug
If a user passes an object as the knex connection option and for some reason the connection failes, the error is masked due to knex-adapter mistaking the connection object as a string

## To Reproduce
```js
# Make sure the DB options are incorrect
  name: PROJECT_NAME,
  adapter: new Adapter({
    knexOptions: {
      client: "postgres",
      connection: {
        host: process.env.POSTGRES_HOST,
        user: process.env.POSTGRES_USER,
        password: process.env.POSTGRES_PASSWORD,
        database: process.env.POSTGRES_DB,
      }
    }
  }),
});
```

## Expected behaviour
Should handle when an object is sent as the knex connection option, and return the error from knex 

## Screenshots
image

## System information

- OS: macOS
- `"@keystone-alpha/adapter-knex": "^4.0.11"`

ashinzekene:
Fixes #1725

### DIFF
--- a/.changeset/slow-radios-retire/changes.json
+++ b/.changeset/slow-radios-retire/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystone-alpha/adapter-knex", "type": "patch" }], "dependents": [] }
--- a/.changeset/slow-radios-retire/changes.md
+++ b/.changeset/slow-radios-retire/changes.md
@@ -1,0 +1,1 @@
+handle errors better when knex connection is an object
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -343,7 +343,10 @@
       "!packages/fields/**/filterTests.js",
       "!packages/fields/types.js",
       "!packages/mongo-join-builder/examples/**"
-    ]
+    ],
+    "moduleNameMapper": {
+      "@keystone-alpha/(.*)$": "<rootDir>/packages/$1"
+    }
```
timleslie: Let's leave this out.

ashinzekene: Done

PR #1727, opened 2019-10-03
All the new Monorepo things

mitchellhamilton:
- Bolt -> Yarn Workspaces which means when you currently run bolt, you should run yarn instead now
- Use Manypkg for enforcing some monorepo rules, see [Manypkg](https://github.com/mitchellhamilton/manypkg) for the reasoning behind all the rules
- Changesets 2 - single file changesets, new config
  - not a Changesets 2 thing but I updated the changelog generator so it includes references to the PR a change was from and who made the change
- Remove some unnecessary dependencies and force a bunch of Babel packages to a certain version so they're not duplicated a bunch(there were like 34 copies of `@babel/parser` iirc)

If you're curious about any of the reasoning for the high level choices, refer to the [Thinkmill Monorepo Style Guide](https://github.com/Thinkmill/monorepo)


CONTRIBUTING.md
```diff
@@ -24,72 +24,66 @@ If you develop custom fields, adapters, apps or any other Keystone feature, (or
 
 KeystoneJS adheres to the [Contributor Covenant Code of Conduct](code-of-conduct.md).
 
-## Version management
+## Repository Setup
 
-Keystone uses @noviny's [@changesets/cli](https://github.com/noviny/changesets) in combination with `bolt` to track package versions and publish packages.
+KeystoneJS follows the [Thinkmill Monorepo Style Guide](https://github.com/Thinkmill/monorepo). For more information on the reasoning behind using certain tooling, please refer to it.
```
MadeByMike: 👏

packages/app-schema-router/package.json
```diff
@@ -9,5 +9,8 @@
   },
   "peerDependencies": {
     "express": "^4.17.1"
+  },
+  "devDependencies": {
```
MadeByMike: I'm sure you've got this right... but for my own education why does this need to be both a peer dep and dev dep? I thought if it is peer it would be installed anyway?

mitchellhamilton: Peer deps aren't ever installed unless you specify to install them in regular deps, optionalDeps or devDeps. They're essentially a way to say to a consumer "Please install this dependency and the version should be within this range and if you don't you'll see a warning that you'll probably ignore".   See my answer from a question on TM's Stack Overflow about peer deps for more details:

> peerDependencies are a way to communicate that the consumer of a package must install certain dependencies. It's very important to emphasize the consumer part of that, peerDependencies are not automatically installed, a consumer has to install them.
> 
> Why would this be a thing people want you might ask? It allows you to enforce that there is only a single copy of a package (sort of) which is very important for packages like React where there can only be a single copy of them otherwise they break.
> 
> So we understand why you want them and what they do but we also need to know how to use them. As a consumer of a package, you need to add any dependencies in the package's peerDependencies section, to your own project's package.json.
> 
> As a developer of a package, peerDependencies are not installed by default (this is sort of not true with how we work in monorepos because all the packages are hoisted to the root node_modules and therefore available everywhere but we will likely switch to more strict package managers that don't hoist everything and therefore require things that depend on things to say they depend on them). To resolve this, we can add our peerDependencies as devDependencies to have them installed locally when developing a package.

PR #1729, opened 2019-10-03
Defer populating many-relationship fields until they're needed.

timleslie:
At the moment, whenever we perform an `_itemsQuery`, we immediately populate any `many` relationship fields. If an object has a field with a large number (>10,000 for e.g.) of related items, this can begin to have a significant performance impact, particularly if those values are never actually used.

This PR changes the way we perform `_itemsQuery` to explicitly exclude many-relationships. We then make the `Relationship` field itself take responsibility for populating these values as and when needed. To facilitate this we add a `from` parameter to `_itemsQuery` which allows us to perform a query on the related list while passing through info about the original item. In this way we can construct special queries specifically for populating many fields on demand.


timleslie:
These benchmark files capture the improvements made.

[benchmark-master.txt](https://github.com/keystonejs/keystone-5/files/3723294/benchmark-master.txt)
[benchmark.txt](https://github.com/keystonejs/keystone-5/files/3723295/benchmark.txt)

The primary aim of this branch was to avoid loading all related items on a query when they weren't needed. We see a massive win for this, particularly when there are a large number of related items:

```
Read single, ignore relationship, users=1 posts=8192

knex: 6.939529267s -> 0.805608559
mongoose: 15.003637423 -> 1.271998579
```

We also see a speed up in general for the `knex` adapter when loading related items.

```
Read single, read relationship, users=1 posts=8192

knex: 12.770623279 -> 7.358432135
```

Issue #1452, opened 2019-07-30
Custom Actions


VinayaSathyanarayana:
# Feature request

There are a number of different use-cases that keep surfacing in relation to the extensibility of the AdminUI. I think these can be roughly grouped into 3 related features: 

  - Custom Actions
  - Admin UI theming and interface extensibility
  - Custom 'views' for fields
 
To give a brief description each: 

## Custom Actions

This refers to actions a user wants to preform on an item or multiple items in a list. Some examples include:

  - Export to PDF/Excel/JSON
  - Duplicate an item
  - Archive an item

This involves a custom function which has access to the list data. We could potentially implement a standard UI for custom actions and completely separate the customisation of the UI into another feature. 

### Admin UI Hooks: theming and interface extensibility

This refers to the ability to customise parts of the UI. Some examples include:

 - Adding custom CSS
 - Overriding the render method of particular UI elements (to hide or replace them)
 - Adding additional custom UI elements to particular sections of the UI
- Modifying data such as a message show to the user

Custom UI element will need to have the ability to call actions (including custom actions) on click or other events.

## Custom "Views" for fields

This refers to using an alternative UI elements for a field. This could potentially use the same mechanism as other UI extensibility but may also have unique requirements or an alternative API for fields may just offer a better UX/DX. Examples include:

 - Displaying read-only text without an input field
 - Alternative options for displaying and selecting relationships
 - A gallery view for images
 - Displaying of computed fields?
 - Formatting of JSON stored as sting or generally improving display of other complex data stored as text

## Describe the solution you'd like

To keep things as simple as possible and leverage familiarity with the existing APIs I'd like to make use the same API as list & field hooks: https://v5.keystonejs.com/guides/hooks

I think we can do everything we want for custom actions and Admin UI extensions with the same API the only difference is it's defined on the AdminUI.

UI hooks should get access to a bunch of common props such as the current path and auth information so that they can be applied to specific pages or user conditions. 

```js
const adminApp = new AdminUIApp({
  hooks: {

    /* These hooks are actions that run at a specific time but do not return anything */
    pageLoad: ({path}) => { 
       /* runs on every adminUI page */ 
       if(location === `/admin/users/`) {
         /* runs on every users list page */ 
       } 
    }
    /* These hooks are filters that replace existing data */
    footer: ({UIObject}) => {
      /* Some actions get access to UI objects with methods that allow us to register new items */ 
[...]

Issue #310, opened 2018-09-13
Bulk data exports


molomby:
The ability to dump data from a Keystone, via the Admin UI (probably the current view). At a technical level there are two broad approches:

**A) Pull the data into Node and produce a file for download**

This gives us full control of the format (Excel file, CSV, JSON, etc) so is nicest for the user. We can pick columns, ordering and reference hydrated (app generated/virtual) values.

However, without some significant work (generating from a worker, batching items) this approach **_does not scale well_**. The prior art (KS4) uses this approach and i's known to have caused outages on large sites.

**B) Leverage DB platform functionality**

Alternatively, we may want to leverage functionality of the underlying DB directly. Specifically...

* `mongodump` can be given conditions to filter the documents dumped from a collection. Issues: single collection only, not hydrated, format is difficult to consume, dumps all fields.
* the pgsql `copy` command can write queries out directly to a CSV. It may also be possible to [dump JSON](https://dba.stackexchange.com/questions/90482/export-postgres-table-as-json) in this way. Less issues than `mongodump`; `copy` can query across multiple tables/views, control columns, order, format values, etc. Also it's _super_ fast. Locally I can dump 80k+ records (20Mb) to a CSV in _under 200 ms_.
* Etc.. for other DB platforms.

These tool have a _much_ higher capacity to scale but at the cost of some flexibility (esp. Mongo). This is also something we'd need to work into the adapter framework.

VinayaSathyanarayana:
Any update on this?

MadeByMike:
I feel like this one could be a good candidate for a 3rd-party package

gautamsi:
for this we would need extension points for adding UI elements in `admin-`ui`

molomby:
> I feel like this one could be a good candidate for a 3rd-party package

@MadeByMike, Yeah possibly, it really needs DB adapter support though. Keystone's job here is probably to define the API and let DB adapters implement it or not, any third party export tools can pick up from there.

@gautamsi is also right.

MadeByMike:
Adding hooks to the Admin UI opens a lot of doors and is becoming a higher priority

MadeByMike:
@molomby are we clear enough now to say that bulk data exports is something that should be handled at the database layer and handled by a separate tool: https://github.com/keystonejs/keystone/discussions/299

With that and the adminUI extension points as well as custom queries it should be possible for people to make a basic bulk export feature in the AdminUI.

[...]

PR #1739, opened 2019-10-06
[POC] [Discussion] Custom actions (extension/plugins)

gautamsi:
ref: #1452 #310 
Tried to work my way to add some extensibility, see this gif and explanation
keystone plugin demo
> click gif to open in new window for better viewing

## Intention

intention is to enable devs write custom actions which can extend Admin UI with custom Buttons for List and List multi item select management. 

### working:
1. You can add List Action to add UI elements after the "Create" button on List page, this can be any Ui element, I just used the way field view loaders are loaded. 
> the example I used is very crude, actual export features can also be written using this way and the Ui may give download link rather than the item in the console log.

2. You can add ui elements on the toolbar when you select items in the list. ( again very simple example but this can achieve complex scenarios as well)

### possibility
immediate possibility is to 

1. add list item actions similar to #1724 Duplicate items from drop down ( this can also be refactored as list item action rather than in build feature)

2. add Item Action which is on the Item detail/Edit page.

I will work on these two after seeing positive feedback for first two actions working.

## Beyond

later this can also extend to import plugins installed dynamically, I was not able to figure out (yet) the webpack stuff with field-view-loader to load from disk without bundling with webpack.

additionally we can also enable composing with other ui elements like 
* custom filtering options which may enable quick filtering with pre-made filters.
* add ui elements in the create/update item modal which may fetch some information based on input and pre-fill some fields
* allow devs replace CreateItem modal (and other models) with custom one.
 

> this PR may not make it to the repo "as is" due to changes in the Blog Demo project, I will modify this to run with test projects and create new PR once it is acceptable to core team otherwise it will become part of [`keystonejs-contrib`](https://github.com/keystonejs-contrib/keystonejs-contrib) project for sure

### DIFF
--- a/demo-projects/blog/admin/plugins/ListActions.js
+++ b/demo-projects/blog/admin/plugins/ListActions.js
@@ -1,0 +1,74 @@
+
+import React, { useState } from 'react';
+import { DesktopDownloadIcon } from '@arch-ui/icons';
+import { IconButton } from '@arch-ui/button';
+import { useQuery } from 'react-apollo-hooks';
+import gql from 'graphql-tag';
+import { Button } from '@arch-ui/button';
+import Confirm from '@arch-ui/confirm';
+import PageLoading from '@keystone-alpha/app-admin-ui/client/components/PageLoading';
+
+
+export default () => {
+  const [showExportModal, toggleExportModal] = useState(false);
+
+  const closeExportModal = () => {
+    toggleExportModal(false);
+  };
+  const openExportModal = () => {
+    toggleExportModal(true);
+  };
+
+  return (
+    <>
+      {showExportModal ? <DownloadModal isOpen onClose={closeExportModal} />  : null}
+      <IconButton
+        appearance="default"
+        icon={DesktopDownloadIcon}
+        onClick={openExportModal}
+      >Export
+      </IconButton>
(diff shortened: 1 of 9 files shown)


MadeByMike:
This is awesome mate! We need to unify the ideas from here with the comments from: #1665 

I'm going to see if I can get the team together to decide how this might all work.

gautamsi:
we can also make use of refined naming I suggested at https://github.com/keystonejs/keystone-5/pull/1665#issuecomment-532194427
both can be unified easily, just have to find spots in admin ui to add or replace the ui elements.

MadeByMike:
Let me get some feedback from everyone this week and then I'll get together with you and we can spec out and build a complete solution for Admin UI hooks.

VinayaSathyanarayana:
Refer: https://github.com/keystonejs/keystone/pull/2101 & https://github.com/keystonejs/keystone-5/issues/1002

MadeByMike:
The custom actions stuff in https://github.com/keystonejs/keystone/pull/2101 should definitely be considered.

MadeByMike:
Admin UI hooks have shipped (although we need to add the appropriate hook points) With this we will be able to deliver custom actions.

PR #1745, opened 2019-10-08
Publish roadmap

MadeByMike:
Publishes the Roadmap to the website docs.


MadeByMike:
We should add some contextual heading around the features to show what type features we are focusing on: 
  - Initial user experience
  - Better error messages
  - Transactions / Migrations / Relationships?

Call out the Beta release and link to this document. 

Show that it is working at scale.

PR #1755, opened 2019-10-09
Update publishing guide + add backporting docs

jesstelford:


CONTRIBUTING.md
```diff
@@ -85,38 +111,127 @@ bolt build && \
...
+### Backporting Fixes
+
+Occasionally a bug goes undetected for a few versions. When a fix is discovered,
+it needs to be applied to all affected versions. This is called _backporting_.
+
+First, find out the oldest version which was affected. This can be done using
+`git blame`, browsing the `CHANGELOG.md`s, etc.
+
+Once we know which version introduced the bug, walk forward through the
+`CHANGELOG.md` noting all the _minor_ and _major_ releases made since.
```
mitchellhamilton: Is this actually a thing that's going to happen for _every_ major and minor release? That seems like a huge time cost, for security problems and really critical bugs it makes sense but for every bug it seems like a lot of work for little gain.

jesstelford: Yeah, it's definitely discretionary!

mitchellhamilton: Ah, cool, do you want to write that in here to make that clear?

CONTRIBUTING.md
```diff
@@ -85,38 +111,127 @@ bolt build && \
...
 ### A quick note on changelogs
 
 The release process will automatically generate and update a `CHANGELOG.md` file, however this does not need to be the only way this file is modified. The changelogs are deliberately static assets so past changelogs can be updated or expanded upon.
 
 In addition, content added above the last released version will automatically be appended to the next release. If your changes do not fit comfortably within the summary of a changelog, we encourage you to add a more detailed account directly into the `CHANGELOG.md`.
 
+### Backporting Fixes
+
+Occasionally a bug goes undetected for a few versions. When a fix is discovered,
+it needs to be applied to all affected versions. This is called _backporting_.
```
jesstelford:
```suggestion
it may need to be applied to all affected versions (depending on the
severity, security considerations, etc). This is called _backporting_.
```

CONTRIBUTING.md
jesstelford: Good idea 👍 Updated the opening paragraph.

PR #1757, opened 2019-10-09
Remove deprecated options

MadeByMike:
Noticed this in my travels and thought I'd do a quick PR.


.changeset/famous-numbers-happen/changes.json
```diff
@@ -0,0 +1 @@
+{ "releases": [{ "name": "@keystone-alpha/app-graphql", "type": "patch" }], "dependents": [] }
```
jesstelford: I don't think this file should exist with the latest changesets. Make sure your branch is from the HEAD of `master` and that you've reinstalled all your dependencies before rerunning the `changeset` command.

mitchellhamilton: Changesets 2 can handle changesets in this format so it's totally fine to leave this btw.

website/.changeset/config.js
```diff
@@ -0,0 +1,87 @@
+/*
```
mitchellhamilton: This file and `website/.changeset/README.md` shouldn't exist

MadeByMike: Ahh I must have run change set in the website dir :(

PR #1758, opened 2019-10-09
Add a production deployment checklist

sarneaud:
https://github.com/keystonejs/keystone-5/issues/1511


docs/guides/production.md
```diff
@@ -0,0 +1,63 @@
...
+
+It's recommended to run production Javascript servers behind a reverse proxy such as [Nginx](https://nginx.org/), [HAProxy](https://www.haproxy.org/), a CDN or a cloud-based application (layer 7) load balancer. Doing that can improve performance and protect against [Slowloris Dos attacks](https://en.wikipedia.org/wiki/Slowloris_(computer_security)).
+
+You'll need to let your Javascript web app know that you're running behind a proxy, so that it can correctly handle `X-Forwarded-For-*` HTTP headers. Here's how you might set up an Express app to work behind one proxy:
+
+```javascript
+const app = express();
+app.set('trust proxy', 1); // behind one reverse proxy
+app.use(keystoneMiddlewares);
+```
```
MadeByMike: Can we explain a little more about how to configure this? Is it a custom server? If we're suggesting everyone needs this maybe we make it easy to set this middleware option on the Keystone constructor?

sarneaud: Maybe you know more about this than me :)  Vocal runs KS on top of Express.  That `app.set()` line is the only thing we need to make Vocal work behind proxies, so everything else is just "normal" KS setup, whatever that's meant to be.

MadeByMike: A 'normal' set-up has no custom app like this :/

Vultraz: Looks like a place where #1683 could come in handy

sarneaud: Thanks, I just pushed an update, but I'll take a look at your PR to see if it's a better option.

Vultraz:
My solution is cleaner, I think, since it allows configuring any Express
setting and avoids having hardcoded options in the Keystone ctor for
specific ones.

Vultraz:
What do you mean? My PR doesn't wrap the API, it just passes all settings along to each middleware's Express instance.

docs/guides/production.md
MadeByMike: The options in #1683 are added to middlewares not the default express server

Vultraz: Oh, sorry. Didn't realize there was a difference...  D: Should I make it set the default one too?

MadeByMike: That's a slightly trickier pattern since the default server is not part of the keystone instance and that's what @sarneaud solved for us with `configureServerApp(app)` where we explicitly pass in the app before applying the options.

We've done this specifically for `trust proxies` because we think it's a narrow and specific case that users may encounter without wanting a custom server.

The correct API for general configuration of the main server's express app remains a custom server. I don't want to introduce an alternative API for this.

I have similar reservations about #1683 and although I can appreciate your perspective but I think the correct API for significant modifications of an app is to create a custom App. I don't really want to open up the options the way we do in #1683. I think if possible having one way to do things is preferable.

MadeByMike:
@sarneaud Approved :) Thanks!

PR #1759, opened 2019-10-09
Remove env vars

MadeByMike:
Reduced reliance on environmental variables in AdminUI.

Experimental features will now be `false` by default in development and must be explicitly turned on with config:

```
 new AdminUIApp({ enableDevFeatures: true })
```

The webpack build mode will still default to `development` unless the environmental variable `NODE_ENV` is set to `production`. However this can now also be configured in the constructor, allowing users more control over how environmental variables are managed. 

```
 new AdminUIApp({ buildMode: 'production' })
```


packages/app-admin-ui/index.js
```diff
@@ -99,6 +103,8 @@ class AdminUIApp {
         adminMeta,
         entry: 'index',
         outputPath: path.join(builtAdminRoot, 'secure'),
+        enableDevFeatures: this._enableDevFeatures,
```
jesstelford: What does this actually enable? I can't find any info in the webpack docs about it.

packages/app-admin-ui/index.js
```diff
@@ -19,6 +19,8 @@ class AdminUIApp {
     enableDefaultRoute = false,
     isAccessAllowed = () => true,
     schemaName = 'public',
+    enableDevFeatures = false,
+    buildMode = process.env.NODE_ENV === 'production' ? 'production' : 'development',
```
jesstelford: This option isn't necessary, it's already passed into the `build` and `prepareMiddleware` methods as the `dev` argument.

packages/app-admin-ui/index.js
MadeByMike: I thought this was you 😆 It enables the 'update' button in the AdminUI kinda like a feature flag. I assume it's because the update button (which does batch updates) is not working for all field types or is an experimental feature or something?

I'd actually rather remove it or enable it across the board but if experimental features is something we want enabling via config is the right approach.

packages/app-admin-ui/index.js
MadeByMike: It's used by the `getWebpackConfig` function. Are you saying that after the `getWebpackConfig` is called this value is then overwritten?

If so what do you want to do? Strip this out and then set 'production' as the default in `getWebpackConfig`?

packages/app-admin-ui/index.js
gautamsi: If Update button does not work on every field type then we should disable those fields when in production mode rather than turning off this important feature. 

Moreover those non supported fields may be updated using custom action  example i have

MadeByMike: Ok so it looks like build should always be `production` and `createDevMiddleware` always `development`. Love it! Deleting code :)

packages/app-admin-ui/index.js
MadeByMike: Ok so it looks like build should always be production and createDevMiddleware always development. Love it! Deleting code :)

packages/app-admin-ui/index.js
MadeByMike: I checked with Joss (who added it) and it seem like indeed it is not 100% function and ready So I've renamed this flag `experimentalFeatures` and I've also removed the other reference to `ENABLE_DEV_FEATURES`.

MadeByMike:
Closed in favour of #1810

PR #1762, opened 2019-10-10
[WIP] GeoLocation Field

hmlamad:


MadeByMike:
@hmlamad thanks for this. There is already an implementation of a location field: https://github.com/keystonejs/keystone-5/tree/master/packages/fields/src/types/Location is there anything here that should be incorporated?

MadeByMike:
@jordanoverbye I'm about to close this as a dupe. Can you have a quick look at the code and see if there is anything worth merging with the other implementation?

PR #1764, opened 2019-10-11
[WIP] Run tests without lockfile

Noviny:
Hey!

Lockfiles are great. They make our lives a lot easier. Working without one would be insane. Unfortunately, they can also mask problems to us as open source maintainers, where code within our repository runs, but a user getting a fresh install from npm will encounter an error in the exact same scenario.

To try and mitigate this 'works-on-my-machine-ness' of this, I'm suggesting we run a 'clean install' at least once in our release loop. This script should do that.

Still to-do:

- [ ] Set this up to run on specific branches, namely this branch (for testing this) and the `version` branch that gets kept open
- [ ] Make it so running this doesn't immediately cause failures 🚎 . This may involve pinning versions of some devDeps so I don't fight through upgrading them.

### DIFF
--- a/scripts/test-unlocked.sh
+++ b/scripts/test-unlocked.sh
@@ -1,0 +1,2 @@
+mv yarn.lock cached-yarn.lock
+(yarn fresh && yarn test && mv cached-yarn.lock yarn.lock) || mv yarn.lock failed-yarn.lock && mv cached-yarn.lock yarn.lock


mitchellhamilton:
Idea on this: what if we used [Renovate](https://renovatebot.com/) but only with the [`lockFileMaintenance`](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) option so we would have PRs that have the lock file updates every day/week/some other period of time which makes it easier to reproduce problems as a result of lockfiles + we would have updated things in master? (I'm trying Renovate on Preconstruct in the standard "update all the things" mode to see if all the PRs are annoying)

Noviny:
@timleslie I feel like we talked about this earlier and decided against such an approach.

I think this is waaay more tenable now I know about the `lockFileMaintenance` mode. This seems notably more sane and I'd be in favour of trying that instead of this.

MadeByMike:
@timleslie @Noviny @mitchellhamilton Can we get a consensus one the direction we are taking with this PR. @timleslie feel free to make an exec decision or defer to whoever has the most experience in this area.

timleslie:
I'm happy for @Noviny to run with this at his discretion. There's no big rush to get it done, but I think any improvement in this general area will be really helpful 👍

Issue #1768, opened 2019-10-14
A select with number at start of value throws an error


MadeByMike:
# Bug report

## Describe the bug

If you create a list with numbers at the start of the value of will throw an error on start.

` message: 'Syntax Error: Expected Name, found Int "0"',`

## To Reproduce

1. Create a list like: 

```
keystone.createList('ExampleList', {
  fields: {
    age: {
      type: Select,
      options: '0-18,19-35,36-50,50+',
    },
  },
});
```
2. Start Keystone.

It doesn't seem to matter if you use the array notation either:
```
keystone.createList('ExampleList', {
  fields: {
    age: {
      type: Select,
      options: [
        { label: '0-18', value: '0-18' },
        { label: '19-35', value: '19-35' },
        { label: '36-50', value: '36-50' },
        { label: '50+', value: '50+' },
      ],
    },
  },
});
```

## Expected behaviour

I can't see a clear reason why this cannot work. There are no limitations on storing this in a text field in the DB or on the possible values in HTML.

## System information

I was using Knex Adapter

## Additional context

Is Monday.

molomby:
I assume this has something todo with things starting with numbers not being valid JS identifiers? Although I can't think why this would _actually_ be a problem here.. 🤔

It's worth noting that error is from the GraphQL layer.. Maybe some Admin UI code is seeing "0-18" and casting it to a Number? Would be useful to see the GraphQL query being run.

Re: Selects; see also #398. Further, given #851 and #1471 maybe this is an opportunity to revamp the Select field type a bit? The current solution/lack of formality seems very "first pass" to me.

molomby:
It's because we're using GraphQL enums so the values are subjected to the limits enforced there; @JedWatson

singhArmani:
@MadeByMike: This issue is related with how we are enforcing to have **Enums** for [creating Graphql query](https://github.com/keystonejs/keystone/blob/0da2ce99b93a171b64c1877399fb173aa03e646c/packages/fields/src/types/Select/Implementation.js#L31) in [Select](https://github.com/keystonejs/keystone/tree/master/packages/fields/src/types/Select) field.  

In specs **Enums** values are [Name tokens](http://spec.graphql.org/June2018/#sec-Appendix-Grammar-Summary.Lexical-Tokens). Which means the regex expression validating enum values is `/^[_A-Za-z][_0-9A-Za-z]*$/`, which prevents starting values with a digit or having single digit values. 
[...]

PR #1769, opened 2019-10-14
Data migration guide

MadeByMike:
Documentation for the website that provides an introduction to migration, some strategies for how to deal with them and defines Keystone's role in the process.

@jesstelford I'd love you to read due to the bug we discussed. I thought my example was really good because it demonstrated a simple scenario where automated migrations would not be possible. I didn't want to show just adding or dropping a field or casting a value to a new datatype because people might think why can't that be automated. I think my example is pretty real-world and not sure what to do about #1768


docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
+<!--[meta]
+section: guides
+title: Migrations
+[meta]-->
+
+## Migrations in Keystone
+
+Migrations are required when you update a List in Keystone after the database has been initialised. **Keystone can't automatically manage these migrations for you** because it doesn't know what to do with the existing data. This guide will explain when and why you need database migrations and offer some basic solutions to get you started. As your project grows to include production deployments, how you manage the seeding and migration of data, will depend on your specific situation. Keystone tries not to limit type of applications you build and therefor cannot offer a "one size fits all" solution to migrations.
```
dcousens: I mean,  the schema's are declarative,  you probably could produce a reasonable migration output,  but,  that effort probably isn't worth it

timleslie: Yeah, that's something we've tossed around as an idea. It falls in the category of PRs welcome for now, but maybe something we look at down the line once we have a better understanding of the different use cases and problems we're trying to solve for

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+In development sometimes it can be helpful to drop the database everytime Keystone is restarted. To do this using the `KnexAdapter` you can pass the `dropDatabase` option to the adapter:
+
+```js
+const { KnexAdapter } = require('@keystone-alpha/adapter-knex');
+const adapter = new KnexAdapter({ dropDatabase: true });
+```
+
+This might be enough for small projects but it won't be ideal for everyone. If you want data to persist in development you need to re-initialise the database or preform migrations. And regardless of the development environment, almost all projects will eventually need to make changes to a List table in production. For this there are a number of solutions.
+
+The most basic solution is to make these changes on the database. You can do this via a command-line, but if you're reading this guide you're probably looking for an application to help you view and manage tables created by Keystone. For Mongo you might try (Studio 3T)[https://studio3t.com/download/] which is free for non-commercial use. For Postgres you can try (pgadmin)[https://www.pgadmin.org/]. There are a lot of other options and you should find one that works for you.
```
sarneaud: That "if you're reading this" could be interpreted in condescending ways.  How about this?  (Also some link fixes.)

```suggestion
The most basic solution is to make these changes on the database. You can do this via a command-line, or you might prefer a graphical DB admin interface. For Mongo you might try [Studio 3T](https://studio3t.com/download/) which is free for non-commercial use. For Postgres you can try [pgadmin](https://www.pgadmin.org/). There are a lot of other options and you should find one that works for you.
```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+
+| table_name  | column_name | data_type |
+| ----------- | ----------- | --------- |
+| ExampleList | id          | integer   |
+| ExampleList | age         | text      |
+
+Keystone cannot make this change for you because there may be existing data with a type of `integer`. Sometimes migration is as simple as converting an `integer` to a `string` but in certian cases, such as this, it requires programatically resolving the appropriate.
+
+## Migration options
+
```
sarneaud: (Adding subsection headers so it's easy to see and evaluate the different options at a glance.)

```suggestion

### Drop and replace

```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+In development sometimes it can be helpful to drop the database everytime Keystone is restarted. To do this using the `KnexAdapter` you can pass the `dropDatabase` option to the adapter:
+
+```js
+const { KnexAdapter } = require('@keystone-alpha/adapter-knex');
+const adapter = new KnexAdapter({ dropDatabase: true });
+```
+
+This might be enough for small projects but it won't be ideal for everyone. If you want data to persist in development you need to re-initialise the database or preform migrations. And regardless of the development environment, almost all projects will eventually need to make changes to a List table in production. For this there are a number of solutions.
+
```
sarneaud:
```suggestion

### Manual migrations

```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+ALTER TABLE "ExampleList"
+ALTER COLUMN "age" TYPE text
+    USING (
+    CASE WHEN age<=18 THEN '0-18'
+        WHEN age<=35 THEN '19-35'
+        WHEN age<=50 THEN '36-50'
+        ELSE '50+'
+    END);
+```
+
```
sarneaud:
```suggestion

### Managed migrations

```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+Keystone database adapters use Mongoose and Knex under the hood, both of which have migrations solutions. Rather than running SQL queries directly, you can write migrations using JavaScript, and tools like the (Knex CLI)[http://knexjs.org/#Migrations] allow you to keep track of which migrations have been applied. You seed the database, run or rollback specific migrations or apply all missing migrations in the correct order.
+
+Knex migrations or similar tools are probably the right solution for a commercial scale project.
```
sarneaud:
```suggestion
Managed migrations are useful when working in a team or with duplicated deployments (e.g., development and production).  They help keep database changes consistent, even when multiple developers are working with multiple databases.
```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+
+For small scale projects you can avoid the need for many migrations with careful planning, drop the database frequently while in development and use GUI tools to apply manual updates.
```
sarneaud: It feels like this paragraph mostly repeats earlier paragraphs.

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+ALTER COLUMN "age" TYPE text
+    USING (
+    CASE WHEN age<=18 THEN '0-18'
+        WHEN age<=35 THEN '19-35'
+        WHEN age<=50 THEN '36-50'
+        ELSE '50+'
+    END);
+```
+
+Keystone database adapters use Mongoose and Knex under the hood, both of which have migrations solutions. Rather than running SQL queries directly, you can write migrations using JavaScript, and tools like the (Knex CLI)[http://knexjs.org/#Migrations] allow you to keep track of which migrations have been applied. You seed the database, run or rollback specific migrations or apply all missing migrations in the correct order.
```
sarneaud: (Link fix)
```suggestion
Keystone database adapters use Mongoose and Knex under the hood, both of which have migrations solutions. Rather than running SQL queries directly, you can write migrations using JavaScript, and tools like the [Knex CLI](http://knexjs.org/#Migrations) allow you to keep track of which migrations have been applied. You seed the database, run or rollback specific migrations or apply all missing migrations in the correct order.
```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+If we restart Keystone after this change it will throw an error. The `data_type` of the `age` column is now a `text` value and it should contain a string representation of the age range. The table should be change to something like the following:
+
+| table_name  | column_name | data_type |
+| ----------- | ----------- | --------- |
+| ExampleList | id          | integer   |
+| ExampleList | age         | text      |
+
+Keystone cannot make this change for you because there may be existing data with a type of `integer`. Sometimes migration is as simple as converting an `integer` to a `string` but in certian cases, such as this, it requires programatically resolving the appropriate.
```
sarneaud: Last sentence needs an edit.

docs/guides/migrations.md
sarneaud: I think the example below explains what you're saying better.  Maybe keep this paragraph as short as possible.  Something like this could be enough:

```suggestion
As your app grows new features, you'll inevitably need to change some of your existing List definitions. When you do, you'll also need to migrate the data and schema in your database backend. 
```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+
+```javascript
+keystone.createList('MigrationExample', {
+  fields: {
+    age: { type: Select, options: '0-18, 19-35, 36-50, 50+' },
+  },
+});
+```
+
+If we restart Keystone after this change it will throw an error. The `data_type` of the `age` column is now a `text` value and it should contain a string representation of the age range. The table should be change to something like the following:
```
sarneaud:
```suggestion
If we restart Keystone after this change it will throw an error. The `data_type` of the `age` column is now a `text` value and it should contain a string representation of the age range. The table should change to something like the following:
```

docs/guides/migrations.md
```diff
@@ -0,0 +1,83 @@
...
+const adapter = new KnexAdapter({ dropDatabase: true });
+```
+
+This might be enough for small projects but it won't be ideal for everyone. If you want data to persist in development you need to re-initialise the database or preform migrations. And regardless of the development environment, almost all projects will eventually need to make changes to a List table in production. For this there are a number of solutions.
+
+The most basic solution is to make these changes on the database. You can do this via a command-line, but if you're reading this guide you're probably looking for an application to help you view and manage tables created by Keystone. For Mongo you might try (Studio 3T)[https://studio3t.com/download/] which is free for non-commercial use. For Postgres you can try (pgadmin)[https://www.pgadmin.org/]. There are a lot of other options and you should find one that works for you.
+
+Once you get familiar with your database GUI you will be able to look at the tables Keystone creates for you, manually drop them, make change to Lists and see how the data-structure differs, and eventually write migration queries such as:
```
sarneaud: I know you're just talking about what we were doing the other day, but other people might interpret the tone differently.  Maybe something like:

```suggestion
Here's how you might do a data migration for the `ExampleList` above in Postgres:
```

There are lots more Postgres examples you can link to here: https://www.postgresql.org/docs/current/sql-altertable.html

docs/guides/migrations.md
MadeByMike: Kinda but I do want to sum up but reminding people to plan changes

mitchellhamilton:
re the snapshots that are failing here: apply changes similar to https://github.com/preconstruct/preconstruct/commit/baf10b9b97036d534a9b59bd6a0ce0baa2c4e4b3 but for build-field-types and they should be fine

MadeByMike:
Reverted and merged the latest master again... let's see how this goes

jesstelford:
Love the example, great work!

This is a great document to point people in the right direction without being prescriptive or trying to solve the problem ourselves 👍

PR #1771, opened 2019-10-14
[WIP] Allow keystone.initialise() to setup Db without connect

MadeByMike:
Splits out the db initialisation from the connect method so that each can be called independently. Ideally you'd not call initialise in prod.

Question for Tim: If you opt to initialise on create this will attempt to connect twice? Any issues with that?

### DIFF
--- a/packages/keystone/lib/adapters/index.js
+++ b/packages/keystone/lib/adapters/index.js
@@ -30,7 +30,7 @@ class BaseKeystoneAdapter {
 
     // Set up all list adapters
     try {
-      const taskResults = await this.initialise();
+      const taskResults = await this._initialise();
       const errors = taskResults.filter(({ isRejected }) => isRejected).map(({ reason }) => reason);
 
       if (errors.length) {
@@ -54,7 +54,7 @@ class BaseKeystoneAdapter {
     }
   }
 
-  async initialise() {}
+  async _initialise() {}
 }
 
 class BaseListAdapter {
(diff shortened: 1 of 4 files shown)


jesstelford:
What's the motivation for this change?

packages/keystone/lib/adapters/index.js
```diff
@@ -17,18 +17,25 @@ class BaseKeystoneAdapter {
...
+    }
+  }
+
+  async initialise({ name }) {
+    await this._connect({ name }, this.config);
 
     // Set up all list adapters
     try {
-      const taskResults = await this.postConnect();
+      const taskResults = await this._initialise();
```
MadeByMike:
```suggestion
      const taskResults = await this.initialise();
```

packages/keystone/lib/adapters/index.js
```diff
@@ -47,7 +54,7 @@ class BaseKeystoneAdapter {
     }
   }
 
-  async postConnect() {}
+  async _initialise() {}
```
MadeByMike:
```suggestion
  async initialise() {}
```

MadeByMike:
@jesstelford Tim's suggestion based on my need to initialise a Keystone database before running a seeding/migration task. I don't want to start a Keystone to do this. 

Tim also felt that connect should not attempt to initialise the db every time the server started but only when explicitly directed.

PR #1775, opened 2019-10-15
Use Official Apollo Hooks for react - Part 1, update dependencies

gautamsi:
part 1 of #1701 , splitting into several smaller PRs. 

* upgrade `react-apollo` to latest and remove `react-apollo-hooks`
* use `react-apollo` instead of `react-apollo-hooks`

next PR will convert some of the `<Query>` and `<Mutation>` to `useQuery` and `useMutation` in packages, then in demo-projects


MadeByMike:
@gautamsi this is failing CI. I think this would be far better to do per package.

I updated the meet-up demo so I could open source this and collect issues: https://github.com/Thinkmill/meetup-alternative At some point I'm going to roll this back into the monorepo but if you want to tackle that is seems like a good isolated chunk.

gautamsi:
BTW those are wired errors, I had random timeout failures locally but no such issue as in CI. I saw similar issues in other PR.

gautamsi:
I will work on merging the meetup-alternative here, re-created this PR to see if those CI issues are gone

PR #1781, opened 2019-10-17
fixing require path in knex adapter documentation

simonswiss:
Fixing the require path to use the `@keystone-alpha` namespace in knex adapter documentation.


simonswiss:
Also updated the path for the `mongoose` adapter (same issue), requesting new review!

PR #1783, opened 2019-10-17
Support configureExpress being returned from an entry file

timleslie:
Reverts #1758 and supersedes #1683.

This PR provides a general mechanism providing the system designer a chance to configure the `app()` used in the `keystone` CLI helper scripts.

The behaviour will be equivalent to:

```js
const { keystone, apps, configureExpress } = require('./index.js');

keystone.prepare().then(({ middlewares }) => {
  const app = express();
  configureExpress(app);
  app.use(middlewares);
  app.listen(3000);
});
```

### DIFF
--- a/.changeset/chilled-ads-learn/changes.md
+++ b/.changeset/chilled-ads-learn/changes.md
@@ -1,1 +1,0 @@
-Add trustProxies config option to Keystone object
--- a/docs/api/hooks.md
+++ b/docs/api/hooks.md
@@ -52,7 +52,6 @@ _Note_: `resolveInput` is not executed for deleted items.
 #### Usage:
 
 <!-- prettier-ignore -->
-
 ```javascript
 const resolveInput = ({
   resolvedData,
@@ -70,7 +69,6 @@ Executed after `resolveInput`. Should throw if `resolvedData` is invalid.
 #### Usage
 
 <!-- prettier-ignore -->
-
 ```javascript
 const validateInput = ({
   resolvedData,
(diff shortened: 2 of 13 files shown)


.changeset/chilled-ads-learn/changes.md
```diff
@@ -1 +0,0 @@
-Add trustProxies config option to Keystone object
```
timleslie: I've reverted the changes made in the `chilled-ads-learn` PR as this solution supersedes them. I've deleted the changeset as we don't want this in the changelog, it's as if the `chilled-ads-learn` never happened :magic-hands:

docs/api/hooks.md
```diff
@@ -52,7 +52,6 @@ _Note_: `resolveInput` is not executed for deleted items.
 #### Usage:
 
 <!-- prettier-ignore -->
-
 ```javascript
```
timleslie: `yarn format` does weird things 🤷‍♂

Vultraz:
If I understand correctly, this should allow using `app.set` *without* a custom server? That is, just export your `configureExpress` setter in your index.js?

timleslie:
Yep, that’s correct.

Vultraz:
:D Ok, this definitely seems like a better solution than my PR. I'll close mine.

gautamsi:
I have a situation, if i use admin-ui with custom server it swallows all the `app.set`, see my example, just run blog demo project which uses custom server similar to v4 style view engines https://github.com/keystonejs-contrib/keystonejs-contrib/pull/1

gautamsi:
I see the configure express may be able to help, i have to test that when i return next week

PR #1792, opened 2019-10-21
Update the website to point to v4 docs

MadeByMike:
keystonejs.com now resolves to the same website as v5.keystonejs.com. This removes the word alpha and provides a link to v4.keystonejs.com for people looking for this.

There is a similar PR for the old website to reference v5.

In a future PR we will update all the links and references to alpha when the packages change.

### DIFF
--- a/website/src/components/Header.js
+++ b/website/src/components/Header.js
@@ -73,7 +73,7 @@ const Logo = () => (
           marginLeft: '0.5em',
         }}
       >
-        v5.x alpha
+        v5.x
       </span>
     </Link>
   </div>
--- a/website/src/components/Header.js
+++ b/website/src/components/Header.js
@@ -73,7 +73,7 @@ const Logo = () => (
           marginLeft: '0.5em',
         }}
       >
-        v5.x
+        v5
       </span>
     </Link>
   </div>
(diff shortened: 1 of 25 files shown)


website/src/components/Header.js
```diff
@@ -73,7 +73,7 @@ const Logo = () => (
           marginLeft: '0.5em',
         }}
       >
-        v5.x alpha
+        v5.x
```
timleslie: Can we make this just `v5`, rather than `v5.x`. Since `v5` isn't a semantic version number, it doesn't really make sense to imply that it is...

MadeByMike:
```suggestion
        v5
```

Vultraz:
Does this mean v5 is moving out of alpha and is ready to start use in production projects?

timleslie:
> Does this mean v5 is moving out of alpha and is ready to start use in production projects?

Yep! We're in the process of moving out of `alpha` in the coming days. There are various housekeeping things to do to make this happen, and the website is one of them. You'll see more PRs like this landing in the coming days and once we're happy that everything is all running smoothly we'll make an announcement. 🎉

PR #1794, opened 2019-10-21
Change keystone.use to app.use

cmosgh:
Keystone doesn't have a use method. We actually have to pass the middlewares to the newly created express app.


jesstelford:
@all-contributors can you add @cmosgh for docs please 🙏

PR #1797, opened 2019-10-22
Doc cleanup - Pass 2

Vultraz:
More things!

* Replaced all v5.keystonejs.com links in `docs/` with relative paths. Before it was a mix of hardcoded and relative. Also fixes certain links at keystonejs.com sending people to v5.keystonejs.com.
* Replaced all other instances of v5.keystonejs.com outside of `docs/` with keystonejs.com. Just did a find/replace on these.
* Few other small things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,11 @@ yarn start
 ```
 
 For more details and system requirements, check out the [5 Minute Quick Start
-Guide](https://v5.keystonejs.com/quick-start/).
+Guide](https://keystonejs.com/quick-start/).
 
 ### API
 
-The [API documentation](https://v5.keystonejs.com/api/) contains a reference for all KeystoneJS packages.
+The [API documentation](https://keystonejs.com/api/) contains a reference for all KeystoneJS packages.
 
 ### Demo Projects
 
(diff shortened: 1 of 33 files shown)


MadeByMike:
@Vultraz We don't want relative paths or it they work in the Readme files :(

Vultraz:
Ah! Should relative paths be converted to full paths?

MadeByMike:
Hang on... looks like you got this right. Relative URLs in the `docs/` directory should be ok. But in Readme.md files they should be absolute.

Vultraz:
I double checked. All paths in readmes are indeed absolute.

jesstelford:
To flesh out @MadeByMike's point - We wanted to retain the clickability of links in all mardown files on GitHub as well as on the docs website.

For example, with this change, the first link on this page (in the opening paragraph) is no longer clickable: https://github.com/keystonejs/keystone-5/blob/847e6c7fa42ffe5ad6909f8fb5a07e0a884292cf/docs/api/access-control.md

Vultraz:
I can go back and prepend the full URLs on all links if you'd prefer. The way it was before was a mix of full URL and relative.

Issue #1737, opened 2019-10-05
Custom Pages context/props


gautamsi:
I see that we render the custom pages without any admin-ui context in it. I see that I can use `useQuery` and `useMutation` to work with AdminUI GraphQL. (useQuery and useMutation from `react-apollo-hooks`, pending PR #1701 for official apollo hooks usage)

I also found that I can also import `useAdminMeta` as follows and do some of the things already possible with admin-ui.
```js
import { useAdminMeta } from '@keystone-alpha/app-admin-ui/client/providers/AdminMeta';
``` 

May I propose that we add adminMeta as props when rendering custom pages along with simplified export of `useAdminMeta` hook in `@keystone-alpha/app-admin-ui/client` or a new package `@keystone-alpha/hooks` for better consumption of admin ui data in the custom page. Having props will also help when anyone writes class component (for their reason).

I will create a PR for this.

MadeByMike:
I'm a little confused about what you're suggesting here. Technically I understand but I haven't done much with custom pages so I don't think I understand why or what this will enable. 

I'm going to add the `needs-clarification` label :) until we get a bit more info. I'm sure the PR will help me understand better but if you can also explain how data from `useAdminMeta` might be used in a custom page I'm sure it would help people less familiar... or just help me. Either way is good :)

VinayaSathyanarayana:
There is a need to document Custom Actions Vs Custom Pages

MadeByMike:
@gautamsi if you want to export `useAdminMeta` from `@keystone-alpha/app-admin-ui` I'll approve that PR.

PR #1805, opened 2019-10-22
export useAdminMeta

gautamsi:
ref #1737 

exporting `useAdminMeta` which can be used with custom pages to get information about the lists. 

how to use this
```js
import { useAdminMeta } from '@keystone-alpha/app-admin-ui/client';

const Component = () => {
const adminMeta = useAdminMeta();
// OR
const { adminPath, apiPath, authStrategy, getListByKey, getListByPath, graphiqlPath, listKeys, name, pageViews, pages, preloadViews, readViews, signinPath, signoutPath } = useAdminMeta();
return <UI />
}

```

@MadeByMike I have used `/client.js` for exporting this to isolate from server side exports or execution. `/client/index.js` does not export anything, instead it is  bootstraping the admin-ui.


MadeByMike:
Looks good but why did you opt to put this in a file called client rather than export it from `app-admin-ui/index.js`?

gautamsi:
root index is mostly server part which sets up webpack.

PR #1806, opened 2019-10-22
add icon to list name to indicate auth list

### LINKED ISSUE #426: Visually indicate which list(s) are used for authentication
Eg, something like (`Users` is the list used for authentication in this example):

screen shot 2018-10-10 at 10 21 25 am

gautamsi:
closes #426

current PR (primary color on hover)
keystone-user-list

considered: 
default:
image

alternate (solid primary color): 
image

@MadeByMike if there is any preference in color, let me know.

I would also like to propose icon name for list config. so that anyone can add icon to the list name for visual indication. either a react fragment or named arch-ui icon.


MadeByMike:
This is good. I'd go with light grey, lighter than the text even and no hover color. We don't want this to steal attention it's a subtle indicator. It also needs more spacing on the left a to be vertically centred with the text.

gautamsi:
updated icon to 
image

gautamsi:
fixed after package org renamed.

PR #1807, opened 2019-10-22
docs: update authentication documentation

arnaud-zg:
During project initialization it's not possible to create the first user if access control is not disabled.


MadeByMike:
Awesome work @arnaud-zg \o/ @allcontributors please add @arnaud-zg for docs.

PR #1810, opened 2019-10-22
Reduced reliance on environmental variables

MadeByMike:
This is the exact same PR as #1759. No matter what I do that branch fails CI. Passes locally and nothing wrong with it I can see. Let's checkout the file on a new branch.

Reduced reliance on environmental variables in AdminUI.

Experimental features will now be `false` by default in development and must be explicitly turned on with config:

```
 new AdminUIApp({ enableDevFeatures: true })
```

The webpack build mode will still default to `development` unless the environmental variable `NODE_ENV` is set to `production`. However this can now also be configured in the constructor, allowing users more control over how environmental variables are managed. 

```
 new AdminUIApp({ buildMode: 'production' })
```


packages/app-graphql-playground/index.js
```diff
@@ -5,9 +5,14 @@ const falsey = require('falsey');
 const { addDevQueryMiddlewares } = require('./lib/devQuery');
 
 class GraphQLPlaygroundApp {
-  constructor({ apiPath = '/admin/api', graphiqlPath = '/admin/graphiql' } = {}) {
+  constructor({
+    apiPath = '/admin/api',
+    graphiqlPath = '/admin/graphiql',
+    disableLogging = falsey(process.env.DISABLE_LOGGING),
```
jesstelford: `disableLogging` should be `enableLogging`. Similarly in `packages/keystone/lib/Keystone/index.js`

MadeByMike: This would be a change. Currently the default for logging is `true` do you want to make the default `false`?

yarn.lock
```diff
@@ -1517,16 +1517,6 @@
     "@emotion/utils" "0.11.2"
     csstype "^2.5.7"
```
timleslie: Yarn lock changed without a package.json change. Is that intentional?

MadeByMike: Yes, intentional (kinda). This came out a merge conflict and now it's passing CI. Don't want to change although I can try adding it back.

PR #1817, opened 2019-10-23
Doc updates and minor functionality improvements for file field/adapters

Vultraz:
EDIT: Summary of current changes:
* `fileAdapter` is now properly treated as mandatory in File fields.
* `src` is now mandatory for `LocalFileAdapter`.
*  `path` now defaults to `src` if not provided for `LocalFileAdapter`.
* Improved documentation for both file adapters and the File field.
* Fixed this warning:
```
DeprecationWarning: File upload property ‘stream’ is deprecated. Use ‘createReadStream()’ instead.
```

### DIFF
--- a/.changeset/moody-hairs-type.md
+++ b/.changeset/moody-hairs-type.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/fields': patch
+'@keystonejs/file-adapters': patch
+---
+
+Doc updates and minor functionality improvements for file field/adapters
--- a/packages/fields/src/types/File/Implementation.js
+++ b/packages/fields/src/types/File/Implementation.js
@@ -98,7 +98,8 @@ export class File extends Implementation {
       return null;
     }
 
-    const { stream, filename: originalFilename, mimetype, encoding } = await uploadData;
+    const { createReadStream, filename: originalFilename, mimetype, encoding } = await uploadData;
+    const stream = createReadStream();
 
     if (!stream && previousData) {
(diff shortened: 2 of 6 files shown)


MadeByMike:
@Vultraz `path` is the public url used to retrieve the image and `src` is the location to upload the file. 

Although they will often match with a static server, they can differ. For example you might have a path configured on your webserver to resolve files from the root `/photo.jpg` even though the source is `/uploads/images/photo.jpg` 

Perhaps you can suggest different names? I'm open to that if it's clearer. Although we should do so in a way that `path` and `src` continue to work.

You are correct in assuming that an adapter is required and maybe this should be enforced with code. I'm also open to PRs on this.

Vultraz:
What about `uploadTo` and `serveFrom`?

MadeByMike:
> What about `uploadTo` and `serveFrom`?

Maybe `uploadDir` and `publicPath` /shrug

Vultraz:
I noticed those keys were actually renamed recently, so I think I'll leave them be.

Anyway, I added a few non-doc commits (made `adapter` mandatory, fixed a deprecation warning).

Vultraz:
Ok why are these tests not running...

Vultraz:
Ok, I've dropped the general prettier commit since this does need some review. Updated the first post with a summary of changes.

MadeByMike:
We can merge these changesets - you only need one

Vultraz:
One changeset it is!

PR #1819, opened 2019-10-23
Upgrade flow

### LINKED ISSUE #1790: Upgrade flow-bin
# Bug report

## Describe the bug

Upgrading to flow-bin 0.92.0 causes it to find new errors in CI.

We want to upgrade `flow-bin` so we need to find and fix these errors. This probably involves changes to come of the components in the AdminUI.

## Additional context
```
packages/arch/packages/day-picker/src/DayPicker/index.js:242:10

Cannot instantiate React ref because class List [1] is incompatible with string [2] in type argument ElementType.

     packages/arch/packages/day-picker/src/DayPicker/index.js
     239│       </Header>
     240│       <div>
     241│         {weekLabels}
     242│         <List
     243│           ref={listRef}
     244│           itemSize={useCallback(
     245│             index => {
     246│               const { weeks } = items[index];
     247│               return weeks.length * DAY_HEIGHT + 26.5;
     248│             },
     249│             [items]
     250│           )}
     251│           itemData={useMemo(
     252│             () => ({
     253│               items,
     254│               selectedDate,
     255│               onSelectedChange,
     256│               observer,
     257│             }),
     258│             [items, selectedDate, onSelectedChange, observer]
     259│           )}
     260│           height={6 * DAY_HEIGHT + 26.5}
     261│           itemCount={years.length * 12}
     262│           width="100%"
     263│         >
     264│           {Month}
     265│         </List>
     266│       </div>
     267│     </Wrapper>
     268│   );
```

w01fgang:
- [x] Upgrade flow-bin to the latest 0.110.1
- [x] Fix new issues
- [x] Remove suppress comments and fix errors
- [x] Types of `sarcastic` and `fixturez` are broken, replace by definition
- [x] Fix prettier issues
- [x] Upgrade `@emation/core` and `@emotion/styled` to version 10.0.22
- [x] Add lazy to flow config to decrease pressure on CPU when using flow with the editor. Now running `yarn flow` will check only changed files.

Closes #1790


flow-typed/npm/@emotion/core_vx.x.x.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module '@emotion/core' {
+  declare module.exports: any;
```
mitchellhamilton: This and the styled-base stubs shouldn’t exist. Emotion already ships flow types

.flowconfig
```diff
@@ -20,6 +25,9 @@
 module.name_mapper='^@keystone\/\(.+\)$' -> '<PROJECT_ROOT>/packages/\1'
 module.name_mapper='.*FIELD_TYPES' -> '<PROJECT_ROOT>/packages/field-views-loader/field-types'
 module.name_mapper='^@arch-ui\/\([^\/]+\)$' -> '@arch-ui/\1/src/index.js'
-server.max_workers=1
+server.max_workers=4
+suppress_comment= \\(.\\|\n\\)*\\$flow:
```
mitchellhamilton: Doing this means that consumers would also have to have this suppression comment in their config which we should definitely avoid.

flow-typed/npm/fixturez_vx.x.x.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'fixturez' {
+  declare module.exports: any;
```
mitchellhamilton: fixturez also ships flow types

flow-typed/npm/@emotion/core_vx.x.x.js
w01fgang: there are flow type errors and there is no way to fix them
could be a temporary fix

flow-typed/npm/fixturez_vx.x.x.js
w01fgang: there are flow type errors and there is no way to fix them
could be a temporary fix

.flowconfig
w01fgang: there is `$FlowFixMe`, but there are cases where errors don't have to be fixed, like in mocked functions, just a descriptive comment is needed
should I use `$FlowFixMe` instead?

mitchellhamilton: Yeah, using $FlowFixMe would be best

flow-typed/npm/@emotion/core_vx.x.x.js
mitchellhamilton: I just published a new version of @emotion/serialize which is the package where there were errors. Could you upgrade the emotion packages and remove these stubs?

w01fgang: great! will do it now!

.flowconfig
w01fgang: done

flow-typed/npm/fixturez_vx.x.x.js
w01fgang: at the end replaced by the type definition

w01fgang:
@mitchellhamilton please, could you check why tests are failing, the only thing I changed is the `@emotion/core` and `@emotion/styled`, the only error I see is `Uncaught TypeError: Cannot read property 'includes' of undefined`

w01fgang:
@mitchellhamilton I found the issue, will fix it now

PR #1839, opened 2019-10-26
Set correct users link

Fabyao:
The authentication guide documentation (https://www.keystonejs.com/guides/authentication) suggests to navigate to http://localhost:3000/users which is a broken link. I believe the correct URL should be http://localhost:3000/admin/users


jesstelford:
@all-contributors please add @Fabyao for docs, thanks!

PR #1855, opened 2019-10-28
Call out that existingItem does not contain Relationship items

jesstelford:


jesstelford:
Actually, this is incorrect. Access Control functions cannot return Promises, so we can't make this function async.

That means it's currently impossible to get any `Relationship` fields on an `existingItem` in an access control method.

jesstelford:
Opened an issue for this: https://github.com/keystonejs/keystone/issues/1856

PR #1860, opened 2019-10-28
docs: update next app documentation

arnaud-zg:
I just setup my project with `keystone-app` and after some investigation, I notice these two points;
1. By default `AdminUIApp` should have `authStrategy`
2. When you just initialize a project, you have this configuration `new AdminUIApp({ enableDefaultRoute: true, authStrategy })`

So I think it may be useful to prevent it.

### DIFF
--- a/packages/app-next/README.md
+++ b/packages/app-next/README.md
@@ -10,8 +10,6 @@ A KeystoneJS App for serving a [Next.js](https://nextjs.org/) application.
 
 ## Usage
 
-/!\ Pay attention, don't forget to disable `enableDefaultRoute` option on `AdminUIApp`.
-
 ```javascript
 const { NextApp } = require('@keystonejs/app-next');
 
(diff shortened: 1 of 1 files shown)


MadeByMike:
Hi @arnaud-zg, "By default AdminUIApp should have authStrategy" - We don't assume this but we do have it well documented [how to set-up](https://v5.keystonejs.com/api/authentication) an `authStrategy`. It not documented on the NextApp because it's a separate concern with many possible strategies. If we did this we'd be documenting if for every app.

Regarding disabling the default route, we do this if you install it via the CLI example Next App. Perhaps we could explicitly show `({enableDefaultRoute: false})` in the `adminUIApp` constructor. Although I still think this is better documented in our getting started guide. There is no reason you even need an AdminUI to use the NextApp - once again, this is very situational. We only include it in the example because it is common.

Given this all this, do you have any thoughts on how to make this "getting started with apps", experience you had better?

arnaud-zg:
@MadeByMike I agree with you about the first part, I update my changes and as you said, we explicitly show `({enableDefaultRoute: false})`.

MadeByMike:
I'm happy with that. Approved.

PR #1868, opened 2019-10-29
Admin UI: made NavIcons section easier to click

Vultraz:
Small UX change. Gets rid of the massive padding around and between the bottom nav buttons so they're easier to click. Also brings them in line with the list items above that extend to the screen edge.

**Before**
buttonlayoutorig

**After**
image

### DIFF
--- a/.changeset/warm-camels-tie.md
+++ b/.changeset/warm-camels-tie.md
@@ -1,0 +1,5 @@
+---
+'@arch-ui/navbar': patch
+---
+
+Made NavIcons section easier to click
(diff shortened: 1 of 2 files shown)


MadeByMike:
Thanks, I'll check this out and have a look today 👍

MadeByMike:
I just had a look at this one and I'd like to do it a different way. Using a child selector is a bit of a hack. When I looked at it, it became apparent that we need an icon element anyway, there is loads of redundant CSS.

https://github.com/keystonejs/keystone/pull/1869

MadeByMike:
@VinayaSathyanarayana could you checkout the new PR and see if you have any comments. Closing this one.

PR #1869, opened 2019-10-30
Vultraz admin ui nav tweak

MadeByMike:
Some change to implementation of #1868.

Also fixes the use of some outdated\bad practice flex properties.

> Gets rid of the margin around the bottom nav buttons so they're easier to click. Also brings them in line with the list items above that extend to the screen edge.


packages/arch/packages/navbar/src/PrimaryNav.js
```diff
@@ -108,6 +107,27 @@ export const PrimaryNavItem = styled(ItemElement)(({ depth, isSelected, mouseIsO
...
+  return {
+    color: mouseIsOverNav ? colors.N70 : colors.N40,
+    padding: PRIMARY_NAV_GUTTER,
+    textDecoration: 'none',
+    flexGrow: 1,
+    flexBasis: '100%',
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    ':hover': {
```
jordanoverbye: I think this key should be `':hover, :focus'` for an easy a11y win

packages/arch/packages/navbar/src/PrimaryNav.js
```diff
@@ -75,16 +73,17 @@ export const PrimaryNavItem = styled(ItemElement)(({ depth, isSelected, mouseIsO
...
     position: 'relative',
     textDecoration: 'none',
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
     transition: 'color 110ms',
 
+    flexGrow: 1,
+    flexBasis: '100%',
+
     ':hover': {
```
jordanoverbye: I think this key should be `':hover, :focus'` for an easy a11y win

MadeByMike:
```suggestion
    ':hover, :focus': {
```

packages/arch/packages/navbar/src/PrimaryNav.js
MadeByMike:
```suggestion
    ':hover, :focus': {
```

PR #1884, opened 2019-10-31
Upgrade memoize-one to 5.1.1

Vultraz:
Fixes this node warning:
warning memoize-one@4.1.0: New custom equality api does not play well with all equality helpers. Please use v5.x

Let's see what CI says....


Vultraz:
Nyet p_p

MadeByMike:
I agree with the error, do you need to commit the lock file? I expect it would have changed.

Vultraz:
@MadeByMike working on it. Been waiting a few hours for `yarn install` to finishing building/installing all the test utils (mongodb-memory-server, cypress, sharp, etc). Wish there were a way to skip those >_<

PR #1894, opened 2019-11-02
Removed stray password auth impl from Keystone

Vultraz:
PasswordAuthStrategy was split into its own package in b06fc55dafb259a95a7f9091eb84d09fad4fac37.
There's no difference between this impl and the impl in the auth-password package, and nothing
was including this one.


Vultraz:
@timleslie done

PR #1901, opened 2019-11-03
Knex adapter does not recognise config

androidi:
Something goes wrong when adapter-knex tries to get config object from BaseKeystoneAdapter.
Either the syntax is wrong or Node.js v.12.13.0 does not understand the notation for default value of an empty object.


androidi:
Not a bug, but just a config error.

PR #1916, opened 2019-11-05
Add linkField and itemLink props to ListTable

gautamsi:
I have earlier exported `ListTable` component in #1880, 

when consuming, I needed to 
* set a different field as `ItemLink` than default `_label_` field. This should also be part of `adminConfig` for lists (I will try that in a commit soon)
* set a different link target for `ItemLink` 

I made a solution to accept `linkField` prop to render `ItemLink` for that field, default to `_label_` and `itemLink` function as prop to render correct link. `itemLink` also has `item` along with `path` and `id` property on parameter, so you can do this 
```jsx
<ListTable ....
  itemLink={({ item, path }) => `/admin/customForm/${item.customFieldpath}`
  .......
  />
```

needed this in combination to #1906 where I want to create pages to admin ui which looks similar to keystone


jesstelford:
Just needs a changeset (`minor`) and it's good to go 👍

PR #1922, opened 2019-11-06
Switched to @rollup/plugin-replace

Vultraz:
Fixes this warning for rollup-plugin-replace:
"This module has moved and is now available at @rollup/plugin-replace. Please update your dependencies. This version is no longer maintained."


Vultraz:
@timleslie changeset ahoy!

PR #1923, opened 2019-11-06
Admin UI: added Home nav button

Vultraz:
Keep finding myself wanting this.
image

### DIFF
--- a/.changeset/silent-lions-roll.md
+++ b/.changeset/silent-lions-roll.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Added a "home" icon to the footer navigation in the admin-ui.
--- a/packages/app-admin-ui/client/components/Nav/NavIcons.js
+++ b/packages/app-admin-ui/client/components/Nav/NavIcons.js
@@ -4,7 +4,7 @@
 import { Fragment } from 'react';
 import { jsx } from '@emotion/core';
 
-import { TerminalIcon, MarkGithubIcon, SignOutIcon } from '@arch-ui/icons';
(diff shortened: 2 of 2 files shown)


packages/app-admin-ui/client/components/Nav/NavIcons.js
```diff
@@ -15,6 +15,10 @@ export function NavIcons() {
   let { graphiqlPath, signoutPath, authStrategy } = useAdminMeta();
   return ENABLE_DEV_FEATURES || authStrategy ? (
     <NavGroupIcons>
+      <NavIcon href="/" title="Home">
```
MadeByMike: This is inside the `ENABLE_DEV_FEATURES` flag which means it will only be shown in dev and in future only when this flag is explicitly set.

Was that intentional?

.changeset/silent-lions-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Added Home nav button
```
MadeByMike: Could we pleas
```suggestion
Added a "home" icon to the footer navigation in the admin-ui.
```

Vultraz:
@MadeByMike fixed both.

MadeByMike:
I hate to do this but i think I'm going to pass on this PR. :( There is nothing wrong with it, it's just that I had a little discussion with some UX people and we think this area needs a general re-think with the addition of multiple schemas. 

We felt like the bottom left is not the right place for a home link (and maybe other links in this footer) Currently title should also link home providing a mechanism for this. 

image

You're right that it needs a little work and I'm going to try and steal some time from design/UX to have a look at the entire sidebar.

Vultraz:
No problem! 😃 I definitely also considered the title as a possible place for the home link (I actually thought it _was_ a home link initially).

PR #1931, opened 2019-11-07
App versioning

timleslie:
Screen Shot 2019-11-07 at 2 46 40 pm
Screen Shot 2019-11-07 at 3 02 30 pm

### DIFF
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -57,6 +57,11 @@ module.exports = class Keystone {
     secureCookies = process.env.NODE_ENV === 'production', // Default to true in production
     cookieMaxAge = 1000 * 60 * 60 * 24 * 30, // 30 days
     schemaNames = ['public'],
+    appVersion = {
+      version: '1.0.0',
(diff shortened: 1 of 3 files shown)


packages/keystone/lib/Keystone/index.js
```diff
@@ -57,6 +57,11 @@ module.exports = class Keystone {
     secureCookies = process.env.NODE_ENV === 'production', // Default to true in production
     cookieMaxAge = 1000 * 60 * 60 * 24 * 30, // 30 days
     schemaNames = ['public'],
+    appVersion = {
+      version: '1.0.0',
+      addVersionToHttpHeaders: true,
```
molomby: I'd love this variable so somehow clarify that the info is added to the headers of _all_, regardless of the access controls configured. Something like `addVersionToHttpHeadersForAllRequests` but less wordy.. :/

timleslie: The docs currently say "If `addVersionToHttpHeaders` is `true` then all requests will have the header `X-Keystone-App-Version` set.". Is there some way we can make that any more clear?

PR #1942, opened 2019-11-09
Made the Content field toolbar fixed instead of floating

Vultraz:
Decided to take a shot at this. Sorry if I screwed something up - still new to React 😬 Also fixes the layering issue with the Blocks button and the incorrect editable area size.
image


MadeByMike:
This has unfortunately been superseded by other changes to the content field. This was useful for thinking bout the overall UI. Sorry it didn't land :(

PR #1949, opened 2019-11-11
Update yarn.lock to silence errors from fsevents build

timleslie:


MadeByMike:
@timleslie looks like we got a conflict with the last merge.

PR #1957, opened 2019-11-14
Updated Keystone GitHub link

thekevinbrown:
Now that Keystone 5 is at /keystone, the link goes to the right place.


timleslie:
Thanks for the fix. Could you quickly run `yarn format` over the code, looks like `prettier` isn't happy about something.

thekevinbrown:
I didn't clone the repo, I did the change in GitHub's editing interface.

MadeByMike:
Done :)

Issue #1958, opened 2019-11-14
"View related items" link breaks with integer id fields


LiamAttClarke:
# Bug report

## Describe the bug

Clicking "View related items" on a Relationship field with `many: true` links to an unfiltered list.

After some invesitigation:
This issue only exists when using KnexAdapter. KnexAdapter uses the IntegerField for ids which does not have an `in` or `not_in` filter option.

## To Reproduce

1. Setup a project with the KnexAdapter
2. Setup a one-to-many relationship field:
```
// User model
{
  fields: {
    posts: { type: Relationship, ref: 'Post.user', many: true }
  }
}
```
3. Associate Posts to a User
4. Click "View related items" beside the User.posts field.
5. :'(

## Expected behaviour

When using KnexAdapter, clicking "View related items" on a one-to-many Relationship field should navigate to a filtered list of items.

timleslie:
Closing this as I have confirmed that the reproduction steps no longer result in an error.

PR #1959, opened 2019-11-14
Added in and not_in filters to IntegerField

LiamAttClarke:
Related issue: https://github.com/keystonejs/keystone/issues/1958


gautamsi:
I think the contributor specs are taken care by a bot, you should not have to add your detail in there.  also deleting `CONTRIBUTING.md` is not necessary for this PR.

packages/fields/src/types/Integer/views/Controller.js
```diff
@@ -1,21 +1,26 @@
...
+    let arg = value.replace(/\s/g, '');
+    if (['in', 'not_in'].includes(type)) {
+      arg = `[${arg}]`;
+    }
+    return `${key}: ${arg}`;
   };
-  getFilterLabel = ({ label }: FilterLabel): string => {
-    return `${this.label} ${label.toLowerCase()}`;
+  getFilterLabel = ({ label, type }: FilterLabel): string => {
+    const suffix = ['in', 'not_in'].includes(type) ? ' (comma separated)' : '';
```
timleslie: I think it's probably easier if we just put this `suffix` explicitly on the labels at line 70 and 75, it'll keep this code nice and simple.

LiamAttClarke: I agree, that would be better. I was trying to keep the code consistent with the UUID and MongoID fields.

LiamAttClarke: Looks like adding '(comma separated)' directly on the field label results in a UI where the suffix is added to the filter drop-down and the active filter pill interface. It seems like the term `label` is overloaded for this context.

For this bug fix, I suggest we leave this implementation as it's consistent with other fields. I suggest refactoring fields to have an option to override the input placeholder value.

The FieldController.getFilterTypes could return a list of types with the included placeholder:
```
getFilterTypes = () => [
    {
      type: 'in',
      label: 'Is one of',
      placeholder: 'Is one of (comma separated)',
      getInitialValue: () => '',
    }
]
```

PR #1963, opened 2019-11-15
Insensitive by default for text field filter

gemscng:


gemscng:
@gautamsi  So is this good to merge?

gautamsi:
To me: YES

@MadeByMike

PR #1965, opened 2019-11-16
[WIP] Calendar day date picker

### LINKED ISSUE #1964: CalendarDay `date: null` when `isRequired: true`
# Bug report

## Describe the bug

When `CalendarDay` is required, creating fails because the GraphQL request sends the field value as `null` regardless of what the user input.

## To Reproduce

1. Define a list with `CalendarDay` `isRequired: true`

```
date: {
    type: CalendarDay,
    format: 'Do MMMM YYYY',
    yearRangeTo: new Date().getFullYear() + 30,
    isRequired: true,
},
```
2. Create a new item in that list and set the date field value

image

3. Click create

image

4. Observe outgoing GraphQL request payload

```js
{"operationName":"create","variables":{"data":{"date":null, ...}}}
```

## Expected behaviour

It should send the date value.

## System information

- OS: MacOS
- Browser: Chrome
- Keystone Versions
- Keystone Database: Postgres

```
    "@keystonejs/fields": "^5.2.0",
    "@keystonejs/keystone": "^5.0.0",
```

Vultraz:
UX improvement for the CalendarDay field. This adds the [Material UI date picker](https://www.npmjs.com/package/@material-ui/pickers) for day selection:

image

I looked around for a good date picker, and the Material one seemed to look the best. It doesn't clash too much with the Admin UI design, and I set its color scheme to match.

Does a few other things:
-  Fixes #1964 
- Upgrades `date-fns` to 2.7. I haven't tested `DateTime` or `DateTimeUTC` yet, though.

Still need to do:

- [ ] Documentation
- [x] Readd support for `yearRangeFrom` and `yearRangeTo`
- [x] Possibly remove `yearPickerType`
- [ ] Implement this in the two `DateTime` fields too.
- [x] Fix `defaultValue` doing nothing
- [x] Fix the default in-field value (the current date) not being saved without user interaction


Vultraz:
@MadeByMike I'm thinking, perhaps we should add `dateRangeFrom` and `dateRangeTo` to allow more granular ranges than just by year.

Abogical:
Why not a simple html5 date input?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

Vultraz:
I did consider that, but I don't see any way to style those, and they look different on every browser. It looks like Firefox on Android uses the Material UI date picker, but it would be even better to have the same thing everywhere. And plus it says support on Safari is lacking.

Vultraz:
Ok, I've added `dateRangeFrom` and `dateRangeTo` that take a date or string. I've just dropped `yearRangeFrom` and `yearRangeTo`... let me know if you want them back.

Vultraz:
Could I get some feedback on this in its current state?

Vultraz:
I'm going to close this for now. I'll keep it locally and might revisit it, but this definitely needs a lot of cleanup. #2012 already fixed a bunch of stuff this did, and the `date-fns` update should be done separately.

PR #1971, opened 2019-11-18
Removed the KnexListAdapter.createAdjacencyTable method.

timleslie:
This is part of a series of refactorings to move towards a more consistent relationship model.

This PR doesn't make any functional changes.

### DIFF
--- a/packages/adapter-knex/lib/adapter-knex.js
+++ b/packages/adapter-knex/lib/adapter-knex.js
@@ -65,6 +65,9 @@ class KnexAdapter extends BaseKeystoneAdapter {
   }
 
   async postConnect() {
+    Object.values(this.listAdapters).forEach(listAdapter => {
+      listAdapter._postConnect();
+    });
     const isSetup = await this.schema().hasTable(Object.keys(this.listAdapters)[0]);
     if (this.config.dropDatabase || !isSetup) {
       console.log('Knex adapter: Dropping database');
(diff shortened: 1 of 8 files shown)


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -102,6 +102,57 @@ class KnexAdapter extends BaseKeystoneAdapter {
     return fkResult;
   }
 
+  async _createAdjacencyTable({ tableName, relationshipFa, leftListAdapter }) {
+    // Create an adjacency table for the (many to many) relationship field adapter provided
+    const dbAdapter = this;
+    try {
+      console.log(`Dropping table ${tableName}`);
+      await dbAdapter.schema().dropTableIfExists(tableName);
```
MadeByMike: Do you don't want to drop the table every time? Won't this loose some relationships?

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -102,6 +102,57 @@ class KnexAdapter extends BaseKeystoneAdapter {
...
+    } catch (err) {
+      console.log('Failed to drop');
+      console.log(err);
+      throw err;
+    }
+
+    // To be clear..
+    const leftPkFa = leftListAdapter.getPrimaryKeyAdapter();
```
MadeByMike: If we're going to be clear can we be clearer? `leftPrimaryKey(what is 'Fa')`

packages/adapter-knex/lib/adapter-knex.js
timleslie: If we're at this point of the code it's assumed that we're recreating the database, so we cleanup the table we're about to create just to be 100% certain we're starting from a clean slate.

packages/adapter-knex/lib/adapter-knex.js
timleslie: The cleanliness of this code is out of scope for this PR. It's all just moving around here, and it's almost definitely going to change in future PRs.

PR #1976, opened 2019-11-19
Remove _populateMany()

timleslie:
Since the refactoring of how relationships are resolved, this code is no longer required.


timleslie:
On second thoughts this might not be where we want to end up... let me think this through a bit longer.

PR #1978, opened 2019-11-19
Virtual fields

### LINKED ISSUE #1117: Virtual Fields (aka; Calculated Fields)
Virtual field allow adding a field to the GraphQL end point, as well as the Admin UI, which are not backed by a direct database field, but instead are calculated at request time based on either the result of the other requested fields, related fields, or external data.

Examples of virtual fields:

- `_label`; a human readable short summary of an item
- `approvedPosts`; A shorthand for:
    ```graphql
    query {
      User {
        approvedPosts: _postsMeta(where: { status: 'Approved' }) {
          count
        }
      }
    }
    ```

Could be implemented as a new Field Type:

```javascript
import { Virtual, Integer } from '@keystone-alpha/fields';

keystone.createList('User', {
  fields: {
    approvedPosts: {
      // Use Integer's views, outputType etc.
      // But override all the functionality with the `Virtual` type.
      type: { ...Integer, Virtual },
      // Specify any data required to calculate this field.
      // Will be mixed into the graphQL query for getting this item's data.
      queryFragment: `
        _postsMeta(where: { status: 'Approved' }) {
          count
        }
      `,
      resolver: item => item._postsMeta.count,
    }
  }
});
```

Or, could be more explicit in setting `views`/`outputType` so it becomes the `Virtual` field's responsibility to take those options and mix them into the type correctly.

```javascript
import { Virtual, Integer } from '@keystone-alpha/fields';

keystone.createList('User', {
  fields: {
    approvedPosts: {
      type: Virtual,
      views: Integer.views,
      outputType: Integer.outputType,
      // Specify any data required to calculate this field.
      // Will be mixed into the graphQL query for getting this item's data.
      queryFragment: `
        _postsMeta(where: { status: 'Approved' }) {
          count
        }
      `,
      resolver: item => item._postsMeta.count,
    }
  }
});
```

To describe a nested object type, may need to also allow overwriting `getGqlAuxTypes()`.

## Notes

- To ensure the correct typing information for the GraphQL API, and also for the display aspect, we need to allow setting the GraphQL return type.
- The field would be displayed in the AdminUI's list & details views
  - This display info is specified like any other field type; via _views_ on the type (but set by options).
- The field would _not_ be displayed in the AdminUI's edit views
- The field would show up in the GraphQL API's queries
  - The output type would be specified as an option.
- The field would _not_ be available as input to the GraphQL API's mutations

## Todo
- [ ] Implement #220: _Optimize database queries to only pull out fields actually requested_
- [ ] Implement query dependencies (#1116 _Enable `_label` field to specify its GraphQL dependencies_)
- [x] Fix #487: _`encodeSortBy` not working with non-virtual fields_

MadeByMike:
Added a new field type 'Virtual' that allows creation of fields that return data computed from other field values or outside Keystone.

Simple example:

```js
  fields: {
    firstName: { type: Text },
    lastName: { type: Text },
    name: {
      type: Virtual,
      resolver: item => (`${item.firstName} ${item.lastName}`)
      };
    },
  },
});
```

Complex example:

```js
  fields: {
    movies: {
      type: Virtual,
      extendGraphQLTypes: [`type Movie { title: String, rating: Int }`],
      graphQLReturnType: `[Movie]`,
      graphQLReturnFragment: `{
        title
        rating
      }`,
      resolver: async item => {
        const response = await fetch('http://example.com/api/movies/');
        const data = await response.json();
        return data.map(({ title, rating }) => ({ title, rating }));
      },
    },
  },
});
```

Computed fields can't be sorted or filtered.
 

Closes #1117


VinayaSathyanarayana:
Assume the computed fields would be saved in the database

MadeByMike:
No the point of a computed field is that it's not saved. If you want it to be in the database use a normal field. Make it read only and apply a hook to set the value on save.

VinayaSathyanarayana:
Can readonly fields be modified by a hook?

MadeByMike:
Update, I've been getting a few pre-req changes in place. And were going to change this to "virtual-fields" so that it's clear the content doesn't persist and is different from `resolveInput` hook

MadeByMike:
> Can readonly fields be modified by a hook?

There is no save mutation available for them so no `resolveInput` hook will ever run. Other than that they look like normal fields so I assume they could trigger appropriate hooks. I haven't tried yet.

packages/fields/src/Implementation.js
```diff
@@ -13,6 +13,7 @@ class Field {
     this.config = config;
     this.isRequired = !!isRequired;
     this.defaultValue = defaultValue;
+
```
MadeByMike:
```suggestion
```

wesbos:
Are virtual fields not published anywhere yet?

madhums:
@wesbos looks like its published https://www.keystonejs.com/keystonejs/fields/src/types/virtual/

PR #1983, opened 2019-11-20
Allow lists to work without fields (or fields with no access)

MadeByMike:
This PR allows you to create lists without any fields (which is PROBABLY a bad idea) but it's nice that you can do it.

The above is really a side-effect of being able to handle lists that have all access control set to 'false' which is plausible for some edge-case situations.

But that is also kinda a side-effect. The true reason for this PR is that it removes items from the schema when lists have no fields with the required access. E.g. no create for lists where all fields have no create access and no update for lists where all fields had no update access. -- This fixes a bug where an invalid schema would be generated.

This edge case becomes particularly important for computed fields that have no create and update access.

### DIFF
--- a/packages/keystone/lib/List/index.js
+++ b/packages/keystone/lib/List/index.js
@@ -152,8 +152,8 @@ module.exports = class List {
     const nonIdFieldNames = Object.keys(fields).filter(k => k !== 'id');
     this.adminConfig = {
       defaultPageSize: 50,
-      defaultColumns: nonIdFieldNames.slice(0, 2).join(','),
-      defaultSort: nonIdFieldNames[0],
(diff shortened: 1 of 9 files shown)


packages/keystone/lib/List/index.js
```diff
@@ -152,8 +152,8 @@ module.exports = class List {
     const nonIdFieldNames = Object.keys(fields).filter(k => k !== 'id');
```
timleslie: Just wondering if we should account for the case of fields with `read: false`, as these probably shouldn't be used as `defaultColumns` or `defaultSort`.

timleslie: On closer inspection we don't have that info available at this stage, so this is fine for now.

PR #1986, opened 2019-11-21
Added S3 file adapter

LiamAttClarke:
Additional file adapter to allow AWS S3 file uploads.


timleslie:
I'll hopefully get a chance to take a look at this, but in the meantime if you run `yarn format` that will sort out the complaints that prettier has on CI 👍

packages/file-adapters/README.md
```diff
@@ -48,3 +48,37 @@ const fileAdapter = new CloudinaryAdapter({
...
+```javascript
+const { S3Adapter } = require('@keystonejs/file-adapters');
+
+const fileAdapter = new S3Adapter({
+  accessKeyId: 'ACCESS_KEY_ID',
+  secretAccessKey: 'SECRET_ACCESS_KEY',
+  region: 'us-west-2',
+  bucket: 'bucket-name',
+  folder: '/cat-photos',
+  publicUrl: ({ id, filename, _meta }) => `https://{CF_DISTRIBUTION_ID}.cloudfront.net/cat-photos/${filename}`,
```
timleslie: Should this be `${CF_DISTRIBUTION_ID}`?

LiamAttClarke: It was meant as a placeholder but I updated it to be a constant.

timleslie:
Hey @LiamAttClarke, sorry this took a while to get around to. This PR has got two things going on in it and I think I'd like to split them out into separate PRs. Could we create a new PR which just has the changes to the `delete` function and then that'll leave this PR with just the changes required to add `S3` support. I think that will make it easier to evaluate each of the changes independently of each other. Thanks!

LiamAttClarke:
> Hey @LiamAttClarke, sorry this took a while to get around to. This PR has got two things going on in it and I think I'd like to split them out into separate PRs. Could we create a new PR which just has the changes to the `delete` function and then that'll leave this PR with just the changes required to add `S3` support. I think that will make it easier to evaluate each of the changes independently of each other. Thanks!

I've removed the promise refactor.

The S3Adapter.delete method takes the file data structure directly instead of assuming the `existingItem` object will contain a key `file`. I think this is a cleaner approach than that currently taken by the LocalFileAdapter.

Issue #1982, opened 2019-11-20
Add an isOrderable flag to all field types to indicate whether they can be used as an orderBy field


timleslie:
Not all field types are necessarily orderable (e.g. Location, Relationship, Password, etc). We should set an explicit field on field types to indicate this and then respect this flag when setting up the `orderBy` enumeration and also when presenting column sort functions in the Admin UI.

c.f. #1981

molomby:
#183 has some related thoughts.

Ultimately where we want to end up is having field types (or field type adapters?) define the "orderings"/"collation" they support. Some, like `Password`, probably shouldn't define any. Others like, say, `CloudinaryImage` might want to define multiple (height, width, file size, ...?).

See also #391 for some relevant complications around ordering `null` values.

PR #1988, opened 2019-11-22
Add isOrderable property to fields

MadeByMike:
Adds an `isOrderable` option to the default field implementation meaning it can be used by all fields.

Sets whether the field can be ordered in the adminUI:

image

@timleslie I checked and there is no way to infer this from `orderingInputFields` most fields implement `gqlQueryInputFields` and if they don't include ordering there is no way to know (without a large refactor). I think it makes sense for this to be a user option as well so the fields that wast to explicitly set no order will not have this and a config option and default to false internally.

Closes: #1982


packages/fields/README.md
```diff
@@ -56,6 +56,7 @@ Fields share some standard configuration options.
 | `defaultValue` | `Boolean` \| `Function`             | `undefined` | A default value of the field.                                                           |
 | `isUnique`     | `Boolean`                           | `false`     | Whether or not the field should be unique.                                              |
 | `isRequired`   | `Boolean`                           | `false`     | Whether or not the field should be mandatory.                                           |
+| `isOrderable`  | `Boolean`                           | `true`      | Whether or not the field can be ordered in the AdminUI.                                 |
```
gautamsi: should this be part of `adminConfig`, I have always been in favor of merging config and adminConfig

gautamsi: is the Relationship field orderable by default?

MadeByMike: Currently all fields are orderable by default

MadeByMike: Changed this so that default is not orderable, and made this an internal property

packages/fields/src/Implementation.js
```diff
@@ -29,15 +30,19 @@ class Field {
...
       schemaNames,
       listKey,
       fieldKey: path,
       defaultAccess,
-      access: access,
+      access,
     });
   }
 
+  parseFieldAccess(args) {
```
timleslie: If we're going to make this change in this PR we should make sure we reference it in the changelog as a `minor`.

packages/fields/README.md
MadeByMike:
```suggestion
```

.changeset/six-fishes-give.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@keystonejs/fields': minor
+---
+
+Adds an `isOrderable` option to field configs that controls whether the field is sortable in the AdminUI.
```
MadeByMike:
```suggestion
Adds an `isOrderable` option to field implementation that controls whether the field is sortable in the AdminUI. Defaults to false and is set to true by sortable field types.

The `parseFieldAccess` method can now also be overwritten in the field implementation class.
```

PR #2005, opened 2019-11-27
Improve blocks UI

MadeByMike:
- Simplifies the positioning of the "Add block" icon
- Update presentation
 - Adds a better menu presentation for the "Insert Image" blocks
 - Add some contextual titles etc... to make clearer what the action is

image


MadeByMike:
> There are a bunch of instances of hardcoded color values here. We should probably be using named colors from `@arch-ui/theme` wherever possible. Other than that I think everthing here is good.

DOne

MadeByMike:
@timleslie And I get a ✅?

gautamsi:
do we plan to release this soon?

jesstelford:
The Unsplash, and Embedly blocks aren't styled the same as the CloudinaryImage one:

Screen Shot 2020-01-13 at 2 05 58 pm

In current `master`, they're horizontally stacked:

Screen Shot 2020-01-13 at 2 11 45 pm

test-projects/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jesstelford: Need to remove this file so it doesn't interfere with the main `yarn.lock`

yarn.lock
```diff
@@ -1226,6 +1226,14 @@
     core-js "^2.6.5"
     regenerator-runtime "^0.13.2"
 
+"@babel/runtime-corejs2@^7.6.3":
```
jesstelford: Is there a reason these dependencies are updated in the lockfile? Or was this just yarn doing its thing to get all the latest things?

MadeByMike: Probably yarn doing its thing. This was a fresh clone.

PR #2008, opened 2019-11-27
Add `operation` arg to hooks

molomby:
Currently, `beforeChange` and `afterChange` hooks receive the `operation` arg but it isn't documented. The _internal_ `_resolveInput` and `_validateInput` list functions get the `operation` but don't pass it though to the configured hooks.

This PR makes this all consistent by just adding the `operation` arg to all hooks (and updating the docs).

### DIFF
--- a/.changeset/real-rats-shave/changes.json
+++ b/.changeset/real-rats-shave/changes.json
@@ -1,0 +1,1 @@
+{ "releases": [{ "name": "@keystonejs/keystone", "type": "minor" }], "dependents": [] }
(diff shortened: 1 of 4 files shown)


gautamsi:
can you also change list-plugins to make use of `operation` as part of this PR, if not I will create a PR soon. 

see L38 and L46 how we detect now.
https://github.com/keystonejs/keystone/blob/5a5905d07c29d6c49dffcc93a576cb2c599c4a65/packages/list-plugins/lib/tracking/atTracking.js#L37-L47

molomby:
@gautamsi lets do a separate PR for that, @MadeByMike's already approved this as is

PR #2009, opened 2019-11-27
Update README.md

virgafox:
Probably a typo in docs


Vultraz:
Can you add an extra space so the table columns line up?

PR #2012, opened 2019-11-27
CalendarDay fixes

### LINKED ISSUE #1964: CalendarDay `date: null` when `isRequired: true`

Vultraz:
Fixes #1964 independent of #1965 and does a few other things. Should be easier to do the date-fns upgrade with everything working properly.

Changes:
- Ensured format, yearRangeFrom, and yearRangeTo defaults are set correctly
- Removed non-functional yearPickerType
- Made displayed date format consistent between list and detailed views
- Moved date range validation to validateInput
- Added some vertical margins to <Alert>


packages/fields/src/types/CalendarDay/views/Field.js
```diff
@@ -1,41 +1,35 @@
...
 
-  render() {
-    const { autoFocus, field, value, errors } = this.props;
-    const htmlID = `ks-input-${field.path}`;
+  return (
+    <FieldContainer>
+      <FieldLabel htmlFor={htmlID} field={field} errors={errors} />
+      <FieldInput>
+        <TextDayPicker
+          id={`ks-daypicker-${field.path}`}
```
MadeByMike:
```suggestion
          id={htmlID}
```

Vultraz: `htmlID` is `ks-input-${field.path}`, though. A different value.

packages/arch/packages/day-picker/src/TextDayPicker.js
```diff
@@ -1,29 +1,35 @@
...
 
   return (
     <Input
       value={value}
       placeholder="Enter a date..."
       onBlur={() => {
-        let newDate = parseDate(value);
+        const newDate = parseDate(value);
         onChange(newDate);
         setValue(formatDate(newDate));
```
MadeByMike: We should validate the date format and if it is valid don't set the value.

packages/fields/src/types/CalendarDay/views/Field.js
Vultraz: Also looks like the tests look for that value (`ks-daypicker-`).

MadeByMike: They need to be the same because we have `htmlFor={htmlID}` and this need to match the id on the input.

My suggestion is to change `const htmlID = `ks-input-${field.path}`;` to match

Vultraz: Ok, changed `htmlID`.

PR #2013, opened 2019-11-27
Corrected assignment to keystone const

dacre-denny:
Documentation shows the `keystone` variable being accessed before it is defined, with the result of `keystone.createList(..)` then assiged to `const keystone`. This is cause for confusion. The proposed adjustment continues the convention of the `keystone` instance being defined `const keystone = // ...`, as is the case elsewhere in the docs

### DIFF
--- a/docs/guides/authentication.md
+++ b/docs/guides/authentication.md
@@ -34,7 +34,9 @@ const { Text, Password } = require('@keystonejs/fields');
 const { PasswordAuthStrategy } = require('@keystonejs/auth-password');
 const { AdminUIApp } = require('@keystonejs/app-admin-ui');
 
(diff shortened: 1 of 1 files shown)


docs/guides/authentication.md
```diff
@@ -34,7 +34,9 @@ const { Text, Password } = require('@keystonejs/fields');
 const { PasswordAuthStrategy } = require('@keystonejs/auth-password');
 const { AdminUIApp } = require('@keystonejs/app-admin-ui');
 
-const keystone = keystone.createList('User', {
+const keystone = // ...
```
MadeByMike:
```suggestion
```

MadeByMike: Better to just delete I think

PR #2023, opened 2019-12-01
Fixed letters being cut off in Admin UI cards

Vultraz:
Before (notice the bottom of the G is cut off):
image

After:
image

Space between the two lines is text is also increased as a result (it's too close with a `line-height` of 1.1), which I think looks better.


packages/app-admin-ui/client/pages/Home/components.js
```diff
@@ -76,7 +75,7 @@ export const Name = styled.span(
   font-size: 1.1em;
```
MadeByMike:
```suggestion
  font-size: 1.1rem;
```

packages/app-admin-ui/client/pages/Home/components.js
```diff
@@ -76,7 +75,7 @@ export const Name = styled.span(
...
   font-weight: 500;
   max-width: 100%;
-  overflow: hidden;
+  overflow-x: hidden;
```
MadeByMike:
```suggestion
  overflow: hidden;
```

packages/app-admin-ui/client/pages/Home/components.js
```diff
@@ -76,7 +75,7 @@ export const Name = styled.span(
...
   text-overflow: ellipsis;
   transition: border-color 80ms linear;
   white-space: nowrap;
```
MadeByMike:
```suggestion
  white-space: nowrap;
  padding-right: 2em;
```

MadeByMike: Allows text ellipsis to work

packages/app-admin-ui/client/pages/Home/components.js
Vultraz: `overflow-y` is what caused the letter bottoms to be cut off. I stuck with `overflow-x` so long labels wouldn't extend past the right edge of the card.

packages/app-admin-ui/client/pages/Home/components.js
Vultraz: Added. Needed to switch to a div for it to work, though (and I still needed `overflow-x`).

packages/app-admin-ui/client/pages/Home/components.js
MadeByMike: It was really the line-height that was causing the issue not the overflow.

Vultraz: Ah, you're right! Done.

packages/app-admin-ui/client/pages/Home/components.js
MadeByMike: Add `box-sizing: border-box` rather than changing it to a div.

Vultraz: Done.

Vultraz:
Ok, I don't understand why the tests are failing now :|

PR #2024, opened 2019-12-01
Fixed signOut codepath being called multiple times in succession

Vultraz:
When implementing the `beforeUnauth` hook I noticed this signout path was actually getting called like 8 times in a row. Turns out it was triggering a sign out every render. 😬


Vultraz:
Tests complaining about #2021.

PR #2028, opened 2019-12-02
Remove flow

mitchellhamilton:
We've discussed it and we're going to remove Flow. Maintaining our usage of Flow has been causing more pain than the benefit we get out of it.


gautamsi:
can we create a branch/tag which has Flow (just before this merge)? it would be helpful to use them when we go for TypeScript (soon hopefully). lot of typings can be copied to TypeScript as is even in jsdoc format.

see #1373

mitchellhamilton:
We don't really need to make a branch since we can look at the git history to get to a version of the repo with flow in it

gautamsi:
A git tag would be helpful and easy to get to

mitchellhamilton:
Just pushed a `flow` tag from what is currently on master.

PR #2034, opened 2019-12-03
Conditional CI

MadeByMike:
This adds conditional logic to running tests on Circle CUI.

If looks at the files that have changed will skip tests if not required.

This adds the ground-work for more conditional logic in future where we only test packages that are change.

Current logic:

 - If the PR only contains markdown or media files, skip CI
 - If the PR is one of a few selected packages ignore all tests other than 'simple tests' 

Future PR:
  - Re-write Cypress GraphQL tests to Jest
  -  Report specific time tests take to Circle
  -  Additional skip logic

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,69 +29,71 @@ x-save-cache: &save-cache
   key: v14-dependencies-{{ checksum "yarn.lock" }}
 
 x-integration-steps: &integration-steps
-  - run:
-      name: Yarn - Install packages
-      command: CYPRESS_CACHE_FOLDER=$CIRCLE_WORKING_DIRECTORY/node_modules/cypress/.cache/ yarn
-  - save_cache: *save-cache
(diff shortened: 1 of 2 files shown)


MadeByMike:
I don't think this needs a changeset?

.circleci/config.yml
```diff
@@ -59,36 +61,42 @@ x-integration-steps: &integration-steps
     - store_artifacts:
         path: /tmp/screenshots
 
+x-simple-tests: &simple-tests
```
mitchellhamilton: We could move this back to the job now since the thing we tried with this didn't work.

MadeByMike: I actually think it reads nicer as an inclusion like this. Makes everything at the bottom neat and easy to read. What do you think?

mitchellhamilton: Sure - sounds good

Issue #2001, opened 2019-11-26
List-level hooks for authentication and deauthentication mutations


molomby:
Recently we've started thinking of the `auth` operations on a list as almost in the same category as create, read, update and delete operations (CRUDA?). This has been [extended to access control](https://github.com/keystonejs/keystone/pull/1627) but not yet been rolled out to hooks.

## Use Cases

* Logging user signins to track active users; eg. updating a `lastSeenAt` Timestamp field (or similar) on the user item
* Adding rate limiting based on request characteristics (eg. IP range), auth ID value (eg. username) or both
* Account lockout (and/or notification) after some number of failed attempted
* Adding additional layers of security to authentication mutations (eg. IP whitelisting) without requiring a custom auth strategy
* Additional logging for security auditing or debug

## Existing Solutions

Note the `auth-passport` packed does already support `onAuthenticated()` and `onError()` hooks but..

* Theres no reason for these to be auth strategy-specific
* Their signatures are express-middleware-like (makes sense for Passport but doesn't necessarily generalise to other auth strategies)
* Not setup to intercede _before_ the authentication is processed (although I guess you could stop the token being returned)

## Proposed Solution

My first pass at a design would be to add three list-level hooks to lists with signatures like..

```js
async beforeAuth(
  authStrategy // The auth strategy to be used (alternatively could be `authType`)
  args // The arguments supplied to the auth mutation (ie. `email`, `password`)
  context // GraphQL context object
  audiences // Array of audience keys included in the session
): {
  // Nothing? Or maybe replace/override..
  authStrategy,
  args,
  audiences
};
```

```js
async afterAuth(
  authStrategy // The auth strategy to be used (alternatively could be `authType`)
  args // The arguments supplied to the auth mutation (ie. `email`, `password`)
  context // GraphQL context object
  audiences // Array of audience keys included in the session
  item // The item being authenticated as
  success // Boolean
  message // Message being returned to caller
  token // The session token
): {
  // Nothing? Or override...
  item, // The item being authenticated as
  success, // Boolean
  message // Message being returned to caller
};
```

```js
async beforeUnauth(
  context
  session
): {}
```

## Open Questions

* Do these hooks make sense in the context of existing hook names, etc.?
* Should these hooks be able to return/replace the values being used (ie. override the auth strategy, return alternate args to use for auth, etc.)...?
* .. Or should they only be allowed to produce side effects?
* Anything else I've missed?

jesstelford:
[...]

PR #2039, opened 2019-12-04
Authentication hooks

molomby:
Implementing authentication hooks to lists as per #2001.

Starting with docs to formalise/get agreement on the spec.

### DIFF
--- a/.changeset/dry-sheep-report.md
+++ b/.changeset/dry-sheep-report.md
@@ -1,0 +1,24 @@
+---
+'@keystonejs/keystone': minor
+'@keystonejs/api-tests': patch
+---
+
+Added [authentication hooks](https://www.keystonejs.com/api/hooks).
+
+You can now customise the behaviour of authentication mutations as follows:
+
+```js
+keystone.creatAuthStrategy({
+  type: PasswordAuthStrategy,
+  list: 'User',
+  hooks: {
+    resolveAuthInput: async (...) => {...},
+    validateAuthInput: async (...) => {...},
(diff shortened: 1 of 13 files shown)


timleslie:
I'm a 👍 on these APIs.

Vultraz:
Quick aside since you're editing that file. The current Apollo Context link in the Hooks API docs is broken. The new link is `https://www.apollographql.com/docs/apollo-server/data/data/#context-argument`.

molomby:
@Vultraz indeed it is! I've fixed them already in [this branch](https://github.com/keystonejs/keystone/commit/048c0d6579575820b376056d86bc88579a432491) but that's all caught up in #2041. I'll fix it here too.

timleslie:
Just making a note there that the new providers API (#2418) might have implications to some of thing things outlined here.

gautamsi:
@timleslie just curious if this list will be another `List type` (Auth list type) along the line of recent list type refactoring?

if that is the case, I guess we can have split page in documentation for each type of list, right?

timleslie:
> @timleslie just curious if this list will be another `List type` (Auth list type) along the line of recent list type refactoring?
> 
> if that is the case, I guess we can have split page in documentation for each type of list, right?

This won't be a separate list type, this will just be adding a `hooks` parameter to `keystone.createAuthStrategy()` method. Under the hood this method uses a separate `provider` to generate the GraphQL, but it's a slightly different thing to a `List type`.

adrianbienias:
I'd love to use those auth hooks. How come I don't see anything in docs about them?

PR #2046, opened 2019-12-06
Removed some workarounds for issues with older graphql versions

Vultraz:
Min graphql version used is 14.4.2 now.


Vultraz:
Ok, looks like this needs some more work...

timleslie:
@Vultraz Hey, have you had a chance to look into the failures here? Let us know if you have something in the works, otherwise probably best for us to close out this particular PR and maybe leave it as a backlog issue instead?

Vultraz:
@timleslie I did look into the failures, but I was hoping for some guidance. The vast majority of the test failures are something like,
```js
    - Expected
    + Received
...

    -       "id": 1,
    +       "id": "1",
```
I can't tell if the tests need to be updated, or if the new behavior is incorrect.

Vultraz:
@timleslie ok, I went back to this and kept one of the workarounds. Now everything passes.

PR #2049, opened 2019-12-06
fix bug that no tracking field is added when create record without data

chochihim:
If I run a mutation without any data:
```
mutation {
  createMyObject(data: {}) {
    id
  }
}
```
No tracking field will be added due to the checking on `originalInput`. I think the checking should be applied for "update mode" only.


molomby:
Shouldn't the `operation` be supplied to that function and checked `=== 'create'` instead? Better to keep these kinds of things explicit rather than badly implying the operation from other 
arguments. It looks like the `operation` is already being passed in too.

molomby:
This fix seems good to me.

PR #2053, opened 2019-12-06
Replace bcrypt with bcryptjs

OskarsPakers:
It is a nightmare to run this project on Windows as it uses bcrypt -> node-gyp which requires Python and Windows build tools.


timleslie:
@molomby I think you're best placed to review this and make an executive decision.

sarneaud:
I like the "just works" feature of using bcryptjs.  The 30% performance loss isn't such a big deal, but the fact that it blocks the main thread is going to be a problem for some sites.

It would be great if we supported both.

gautamsi:
Usually the deployment to cloud is on linux based servers (cheaper than windows counterparts)

Dev environments on windows may use the instructions rather than the loss in performance. 

Native modules are higher performance and usually non blocking. Nodejs is single threaded traditionally, we should avoid adding overhead. 

We already have multiple overhead in process like ODM (knex/monfoose), graphql, keystone.

OskarsPakers:
Maybe somebody should contribute to bcryptjs and make it not blocking the main thread if that is possible?

No doubt native modules are more performant. 
Where ia the border line when NOT to use a native module?
Here the cost of this performance gain is that I almost lost interest in keystone project. It might or might not be worth it.

gautamsi:
I have two internal sites running on windows, no issue with that. 

Just install windows build tools and run cmd with admin while installing dependencies or use developer mode in win10 to help you with windows setup.

MadeByMike:
@OskarsPakers any reason for not using WSL. It's pretty much built-into Windows and makes these things a lot less painful. It's why WSL exists.

MadeByMike:
I say this as a long time Windows user.

Vultraz:
FTR, I'm also using Windows to run KS and I've had no issues (though granted, I do have a dev environment set up). I'm not using WSL.

OskarsPakers:
Perhaps I was doing something really wrong, but that was a journey to the successful npm install.

- So I did npm install

```
> bcrypt@3.0.7 install C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt
> node-pre-gyp install --fallback-to-build

node-pre-gyp WARN Using needle for node-pre-gyp https download
node-pre-gyp WARN Tried to download(404): https://github.com/kelektiv/node.bcrypt.js/releases/download/v3.0.7/bcrypt_lib-v3.0.7-node-v72-win32-x64-unknown.tar.gz
node-pre-gyp WARN Pre-built binaries not found for bcrypt@3.0.7 and node@12.13.1 (node-v72 ABI, unknown) (falling back to source compile with node-gyp)

 Error: Could not find any Python installation to use
```
 - Found out windows build tools can help with Python (installed from elevated PowerShell)
 ` npm i windows-build-tools -g`
- npm install again
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\Common7\IDE\VC\VCTargets\Microsoft.Cpp.WindowsSDK.target
s(46,5): error MSB8036: The Windows SDK version 10.0.17763.0 was not found. Install the required version of Windows SDK
 or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution". [C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt\build\bcrypt_lib.vcxproj]
gyp ERR! build error
gyp ERR! stack Error: `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin\MSBuild.exe` failed with exit code: 1
```
- Found out that some build tools I had were not the with the correct version.
Installed Microsoft visual studio community 2017 with Workload "Desktop development with C++" https://www.visualstudio.com/thank-you-downloading-visual-studio/?sku=Community&rel=15 of size 1.58 GB

```
npm i

> bcrypt@3.0.7 install C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt
> node-pre-gyp install --fallback-to-build

node-pre-gyp WARN Using needle for node-pre-gyp https download
node-pre-gyp WARN Tried to download(404): https://github.com/kelektiv/node.bcrypt.js/releases/download/v3.0.7/bcrypt_lib-v3.0.7-node-v72-win32-x64-unknown.tar.gz
node-pre-gyp WARN Pre-built binaries not found for bcrypt@3.0.7 and node@12.13.1 (node-v72 ABI, unknown) (falling back to source compile with node-gyp)
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
  blowfish.cc
  bcrypt.cc
  bcrypt_node.cc
  win_delay_load_hook.cc
     Creating library C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt\build\Release\bcrypt_lib.lib and object C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt\build\Release\bcrypt_lib.exp
  bcrypt_lib.vcxproj -> C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt\build\Release\\bcrypt_lib.node
  Copying C:\Users\o.pakers\Downloads\temidio\node_modules\bcrypt\build\Release\/bcrypt_lib.node to C:/Users/o.pakers/Downloads/temidio/node_modules/bcrypt/lib/binding\bcrypt_lib.node
          1 file(s) copied.
```
Success!

And I am not arguing that it is not possible to install keystone on Windows, but this is far from a simple npm install. And as bcrypt is the problem, switching to bcryptjs would make it easier.

@MadeByMike I actually have subsystem for Linux and I use it, however, not all of my colleagues have it. Also switching to Linux is just running away from the problem.

timleslie:
Another issue with using `bcrypt` arises when trying to deploy to an AWS lambda. Deployment to a lambda requires building the `node_modules` directory locally and then supplying it to the lambda function to use. If the build is done on OSX and the lambda is deployed on a linux system then the `bcrypt` package will not work, as OSX uses a different binary file format to linux.

timleslie:
Closed by #2523

PR #2070, opened 2019-12-09
Generate initial user for CLI starter temaplte

MadeByMike:
New users are having trouble with the starter template. Doc's and instructions in code did not seem to be helping as intended so I think a script to generate an initial user is required.

In the long term this should be changed to follow whatever method we come up with for seeding.


packages/create-keystone-app/example-projects/starter/initial-data.js
```diff
@@ -0,0 +1,51 @@
+// This script will check if the User table is empty and if
+// required seed the table with an initial Admin user.
+const randomString = () =>
+  Math.random()
+    .toString(36)
+    .substring(2, 6) +
+  Math.random()
+    .toString(36)
+    .substring(2, 6);
```
sarneaud: `Math.random()` is not very good.  How about something like this?

```
const crypto = require('crypto');
const randomString = () => crypto.randomBytes(6).hexSlice();
```

MadeByMike: This is not intended to be a secure password by any measure. I thought about using `12345678` and honestly that might even be better because I want people to change it.

In the end I went random just incase someone accidentally deploys with this.

MadeByMike: Now having defended my decision I realise `crypto` is native node and the code is shorter... so why not have it be better.

packages/create-keystone-app/example-projects/starter/initial-data.js
```diff
@@ -0,0 +1,44 @@
...
+    `query {
+      _allUsersMeta {
+        count
+      }
+    }`
+  );
+
+  if (count === 0) {
+    const password = randomString();
+    const email = 'admin@keystonejs.com';
```
jesstelford: Can we make this `@example.com` to avoid any issues with people in the future accidentally sending emails, etc?

PR #2102, opened 2019-12-13
Clarify that DateTimeUTC is a separate package

wesbos:


Vultraz:
I have been wondering why that's a separate package and not an option in the regular `DateTime` field 🤔

PR #2113, opened 2019-12-17
Remove pagination from Admin List page

julionav:
* Add prop `showAllPages` to arch-ui Pagination element. When this prop is false, the element that shows/hides all the pages by number won't be displayed.
* Used the new prop to hide the pagination by number from the Admin Table List

### DIFF
--- a/.changeset/afraid-seas-develop.md
+++ b/.changeset/afraid-seas-develop.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@arch-ui/pagination': minor
+---
+
+Removed pagination per page from Admin List UI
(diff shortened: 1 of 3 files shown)


MadeByMike:
I think this should be a hook

julionav:
> I think this should be a hook

What exactly should be a hook? 🤔

MadeByMike:
Sorry Julio. I replied on my phone and that was more a bookmark for more detail later. Basically we've realised that there are many types of UI customisation users might want in the Admin App and the best way to provide this is with "hooks" that allow them to replace pages or components.

Adding options for every component could grow into a huge configuration API whereas allowing people to  hook into the Admin UI at different points is one API and allows greater customisation.

Issue #2114, opened 2019-12-18
Update doc site publishing process to convert link format


molomby:
Basically, pages within the [Keystone docs site](https://www.keystonejs.com/quick-start/) are available a different URL paths than the Markdown file paths from which they're generated. This is a Good Thing -- the docs site and the repo serve different purposes so, naturally, will tend to organise their information differently. However, when publishing the docs site, we need to convert from one link schema to the other. **This issue represents the work to update the publishing process to perform this conversion.**

See [the guidelines](https://github.com/keystonejs/keystone/issues/2041#issuecomment-566802733) in issue #2041 for how links are expected to be represented in the Markdown files. The `meta` headers included in most docs should give us enough information to create a mapping between the two link schemas.

molomby:
@Noviny and/or @mitchellhamilton -- either of you want to take this on? The spec should now be clear enough to action.

Issue #2041, opened 2019-12-04
Internal links in docs either work on the docs site OR GitHub (but not both)


molomby:
Right now a lot (all?!) links in the project docs that reference _other internal_ docs or packages will either work on [the docs site](https://www.keystonejs.com) **OR** work locally and on GitHub. Eg..

| Doc site URL | Local path and GitHub URL |
|-----|-----|
| `/guides/authentication` | `/docs/guides/authentication.md` |
| `/keystonejs/auth-password` | `/packages/auth-password/` |
| `/api/access-control` | `/docs/api/access-control.md` |
| `../../guides/access-control` | `../../docs/guides/access-control.md` |

This is due to a divergence between the physical path of the Markdown files and the published URLs. The problem effects relative and absolute paths. File extension and some edge cases around deep linking come into it too. (The local path and GitHub URL also diverge slightly due to GitHub defaulting to `README.md` as an "index" file.)

I see this as a pretty serious problem. Obviously we want a nice docs site to keep Keystone accessible but surely having these links make sense when you're in the code or on GitHub is critical.

It's has been picked up before (#902) and patched in favour of the docs site (but maybe the full implications weren't realised at that point?). More recently, I went though ["fixing" a stack of links](https://github.com/keystonejs/keystone/commit/048c0d6579575820b376056d86bc88579a432491) before realising the implications 😔

## Solutioning

We need to pick a _canonical_ path schema for our links and I think it has to be based on the physical file paths. The docs site is generated, which gives us the chance to map physical paths to "doc site URLs". Having the file paths in the raw Markdown is the only way to solve for Keystone developers who are seeing the links either locally in their code editor or on GitHub. I also suspect that _without_ file paths in Markdown files, automated testing of Markdown links (#892) will be much more difficult.

Beyond that, there's clearly a mapping between the file paths and the doc URLs. Whether this is extracted from the existing `[meta]` headers or stored elsewhere, it should be applied to links during the publishing process.

timleslie:
>We need to pick a canonical path schema for our links and I think it has to be based on the physical file paths

Strongly agree! ☝️

molomby:
Reframing this slightly... our Markdown files are currently consumed in several way:

* As raw Markdown in a local dev environment (eg. in editor, search/grep'ing, etc.)
* Rendered on the docs site
* Rendered on GitHub
* Rendered on NPM (package `README.md`s only)
[...]

PR #2115, opened 2019-12-18
Fixing links in Markdown docs to use file paths

molomby:
Links in Markdown documents should follow [the guidelines documented in #2041](https://github.com/keystonejs/keystone/issues/2041#issuecomment-566802733). This brings a swathe of existing docs inline with these decisions. These fixes are not exhaustive. 

See also #2114, which should possibly be resolved either before this PR is merged or before the docs site is next updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ If you develop custom fields, adapters, apps or any other Keystone feature, (or 
 
 ## Code of Conduct
 
-KeystoneJS adheres to the [Contributor Covenant Code of Conduct](code-of-conduct.md).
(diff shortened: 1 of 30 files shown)


MadeByMike:
This breaks links on the website. We need to fix/re-implement a solution for that first.

molomby:
@MadeByMike yep, seems like a good idea 👍🏼 Should we slap a high priority on #2114 then? Maybe you can prod the relevant people?

Alternatively, how does the site currently get published? Are there docs on the process? (Meta-docs?! 🙃)

timleslie:
@molomby Now that #2160 is complete could you please review whether there are any useful changes left in this PR?

mitchellhamilton:
I based #2160 on this so we can close this now

PR #2119, opened 2019-12-18
Cors & pinoOptions for default server

jesstelford:
Previously it was possible to configure cors when instantiating a GraphQL App, but since [the `cors` option was moved to `keystone.prepare()](https://github.com/keystonejs/keystone/pull/1618), it was no longer possible to set it from a default server:

```javascript
module.exports = {
  keystone: /* ... */,
  apps: /* ... */,
  cors: { /* NEW set cors here */ },
  pinoOptions: { /* NEW set pinoOptions here */ },
};
```

This PR brings that functionality back and includes some docs.

### DIFF
--- a/.changeset/chilly-berries-yell.md
+++ b/.changeset/chilly-berries-yell.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': minor
+---
+
+Allow configuring cors & pinoOptions for default server by adding them to the export in index.js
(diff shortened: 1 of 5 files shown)


MadeByMike:
@jesstelford isn't this why we added `configureExpress` ?

jesstelford:
Yeah, I thought about `configureExpress`, but while the app could setup cors in there manually, keystone is still hard-coding cors again regardless of what's put in `configureExpress`.

So we'd need some option to turn off the hard-coded use of cors so the one the app sets up in `configureExpress` is the only one.

But then we're back to having a special cors option again...

PR #2123, opened 2019-12-19
Fix SyntaxError

marxvn:
SyntaxError: Cannot use import statement outside a module


jesstelford:
@all-contributors can you add @marxvn for bugs, thanks!

PR #2132, opened 2019-12-20
Fix content field view in Admin UI

sarneaud:
The view wouldn't show properly (or even break the page) because not all
its loaders were running.

This is just a quick fix to get the UI working again.


sarneaud:
It feels :/ to add more coupling between the Admin UI and the fields.  Let's look at this.

packages/app-admin-ui/client/providers/AdminMeta.js
```diff
@@ -59,8 +59,11 @@ function readAdminMeta() {
     });
 
     Object.values(listViews).forEach(list => {
-      Object.values(list).forEach(({ Controller }) => {
+      Object.values(list).forEach(({ Controller, blocks }) => {
         viewsToLoad.add(Controller);
+        // This is a special case for content fields
+        // FIXME: maybe we'd have less coupling if we just let each field asynchronously render its own view component
+        if (blocks) blocks.forEach(l => viewsToLoad.add(l));
```
jesstelford: 🤔 This code (or something very very similar) _used_ to live in the admin ui... Why is it now necessary to add here? Did something change recently?

MadeByMike:
Closed as I think this is stale

PR #2139, opened 2019-12-21
remove whitespaces

gautamsi:
result of yarn manypkg fix,
unintended whitespaces cleaned


timleslie:
Fixed in another branch 👍

Issue #2150, opened 2019-12-23
Fields with `access: { update: false }` disappear from Admin UI Details View


jesstelford:
# Bug report

## Describe the bug

When a field is marked with access as `update: false`, it wont be rendered on the details page, even if `read: true`.

## Expected behaviour

Ideally we still render the field, but in a "Read Only" mode.

## Additional context

I had a quick play and it doesn't look like the `Content` field supports any kind of "Read Only" mode just yet, so we'd need to add that ability also.

MadeByMike:
This will now rendered as readOnly \o/

PR #2151, opened 2019-12-23
Correctly load Content field views when field access set to update: false

jesstelford:
This was an existing bug which would only occur in some cases due to race conditions (code & data loading). The changes in #2135 surfaced this bug so it happens every time (this is a good thing).

There is another bug here which this PR doesn't address but which became clear after poking at the code; When a field is marked access as `update: false`, it wont be rendered on the details page, even if `read: true`. I've created issue #2150 to track this.

This PR is best viewed [with whitespace changes disabled](https://github.com/keystonejs/keystone/pull/2151/files?utf8=%E2%9C%93&diff=unified&w=1)


packages/app-admin-ui/client/pages/Item/index.js
```diff
@@ -413,10 +412,13 @@ const ItemPage = ({ list, itemId, adminPath, getListByKey }) => {
...
-      .filter(({ isPrimaryKey }) => !isPrimaryKey)
-      .filter(({ maybeAccess }) => !!maybeAccess.update)
-      .map(field => () => field.initFieldView()),
+    // NOTE: We should really filter out non-renderable fields here, but there's
+    // an edgecase where deserialising will include all fields (even
+    // non-renderable ones), so we have to ensure the field views are all loaded
+    // even if they may not be used. This is particularly salient for the
+    // Content field which needs its views loaded to correctly deserialize.
+    // Ideally we'd use getRenderableFields() here.
+    ...list.fields.map(field => () => field.initFieldView()),
```
jesstelford: The reason it's important to do this here is later on line 452 we call `deserializeItem` which internally calls `deserialize`, which relies on the fields views having been loaded.

.changeset/swift-gorillas-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Correctly load Content field views when field access set to update: false
```
MadeByMike: Is this the description for the other PR?

jesstelford: Nope, that's correct. Without this fix, the Content field will not load if you set `update: false` - it'll throw an error about deserialisation being called before it was loaded.

The _other_ issue is https://github.com/keystonejs/keystone/issues/2150 which is how fields with `update: false` on them will just disapear completely from the Admin UI

Issue #1887, opened 2019-11-01
Secure cookies over reverse proxy with https


arnaud-zg:
# Feature request

Actually, it's not possible to set secure cookies over reverse proxy with https.

## Is your feature request related to a problem? Please describe.

As mentioned on this issue https://github.com/keystonejs/keystone/issues/1842#issuecomment-548743146, app is not able to set a secure cookie when there is a proxy.

## Describe the solution you'd like

I'm not sure, I think if we enabled `proxy: true` on this file https://github.com/keystonejs/keystone/blob/master/packages/session/lib/session.js#L54, it should work.

## Describe alternatives you've considered

Nothing else at the moment ...

## Additional context

It was tested with Heroku.

Vultraz:
The [docs][(https://www.npmjs.com/package/express-session) mention `proxy` takes its value from Express's `trust proxy` setting if the former is undefined. 🤔

arnaud-zg:
It seems like we need to call `app.set('trust proxy', 1)`, I'm a little bit confused about where should I put this configuration, it seems like express instance is not accessible on my side (app).

Vultraz:
You can export a `configureExpress` function from your entry file. See [here](https://www.keystonejs.com/guides/custom-server#you-might-not-need-a-custom-server-if).

arnaud-zg:
Oh nice, I'll try it

arnaud-zg:
I just deployed my app with this configuration, I still have the issue ...

```
module.exports = {
  keystone,
  apps: [
    new GraphQLApp(),
    new AdminUIApp({ authStrategy }),
    new NextApp({ dir: NEXT_APP_PATH }),
  ],
  configureExpress: app => {
    app.set('trust proxy', 1)
  },
}
```

Vultraz:
Ok, so you're still getting the signin page loop with `secureCookies` set to `true` in the keystone ctor *and* `trust proxy` set in Express?

arnaud-zg:
huh by default `secureCookies` is set to `true`, but I didn't add it in ctor explicitly.

arnaud-zg:
I just deployed a version with `secureCookies`, still have the issue;

![image](https://user-images.githubusercontent.com/10991546/68069772-325d2480-fd65-11e9-8f52-650a607f5ed8.png)

Vultraz:
[...]

PR #2152, opened 2019-12-23
Fix configureExpress()

sarneaud:
`configureExpress()` was run on an Express app, and routes were added to
an Express router, but the two weren't connected together.

This version ditches the extra router and just uses the middleware by
itself to make the loading screen intercept.

Addresses https://github.com/keystonejs/keystone/issues/1887


Vultraz:
This might actually fix various problems, including the issues with secure cookies and with templating engines not working.

gautamsi:
I can confirm this does not fix the view engine stuff. but I managed to update `@keystonejs-contrib/view` without the need of this. thanks to `@Whosy`'s finding (slack).

here is the code 
https://github.com/keystonejs-contrib/keystonejs-contrib/blob/4f2acb35a8dd470ba5dcfdeb6758dd30b5e653d4/examples/blog-view/index.js#L63-L84


```js
const blogApp = {
  prepareMiddleware: ({ keystone }) => {
    const app = express();
    app.use(bodyParser.urlencoded({ extended: true }));
    app.set('views', './templates');
    app.set('view engine', 'pug');
    initRoutes(keystone, app);

    return app;
  }
}

module.exports = {
  keystone,
  apps: [
    new GraphQLApp(),
    new StaticApp({ path: staticPath, src: staticSrc }),
    adminApp,
    blogApp,
  ],
  distDir,
};
```

Vultraz:
Ok, but the view engine stuff still needs to be fixed.

gautamsi:
possible if we change few lines in this PR like this

https://github.com/keystonejs/keystone/pull/2152/files#diff-ac38b0d61846d7008d188f168fb1f6caR98

```js
configureExpress = () => null,
```

https://github.com/keystonejs/keystone/pull/2152/files#diff-ac38b0d61846d7008d188f168fb1f6caR103
```js
const customApp = configureExpress(express());
```
and 
https://github.com/keystonejs/keystone/pull/2152/files#diff-ac38b0d61846d7008d188f168fb1f6caR119

```js
app.use([...middlewares, ...(customApp ? [customApp] : []) ]);
```

gautamsi:
@timleslie should I create a PR for changes I suggested? Or this PR should accommodate them?

timleslie:
> @timleslie should I create a PR for changes I suggested? Or this PR should accommodate them?

A separate PR will be better, I think your proposed changes are related but different to the specific issue being addressed in this PR 👍

PR #1927, opened 2019-11-06
Admin UI Hooks/Custom Actions

MadeByMike:
This creates a framework for Admin hooks and Custom Actions. 

It doesn't set-up any hooks at the moment except converting the current custom pages mechanism over to using this while retaining compatibility with the old API.

Hooks now work by creating a folder in your project called `admin-ui` or specifying a hooks path in the `AdminUIApp` constructor:

```js
new AdminUIApp({
  hooks: require.resolve('./admin/'),
});
```
The index file in the `admin-ui` directory should export a hooks object:

```
export default {
  pages: () => [...],
  futurehookname: () => { ... },
};
```

All hooks are functions and should get useful context information. This isn't implemented yet but I'd like to add the `authContext` and page context. I'd like feedback on the variables to provide to hooks and implementation advice.

Since pages doesn't currently use this it's safe to add these variables in a future PR when we add additional hooks.

I've tested this mechanism for a number of use-cases and was able to implement crude custom actions and override global styles. However, the scope of this PR is not to add new hooks.

Oh, this PR also changes the way the blog demo custom pages work.

PR #2155, opened 2019-12-24
Consume field hooks

gautamsi:
continuing  from #1927 
* enable use of Field UI hooks from hooks config

```js
// hooks config for list based per field views
lists: {
    User: {
      views: {
        email: {
          Field: TextFieldEx,
        },
      },
    },
  },
```

Example of copying the Text Field view and using it with background Changed in Email field in blog demo 
(ideally we should have a way to add theming hook per field)

image

> Still need to figure out how to provide correct example with blog demo or add it to test projects for reference. Not sure of the testing scenario either. @MadeByMike

### DIFF
--- a/.changeset/thick-oranges-change.md
+++ b/.changeset/thick-oranges-change.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@keystonejs/fields': minor
+---
+
+Enabled use of cuatom field views. Custom Field, Cell and Filter views can now be configured per field in list.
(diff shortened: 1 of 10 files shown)


.changeset/thick-oranges-change.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@keystonejs/fields': minor
```
gautamsi: should this be `major`?

gautamsi:
How do I add a list hook for all List?

should I add a key `list` which would be for all list? or some other name inside `lists` like `__list__`

https://github.com/keystonejs/keystone/blob/f0235b7df9d5ee3118c40a056a559d59f186a518/packages/field-views-loader/index.js#L43-L68

gautamsi:
fixed merge conflicts

PR #2160, opened 2019-12-27
Fix doc links

### LINKED ISSUE #2041: Internal links in docs either work on the docs site OR GitHub (but not both)

### LINKED ISSUE #2114: Update doc site publishing process to convert link format

mitchellhamilton:
Closes #2114
Closes #2041


mitchellhamilton:
This is now working and throws if a link to another page with or without heading doesn't exist/isn't linked in the absolute form.

All of the docs are now correct except for one, `docs/guides/production.md` links to `/packages/keystone/README.md#trustproxies` but the trust proxies heading doesn't exist and I'm not sure what the correct heading is.

PR #2177, opened 2020-01-02
Remove redundant word

bartduisters:
Unless 'solution developer' is a special kind of developer, then this PR can be ignored.


timleslie:
Yeah, a `solution developer` is special kind of developer. In particular, they're a developer *using* keystone to developer a solution to their own particular problem. This is in distinction to "developers" who are the people working on keystone itself.

Issue #2164, opened 2019-12-28
npm init keystone-app my-app fails


bothwellw:
Hi

I am following the getting started documentation and I am getting this error when I run 

npm init keystone-app my-app

npx: installed 61 in 15.448s
(node:10908) UnhandledPromiseRejectionWarning: TypeError: Cannot read property '0' of undefined
    at Colorize (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\node_modules\cfonts\lib\index.js:328:69)
    at AddLetterSpacing (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\node_modules\cfonts\lib\index.js:412:17)
    at Render (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\node_modules\cfonts\lib\index.js:788:14)
    at Object.Say [as say] (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\node_modules\cfonts\lib\index.js:884:15)
    at showWelcomeMessage (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\lib\show-welcome-message.js:4:10)
    at C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\bin\cli.js:14:3
    at Object.<anonymous> (C:\Users\dev\AppData\Roaming\npm-cache\_npx\10484\node_modules\create-keystone-app\bin\cli.js:22:3)
    at Module._compile (internal/modules/cjs/loader.js:959:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:995:10)
    at Module.load (internal/modules/cjs/loader.js:815:32)
(node:10908) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:10908) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

masbossun:
experienced same issue using `yarn create`

zoltanradics:
Same for me. Neither npm or yarn install works at the moment.

dufemeng:
I also have this problem.

[...]

PR #2188, opened 2020-01-05
bump cfonts version to ^2.4.8

jeffdhooton:
Fix for [#2164 ](https://github.com/keystonejs/keystone/issues/2164).

Bumps cfonts version inside create-keystone-app to ^2.4.8


jeffdhooton:
> @jeffdhooton This will need a `changeset` to make sure the package gets released correctly. Could you please run `yarn changeset` on this branch. Thanks!

Added - I can take out the manual minor version bump if you'd like, just noticed it in the changeset and figured it should probably be bumped.

timleslie:
@jeffdhooton Yeah, please take out the version bump, that will all be handled by changesets when we do the next release (which I'm planning for later today).

PR #2214, opened 2020-01-11
Remove first sentence

bartduisters:
After reading the quick start pages, the first page is this one. It talks about KeystoneJS being 3 distinct pieces and then shows some arrow function. However, it is unclear what this is doing here. The schema describes the data model (which is the title of this page) and then it talks about AdminUI and GraphQL ... which, as far as I understand are two separate apps that can be wired in.

This piece is confusing to me and I can't place it anywhere in the context, hence the PR to delete it.


MadeByMike:
Yeah fair enough.

PR #2216, opened 2020-01-11
en-US spelling

bartduisters:
I noticed on other pages that en-US was used instead of en-UK. Made it consistent on this page.

### DIFF
--- a/docs/guides/hooks.md
+++ b/docs/guides/hooks.md
@@ -12,7 +12,7 @@ The [Hooks API docs](/docs/api/hooks.md) describe the specific arguments and usa
 
 ## Conceptual Organisation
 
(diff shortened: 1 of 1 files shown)


MadeByMike:
We're not going to stress over regional spelling differences. More of the team are from Australia so we'll probably stick with this unless it's confusing in the case of an API.

molomby:
According to the [style guide](https://github.com/keystonejs/keystone/blob/master/STYLE_GUIDE.md) we favour UK/AU spelling (but I've heard that contradicted at times).

bartduisters:
I noticed it was en-UK after a couple of PR. Will not include en-US changes in future pull requests.

PR #2220, opened 2020-01-11
Corrected a spelling mistake + question

bartduisters:
In the GraphQL Schema represented on this page, there is a reference to 'UserToOneRelationshipInput', which is not defined in the schema. Is this left out on purpose?

### DIFF
--- a/docs/guides/graphql-philosophy.md
+++ b/docs/guides/graphql-philosophy.md
@@ -7,7 +7,7 @@ order: 3
 
 # GraphQL Philosophy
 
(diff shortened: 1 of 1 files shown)


bartduisters:
@MadeByMike I notice it is closed, but I still think it is 'thinks' and not 'think'.

"This is how the team thinks about it."
"This is how the team think about it."

I'm also wondering what I commented before:
In the GraphQL Schema represented on this page, there is a reference to 'UserToOneRelationshipInput', which is not defined in the schema. Is this left out on purpose?

PR #2235, opened 2020-01-14
Update Content Field README

justintemps:
The README for Content Field should point out that you have to install the `field-content` package separately, the same way the Wysiwig Field docs do.


jesstelford:
@all-contributors please add @justintemps for docs <3

PR #2240, opened 2020-01-15
Simplify query parsing in mongoose adapter

timleslie:
No functional changes, just an internal simplification and updated inline docs.

### DIFF
--- a/packages/adapter-mongoose/lib/adapter-mongoose.js
+++ b/packages/adapter-mongoose/lib/adapter-mongoose.js
@@ -224,14 +224,19 @@ class MongooseListAdapter extends BaseListAdapter {
     // { where: { a: 'A', b: { where: { c: 'C' } } }, skip: 10 }
     //       => { a: 'A', b: { c: 'C' }, $skip: 10 }
     const graphQlQueryToMongoJoinQuery = ({ where, ...modifiers }) => ({
-      ...mapKeys(where, whereElement =>
+      ...mapKeys(where || {}, whereElement =>
         getType(whereElement) === 'Object' && whereElement.where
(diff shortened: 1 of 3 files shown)


packages/adapter-mongoose/lib/adapter-mongoose.js
```diff
@@ -216,35 +216,22 @@ class MongooseListAdapter extends BaseListAdapter {
...
-        // recurse on object (ie; relationship) types
-        return graphQlQueryToMongoJoinQuery(field);
-      });
-    }
-
-    let query;
-    try {
-      query = graphQlQueryToMongoJoinQuery(args);
-    } catch (error) {
-      return Promise.reject(error);
```
jesstelford: 🤔 Why was this catch here? Looking over the original code, I can't see a reason why it might have thrown. But now I'm anxious that we're going to have an edge case we'll no longer be handling safely.

timleslie: Yeah, I looked really closely and I couldn't see any reason for having this error handling pattern here. I can't see any good reason to try, catch, re-throw rather than just letting the original error propagate. Nothing about the parsing looked like it was worth having this there so I decided to clear it out 🤷‍♂

jesstelford: The difference here is that we converted the synchronous `Error` into an asynchronous _rejected_ `Promise`, so letting it propagate would change the behaviour of all the calling code when it errors.

timleslie: Mmm, good point, and now I'm seeing a bunch of failures which are perhaps related? Why must programming be so hard!

PR #2250, opened 2020-01-16
Added getFilename to S3Adapter

LiamAttClarke:
Allows users to override the filename before uploading. This mirrors the behaviour of the LocalFileAdapter's `getFilename` option.


packages/file-adapters/lib/s3.js
```diff
@@ -32,23 +36,30 @@ module.exports = class S3Adapter {
 
   save({ stream, filename, id, mimetype, encoding }) {
     return new Promise((resolve, reject) => {
+      const fileData = {
+        id,
+        originalFilename: filename,
+        filename: this.getFilename({ id, originalFilename: filename }),
```
timleslie: This will fail if `getFilename` has not been defined. We should have a default function in the same way that `LocalFileAdapter` does.

LiamAttClarke: I defined the default as a class method 😬 

I did a similar thing with `publicUrl`.

If this isn't kosher I can rewrite it.

timleslie: Oh, I totally missed that, sorry! This all looks good then I think 👍

PR #2258, opened 2020-01-17
Show read only field by access in admin ui

gautamsi:
This renders readonly fields in admin-ui Item Detail/Edit page, does not affect CreateItem modal as any readonly fields do not need to be shown on create as they have no content anyways (mostly).

earlier those details are hidden and you have no ways to see the field where you have read access but no write access. 

this renders them as disabled

also added `isReadOnly` option in field's admin config, this one is assumed in some places but the actual implementation was missing

this is how basic test project looks like when all the fields are set to `isReadOnly: true`

user item:
image

post item:
image
> some fields are set to read only as that is the option available in external component like codemirror and slate block editor

originally if user did not have access (instead of Read Only), all of this would be removed. 

this is useful in where you don't want anyone to update the content of some field (access or read only) but  still render the content. simple example would be when we use `atTracking/byTracking` plugin, those contents are only visible when you select them in ListTable.

example:

```js
keystone.createList('Todo', {
  fields: {
    name: { type: Text, isRequired: true },
    someReadOnlyField: {
      type: Text,
      adminConfig: {
        isReadOnly: true,
      },
      defaultValue: 'Some default value',
    },
  },
});

Tests added for read only fields being disabled.

### CODE REVIEW

**Comment by LiamAttClarke** [comment]
This branch needs more ❤️.

**Comment by gautamsi** [comment]
unless there is any comment from maintainers I can not make any changes. already tried fixing things earlier. I am using this branch in my projects already

**Comment by gautamsi** [comment]
@LiamAttClarke I have fixed the conflict once more and tried to resolve failing test one last time.

**Comment by lucatk** [comment]
@gautamsi As you said you're using your own branch with this change in your projects already, could you give me a quick info on how you achieved that? When I try to use any local version of app-admin-ui (adding them locally), I always receive Module parse failed errors.
(see my issue #2836)

**Comment by gautamsi** [comment]
they are small business companies and did not need a high performance production build so I deployed the whole project. they needed it for internal management of some processes. 

I had [another PR long ago](https://github.com/keystonejs/keystone/pull/1832) which gives you a `projects` folder in keystone repo which is ignored in git. 

I create a repo for my project and deploy the whole keystone repo including my projects folder there.

File: packages/fields/src/Controller.js [hunk 1]
```diff
@@ -8,6 +8,7 @@ export default class FieldController {
     this.type = config.type;
     this.maybeAccess = config.access;
     this.isPrimaryKey = config.isPrimaryKey;
+    this.isReadOnly = config.isReadOnly;
```
MadeByMike, line 11: Move this onto adminConfig: {
  isReadOnly: true
}

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -222,6 +222,7 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate }) => {
                           warnings={validationWarnings[field.path] || []}
                           onChange={onChange}
                           renderContext="dialog"
+                          isReadOnly={field.isReadOnly}
```
gautamsi, line 225: @MadeByMike this disables the field on Create dialog. this will break anyone who has it required and also sets them as read only.

packages/fields/src/Controller.js
gautamsi: this is now works as adminConfig only.

packages/app-admin-ui/client/components/CreateItemModal.js
MadeByMike: Yeah... that makes sense. People can just not do that... or initialise with a hook.

packages/fields/README.md
```diff
@@ -81,6 +81,28 @@ A description of the field used in the AdminUI.
 
 A description of the field used used in the GraphQL schema.
 
+### `adminConfig`
+
+Additional field configs affecting field rendering or display in `admin-ui`.
+
+#### `adminConfig.isReadOnly`
+
+This admin config makes the field read only in `admin-ui`, Field with `isReadOnly` set to `true` will be rendered as disabled and you can not modify them in Item Details page or in Create item modal.
```
MadeByMike, line 90:
```suggestion
Fields with `isReadOnly` set to `true` will be disabled preventing users from modifying the min the Admin UI. This does not affect access control and fields can still be updated via GraphQL.
```

packages/fields/src/types/Checkbox/views/Field.js
```diff
@@ -2,7 +2,13 @@
...
 import { jsx } from '@emotion/core';
 
-import { FieldContainer, FieldLabel, FieldDescription, FieldInput } from '@arch-ui/fields';
+import {
+  FieldContainer,
+  FieldLabel,
+  FieldDescription,
+  FieldInput,
+  isReadOnly,
+} from '@arch-ui/fields';
```
timleslie, line 11: This doesn't seem right?

packages/fields/README.md
```diff
@@ -81,6 +81,28 @@ A description of the field used in the AdminUI.
 
 A description of the field used used in the GraphQL schema.
 
+### `adminConfig`
+
+Additional field configs affecting field rendering or display in `admin-ui`.
+
+#### `adminConfig.isReadOnly`
+
+Fields with `isReadOnly` set to `true` will be disabled preventing users from modifying the min the Admin UI. This does not affect access control and fields can still be updated via GraphQL.
```
timleslie:
```suggestion
Fields with `isReadOnly` set to `true` will have editing disabled in the Admin UI, preventing users from modifying their values.
This does not affect access control and fields can still be updated via GraphQL.
```

packages/app-admin-ui/client/components/CreateItemModal.js
timleslie: I'm wondering if it would make more sense to completely remove any field that has `isReadOnly` set? Is there any reason to show an empty, disabled form element? In the update case I can see it being useful in order to see the value, but I'm not sure I see any value in the `create` case?

packages/fields/README.md
jesstelford, line 90:
```suggestion
Fields with `isReadOnly` set to `true` will be disabled preventing users from modifying them in the Admin UI. This does not affect access control and fields can still be updated via GraphQL.
```

.changeset/blue-ways-work.md
```diff
@@ -0,0 +1,30 @@
...
+'@keystonejs/fields-markdown': minor
+'@keystonejs/fields-wysiwyg-tinymce': minor
+'@keystonejs/fields': minor
+'@keystonejs/cypress-project-access-control': patch
+---
+
+* Added `isReadOnly` option on field's `adminConfig`. This allows to show them as disabled in `admin-ui` item detail page.
+* This PR also enables you to see the fields (disabled) in item detail page when you have no `update` access to field.
+
+example:
```
timleslie, line 13:
```suggestion
Example:
```

packages/fields/src/types/Checkbox/views/Field.js
MadeByMike: Nope

packages/fields/src/types/Checkbox/views/Field.js
```diff
@@ -2,7 +2,13 @@
 
 import { jsx } from '@emotion/core';
 
-import { FieldContainer, FieldLabel, FieldDescription, FieldInput } from '@arch-ui/fields';
+import {
+  FieldContainer,
+  FieldLabel,
+  FieldDescription,
+  FieldInput,
+  isReadOnly,
```
MadeByMike, line 10:
```suggestion
```

.changeset/blue-ways-work.md
```diff
@@ -0,0 +1,30 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@keystonejs/field-content': minor
+'@keystonejs/fields-markdown': minor
+'@keystonejs/fields-wysiwyg-tinymce': minor
+'@keystonejs/fields': minor
+'@keystonejs/cypress-project-access-control': patch
+---
+
+* Added `isReadOnly` option on field's `adminConfig`. This allows to show them as disabled in `admin-ui` item detail page.
```
timleslie, line 10:
```suggestion
* Added `isReadOnly` option on field's `adminConfig`. Fields with this option set will be excluded from the `create` form, and set as disabled in the `update` form in the Admin UI.
```

.changeset/blue-ways-work.md
```diff
@@ -0,0 +1,30 @@
...
+* This PR also enables you to see the fields (disabled) in item detail page when you have no `update` access to field.
```
timleslie, line 11:
```suggestion
* Updated the item detail page to include fields with access `{ update: false }` in a disabled state, rather than excluded the form.
```

MadeByMike:
@timleslie I think all feedback is now addressed

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -192,6 +192,8 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate }) => {
             {() => {
               const creatable = list.fields
                 .filter(({ isPrimaryKey }) => !isPrimaryKey)
+                // Remove read-only fields from the create dialog
```
MadeByMike, line 195:
```suggestion
```

gautamsi:
finally, thanks all. 
(that was my half century too)

timleslie:
Huzzah, congrats! 🎉

PR #2264, opened 2020-01-20
admin-ui: Fix bug where the search input on the List page would gain focus when the user pressed SHIFT

jordanoverbye:


packages/app-admin-ui/client/pages/List/Search.js
```diff
@@ -32,14 +31,11 @@ export default function Search({ isLoading, list }) {
   };
 
   const handleClear = () => {
+    if (inputRef.current) inputRef.current.focus();
     setValue('');
     onClear();
   };
 
-  useEffect(() => {
-    if (inputRef.current) inputRef.current.focus();
```
jesstelford: This would previously focus the field on first render too.

While I don't mind it not being focused on first render, is it a significant enough change to warrant perhaps a `major` bump? (I'm thinking about the expert users who navigate with keyboard here, and expect the search box to be focused on load)

jordanoverbye: Hey @jesstelford,
This `useEffect` was only introduced a couple of weeks ago in #2198, which was a minor release, but it has a couple of bugs so I've reverted to some of the original functionality

PR #2279, opened 2020-01-23
Convert parts of the Admin UI createItemModal to hooks

MadeByMike:
Sorry, I wanted this to be a smaller PR but much of the AdminUI is interconnected. If you've got suggestions for splitting this into pieces I'd love to hear it but I'm nervous about it not working without everything. Hopefully this will pave the way for smaller PRs in future.

Notable changes include how the CreateItemModal works internally and adding a pre-construct step so that AdminUI components can be consumed by the relationship field.


gautamsi:
This looks great. Indeed a big PR, but I am not sure also how to split them.

one way to split this is to convert some of the code from React Class to functional component first.

probably another PR can add other models which will be used by UI Hooks, Delete, DeleteMany, UpdateMany (as in *ItemModel)

demo-projects/todo/index.js
```diff
@@ -12,6 +12,13 @@ const keystone = new Keystone({
 
 keystone.createList('Todo', {
   schemaDoc: 'A list of things which need to be done',
+  fields: {
+    name: { type: Text, schemaDoc: 'This is the thing you need to do', isRequired: true },
+    dsafg: { type: Relationship, ref: 'Related' },
+  },
+});
+
+keystone.createList('Related', {
```
mitchellhamilton: Do we want to keep this?

packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -34,22 +35,10 @@ const HeaderInset = props => (
...
-};
-
-export function ListLayout(props: LayoutProps) {
-  const { adminMeta, items, itemCount, queryErrors, list, routeProps, query } = props;
-  const [showCreateModal, toggleCreateModal] = useState(false);
+export function ListLayout(props) {
+  const { adminMeta, items, itemCount, queryErrors, routeProps, query } = props;
   const measureElementRef = useRef();
-
+  const { list, openCreateItemModal } = useList();
```
mitchellhamilton: I'm not a fan of the `openCreateItemModal` thing because I'm not sure how to answer the question of "What modal will open when I call this function?" If I render two modals, will they both open? Will it open modals that are in totally different parts of the react tree

demo-projects/todo/index.js
MadeByMike: Nope!

MadeByMike: Reverted change to test demo projectrs

packages/app-admin-ui/client/pages/List/index.js
MadeByMike: It only open the modal inside the list provider. I think the pattern can be improved but I don't love unrelated wrapper components controlling the state of the modal by passing props around. This really creates a dis-connect with the modal and it's state. 

I think perhaps the right pattern is to either import the modal from the list provider so it's clear there is one instance of a modal per list provider. 

An alternative is to create a modal provider.

gautamsi: should there be a `useItemModal({...listConfig})` which let us use modals when needed, we would also need some models like `DeleteItemModal` `DeleteManyItemsModel`, `updateManyItemsModel` and perhaps a new `updateItemModel` (when we really have a list item detail page rather than a pure editing experience)

MadeByMike: Yep, I want to move towards something like that but I consider this a step on that path.

Issue #731, opened 2019-02-19
Add a description fields in the AdminUI


jesstelford:
Updated based on comments below...

`schemaDoc` is for developer docs and `adminDoc` is for AdminUI.

The action here is to allow a `adminDoc` key on lists:

```
fields: {
  name: {
    type: Text,
    schemaDoc: "This will show up in GraphiQL",
    adminDoc: "This will show up in the Admin UI",
  }
}
```

MadeByMike:
Can we do this in the AdminUI in a standard way without requiring every field implementation to be updated?

gautamsi:
I think admin ui should not be making description on its own.

Basic implementation should be quick one. 
Should this be a "good first issue"

MadeByMike:
@gautamsi this just means that the admin UI would look for a description and display help text in a consistent way rather than having this be a field concern. The adminUI already makes thousands of decisions like this in how it displays any information about a list or field... this is far less opinionated than most of them.

timleslie:
Fixed in #2287

PR #2287, opened 2020-01-24
Added adminDoc option to all fields

LiamAttClarke:
Related issue: https://github.com/keystonejs/keystone/issues/731

This depends on changes made here: https://github.com/keystonejs/keystone/pull/2286

Screen Shot 2020-01-24 at 4 20 25 PM


MadeByMike:
@LiamAttClarke this is great but this might suddenly show developer documentation in the adminUI for existing applications.

Can I suggest that we make a new field... called "description" and use this? If the description is set but there is not `schemaDoc`, I also suggest that we set the value of `schemaDoc` to the value of `description`. 

I think it make more sense to have a UI description appear in GraphQL when there is no other documentation than the other way around. I also think it is very possible that not all fields need descriptions but may need developer docs.

This should give the most flexibility and not change anything in existing apps.

Issue #1654, opened 2019-09-17
Knex adapter doesn't implement the `search` condition


molomby:
There's a (somewhat half-baked) implementation of the `search` condition [in the Mongoose adapter](https://github.com/keystonejs/keystone-5/blob/master/packages/adapter-mongoose/lib/adapter-mongoose.js#L29
). 

This filter is used by the AdminUI when selecting a related item for a Relationship field in the create and edit forms. The GraphQL queries look like this:

```graphql
query RelationshipSelect($search: String!, $skip: Int!) {
   allUsers(first: 10, search: $search, skip: $skip) { id _label_ __typename }
   _allUsersMeta(search: $search) { count __typename }
}
# Eg. variables: {search: "molo", skip: 0}
```

**The Knex adapter doesn't implement this filter**. This partially breaks the Relationship field UI.

It actually looks like the UI itself re-applies the filter to the results it receives from the backend it doesn't break quite as hard as it could/should. The downside of this is that, this filtering is applied to the `_label_` field where as, on the backend, the filtering is applied to the `name` field. Usually this is the same value but they can diverge. This will become more of a problem when we [fix the fields the `search` filter is applied to](https://github.com/keystonejs/keystone-5/issues/640).

The short term fix is probably just to mirror the functionality the Mongoose adapter implements across to the Knex adapter. In the long term, list need a much more configurable filtering solution (as described in #319). We should also remove that client-side filtering.

jesstelford:
I think we should remove searching across the board and instead open the Relationship filter to be a proper nested set of filters instead of a react-select input field thing.

molomby:
> I think we should remove searching across the board and instead open the Relationship filter to be a proper nested set of filters instead of a react-select input field thing.

Why? Surely just typing someones name and getting a filtered set of result to pick from provides a faster/easier/more intuitive user experience in some cases. Not to mention that, free form text search _was_ a list-level filter in KS4 -- something we've always planned to restore.

There's no reason this can't be made performant on both Mongo and Relational DBs, we just need to support some full-text indexing at a list level.

I agree that there are other interfaces that should be available to manage relationship field (including a sort of "full, embedded list view with filters" as you suggest) but which interface is appropriate depends on the properties of the relationship. [The way I think about it](https://gist.github.com/molomby/0b3fa2a22747ca053199e75945334da3#admin-ui), the interface used should be a function of...

1. How many items you need to select (ie. one, a few items or _heaps_ of items)
[...]

Issue #640, opened 2019-01-23
Relationship Select only queries a list's `name` field


nathsimpson:
Relationship Select doesn't seem to be surfacing any options when filtering.

This appears in my own Keystone project (will show you the code if you need), and also the Blog demo project.

![jan-24-2019 08-59-00](https://user-images.githubusercontent.com/12689383/51640025-8b1db680-1fb6-11e9-94d4-4161d0bba916.gif)

jesstelford:
The way searching within a Relationship dropdown _should_ work (IMO) is to filter based on the `_label_` field. But that's not a query we can actually execute because it's not generated until _after_ data is returned.

The next best thing is to search across all fields that have text-ish values and combine them in an `OR` query:

```
query relationshipSearch($query: String) {
  Users(where: { OR: [
    { name_contains_i: $query },
    { bio_contains_i: $query },
    { username_contains_i: $query },
  ] }) {
    id
    _label_
  }
}
```

Issue #343, opened 2018-09-20
Implement configurable search fields


timleslie:
As a stepping stone to the full search implementation (#320) we should at least add the ability to specify which fields to search on.

https://github.com/keystonejs/keystone-test-project/search?q=searchfields&unscoped_q=searchfields

PR #2292, opened 2020-01-28
Fix Knex search

### LINKED ISSUE #1939: Search is not working
# Bug report

## Describe the bug
Keystone list level 'Search' is not working even after adding filters

## Expected behaviour
With filters and without filters List level Search should work checking the entire items in the list.

## System information

- OS: [ macOS, Windows]
- Browser [chrome, safari]

MadeByMike:
This resolves #1939 and #1654 by making the search work the same as the Mongoose adapter.

It's not a good search and more work is needed. See: #343 and #640


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -507,6 +511,13 @@ class QueryBuilder {
 
     this._addWheres(w => this._query.andWhere(w), listAdapter, where, baseTableAlias);
 
+    // TODO: Implement configurable search fields for lists
+    const searchField = listAdapter.fieldAdaptersByPath['name'];
+    if (search !== undefined && searchField && searchField.fieldName === 'Text') {
+      this._query.whereRaw('true');
```
timleslie: We don't need the `whereRaw('true')` here, as we're now in the phase of the query builder where we're chaining together `andWhere` calls.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -507,6 +511,13 @@ class QueryBuilder {
...
+      this._query.andWhere(`${baseTableAlias}.name`, `LIKE`, `%${search}%`);
```
timleslie: We need to apply `escapeRegExp` to `search` to ensure special characters in the search don't get interpreted as regex controls.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -507,6 +511,13 @@ class QueryBuilder {
 
     this._addWheres(w => this._query.andWhere(w), listAdapter, where, baseTableAlias);
 
+    // TODO: Implement configurable search fields for lists
+    const searchField = listAdapter.fieldAdaptersByPath['name'];
+    if (search !== undefined && searchField && searchField.fieldName === 'Text') {
```
timleslie: I'm not sure that `searchField.fieldName === 'Text'` will actually work? TBH I'm not sure, but we should really be checking `searchField.field instanceof Text` to make this more explicit.

MadeByMike: I tested it with different field types but happy to go with instanceof

PR #2313, opened 2020-01-31
fix password message when no secret is set

### LINKED ISSUE #2301: Password auth strategy should check a hash exists before comparing
## Description

Right now, if you're using password auth but a user account has no hash set, attempting to auth as that user will result in a `data and hash arguments required` error, thrown by the `bcrypt` package itself. This comes up a fair bit if you're using multiple auth strategies, eg. password _and_ social auth. If someone signs up with social then, later, attempt to sign in with a password, you'll get this error.

The downsides of this are:

* The error isn't very helpful. Unless you understand how passwords are stored and/or are familiar with the argument names `bcrypt.compare()` uses, the message doesn't mean much.
* It leaks information about the existence and state of user records and, in doing so, violates the promise of the [`protectIdentities` config flag](https://github.com/keystonejs/keystone/blob/master/packages/auth-password/README.md#protectidentities).

## To Reproduce

1. Fire up a Keystone app that uses the password auth strategy for a list
2. Null out the password hash for a user account
3. Call the authentication mutation, specifying the identifier for that user account (and any password)

## Expected behaviour

If `protectIdentities` is false, the strategy should return a helpful key and message -- something like `[passwordAuth:secret:notSet] The item identified has no secret set so can not be authenticated`.

If `protectIdentities` is true, the standard `[passwordAuth:failure] Authentication failed` key/message should be returned, with an ignored call to `generateHash()` or `compare()` to obfuscate the existence and state of the user.

## Additional context

Not withstanding #2300 and #2053.

gautamsi:
closes #2301


packages/auth-password/index.js
```diff
@@ -74,7 +74,8 @@ class PasswordAuthStrategy {
       // TODO: This should call `secretFieldInstance.compare()` to ensure it's
       // always consistent.
       // This may still leak if the workfactor for the password field has changed
-      await secretFieldInstance.generateHash('password1234');
+      const hash = await secretFieldInstance.generateHash('password1234');
```
gautamsi: @MadeByMike also tried to patch on TODO comment in line 74, if this is good, can remove the comment.

packages/auth-password/index.js
```diff
@@ -95,8 +96,20 @@ class PasswordAuthStrategy {
 
   async _matchItem(item, args, secretFieldInstance) {
     const { secretField } = this.config;
+    let hash = item[secretField];
     const secret = args[secretField];
-    return await secretFieldInstance.compare(secret, item[secretField]);
+    let message = `[passwordAuth:secret:mismatch] The ${secretField} provided is incorrect`;
+    if (!hash) {
```
MadeByMike: It looks like the logic has changed here? Previously we were accepting that `item[secretField]` contained the hash? Now if it does't exist we're generating a new one?

packages/auth-password/index.js
```diff
@@ -95,8 +96,20 @@ class PasswordAuthStrategy {
...
+      hash = await secretFieldInstance.generateHash('password1234');
+      await secretFieldInstance.compare(secret, hash);
+      return {
+        success: false,
```
MadeByMike: If we are returning false here why are we bothering checking the `secretFieldInstance` above?

gautamsi: this was due to comment earlier that we should have `work factor` factored in for this. comparing password would keep the system busy for that time and guess work for time will not be easy.

We are returning only the password incorrect error when `protectIdentities = true` this means we have to use the cpu cycle before returning false.

packages/auth-password/index.js
gautamsi: `work factor` factored in by generating new hash, see line 77 for previous instance of such computation to prevent guess work for time based attack.

packages/auth-password/index.js
```diff
@@ -95,8 +96,20 @@ class PasswordAuthStrategy {
 
   async _matchItem(item, args, secretFieldInstance) {
     const { secretField } = this.config;
+    let hash = item[secretField];
     const secret = args[secretField];
-    return await secretFieldInstance.compare(secret, item[secretField]);
+    let message = `[passwordAuth:secret:mismatch] The ${secretField} provided is incorrect`;
```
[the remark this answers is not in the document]
gautamsi: user still sees this in the admin-ui, this is only for graphql which gives more information for devs to fix any bug.

if you still want, I can remove this.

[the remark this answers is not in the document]
MadeByMike: If it's only in GraphQL I don't mind

packages/auth-password/index.js
MadeByMike: I guess my understanding of work factor is something different. I thought this was just a value given to the hashing algorithm. I still don't understand why we are doing unnecessary checks for the sake of increasing CPU cycles. Is the intent to slow down automated brute force attacks?

gautamsi: yes, some existing comments from file.

```js
   // If we failed to match an identity and we're protecting existing identities then combat timing
    // attacks by creating an arbitrary hash (should take about as long has comparing an existing one)
    if (results.length !== 1 && this.config.protectIdentities) {
      // TODO: This should call `secretFieldInstance.compare()` to ensure it's
      // always consistent.
      // This may still leak if the workfactor for the password field has changed
```

we may have to revisit in future to add unnecessary delay in the response for every wrong password attempt from same client, or even option to integrate re-captcha to password authentication

packages/auth-password/index.js
```diff
@@ -52,12 +52,12 @@ class PasswordAuthStrategy {
     // Verify the secret matches
     const match = await this._matchItem(item, args, secretFieldInstance);
 
-    if (!match) {
+    if (!match.success) {
```
timleslie:
```suggestion
    const { success, message } = await this._matchItem(item, args, secretFieldInstance);

    if (!success) {
      return {
        success,
        message: this.config.protectIdentities
          ? '[passwordAuth:failure] Authentication failed'
          : message,
      };
    }
    return { success, list, item, message };

File: packages/auth-password/index.js [hunk 6]
```diff
@@ -95,8 +96,20 @@ class PasswordAuthStrategy {
 
   async _matchItem(item, args, secretFieldInstance) {
     const { secretField } = this.config;
+    let hash = item[secretField];
```
timleslie:
```suggestion
    const { secretField } = this.config;
    const secret = args[secretField];
    if (item[secretField]) {
      const success = await secretFieldInstance.compare(secret, item[secretField]);
      return {
        success,
        message: success
          ? 'Authentication successful'
          : `[passwordAuth:secret:mismatch] The ${secretField} provided is incorrect`,
      };
    } else {
      hash = await secretFieldInstance.generateHash('password1234');
      await secretFieldInstance.compare(secret, hash);
      return {
        success: false,
        message:
          '[passwordAuth:secret:notSet] The item identified has no secret set so can not be authenticated',
      };
    }
```

gautamsi: I have committed suggestions but had to fix them as well.

molomby:
@timleslie hey bud, this takes twice as long as it should and breaks our timing attack protection :sad-face:

timleslie:
@molomby That is sad-face, should definitely look into that.

PR #2314, opened 2020-01-31
Make more configs accessible to admin-ui from Keystone/List/Field

gautamsi:
* `field` object accessed through `adminMeta` in `admin-ui` now has all field configs including custom one
*  `list` object accessed through `adminMeta` in `admin-ui` now has all `adminConfig` values including custom one
* you can add `publicConfig` option in `Keystone` constructor which can be accessed using `adminMeta.publicConfig`.

all of this can be used outside for `admin-ui` in custom apps (NextJs/React) from result of `useAdminMeta()` hook.


gautamsi:
fixed conflict

gautamsi:
@Vultraz has a point that we should not expose everything like `file-adapter` `route` and `src` config. (I saw another PR which specifically changed to named parameter in past).
To overcome this I propose
* add `adminConfig` option to field (which will only be used for people extending admin UI or future improvements)
* List is already using `adminConfig`, so no changes (just pass extra config)
* add `adminMeta` on `AdminUIApp` constructor as you suggested.
* remove `publicConfig` from keystone.

PR #2322, opened 2020-02-02
Added way to use keystone from Next App

dijs:


MadeByMike:
Why not export Keystone from your index file and import it into your NextJS app? This feels better than passing it via the request object.

dijs:
@MadeByMike I tried that first. The `index` is ES5 and Next is ES6. I did not play well together. Just like anytime you try to import things outside of a webpack'ed `src` directory.

wuichen:
@dijs  i think its a great way to expose keystone thru req. what about adminmeta object and field view loaders? do you have a better idea on injecting those into next.config.js?

MadeByMike:
What do you mean by not playing nice together? You can import es5 into an es6 project, it's only the other way around that is problematic. Perhaps Next or Webpack is the issue? Either way I'm not sure passing Keystone through a request object is a pattern we want endorse.

dijs:
@MadeByMike You're right, I am not sure exactly what I was trying to accomplish before, but now when I try to import and use `keystone` from a next handler, it works fine.

PR #2324, opened 2020-02-02
Updated Meetup demo to use Apollo hooks

Vultraz:
Still need to refactor out the use of the `withApollo` HOC but I'll do that separately.


Vultraz:
@timleslie rebased

yarn.lock
```diff
@@ -92,7 +92,7 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
-"@apollo/react-ssr@^3.1.3":
+"@apollo/react-ssr@3.1.3":
```
timleslie: Is this the correct yarn.lock file? I don't understand why this line would have changed?

Vultraz: My bad! Fixed.

Vultraz:
@timleslie fixed another merge conflict

PR #2332, opened 2020-02-04
Added headers options for ApolloClient instance for server-side requests

### LINKED ISSUE #1541: meetup demo broken for profile page
Profile page has issues with GraphQL query when called from `getInitialProps`

It calls `authenticatedUser` query from server side where the response is always null, seems session is not preserved. It redirects to `signin` page. which then again redirects to `home` page as the app is already authenticated..

singhArmani:
For the profile page server side request, the **authenticated** data was coming back as **null**.
This was happening as we were not configuring the ApolloClient instance with headers option.
Adding headers options would forward the cookie into the server side request.

closes #1541


MadeByMike:
Looks great! Do you know how to add a changeset?

MadeByMike:
It also looks like ci failed because of lining. If you run prettier on save you should be able to always avoid these.

singhArmani:
> Looks great! Do you know how to add a changeset?

@MadeByMike: Do I need to add changeset for this PR?

PR #2344, opened 2020-02-06
Create SECURITY.md

VinayaSathyanarayana:
Created Initial SECURITY.md file

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
+At this point only the latest version is supported 
+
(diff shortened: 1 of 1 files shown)


MadeByMike:
I support the idea, but I think if we are going to create one, we need to add a little more than this. 

I'm probably not likely to look into this for a few weeks but if someone wants to look at some examples and put forward a draft I'm happy to review it.

timleslie:
I don't think we're have any specific security reporting/info beyond what's already available through the existing mechanisms. I think we should hold off adding this file until such time as we have significant security-specific information to convey.

PR #2345, opened 2020-02-06
Don't call postRead hooks on the result of the internal delete function

molomby:
Ok, this is a fun one.

When the abstract `BaseListAdapter` class invokes the `_delete()` operation on its child class (the concrete database adapter), it doesn't need to then run the `postRead` hooks (contributed by the field types used).

Delete's aren't reads. The delete operation(s) _do_ read the item(s) they're about to delete but if you look at the implementation in the `List` class, you'll see this happens before the delete itself kicks off -- the item read is needed for things like the `beforeDelete` hooks, etc. so it's done first thing. The `postRead` hooks on the field types are correctly applied at this time.

When we come to actually remove the item from the DB, there's no reason to re-read the item, so no reason to call the `postRead` hooks.

---

Right now, the Mongoose adapters internal `_delete()` function _does_ return the item...

```js
return this.model.findByIdAndRemove(id);
```

While the Knex adapter function for the same _does not_...

```js
return this._query().table(this.tableName).where({ id }).del();
```

Because the `BaseListAdapter` invokes the `postRead` hooks on the result (which in Knex's case will be the number of records affected), **items from lists containing field types that use `postRead` hooks can't be deleted on Knex**. 

The practical example of this is the core `DateTime` field type, which uses a `postRead` hook to re-pack its UTC value and offset into a single value. This surfaces as an error like:

```
Cannot use 'in' operator to search for 'updatedAt_utc' in 1
```

### DIFF
--- a/packages/keystone/lib/adapters/index.js
+++ b/packages/keystone/lib/adapters/index.js
@@ -113,7 +113,7 @@ class BaseListAdapter {
   }
 
   async delete(id) {
-    return this.onPostRead(this._delete(id));
+    return this._delete(id);
   }
 
   async update(id, data) {
--- a/.changeset/lucky-apples-warn.md
+++ b/.changeset/lucky-apples-warn.md
@@ -2,4 +2,4 @@
 '@keystonejs/keystone': patch
 ---
 
-Removing unnecessary calls to field type postRead hooks on delete operations
+Removing unnecessary calls to field type postRead hooks on delete operations. The internal _delete() functions provide by the DB adapter now return a count of the records removed.
--- a/.changeset/lucky-apples-warn.md
+++ b/.changeset/lucky-apples-warn.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Removing unnecessary calls to field type postRead hooks on delete operations
--- a/packages/adapter-mongoose/lib/adapter-mongoose.js
+++ b/packages/adapter-mongoose/lib/adapter-mongoose.js
@@ -181,7 +181,7 @@ class MongooseListAdapter extends BaseListAdapter {
   }
 
   _delete(id) {
-    return this.model.findByIdAndRemove(id);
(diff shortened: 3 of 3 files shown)


packages/keystone/lib/adapters/index.js
```diff
@@ -113,7 +113,7 @@ class BaseListAdapter {
   }
 
   async delete(id) {
-    return this.onPostRead(this._delete(id));
+    return this._delete(id);
```
jesstelford: Given the differences between adapters, this is a good opportunity to normalise the possible return value to be just the `id`:

```suggestion
   await this._delete(id);
   return id;
```

molomby:
Incidentally, for the sake of constancy, we should probably either update the internal Mongoose `_delete()` to return the deleted count (like Knex)...

```js
return this.model.deleteOne({ id }).then(result => result.deletedCount);
```

_Or_ update the Knex version to return the raw values (like Mongoose)...

```js
return this._query().table(this.tableName).where({ id }).del().returning('*');
```

At least then they'll act more or less the same.

packages/keystone/lib/adapters/index.js
molomby: Ah, was just typing something similar vv.

But yes, or this ^^

molomby: Or not return anything..

molomby:
@jesstelford, I've aligned the adapters so both `_delete` functions return the count of the items deleted. Pretty standard pattern for delete-like operations. This is in favour of returning the `id` as you suggested because, being a standard field type, the `id` field too can potentially have `postRead` hooks (eg. the `uuid` field type adds a post read hook). In this scenario we'd be back to running post read hooks _twice_ for some fields during a _delete_ operations (which just seems wrong).

I don't love the idea of any hooks (other than the explicit `delete` hooks) being called after an Item is deleted. I can just imagine it opening up some really unexpected behaviour/errors. Hook authors should be able to assume the item they're executing on, you know, exists. _Especially_ for `read` hooks.

PR #2356, opened 2020-02-11
Implemented default OEmbed type 'link' for invalid/inaccessible urls

SeanDoyleGit:
Setting the default type for oEmbeds to be `link` instead of `undefined`.

OEmbeds with invalid or inaccessible urls such as `https://hi` will be saved with a type of `undefined`. graphql will throw the following error when users try to load content containing an oEmbed with a type of `undefined`. 

```
Abstract type OEmbed must resolve to an Object type at runtime for field _Block_Post_oEmbed.embed with value { originalUrl: "https://hi", author: null, provider: null, thumbnail: null, __typename: undefined }, received "undefined". Either the OEmbed type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function.
```

I am setting the default type to be link as it only implements the base fields for oembeds. It will also allow users to follow the link to the resource instead of the oEmbed not rendering at all. 

Errors returned from the oEmbed adapter will also be saved into the embed object.


packages/fields/src/types/OEmbed/Implementation.js
```diff
@@ -215,7 +216,7 @@ export class OEmbed extends Implementation {
     // Convert them into a more GraphQL friendly format
     return {
       originalUrl: inputUrl,
-      type,
+      type: type || 'link',
```
molomby:
```suggestion
      type: type || 'url',
```

Slightly more accurate I think? We're not really storing/returning a "link" -- links are things rendered in HTML documents. We're storing the URL that a link might use.

jesstelford: `link` is a specific type name used by the OEmbed spec, and matches a specific GraphQL type defined above. `url` in this case may be more accurate, but it's of type "OEmbed Link"

PR #2373, opened 2020-02-15
Replaced `react-focus-marshal` dependency with `focus-trap-react`.

Vultraz:
Both are wrappers around `focus-trap`, but the latter has been updated more recently and no longer uses deprecated lifecycle methods.

Fixes this:
image


Vultraz:
Well, crap. This breaks the Create popup :(

Vultraz:
Ok... but *now* why are they failing...

Vultraz:
@timleslie done

PR #2397, opened 2020-02-19
Converted most field React components to functional ones

Vultraz:
Still need to do the CloudinaryImage and File fields, but those are more complicated.

### DIFF
--- a/packages/fields/src/types/Select/views/Filter.js
+++ b/packages/fields/src/types/Select/views/Filter.js
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 
 import { jsx } from '@emotion/core';
-import { Component, Fragment } from 'react';
+import { Fragment } from 'react';
 import Select from '@arch-ui/select';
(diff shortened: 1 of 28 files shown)


packages/fields/src/types/Select/views/Filter.js
```diff
@@ -18,54 +18,51 @@ const EventCatcher = props => (
 );
 const SelectWrapper = props => <div css={{ marginTop: gridSize * 2 }} {...props} />;
 
-export default class SelectFilterView extends Component {
-  state = { inverted: this.props.value.inverted };
```
Vultraz: This state was unused.

packages/fields/src/types/Virtual/views/Field.js
```diff
@@ -34,24 +33,19 @@ export const FieldLabel = props => {
   );
 };
 
-export default class VirtualField extends Component {
-  onChange = event => {
```
Vultraz: This callback was unused.

Vultraz:
Ok, what to do about these tests...

PR #2402, opened 2020-02-20
Fixed invalid argument dereference in Virtual Field

LiamAttClarke:
`value` was being used before it was defined resulting in a broken admin-ui.


LiamAttClarke:
> Nice catch!

That was fast :0

cowjen01:
@timleslie it would be possible to release this fix ASAP please? It really makes all models contain virtual field unmanageable :(

MaciejPara:
@timleslie Waiting for the realease

PR #2403, opened 2020-02-21
Admin UI: fixed some minor issues when deleting items

Vultraz:
Resolved two history issues when deleting items:
- Fixed being able to go back to a now-invalid item entry
- Fixed being able to remain on an item page even after it was deleted if it had unsaved data at the time.


Vultraz:
@timleslie rebased, this one's ready

Issue #2159, opened 2019-12-26
Trouble running starter


justinmoon:
I attempted to follow [these instructions](https://www.keystonejs.com/quick-start/) to try out Keystone:

```
$ npm init keystone-app keystone-app
(choose postgres and nuxt app)
$ cd keystone-app
$ yarn start
...
**If this is the first time you've run Keystone, you can create your database with the following command:
createdb keystone_app
✖ Connecting to database
Error: database "keystone_app" does not exist
...
```

This was my first problem. It was easy enough to run `createdb keystone_app` but this wasn't in the [instructions](https://www.keystonejs.com/quick-start/). You should probably mention it in the instructions.

I tried again and I ran into Vue errors

```
$ yarn dev
yarn run v1.19.1
$ cross-env NODE_ENV=development DISABLE_LOGGING=true keystone dev
ℹ Command: keystone dev
✔ Validated project entry file ./index.js
✔ Keystone server listening on port 3000
✔ Initialised Keystone instance
⠋ Connecting to databaseℹ Preparing project for development                                                                                                                                                               16:39:23
ℹ Initial build may take a while                                                                                                                                                                  16:39:23
⠹ Connecting to database✔ Builder initialized                                                                                                                                                                             16:39:23
⠴ Connecting to database✔ Nuxt files generated                                                                                                                                                                            16:39:24

✖ Client
  Compiled with some errors in 12.85s

✔ Server
  Compiled successfully in 10.00s


 ERROR  Failed to compile with 5 errors                                                                                                                                           friendly-errors 16:39:38

These dependencies were not found:                                                                                                                                                friendly-errors 16:39:38
                                                                                                                                                                                  friendly-errors 16:39:38
* core-js/modules/es6.regexp.to-string in ./dist/utils.js, ./dist/components/nuxt.js                                                                                              friendly-errors 16:39:38
* core-js/modules/es7.promise.finally in ./dist/client.js                                                                                                                         friendly-errors 16:39:38
* core-js/modules/es7.symbol.async-iterator in ./dist/client.js, ./dist/components/nuxt-link.client.js                                                                            friendly-errors 16:39:38
                                                                                                                                                                                  friendly-errors 16:39:38
To install them, you can run: npm install --save core-js/modules/es6.regexp.to-string core-js/modules/es7.promise.finally core-js/modules/es7.symbol.async-iterator               friendly-errors 16:39:38
ℹ Waiting for file changes                                                                                                                                                                        16:39:38
[...]

Issue #2133, opened 2019-12-20
User not created when setting up Keystone with Postgres


martinheidegger:
# Bug report

I tried setting up Keybase following the getting started guide but on `npm run dev` it starts keystone but no user is created.

## To Reproduce

Steps to reproduce the behaviour. Please provide code snippets or a repository:

1. init Keystone
2. install a Postgres database
3. set the user confirguration for the postgres db
4. start the server and see that no user is created

## Expected behaviour

The user should be created immediately or there should be a screen that shows to create a user when starting the app like #1782.

## Additional context

I tried to debug `initial-data.js` and changing following:

https://github.com/keystonejs/keystone/blob/06bff4119a44769b359006e8e6b9e07c32b21a8e/packages/create-keystone-app/example-projects/starter/initial-data.js#L6-L10

to:

```javascript
  const res = await keystone.executeQuery(
    `query {
      _allUsersMeta {
        count
      }
    }`
  )
  console.log(res)
  // Count existing users
  const {
    data: {
      _allUsersMeta: { count },
    },
  } = res;
```
I got following error:

```
{
  errors: [
    error: relation "public.User" does not exist
        at Connection.parseE (./keystonejs/my-app/node_modules/pg/lib/connection.js:604:13)
        at Connection.parseMessage (./keystonejs/my-app/node_modules/pg/lib/connection.js:403:19)
        at Socket.<anonymous> (./keystonejs/my-app/node_modules/pg/lib/connection.js:123:22)
        at Socket.emit (events.js:210:5)
        at addChunk (_stream_readable.js:308:12)
        at readableAddChunk (_stream_readable.js:289:11)
        at Socket.Readable.push (_stream_readable.js:223:10)
        at TCP.onStreamRead (internal/stream_base_commons.js:182:23)
    From previous event:
        at Client_PG._query (./keystonejs/my-app/node_modules/knex/lib/dialects/postgres/index.js:236:12)
        at Client_PG.query (./keystonejs/my-app/node_modules/knex/lib/client.js:168:17)
        at Runner.query (./keystonejs/my-app/node_modules/knex/lib/runner.js:133:36)
        at ./keystonejs/my-app/node_modules/knex/lib/runner.js:39:23
        at ./keystonejs/my-app/node_modules/knex/lib/runner.js:253:24
        at processImmediate (internal/timers.js:439:21)
    From previous event:
[...]

PR #2414, opened 2020-02-23
Bugfix/2159 create tables

agneym:
Solves #2159.

Related to #2133

On Knex adapter, tables where not being created when `dropDatabase` was not enabled and hence the ToDo starter could only be run with `dropDatabase: true` for the first time. 

This PR calls `createTables` on each `postConnect` but creates tables only when they do not already exist. This should make the starter Postgres adapters run smoothly without the `dropDatabase` option.


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -78,8 +78,6 @@ class KnexAdapter extends BaseKeystoneAdapter {
         console.log('Knex adapter: Dropping database');
       }
       await this.dropDatabase();
-    } else {
-      return [];
     }
     return this._createTables();
```
MadeByMike: Am I wrong in reading that this return statement will never be hit?

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -282,9 +280,12 @@ class KnexListAdapter extends BaseListAdapter {
 
   async createTable() {
     // Let the field adapter add what it needs to the table schema
-    await this._schema().createTable(this.tableName, table => {
-      this.fieldAdapters.forEach(adapter => adapter.addToTableSchema(table));
-    });
+    const exists = await this._schema().hasTable(this.tableName);
```
MadeByMike: @timleslie thoughts on this?

packages/adapter-knex/lib/adapter-knex.js
agneym: With the `return` on 81 and 82 this was indeed the case. The `return` for `createTables` was never being hit and hence the issue where no tables where created for Postgres people.

raulrpearson:
I'm really looking forward to this PR being merged. I've been tearing my hair out for several days thinking I was [doing something wrong on deployment](https://github.com/keystonejs/keystone/issues/2488#issuecomment-596985380), but I think the culprit was the problem addressed in this PR. I'm also using the knex adapter with Postgres.

abhijithvijayan:
any status on merging this PR?

agneym:
I would have thought this was an important issue because people using Postgres could not even start their demo project without running into an error. But a month after this PR, not so much.

kyleoliveiro:
Was this issue addressed or is Knex still unusable in production?

borg:
New user, followed latest getting started with postgres. Tables are still not being created by adding lists. Have to run 
```
npm run create-tables
```
which will complain unless db dropped. 
Using  "@keystonejs/adapter-knex": "^10.0.1"

In what version should it be resolved?

PR #2424, opened 2020-02-24
Improved TextDayPicker interface

Vultraz:
This makes it so the raw ISO8601 date is shown when the field is focused.
Also fixes a bug with the CalendarDay field where the set date could change when the field lost focus.
This was because chrono was trying to parse the formatted date instead of the raw input.

Unfocused (formatting is what I was testing with):
image

Focused:
image


Vultraz:
@timleslie could this get merged? It fixes a long-standing bug with the CalendarDay field.

omarchehab98:
Thanks!

PR #2435, opened 2020-02-25
Update Documentation: Temporarily add schemaNames param to allow user success

MichaelZaporozhets:
This function no longer works without schemaName/s passed in. This change will allow users to successfully run the `dumpSchema` command per the documentation.

This should be superseded by a change to `keystone.dumpSchema` to read this value by default.

### DIFF
--- a/docs/guides/performance.md
+++ b/docs/guides/performance.md
@@ -74,7 +74,7 @@ keystone.createList('Foo', {
 
 // ---- Schema dumping ----
 if (typeof process.env.DUMP_SCHEMA === 'string') {
-  keystone.dumpSchema(process.env.DUMP_SCHEMA);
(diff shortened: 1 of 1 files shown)


MichaelZaporozhets:
@timleslie I had it setup that way, passing down the schemaNames property seems to work too though...

timleslie:
Hmm, if passing through the schemaNames works then it's definitely working by coincidence! I'd have to trace the logic through, but we definitely should not advocate passing through a list of schema names when the method is expecting a single schema name.

MichaelZaporozhets:
> Hmm, if passing through the schemaNames works then it's definitely working by coincidence! I'd have to trace the logic through, but we definitely should not advocate passing through a list of schema names when the method is expecting a single schema name.

lol. yay javascript.

image

PR #2440, opened 2020-02-26
Fixed filter for Select Field multiple value filter

gautamsi:
UI can only allow selection of single value in Matches or not matches, it should actually be multi select 

this PR only fixes generation of graphql query when setSearch is used with multiple values manually, another PR may fix the UI part,

example on how it works

image

Update: Also fixed multi select for the Filter popout

image

* Can now select multiple value for the filter for Select field. 
* Can hit apply with no value selected which does nothing (equivalent to cancel)
* Can not deselect all options to prevent incorrect behavior, idle situation would be to have filter error state and disable "Apply" button, but that is large change. 
  - you can still cancel out or clear filter using pill clear button.

### DIFF
--- a/packages/fields/src/types/Select/views/Filter.js
+++ b/packages/fields/src/types/Select/views/Filter.js
@@ -25,9 +25,10 @@ const SelectFilterView = ({ innerRef, field, value, onChange }) => {
   };
 
   const handleSelectChange = newValue => {
-    if (newValue.length == 0) return;
-    const options = [].concat(newValue); // ensure consistent data shape
-    onChange({ ...value, options });
+    if (newValue.length) {
+      const options = [...newValue]; // ensure consistent data shape
+      onChange({ ...value, options });
+    }
   };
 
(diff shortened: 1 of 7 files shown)


packages/fields/src/types/Select/views/Filter.js
```diff
@@ -25,6 +25,7 @@ const SelectFilterView = ({ innerRef, field, value, onChange }) => {
   };
 
   const handleSelectChange = newValue => {
+    if (newValue.length == 0) return;
```
MadeByMike, line 28:
```suggestion
    if (!newValue.length) return;
```
This is a little confusing, do we want to return if the length is exactly `0` vs undefined? Or are we ok to just check we have a non-empty array by checking for length? 

If just checking for a non-empty array I'd suggest we make it a non-inverted check and just wrap the next 2 lines.
A positive check is alway easier to read and since we have now checked it's an array do we need to concat or just make a copy? 
```
if (newValue.length) {
  onChange({ ...value, options: [...newValue] });
}

PR #2441, opened 2020-02-26
Added listView option for persisted url state

gautamsi:
Internal changes on key name for persisted config names. 

in general no one should be affected unless composing custom pages based on admin-ui components

### **Why do I need it**
I have followed a strategy where I compose custom pages by using existing admin-ui components. I have multiple custom pages which is kind of filtered view for same list. having same persisted search/filter/url-state/config is frustrating, as soon as you change a filter on one page it changes everywhere. 

with this change , default behavior does not changes unless you add `listView='name'` for different component affected and/or the list data-hooks. when you add a value for `listView` then it appends this string in local storage and no more same list state on all pages.

I named it `listView`, please suggest better name if needed.

### DIFF
--- a/.changeset/fuzzy-apples-perform.md
+++ b/.changeset/fuzzy-apples-perform.md
@@ -1,0 +1,7 @@
+---
+'@keystonejs/app-admin-ui': minor
+---
+
+Made internal changes to allow separate key for persisted list ui state (search/filter/column etc.) when composing custom pages using admin-ui components.
+
+when you compose using `List` component, you can now add `listView='some string'` to separate config from one rendering to another.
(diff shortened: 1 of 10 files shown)


MadeByMike:
I don't like the way this adds `listView` to so many parts of the code. There must be a better way to manage this state. Could you give me an example so I can reproduce the problem and maybe we can come up with a simpler way to do this.

gautamsi:
I have a custom admin page which is composed of the admin ui components (similar to List page) to filter the list based on customer need, there are several of those pages which makes use of filtered list view for different approvals and actions. 

given that it saves all the filter and search for the list in localstorage, it is applicable to all my custom pages. I apply filter on one page and it is replicated to other page as well.  Which I want to override.

MadeByMike:
@gautamsi I need to see a code example to understand this. What's the minimum custom page + list config to get reproduce this problem?

gautamsi:
@MadeByMike example at https://github.com/gautamsi/test-listview-keystone

I have added a bare example without any business logic. I would want to show even further customized list view with pre-defined filter etc. current challenge is that you apply any filter, search and other view changes to one page, it replicates to all other pages.

gautamsi:
@timleslie FIXED CONFLICT

PR #2448, opened 2020-02-28
Switched to official TinyMCE React integration.

Vultraz:
No need to use a custom component.


packages/fields-wysiwyg-tinymce/src/views/Field.js
```diff
@@ -1,35 +1,70 @@
...
+        border-radius: 5px !important;
+        border-color: #c1c7d0 !important;
+      }
+    `}
+  />
+);
+
+const WysiwygField = ({ onChange, autoFocus, field, errors, value: serverValue }) => {
+  const handleChange = value => {
+    if (typeof value === 'string') {
```
Vultraz: Removed the check against `serverValue` since `ItemDetails` does that check now. It might be more performance-friendly to add an option to disable that check there, though, since these posts can get large.

ra-external:
@gautamsi -- I actually believe I tried this when it came out and for some reason neither I nor @Vultraz  could determine it wasn't detecting the `images_upload_handler` message passed in to `init`.  When I use @tinymce/tinymce-react` in a custom field, it does though. 

I was able to get around the issue with the `async` throwing it into a loop by using promises instead.  I don't know why promises work and async not, I assume it's some internal transpilation issue that's throwing an error an then KeystoneJS is rebooting continuously.  However, I couldn't upload the GraphQL API on KeystoneJS -- see https://github.com/keystonejs/keystone/issues/2516 -- and it seems to be a server issue, so I've abandoned that route for the moment.

PR #2449, opened 2020-02-28
Allow field configuration of the WYSIWYG editor

Vultraz:
Built on top of #2448.

Adds an `editorConfig` key to the WYSIWYG field so users can override the default TinyMCE options:
```js
    body: {
      type: Wysiwyg,
      editorConfig: {
        toolbar: 'formatselect'
      }
    },
```


MadeByMike:
Sounds reasonable. Was there a reason you deleted the `./Editor` file? It seems that this was a sensible place to split the code.

MadeByMike:
OK, I see the other PR now

Vultraz:
@MadeByMike yeah, I just built this on top of the other PR since I was working on both at the same time. If the other one isn't wanted for some reason I'll split this out separately.

Issue #2418, opened 2020-02-24
[RFC] GraphQL Provider Framework


timleslie:
# Background

One of the core responsibilities of Keystone is to generate a GraphQL API. In the common use case, this API consists of **CRUD operations** over a collection of lists. It also provides **authentication mutations**, **custom types/queries/mutations**, as well as an **application version** query.

The current implementation of each of these features is tightly woven into the core of [Keystone](https://github.com/keystonejs/keystone/blob/master/packages/keystone/lib/Keystone/index.js). This tight coupling is a problem, as it blurs the lines between each of the concerns, as well as making it hard to introduce new features into the GraphQL API.

# Proposal

I would like to propose the introduction of a `GraphQL Provider Framework`. While this sounds big and impressive, it can be easily summarised by the following class API definition:

```javascript
class Provider {
  constructor() {}

  getTypes({ schemaName }) {
    return [];
  }
  getQueries({ schemaName }) {
    return [];
  }
  getMutations({ schemaName }) {
    return [];
  }

  getTypeResolvers({ schemaName }) {
    return {};
  }
  getQueryResolvers({ schemaName }) {
    return {};
  }
  getMutationResolvers({ schemaName }) {
    return {};
  }
}
```

Each feature of Keystone which generates graphQL API surface area would be encapsulated in a `Provider` class, which has methods for generating the type definitions and resolvers required by Apollo.

Keystone would keep track of an array of `providers`, and iterate over these to generate the combined set of type definitions and resolvers which make up the complete graphQL API.

# Discussion

### How would this impact the current implementation?

The code which currently lives in `Keystone.getAdminSchema()` and `Keystone.getTypeDefs()` would be greatly simplified to iteration over the list of providers. All the functionality would be shifted into four separate providers:

 * `ListCRUDProvider`
 * `ListAuthProvider`
 * `AppVersionProvider`
 * `CustomProvider`

### How would I add my own `Provider` to Keystone?

A system writer would be able to implement their own `Provider` class and then call `Keystone.addProvider({ provider })`.

### Why don't we just have `getTypedefs` and `getResolvers` as the `Provider` API?

The way Keystone currently combines the different pieces of the graphQL API means that the types, queries and mutations all need to be handled individually. We could definitely write some kind of `BaseProvider` class which had `getTypedefs()` and `getResolvers()` and leveraged the other methods to generate valid results which could be passed directly into `ApolloServer`. I'm 50/50 on whether we want to make this the default or not.

[...]

PR #2457, opened 2020-03-02
Providers 6/n: Auth list provider

timleslie:
c.f. #2418 

This PR moves the responsibility for generating and resolver authentication queries/mutations out of the `List` class and into a `ListAuthProvider` class.

This will hopefully make it much easier to understand the behaviour of these mutations and to decouple the CRUD API from the Auth API.

This will be a *major* breaking change, as you can no longer call `createAuthStrategy` before defining the associated. This is probably for the best, as it explicitly prevents people setting up an auth strategy against a non-existent list.

### DIFF
--- a/.changeset/plenty-pandas-smoke.md
+++ b/.changeset/plenty-pandas-smoke.md
@@ -1,0 +1,17 @@
+---
+'@keystonejs/keystone': major
+---
+
+Introduced `ListAuthProvider` to generate authentication mutations and queries. This introduces the following breaking changes:
+ * `Keystone.createAuthStrategy()` must be called *after* the associated `List` has been created.
+ * `List.hasAuth()` has been removed.
+ * `List.getAuth()` has been removed.
+ * `List.gqlNames.authenticatedQueryName` has been removed.
+ * `List.gqlNames.authenticateMutationPrefix` has been removed.
+ * `List.gqlNames.unauthenticateMutationName` has been removed.
+ * `List.gqlNames.authenticateOutputName` has been removed.
+ * `List.gqlNames.unauthenticateOutputName` has been removed.
+ * `List.checkAuthAccess()` has been removed.
+ * `List.authenticatedQuery()` has been removed.
+ * `List.authenticateMutation()` has been removed.
+ * `List.unauthenticateMutation()` has been removed.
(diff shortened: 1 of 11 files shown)


.changeset/plenty-pandas-smoke.md
```diff
@@ -0,0 +1,17 @@
...
+ * `List.getAuth()` has been removed.
+ * `List.gqlNames.authenticatedQueryName` has been removed.
+ * `List.gqlNames.authenticateMutationPrefix` has been removed.
+ * `List.gqlNames.unauthenticateMutationName` has been removed.
+ * `List.gqlNames.authenticateOutputName` has been removed.
+ * `List.gqlNames.unauthenticateOutputName` has been removed.
+ * `List.checkAuthAccess()` has been removed.
+ * `List.authenticatedQuery()` has been removed.
+ * `List.authenticateMutation()` has been removed.
+ * `List.unauthenticateMutation()` has been removed.
```
MadeByMike: This looks significant and scary. Although I doubt the impact outside this team is large can we add a note about how you migrate code that uses these methods?

demo-projects/blog/index.js
```diff
@@ -24,16 +24,16 @@ const keystone = new Keystone({
...
-  type: PasswordAuthStrategy,
-  list: 'User',
-});
-
 keystone.createList('User', User);
 keystone.createList('Post', Post);
 keystone.createList('PostCategory', PostCategory);
 keystone.createList('Comment', Comment);
 
+const authStrategy = keystone.createAuthStrategy({
```
MadeByMike: For the sake of documenting the discussion If we are going to rely on a specific order of execution for these I feel like maybe rather than passing a string to createAuthStrategy perhaps we should pass the list itself?

```js
const list = keystone.createList('User', User);
  type: PasswordAuthStrategy,
  list
});
```

Would you pass a promise or await createList?

.changeset/plenty-pandas-smoke.md
timleslie: These methods can now be found on the `AuthListProvider` object, however they generally should not be considered part of the Keystone public API. Users who wish to access them can still find them in `keystone._providers`.

demo-projects/blog/index.js
timleslie: `createList` isn't async, it simply returns a `List` object, which could be passed in as an option to `createAuthStrategy` in a hypothetical future change to this API.

I would like to defer making such an API change to a seperate PR and discussion.

PR #2458, opened 2020-03-02
Handle missing oEmbeds in field-content document

SeanDoyleGit:
This PR updates the field-content controller to filter out missing oEmbeds from the document object.  

When content contains an oEmbed the document object will contain a node that points to the oEmbed via `_joinIds`. An oEmbed with this Id should then be included in the oEmbeds array e.g. 
```
content: {
  document: {"object":"document","data":{},"nodes":[{"object":"block","type":"oEmbed","data":{"_joinIds":["5e5c6bb37d65fb1a0b6c210d"]},"nodes":[{"object":"text","text":"","marks":[]}]}]},
  oEmbeds: [{ id: "5e5c6bb37d65fb1a0b6c210d", embed: {type: "rich", originalUrl: "https://google.com", version: "1.0", title: "Leap Day 2020",…}]
}
```

If the oEmbed is missing from the content oEmbeds array e.g.
```
content: {
  document: {"object":"document","data":{},"nodes":[{"object":"block","type":"oEmbed","data":{"_joinIds":["5e5c6bb37d65fb1a0b6c210d"]},"nodes":[{"object":"text","text":"","marks":[]}]}]},
  oEmbeds: []
}
```
The Admin UI will throw the following error `Uncaught TypeError: Cannot read property 'embed' of undefined`

By adding the filter to the field-content controller, the missing oEmbed is removed from the document and is not rendered to the user. When the content is then saved the missing oEmbed will have been removed from the document object.


.changeset/slow-boats-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/field-content': patch
+---
+
+Handle missing oEmbeds in field-content document
```
MadeByMike:
```suggestion
Changed the `deserialize` method of the content field to filter out missing oEmbeds
```

PR #2463, opened 2020-03-03
Admin UI: replaced ContainerQuery component with useResizeObserver

Vultraz:
*sweep sweep sweep*


MadeByMike:
🤔 I don't even want to own this code. Can we just add something like: https://github.com/ZeeCoder/use-resize-observer

Vultraz:
@MadeByMike ah, good idea. Done!

Vultraz:
Agh, dammit, now that `Uncaught Error: ResizeObserver loop limit exceeded` error shows up again. >_<

packages/app-admin-ui/client/pages/Home/index.js
```diff
@@ -56,6 +24,17 @@ const HomepageListProvider = () => {
   const query = gqlCountQueries(lists);
   const { data, error } = useQuery(query, { fetchPolicy: 'cache-and-network', errorPolicy: 'all' });
 
+  const [cellWidth, setCellWidth] = useState(3);
+  const { ref: measureElement } = useResizeObserver({
+    onResize: ({ width }) => {
+      // requestAnimationFrame works around a weird 'ResizeObserver loop limit exceeded' error.
```
MadeByMike: `use-resize-observer` has a throttle\debounce method for this: https://github.com/ZeeCoder/use-resize-observer#throttle--debounce

I think we want to throttle it to about 15 fps ie around 60ms

Vultraz: Switched to using `lodash.throttle` as the docs suggest.

Vultraz: Looks like `requestAnimationFrame` was still required after all.

Vultraz: Ok, `throttle` doesn't fix the problem (see https://github.com/WICG/ResizeObserver/issues/38). `requestAnimationFrame` was needed after all.

timleslie:
@Vultraz Could you fix up the merge conflict?
@MadeByMike Could you re-review this and give it a ✅ if you're happy with it?

Vultraz:
@timleslie done

MadeByMike:
Nice! Thanks @Vultraz

timleslie:
I'm trying to work out why these tests are failing. In the meantime I'm gonna keep re-triggering until this goes green :-) If nothing else it will give me insight into the tests and how and why they might be failing!

PR #2469, opened 2020-03-03
Add more custom actions

gautamsi:
adding more hooks like `itemHeaderActions` in https://github.com/keystonejs/keystone/pull/2279

- `listHeaderActions` - this adds button in place of `Create` item button on list page.
image
- `listManageActions` - this adds button in place of multi select update on list page (Update Many and Delete Many).
image
- `listItemActions` - this adds dropdown buttons for each item in list. caveat: you have to recreate the DropDown component, can be copied from the source.
image

All of the above is composable from the existing components on page. the react component should accept the components as props and render them if they want to keep the original component.


packages/app-admin-ui/client/components/ListTable.js
```diff
@@ -134,6 +135,63 @@ class SortLink extends React.Component<SortLinkProps> {
...
+    <Fragment>
+      <DeleteItemModal
+        isOpen={deleteModalIsVisible}
+        item={item}
+        list={list}
+        onClose={() => setDeleteModal(false)}
+        onDelete={handleDelete}
+      />
+      {listItemActions ? (
+        listItemActions({ list, item, link, onDelete, ...props, items })
```
MadeByMike: I'm sorry, I've done a poor job of communicating how AdminUI hooks are intended to work and this has evolved as we have discussed and evaluated the pattern. 

Rather than passing a list of items to hooks like this: 

We have decided that we want to make the components a part of the API, this makes it easier for people to add, remove, replace, and reorder parts as they need.

I know the current example in the codebase is a bad example as it is implemented in the same way as this:

```js
...
{itemHeaderActions ? (
  itemHeaderActions({ ItemId, AddNewItem })
  ) : (
    <div>
      <ItemId />
      <AddNewItem />
    </div>
)}
```

This should be refactored to just:

```
{itemHeaderActions ? (
  itemHeaderActions(ctx)
  ) : (
    <div>
      <ItemId />
      <AddNewItem />
    </div>
)}
```

Your example is also exporting event callbacks which I assume is so that the consumer can manually re-apply these props to the items? This in particular is something we want to avoid.

In my example `ctx` is the __as yet__ undetermined context variables such as the route and any other contextual information sortOrder etc... that may not be available through context. I began work creating a spec for each of these and what I think the context variables should be. I'll chase that down and share it here.

For this approach to work it's important that the items rendered inside a hook are simple stand-alone components without props: `<ItemId />` and `<AddNewItem />` have been refactored so that all the essential information is available from context providers. 

For this to work, `AddNewItem` and `ItemId` both need to be exported from `packages/app-admin-ui/components/index.js`. This means consumers could write a hook like this:

```js
// These imports are the API
import { AddNewItem, ItemID } from "@keystonejs/app-admin-ui/components";

export default {
  itemHeaderActions: (ctx) => (
    <>
     <ItemId />
     <p>Hello Admin Hooks</p>
     <AddNewItem />
    </>
  )
}
```

So... I know this is a bit of a pain but for this PR to be accepted I need it to conform with the intended direction. This means refactoring the components in `listItemActions` to be standalone and exported from `packages/app-admin-ui/components/index.js`. No props passed to `listItemActions` for now.

I'd also like to see one hook per PR for now as I want to be sure we are getting the API right before we add a bunch in one hook.

gautamsi: > In my example ctx is the as yet undetermined context variables such as the route and any other contextual information sortOrder etc..

I think we can just use blank/no ctx for now and decide when the need come.

gautamsi: I have create 4 new PR for this #2478 #2479 #2482 and #2483 

currently no context is passed in custom action but added a provider for passing around specific details needed for making components stateless. more state/context info may be added as needed

gautamsi:
@MadeByMike seems like it is still not finalized 😞

MadeByMike:
@gautamsi what's the state of this PR? I feel like the other pull requests related to this are in better shape and maybe cover some of what is here. I'm not going to focus on this one today because I think that will be a distraction.

Is it worth factoring out the good bit from here and closing this?

gautamsi:
this was split into #2478, #2479 which are merged already. 

remaining 2 PR can be reviewed separately #2482 and #2483, I will fix their conflict now.

PR #2471, opened 2020-03-03
Admin UI: redesigned signout page

Vultraz:
The signout page has always looked like a WIP design, so I redid it.

While signing out:
image

After signing out:
image


packages/app-admin-ui/client/pages/Signout.js
```diff
@@ -86,24 +70,28 @@ const SignedOutPage = () => {
...
+          <Caption>Signing out.</Caption>
+        </Fragment>
+      ) : (
+        <Fragment>
+          <Animation name="pulse" duration="500ms">
+            <CheckIcon css={{ color: colors.primary, height: '3em', width: '3em' }} />
+          </Animation>
+          <Caption>You have been signed out.</Caption>
+          <FlexBox>
+            {/* TODO: is it ok to use '/' as a path here? */}
```
MadeByMike: Not necessarily, they might not even have a front-end application.

MadeByMike: I'd remove this as we can't make and assumption and don't want to cross concerns between apps

MadeByMike:
I'd remove the home link ... although I would like a better way for sharing information across the 'app stack' I'd love for the adminUI to be able to query something and get the base URL. Similarly we need a better way for sharing graphQL end points with the AdminUI currently it's in the initialisation of the admin app and only 1.

packages/app-admin-ui/client/pages/Signout.js
Vultraz: What about a UI hook (and if so, should a Home button be included or excluded by default)?

Vultraz:
@MadeByMike I commented out the Home button for now. I'll deal with adding a UI hook separately.

MadeByMike:
@Vultraz could you remove it? I don't like merging commented out functional code. Issues are the right to document this if you think it needs to be captured as as feature.

packages/app-admin-ui/client/pages/Signout.js
```diff
@@ -86,24 +70,29 @@ const SignedOutPage = () => {
...
+          <Animation name="pulse" duration="500ms">
+            <CheckIcon css={{ color: colors.primary, height: '3em', width: '3em' }} />
+          </Animation>
+          <Caption>You have been signed out.</Caption>
+          <FlexBox>
+            {/* TODO: handle with UI hook
+            <SignOutPageButton appearance="primary" href="/">
+              Home
+            </SignOutPageButton>
+            */}
```
MadeByMike:
```suggestion
```

MadeByMike:
I've remove it :)

Vultraz:
@MadeByMike anything else I should do on this?

timleslie:
@MadeByMike Could you re-review this now?

PR #2473, opened 2020-03-03
Admin UI: fixed webpack error when using tracking plugins

### LINKED ISSUE #2472: byTracking plugin causes webpack error
# Bug report

## Describe the bug

A webpack error is thrown when I create new list with `byTracking` plugin. Without this plugin there is no error.

It seems like .ts type definition is in .js file, and there should not be, in line 102 of file:
`@keystonejs/list-plugins/node_modules/@keystonejs/app-admin-ui/client/components/ListTable.js`

## To Reproduce

1. Init users and auth sample project
2. Add new list with byTracking plugin:

```javascript
const { atTracking, byTracking } = require('@keystonejs/list-plugins');
const { Text } = require('@keystonejs/fields');

// ... keystone int etc.

keystone.createList('Lesson',{
  fields: {
    name: { type: Text },
  },
  plugins: [
    atTracking(),
    byTracking(),
  ],
});

// ... module exports

```
3. Run app.
4. See error after admin login.

## Expected behaviour

No error.

## Screenshots

admin-error
code-error

## System information

- OS: Pop_OS! (Ubuntu)
- Browser: Chrome

Vultraz:
Should resolve #2472. I couldn't reproduce it, so I'm just going by what the report said.

### DIFF
--- a/.changeset/slow-pianos-occur.md
+++ b/.changeset/slow-pianos-occur.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Fixed webpack error when using tracking plugins.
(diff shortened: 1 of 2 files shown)


Vultraz:
Ok, this doesn't fix the problem.

PR #2479, opened 2020-03-04
Enable listHeaderActions hook

gautamsi:
ref #2469 

added `listHeaderActions` hook and exported `CreateItem` button component.

 `listHeaderActions` - this adds button in place of `Create` item button on list page.
image

PR #2482, opened 2020-03-04
Enable listManageActions hook

gautamsi:
ref #2469 
this PR is on top of #2479 PR due to underlying changes.

[see this for only changed files](https://github.com/gautamsi/keystone/compare/enable-listHeaderActions-hook...gautamsi:enable-listManageActions)

Enables use of `listManageActions` hook.

this PR also adds a `useListData` hook which passes out the context around this hook.

 `listManageActions` - this adds button in place of multi select update on list page (Update Many and Delete Many).
image


packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -91,169 +85,157 @@ export function ListLayout(props) {
...
-              <FlexGroup align="center" growIndexes={[0]}>
-                <div
-                  css={{
-                    alignItems: 'center',
-                    color: colors.N40,
-                    display: 'flex',
-                    marginTop: gridSize - 2,
+    <ListDataProvider {...{ query, items, selectedItems, onSelectChange }}>
+      <main>
+        <div ref={measureElementRef} />
```
MadeByMike, line 94: I thought this was factored out?

packages/app-admin-ui/client/providers/List.js
```diff
@@ -38,3 +38,19 @@ export const ListProvider = ({ list, children }) => {
     </ListContext.Provider>
   );
 };
+
+const ListDataContext = createContext();
+export const useListData = () => useContext(ListDataContext);
+
+export const ListDataProvider = ({ query, items, selectedItems, onSelectChange, children }) => (
```
MadeByMike, line 45: I'm reluctant to add another provider if it can at all be avoided. I see why it might be needed but wonder what other patterns you considered. One thing that is not clear to me is why `onSelectChange` is managed by the `ListDataProvider` but `handleDelete` isn't.

If we are going to have one, I'd like to think about what it should be responsible for. For example, can it handle fetching and re-fecting data and just provide events back? I'm not sure how it should work.

MadeByMike:
We can delete one of those changesets and make sure the remaining one is a complete sentence in past tense and a usage example: E.g.

Added hook `example` to the Admin UI.  
Usage:
```js
// code example here
```

packages/app-admin-ui/client/pages/List/index.js
gautamsi: yes, this is gone in rebase

MadeByMike:
Looks great, I'm pretty happy with everything other than the data provider. I'm singing off for the evening soon so I'll pick this up tomorrow. 👍 Thanks for the help.

packages/app-admin-ui/client/providers/List.js
gautamsi: on digging further I see that `query` and `items` can be used from `useList()`.

`onSelectChange` is the behaviour which  is passed down from `useListSelect()`, unfortunately this is not using s shared state, every time you have `useListSelect(items)` even though the items are same but the internal tracking of selection is different. I have tried using `useListSelect()` in components but they do not get selected items.

We can merge this with `useList()` and that should help consolidate this. let me try do that.

gautamsi: https://github.com/keystonejs/keystone/pull/2482/commits/54b208e4f604f600f8b2cd63f16ba74bc1cbb0cd removes the extra ListDataProvider but the commit does not work due to `useListSelect` not being shared for list.

I had to merge `useListSelect` with `useList` so that `useList` now also returns `selectedItems` and `onSelectChange` (should be renamed to `setSelectedItems` really).

MadeByMike:
We need to be as conservative as possible when it comes to exporting components since once we do we can't change them easily. 

I can't see any reason to expose `useKeyDown` or `useListSelect` I think `selectedItems` should be context passed to the hook.

PR #2487, opened 2020-03-05
update manypkg

Noviny:
It looks like a lot of changes, but mainly I updated manypkg and then ran manypkg fix.

Reason I took the time to do this, manypkg now enforces having a `repository` field, and the absence of it broke changelogs.xyz showing the changelogs once we switched to Algolia

I'm looking at approaching this from the other side where algolia finds changelogs that are published to npm as well as things just from repos, but this seemed fine to do anyway.


Noviny:
The other notable change is internal devDependencies are now enforced to be `*` range - this means they always symlink, updating packages never causes them to leave range.

jesstelford:
Do the new packages need to be published now for changelogs.xyz to pick up the repository field properly?

Noviny:
Yeah until they’re published they won’t be picked up.

jesstelford:
@all-contributors add @Noviny for maintenance & tools please 🙏

PR #2499, opened 2020-03-09
fix chrome mac print bug

mikehazell:
## Fixes an issue with printing docs in Chrome on Mac
Also removes the background colour when printing.

### Old print:

image

### New print:

image

### DIFF
--- a/packages/arch/packages/theme/src/index.js
+++ b/packages/arch/packages/theme/src/index.js
@@ -10,6 +10,10 @@ export const fontFamily = `
   "Segoe UI Emoji",
   "Segoe UI Symbol"
 `;
+
+// BlinkMacSystemFont produces very strange characters when printing from Chrome on Mac.
+const printFontFamily = fontFamily.replace('BlinkMacSystemFont,', '');
+
 export const borderRadius = 6;
 export const gridSize = 8;
 export const fontSize = 16;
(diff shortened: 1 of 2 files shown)


packages/arch/packages/theme/src/index.js
```diff
@@ -33,6 +37,10 @@ export const globalStyles = {
     MozFontFeatureSettings: "'liga' on",
     MozOsxFontSmoothing: 'grayscale',
     WebkitFontSmoothing: 'antialiased',
+    '@media print': {
+      backgroundColor: 'white',
+      fontFamily: printFontFamily,
```
MadeByMike:
```suggestion
      fontFamily: 'unset',
```

packages/arch/packages/theme/src/index.js
```diff
@@ -10,6 +10,10 @@ export const fontFamily = `
   "Segoe UI Emoji",
   "Segoe UI Symbol"
 `;
+
+// BlinkMacSystemFont produces very strange characters when printing from Chrome on Mac.
+const printFontFamily = fontFamily.replace('BlinkMacSystemFont,', '');
```
MadeByMike:
```suggestion
```

packages/arch/packages/theme/src/index.js
mikehazell: `unset` will set it back to a serif font. I don't think thats better.

image

PR #2500, opened 2020-03-09
[RFC] New API doc formatting proposal

Vultraz:
Using the main `keystone` README for this example.

I've been thinking for awhile the use of non-wrapping tables for the API docs isn't the best UX. It serves well when describing a few simple options, but for something like the `keystone` class docs it doesn't look that great.

Issues:
* You can end up with an ugly horizontal scrollbar
* Split concerns. A new user has to check both the table and down below for more details (if any).

Proposal
* The initial Usage section would list *all* options accepted by that class in alphabetical order.
* The Config section would list each option in the same order with this format (based on the current [DB adapter](https://www.keystonejs.com/keystonejs/adapter-knex/) docs:
```
### `option: type`

_**Default:**_ `default value`

Description

#### Usage

More detailed usage example if necessary
```

Benefits:
* Simple at-a-glance list of all options is retained, but cleaner
* Option info is easy to find due to alphabetical ordering
* Only need to worry about updating option details in one place.

### DIFF
--- a/.changeset/silver-tables-build.md
+++ b/.changeset/silver-tables-build.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': minor
+---
+
+Add Mutation.updateAuthenticated<List>, commonly used as Mutation.updateAuthenticatedUser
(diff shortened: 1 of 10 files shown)


timleslie:
Interesting that you should raise this right now, as I've literally just printed off every page of the Keystone docs and I'm going through it all with a red pen. I suspect this exercise will result in some structural as well as superficial changes, so I'll definitely keep this proposal at the front of mind as I'm reading through everything 👍

PR #2534, opened 2020-03-16
Feature/passport add authenticate options

frank10gm:
Add passport authenticate options (eg. prompt: 'consent' to receive Google refresh_token).


jesstelford:
@all-contributors Please add @frank10gm for code ♥

PR #2538, opened 2020-03-17
Updated mongo dependencies to latest version

Vultraz:
Housekeeping


packages/adapter-mongoose/package.json
```diff
@@ -14,7 +14,7 @@
     "@keystonejs/mongo-join-builder": "^6.1.2",
     "@keystonejs/utils": "^5.2.2",
     "@sindresorhus/slugify": "^0.6.0",
-    "mongoose": "^5.8.4",
+    "mongoose": "^5.9.5",
```
gautamsi: isn't this included in `yarn upgrade`, do we need to update dependency versions without actual gain?

PR #2549, opened 2020-03-18
Remove useless existingItem field from defaultValue field function

jesstelford:
Because `getDefaultValue` (and hence the field's `defaultValue`) function is only ever called during create, there can never be an `existingItem`.

### DIFF
--- a/.changeset/plenty-donkeys-draw.md
+++ b/.changeset/plenty-donkeys-draw.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/fields': patch
+'@keystonejs/keystone': patch
+---
+
+Remove unused 'existingItem' parameter from calls to .defaultValue() methods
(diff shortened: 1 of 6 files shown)


.changeset/plenty-donkeys-draw.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/fields': patch
+'@keystonejs/keystone': patch
```
jesstelford: `patch` level change because the value was hard coded to `undefined` before anyway, so now it will still be `undefined`:

```javascript
const { existingItem } = { foo: 'bar' };
console.log(existingItem);
// undefined
```

.changeset/plenty-donkeys-draw.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Remove unused 'existingItem' parameter from calls to .defaultValue() methods
```
timleslie:
```suggestion
Removed unused 'existingItem' parameter from calls to .defaultValue() methods.
```

jesstelford:
Good catch! Done 👍

PR #2555, opened 2020-03-20
Add _getNearFar method to knex adapter

timleslie:
Another refactor to bring us closer to our end goal of fixed relationships.


timleslie:
No need for this now I don't think.

PR #2557, opened 2020-03-20
[WIP] New Image service

JedWatson:
Opening this PR for visibility, it's not ready to merge yet.

We're working on a new local image service for Keystone, built on sharp. Consider it experimental for now! There are some big caveats to document, primarily that it's not for production use.

It won't scale, or play nicely with ephemeral hosting environments (like heroku or serverless), and it also won't work in a multi-server production setup. Answers for production use will come later.

But it should be a great way to start developing with Keystone locally, which is an experience we're putting increasing focus on.


packages/images/src/service.js
```diff
@@ -0,0 +1,199 @@
...
+        const md5sum = crypto.createHash('md5');
+        md5sum.update(JSON.stringify(resizeOptions));
+        const hash = md5sum.digest('hex');
+
+        suffix = `-${hash}`;
+      }
+
+      let transformedFilename = nodePath.join(this.transformsPath, `${id}${suffix}.${format}`);
+
+      if (!(await fs.exists(transformedFilename))) {
```
irwin-r: IMO we should inverse this;  the simpler condition can terminate earlier.

packages/images/src/service.js
```diff
@@ -0,0 +1,199 @@
...
+
+    this.app.post('/upload', upload.single('image'), async (req, res) => {
+      const { originalname, path } = req.file;
+
+      res.status(201).json(this.storeImageInDatabase({ path, originalname }));
+    });
+
+    this.app.get('/image/:filename', async (req, res) => {
+      const { filename } = req.params;
+      const [id, format] = filename.split('.');
```
MadeByMike: Very plausible for filenames to have `.` maybe we should: 
```
.substring(filename.lastIndexOf('.')+1, filename.length))
```

irwin-r: I haven't adequately looked into how files are stored, but is it possible to have a file with `.`s in its name?

packages/images/src/service.js
```diff
@@ -0,0 +1,199 @@
...
+
+    this.app.listen(port, () => {
+      console.log(`🌠 KeystoneJS Image Service ready on port ${port}`);
+    });
+  }
+  getSrc(id, { format, resize = {} }) {
+    let url = `${this.protocol}://${this.host}:${this.port}/image/${id}.${format}`;
+
+    let searchParams = new URLSearchParams();
+    for (let key in resize) {
```
irwin-r: I generally try and avoid using `for (.. in ..)` [due to the potential to iterate through inherited values](https://eslint.org/docs/rules/guard-for-in).

Might doing `resize.forEach` be slightly better here?

packages/images/src/service.js
```diff
@@ -0,0 +1,199 @@
...
+      searchParams.set(key, resize[key]);
+    }
+
+    let stringifiedSearchParams = searchParams.toString();
```
irwin-r: Should be a `const`.

packages/images/src/service.js
JedWatson: Not sure what you mean by this?

packages/images/src/service.js
JedWatson: It's fine (for now)

We control the filenames that are created, so until they are made customisable on upload, this is not going to create a problem

packages/images/src/service.js
JedWatson: These days I tend not to code so defensively, since this code is going to be called from our own code anyway (see `Implementation.js`) so unless somebody's extending object globals in this process, it's going to be fine. We could go nuts everywhere, or just call that an unsupported use case.

If it's public API, I'd agree more

khalwat:
Something to check out -> https://github.com/venveo/serverless-sharp

JedWatson:
Making a note here to come back to this and probably integrate the solution into the `<KeystoneImage>` component (making sure it works well with the service)

https://joshwcomeau.com/performance/embracing-modern-image-formats/

PR #2563, opened 2020-03-21
Update usage of Apollo in Blog demo

jesstelford:
Depends on: https://github.com/keystonejs/keystone/pull/2570

The motivator here was a bug with the Demo blog which wasn't correctly passing cookies when doing SSR. So, I updated to the latest Apollo pattern for Next.js which allows simply disabling SSR.

See # for where this is used.


demo-projects/blog/package.json
```diff
@@ -33,7 +33,9 @@
     "@keystonejs/keystone": "^6.0.1",
     "@keystonejs/oembed-adapters": "^5.1.2",
     "apollo-boost": "^0.4.7",
+    "apollo-cache-inmemory": "^1.6.5",
     "apollo-client": "^2.6.8",
+    "apollo-link-http": "^1.5.16",
```
JedWatson: I can't see why this has been added to the dependencies?

demo-projects/blog/app/lib/apollo.js
```diff
@@ -0,0 +1,62 @@
...
+  return globalApolloClient;
+};
+
+/**
+ * Creates a withApollo HOC
+ * that provides the apolloContext
+ * to a next.js Page or AppTree.
+ * @returns {(PageComponent: ReactNode) => ReactNode}
+ */
+export const withApollo = () => PageComponent => {
```
JedWatson: [1/2] This may be straight from the next.js / apollo examples, but in the context of our examples what's the point of the additional function here? there are no arguments, and it kind of ends up as syntax soup. imo it'll make the example easier to follow without the unnecessary function.

demo-projects/blog/app/pages/index.js
```diff
@@ -41,7 +42,7 @@ const Post = ({ post }) => {
   );
 };
 
-export default function() {
+export default withApollo()(() => {
```
JedWatson: [2/2] this is the syntax soup I'm talking about, `withApollo()(()` could be simplified to `withApollo(()` because there's nothing to the first function invocation

demo-projects/blog/package.json
jesstelford: Good catch!

demo-projects/blog/app/lib/apollo.js
jesstelford: Oh yeah, good point! This _used_ to have config options, but I removed them all to simplify it for this demo. Didn't think to remove the extra function. Will remove it 👍

Issue #2520, opened 2020-03-14
Instantiating Content Field with CloudinaryImage.blocks.image throws an error


justintemps:
The docs for [Content Field](https://www.keystonejs.com/keystonejs/field-content/) suggest you should be able to use CloudinaryImage.blocks.image when instantiating a Content Field like this:

```
const { Content } = require('@keystonejs/field-content');

keystone.createList('Post', {
  fields: {
    body: {
      type: Content,
      blocks: [
        CloudinaryImage.blocks.image
      ],
    },
  },
});
```

However, doing so throws the following error: `Error: No file adapter provided for File field.`

Looking into the implementation of `Cloudinary.blocks.image`, I can see that it's a class whose constructor wants the following arguments:

`({ adapter }, { fromList, joinList, createAuxList, getListByKey })`

`adapter` is obviously an instance of `CloudinaryAdapter` but no idea what kind of object the second set of arguments are from.

Wkasel:
Same error

Wkasel:
(fix) 
```javascript
keystone.createList('Post', {
...
      blocks: [
        CloudinaryImage.blocks.image
      ],
   ...
// change to 
      blocks: [
        [CloudinaryImage.blocks.image, { adapter: cloudinaryAdapter }]
      ],
});
```

justintemps:
Thanks @Wkasel ! It seems like this should be explained in the docs. I'll try to make a quick edit later today.

PR #2568, opened 2020-03-22
Update Content Field docs

justintemps:
The current example of how to instantiate a Content Field with `CloudinaryImage.blocks.image` doesn't work, because that block needs to be instantiated as an array with options. 

A functioning example of how to use the CloudinaryImage block together with a general explanation of how to pass options to blocks should help avoid issues like #2520


MadeByMike:
I've created a PR that includes these changes and documents other custom blocks

justintemps:
@MadeByMike great, then I'll close this.

PR #2569, opened 2020-03-23
Update relationships.md

MadeByMike:


docs/discussions/relationships.md
```diff
@@ -127,7 +127,7 @@ keystone.createList('User', {
 });
 ```
 
-The only relationship configuration not currently supported is having a field reference _itself_, e.g. `friends: { type: Relationship, ref: 'User.friends', many: true }`.
+One relationship configuration not currently supported is having a field reference _itself_, e.g. `friends: { type: Relationship, ref: 'User.friends', many: true }`.
```
timleslie: In the context of what we have defined a relationship to be, this is the only one that is missing. That is to say, we *could* implement it within the current framework, we have just chosen not to. In that sense *only* is correct, and also stops people from wondering what the others might be (answer: there are none (and if there are, we should list them here too!)).

timleslie:
I have manually integrated the changes from this branch, thanks for the review, lots of nice improvements!

PR #2571, opened 2020-03-23
Upgrade prettier

jordanoverbye:
[Prettier v.2.0 was recently released](https://prettier.io/blog/2020/03/21/2.0.0.html) so I have upgraded this dependency, updated the prettier config and also run `yarn format`


jesstelford:
Oof, that's a lot of changes 😞

irwin-r:
As discussed with @timleslie, I'm not sure if some of the changes in Prettier 2.0 necessarily improve legibility with respect to method chaining.

It might be worth having some additional discussion regarding these changes 🙂

Context:

image

Issue #315, opened 2018-09-14
JSON/JSONB field type


molomby:
The storage of JSON blobs in the data layer. 

This is often useful when dealing with, for example, external JSON APIs since the raw response can be logged but remain queryable. JSONB values are distinct from Mongo sub documents and subtly different than storing JSON blobs in a text field. The former provides a level of validation (ie. should only accept well-formed values) and it's not expected to be a character-perfect reproduction (eg. key order in objects).

This should present in the interface as a "code" field, that is, preformatted in a fixed-width font.

molomby:
Many field types already persist json values internally, eg. `File` and `Unsplash`.

molomby:
While we're doing this, lets remember that `JSON.parse` and `JSON.stringify`...

* Are synchronous and blocking
* Scale at `O(n)` for the input length

One thing we can do to mitigate this is to make sure that, when being read, **the value is lazily deserialised**.

That is, rather than reading the JSON string from the DB, deserialising it to JS objects and returning that, the field type implementation should create a getter that closures over the string value and deserialises it when accessed (caching the deserialised value internally, for subsequent reads). If done correctly this should be (almost) completely transparent to the consumer.

This is especially important give that we currently (and for the foreseeable future) [load all item fields, regardless of the fields selected by the GraphQL query](https://github.com/keystonejs/keystone/issues/220). Eg. if I have historical info for each of my `Post` items in a 1 MB JSON blob, I don't want the server parsing and creating hundreds of JS objects just so I can read a post's name field (and certainly not when I query the name of 50 post items).

As soon as you have user submittable data that finds its way into a JSON field this becomes a pretty easy way to DOS a site while staying well within any kind of query limit (#1571).

molomby:
More fun here...
[8.14. JSON Types](https://www.postgresql.org/docs/12/datatype-json.html)

[...]

PR #2574, opened 2020-03-23
Replaced knex `.jsonb()` with `.json()` for better db compatibility

LiamAttClarke:
**Related Knex documentation:** http://knexjs.org/#Schema-json, http://knexjs.org/#Schema-jsonb

Although the Knex documentation claims that `.jsonb()` works similarly to `.json()`, it does not offer backwards compatibility for databases that do not support JSONB.

`.json()` uses the built-in JSON type for PostgreSQL, MySQL and SQLite, defaulting to a text column in older versions or in unsupported databases.


molomby:
Huh. Using the native type and falling back to text is indeed the behaviour we were intending. As you note, the knex docs are pretty ambiguous... they imply this should happen for both `json` and `jsonb` fields.

I'm gonna do some testing, make sure this works. We'll want to update the `Location`, `OEmbed` and `Unsplash` field types also. They currently use the knex `jsonb` type.

(Long term we should build a proper json type and extend it, as per #315, but that's a problem for another day.)

molomby:
Just blindly using `jsonb` isn't ideal at all, for the reasons you mention but using the Knex `json` function results in the string based `json` field type on Postgres. Unfortunately, we're prioritising pg support at the moment and `jsonb` fields offer some significant advantages there.

PR #2575, opened 2020-03-23
Create tables

MadeByMike:
Adds a cli script `keystone create-tables` that will initialise an empty database with the Knex Adapter.


packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+      } else {
+        console.log(e);
+      }
+      errors = true;
+    }
+  });
+  if (!errors) {
+    spinner.succeed(chalk.green.bold(`Tables created`));
+  }
+  process.exit(0); // Stop Keystone
```
irwin-r: IMO the process exit code should be non-zero when we experience an error relating to table creation.

MadeByMike: Yeah sure!

MadeByMike:
```suggestion
    spinner.succeed(chalk.green.bold(`Tables created`));
    process.exit(0);
  }
  process.exit(1);
```

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+  // Allow the spinner time to flush its output to the console.
+  await new Promise(resolve => setTimeout(resolve, 100));
+  const { keystone } = require(path.resolve(entryFile));
+  await keystone.connect(); // Need to do _createTables post connect so awaiting connect
+  keystone._consolidateRelationships();
+  let errors = false;
+  await asyncForEach(Object.values(keystone.adapters), async adapter => {
+    if (!adapter._createTables) {
+      spinner.info(chalk.yellow.bold(`create-tables is only required for KnexAdapter`));
+      return null;
```
irwin-r: /nit -- given that we're electing to not return a value at any other point in the application, this should probably just be a `return;`.

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+    try {
+      console.log('Creating tables...');
+      await adapter._createTables();
+    } catch (e) {
```
irwin-r: /nit -- Bob Martin insists that we call this `error` 😋

MadeByMike: I don't know or care who Bob is

packages/keystone/bin/commands/create-tables.js
MadeByMike: Sure

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+      if (e.message.includes('already exists')) {
+        spinner.fail(chalk.red.bold(`Table already exists`));
+        console.log('Create tables should be used on an empty database');
```
irwin-r: /nit -- this could be a `console.warn` as it will output to `stderr` given the contents of the message.

packages/keystone/bin/commands/create-tables.js
MadeByMike:
```suggestion
      return;
```

irwin-r:
```suggestion
      return;
```

packages/keystone/bin/commands/create-tables.js
irwin-r:
```suggestion
        console.warn('Create tables should be used on an empty database');
```

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+        console.log(e.message);
```
irwin-r:
```suggestion
        console.warn(e.message);
```

packages/keystone/bin/commands/create-tables.js
MadeByMike: Sure

MadeByMike:
```suggestion
        console.warn('Create tables should be used on an empty database');
```

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+      } else {
+        console.log(e);
```
irwin-r:
```suggestion
        console.error(e);
```

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+        spinner.fail(chalk.red.bold(`Table already exists`));
+        console.warn('Create tables should be used on an empty database');
+        console.warn(e.message);
+      } else {
+        console.error(e);
+      }
+      errors = true;
+    }
+  });
+    spinner.succeed(chalk.green.bold(`Tables created`));
```
irwin-r: @MadeByMike I might be misreading it, but I think we have a syntax error here now?

We'd also be unconditionally doing a `process.exit(0);`?

packages/keystone/bin/commands/create-tables.js
```diff
@@ -0,0 +1,56 @@
...
+const chalk = require('chalk');
+const { DEFAULT_ENTRY } = require('../../constants');
+const { getEntryFileFullPath } = require('../utils');
+const { asyncForEach } = require('@keystonejs/utils');
+const createTables = async (args, entryFile, spinner) => {
+  // Allow the spinner time to flush its output to the console.
+  await new Promise(resolve => setTimeout(resolve, 100));
+  const { keystone } = require(path.resolve(entryFile));
+  await keystone.connect(); // Need to do _createTables post connect so awaiting connect
+  keystone._consolidateRelationships();
```
timleslie:
```suggestion
```
We don't need to do this, it's already done in `keystone.connect()`

PR #2577, opened 2020-03-24
(WIP) Added MySQL support to KnexAdapter

LiamAttClarke:
This PR serves as a reference for the changes needed to get MySQL working. I recommend refactoring the KnexAdapter and moving all SQL query building to a separate class (perhaps called QueryBuilder) that is injected on construction of the KnexAdapter. This approach will let the Keystone team as well as end users add additional databases as needed.

The following PR is also required to get File fields working: https://github.com/keystonejs/keystone/pull/2574


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -203,10 +224,20 @@ class KnexAdapter extends BaseKeystoneAdapter {
...
+      return this.knex.raw(`DROP TABLE IF EXISTS ${tables} CASCADE`);
+    } else if (this.isClientMySQL) {
+      const tableNames = Object.values(this.listAdapters).map((listAdapter) => listAdapter.tableName);
+      return this.knex.transaction(async (transaction) => {
+          await transaction.raw('SET FOREIGN_KEY_CHECKS = 0');
+          await transaction.raw(`DROP TABLE IF EXISTS ${tableNames.join(',')}`);
+          return transaction.raw('SET FOREIGN_KEY_CHECKS = 1');
+      });
+    }
+    return Promise.reject(new Error(`KnexAdapter.dropDatabase does not support the '${this.client}' client.`));
```
irwin-r: This should be a `throw`.

```suggestion
    throw new Error(`KnexAdapter.dropDatabase does not support the '${this.client}' client.`);
```

Nikodermus:
@timleslie Seems like #2000 has completed its way, anything blocking this?

timleslie:
@Nikodermus This effort is being lead by @LiamAttClarke. Liam, have you had a chance to loop back and work on this since #2000 landed?

rmalvarado1805:
First of all I would like to thank you all, for the effort being done here.

I have some comments regarding the code that is being implemented in this PR:

- I don't think adding a check, in every method for the client is a very good idea and the fact that the code only considers MySQL and PostgreSQL. What about SQLite or MS-SQL? do we need to add a new check for every other database engine? (_I don't think that is a good idea_)

- Inside the MySQL implementations the code uses the **LIKE** operator. Why not use that by default regardless of the client? That way you can eliminate the check for the client, because SQLite, MS-SQL, MySQL, and PostgreSQL all support it.

- Now, the operators (**~\***, **!~\***, **~**, **!~**) used by the original authors are intended to match a regular expression, which are only available in PostgreSQL, I assume there has to be a good reason for this, maybe there are complex matching patterns that wouldn’t be possible using the **LIKE** operator, (_I haven't explored the code enough to be sure_) but if we want to keep using regular expressions we need to explore other solutions like injecting a dictionary of operators corresponding to the client in use (_the equivalent in MySQL and SQLite would be **REGEXP**_).

**Conclusion**: I think the cleanest way to implement this is using the **LIKE** and replace the regular expression for the corresponding **LIKE** expressions, the considerations to take into account are:

- Case sensitive or insensitive matching
- Is the **LIKE** operator is a good fit for the current matching patterns implemented?.

Those are my two cents. I would like to know what you do think and how can I contribute either by sharing some code samples of my ideas or submitting a new PR.

LiamAttClarke:
> First of all I would like to thank you all, for the effort being done here.
> 
> I have some comments regarding the code that is being implemented in this PR:
> 
> * I don't think adding a check, in every method for the client is a very good idea and the fact that the code only considers MySQL and PostgreSQL. What about SQLite or MS-SQL? do we need to add a new check for every other database engine? (_I don't think that is a good idea_)
> * Inside the MySQL implementations the code uses the **LIKE** operator. Why not use that by default regardless of the client? That way you can eliminate the check for the client, because SQLite, MS-SQL, MySQL, and PostgreSQL all support it.
> * Now, the operators (**~***, **!~***, **~**, **!~**) used by the original authors are intended to match a regular expression, which are only available in PostgreSQL, I assume there has to be a good reason for this, maybe there are complex matching patterns that wouldn’t be possible using the **LIKE** operator, (_I haven't explored the code enough to be sure_) but if we want to keep using regular expressions we need to explore other solutions like injecting a dictionary of operators corresponding to the client in use (_the equivalent in MySQL and SQLite would be **REGEXP**_).
> 
> **Conclusion**: I think the cleanest way to implement this is using the **LIKE** and replace the regular expression for the corresponding **LIKE** expressions, the considerations to take into account are:
> 
> * Case sensitive or insensitive matching
> * Is the **LIKE** operator is a good fit for the current matching patterns implemented?.
> 
> Those are my two cents. I would like to know what you do think and how can I contribute either by sharing some code samples of my ideas or submitting a new PR.

I completely agree with you. As my description mentions, this PR serves as a reference for what changes are needed to get MySQL working.

I think we should break out all the SQL building into a separate class and override methods when needed for particular database engines.

LiamAttClarke:
> @Nikodermus This effort is being lead by @LiamAttClarke. Liam, have you had a chance to loop back and work on this since #2000 landed?

I haven't had the opportunity to work on this further. I'm backed up with unrelated work for the foreseeable future.

rmalvarado1805:
After researching a bit more about how to accomplish a wider SQL support, I found out that the problem is more complicated than what I expected. the following are my findings:

- **LIKE** operator in MySQL is case insensitive. [[1](https://dev.mysql.com/doc/refman/8.0/en/pattern-matching.html)]
- **LIKE** operator in SQLite is case insensitive by default. [[2](https://www.sqlite.org/lang_expr.html)]
- **LIKE** operator in PostgreSQL is case sensitive. [[3](https://www.postgresql.org/docs/8.0/functions-matching.html)]

Given the previous statements, we can't use the LIKE operator interchangeably as they behave differently depending on the engine.

- There is an **ILIKE** for PostgreSQL which is the case insensitive version of **LIKE**. [[3](https://www.postgresql.org/docs/8.0/functions-matching.html)]
- To make a case sensitive search in MySQL we need to prepend the **LIKE** operator with the **BINARY** operator. [[1](https://dev.mysql.com/doc/refman/8.0/en/pattern-matching.html)]
- The **BINARY** operator is not available in SQLite. To make this possible in SQLite, we need to run a **PRAGMA** query before executing the actual query to change the behavior of the **LIKE** operator. [[4](https://www.sqlite.org/pragma.html#pragma_case_sensitive_like)]

And I'm sure there are more inconsistencies across the SQL engines, that I haven't got to.

**Conclusion**:
The suggestions I made earlier are I think are not a good fit due to the new information I have found described above.
My new suggestion would be to create an adapter for each SQL engine exposing the same public API, but with their corresponding implementations, the only drawback I see is the fact that any changes to the public API, and new functionality would require updating every adapter, and this will increase the maintainability cost. But there might be a way to share common functionality across engines (_orderingConditions_ for example, works as expected for every engine). 

let me know what you think, and which approach adheres more to the project goals.

timleslie:
Closing this for now as MySQL support in the `knex` adapter is not something we intend to support. MySQL support may be introduced in the `Prisma` adapter in the future.

PR #2580, opened 2020-03-24
noviny review notes

Noviny:
@timleslie asked me to review some text and honestly it's easier if I have a PR for it so I'm making one and you can't stop me


Noviny:
Random clash thing, all examples don't explain where `keystone`, `Text` etc variables come from. Can we put somewhere where or how I import things?

> there is only one relationship between the lists

You say this twice and I'm not sure why I care that this be the case.

I do not know the word cardinality properly making me unsure of it as the heading for a title. I know the first sentence defines it, I'm just... unsure. I suspect this is a fight I have already missed though.

The section on cardinality with all the possible configurations with examples is *chef kiss* though.

Overall really good. Most things I disliked I ended up liking after reading the whole docs

timleslie:
Thanks for the feedback, very useful!

PR #2587, opened 2020-03-24
Allow providers to be lazily instantiated

jesstelford:


jesstelford:
Closing as this problem can be solved with a better error message 👍

PR #2599, opened 2020-03-29
use manypkg run over yarn --cwd

Noviny:
Small switch with two reasons:

1) I wanted to dogfood using manypkg run, and this seemed a good way to do it.
2) I *think* this gives us a better pattern for how to quickly and easily run scripts in a chosen package. Here's its main docs selling point:

```
manypkg run <partial package name or partial directory> <script>
```

-> run the script in a matching package (if one is found)

This solves problems with either long dirs to run the scripts, or going off full package names and leaving you to need to type `@keystonejs/keystone` to execute a script in it. The shortest text with only one match is fine.

There's a [new issue I made](https://github.com/Thinkmill/manypkg/issues/58) based on using it in keystone, but I think it's fine as is.

CONTROVERSIAL THING: I added `rune` as a "clever" script, so you can do `yarn rune` instead of `yarn manypkg run` when leverage the new script. Your mileage may vary on whether you think this is a great or terrible idea.


package.json
```diff
@@ -8,10 +8,11 @@
   "repository": "https://github.com/keystonejs/keystone",
   "homepage": "https://github.com/keystonejs/keystone",
   "scripts": {
+    "rune": "manypkg run",
```
timleslie: I think I'm -1 on this, mostly because it adds one more layer of abstraction for people to understand without really giving us any significant wins.

Noviny: Since one of the complaints about was amount of typing. Will remove but would also appreciate an alternative.

Noviny:
All the magic has been removed @timleslie

PR #2600, opened 2020-03-29
[WIP] add changelog links

Noviny:
So, I started this process of adding changelog links, got to auth-password. Auth-password has a readme written almost entirely to be displayed on the docs site, and doesn't really reference itself as a package.

This left me unsure about just adding in the badge to those packages, so asking for a thumbsup, thumbsdown, or a call for nuance before I finish this.

Cheers!


timleslie:
How do the badges look when they're rendered in contexts other than Github? E.g. on the docs site or npm website? I think if they work and look fine in those contexts we should continue to add them to all packages.

Noviny:
It’s just markdown displaying an image: https://deploy-preview-2600--v5keystonejs.netlify.com/keystonejs/auth-passport/ so just renders.

Will finish this up tonight then.

MadeByMike:
Yep I say go ahead. It does look odd but the package readme should be the API docs for that package. We probably need to adapt the content.

JedWatson:
I think there's a case to be made for excluding these from readmes that are rendered as content pages on the website, because it's seriously inconsistent and sometimes also unclear which package is being documented. But that's all a website issue, on a standalone basis this get a 👍 from me

Noviny:
Jed no, this PR wasn’t done yet. Will get the remaining packages up soon(tm)

PR #2601, opened 2020-03-29
[WIP] Added workaround for crash/error on file upload

### LINKED ISSUE #2101: Cloudinary field doesn't work on Node 13.3.2
# Bug report

Not sure if this is upstream or not..

Trying to use the cloudinary field on Node 13.3.2. It gives the error `(node:50514) [DEP0135] DeprecationWarning: ReadStream.prototype.open() is deprecated` in the console and a 'nested error' in the Admin UI.

Reverting back to node 12.13 fixes it for now

## System information

- OS: Mac OS Catalina

Vultraz:
Fixes #2101.

Currently, `apollo-server-express` depends on `apollo-server-core` which depends on `graphql-upload` ^8.0.2. `graphql-upload` 9.0.0 updated its own `fs-capacitor` dependency to a newer version that supports Node 13, but `apollo-server-core` hasn't updated its own `graphql-upload` dependency yet. Solve this by just forcing a compatible version of `graphql-upload` (currently the latest 10.0.0) to be used.

This also fixes your server crashing when trying to upload a file using the local file adapter (the cause was the same) on Node 13.

### DIFF
--- a/package.json
+++ b/package.json
@@ -202,5 +202,8 @@
       "!packages/fields/types.js",
       "!packages/mongo-join-builder/examples/**"
     ]
+  },
+  "resolutions": {
+    "graphql-upload": "^10.0.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5936,13 +5936,12 @@ builtins@^1.0.3:
   resolved "https://registry.yarnpkg.com/builtins/-/builtins-1.0.3.tgz#cb94faeb61c8696451db36534e1422f94f0aee88"
   integrity sha1-y5T662HIaWRR2zZTThQi+U8K7og=
 
-busboy@^0.2.14:
-  version "0.2.14"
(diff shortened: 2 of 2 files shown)


Vultraz:
Ok, it just occurred to me - this should probably be in a package so it can actually be published 😬

timleslie:
Is this a known upstream issue? Can we perhaps escalate it to `apollo-server` so that they can update their dependencies? Solving the root problem may end up being easier than getting a workaround to play nicely?

Vultraz:
There is an apollo-server PR to bump their graphql-upload dependency, but that hasn't been merged yet despite being opened since the end of January. Seems they're a little concerned about breaking changes: https://github.com/apollographql/apollo-server/pull/3723#issuecomment-590291980

Vultraz:
I'm going to close this since #2780 documents the issue and gives this fix as a workaround. I don't think it's something we can deploy on this end. I'll keep an eye on the apollo packages and update them if the dep bumps ever lands in 2.x (or 3).

PR #2607, opened 2020-03-30
Made meetup demo functional again

Vultraz:
It was completely broken when I launched it, which is 100% my fault. I didn't test #2324 before opening it. 😬 

Fixed everything, including a few other issues I found (Dashboard link disappearing, React error about a missing `key`, incorrectly-forwarded ref, and an error with `fetch`).

Also refactors out the last use of the `withApollo` HOC, so we don't depend on `@apollo/react-hoc` anymore. And I did make sure to thoroughly test it. 😝

### DIFF
--- a/demo-projects/meetup/site/lib/initApollo.js
+++ b/demo-projects/meetup/site/lib/initApollo.js
@@ -19,8 +19,7 @@ function create(initialState, req) {
     link: createUploadLink({
       uri: `${serverUrl}/admin/api`, // Server URL (must be absolute)
       credentials: 'same-origin', // Additional fetch() options like `credentials` or `headers`
-      // Use fetch() polyfill on the server
-      fetch: !process.browser && fetch,
+      fetch,
       headers: req && req.headers,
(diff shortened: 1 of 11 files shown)


demo-projects/meetup/site/lib/initApollo.js
```diff
@@ -19,8 +19,7 @@ function create(initialState, req) {
     link: createUploadLink({
       uri: `${serverUrl}/admin/api`, // Server URL (must be absolute)
       credentials: 'same-origin', // Additional fetch() options like `credentials` or `headers`
-      // Use fetch() polyfill on the server
-      fetch: !process.browser && fetch,
+      fetch,
```
Vultraz: Made this use `fetch` unconditionally since that's what the Blog demo did. I'm not sure why it was made conditional, but it was causing an error in-browser. See https://github.com/jaydenseric/apollo-upload-client/issues/167 I figured it couldn't hurt to match what the Blog does.

PR #2618, opened 2020-03-31
Document custom access controls

Vultraz:
Also includes some more accurate default config values for `Keystone`.

I left out any mentions of declarative access since I can't figure out how that's supposed to work on custom schema. 🤔


MadeByMike:
Looks like https://github.com/keystonejs/keystone/pull/2625 was based off this... I'm going to merge #2625 then compare what's left.

Vultraz:
@MadeByMike fixed a merge error, should be good to go

PR #2622, opened 2020-04-01
Docs: Cleanup table of contents

jossmac:
- resolve out-of-sync IDs issue (`github-slugger`)
- let ToC items wrap, without looking like separate headings

Replaces #2619

PR #2623, opened 2020-04-01
Update @sindresorhus/slugify

maryam-mv:
Seeing some weird behaviour in this dependency @sindresorhus/slugify of the @keystone/adapter-knex package. At the current specified version 0.6.0 the slugify function processes “HELLO3 Website” to “hell_o3_website”. That’s the minimum version in the package.json for that package https://github.com/keystonejs/keystone/blob/master/packages/adapter-knex/package.json. As of v0.10.0 it processes “HELLO3 Website” as “hello_3_website”. Take a look here and switch the version of the dependency https://codesandbox.io/s/youthful-blackburn-wvk5n

### DIFF
--- a/.changeset/slow-bikes-wonder.md
+++ b/.changeset/slow-bikes-wonder.md
@@ -6,3 +6,5 @@
 ---
 
 Updated @sindresorhus/slugify to fix a problem where it was producing unexpected output, eg. adding unexpected underscores: 'NAME1 Website' => 'nam_e1_website'. The slugify output for db name may be different with this change. For the above example, the output will now be 'name_1_website' for the same string.
+
+If your database name changes unexpectedly, add an environmental variable called `DATABASE_URL` with a full path to the database.
--- a/.changeset/slow-bikes-wonder.md
+++ b/.changeset/slow-bikes-wonder.md
@@ -1,0 +1,9 @@
+---
+'@keystonejs/adapter-knex': major
+'@keystonejs/adapter-mongoose': major
+'create-keystone-app': major
+'@keystonejs/fields': major
+'@keystonejs/website': major
+---
+
+Updated @sindresorhus/slugify to fix a problem where it was producing unexpected output, eg. adding unexpected underscores: 'NAME1 Website' => 'nam_e1_website'. The slugify output for db name may be different with this change. For the above example, the output will now be 'name_1_website' for the same string.
--- a/.changeset/slow-bikes-wonder.md
+++ b/.changeset/slow-bikes-wonder.md
@@ -7,4 +7,4 @@
 
 Updated @sindresorhus/slugify to fix a problem where it was producing unexpected output, eg. adding unexpected underscores: 'NAME1 Website' => 'nam_e1_website'. The slugify output for db name may be different with this change. For the above example, the output will now be 'name_1_website' for the same string.
 
-If your database name changes unexpectedly, add an environmental variable called `DATABASE_URL` with a full path to the database.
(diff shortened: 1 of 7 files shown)


jossmac:
Somewhat related to #2622. I swapped out `@sindresorhus/slugify` for `github-slugger` in the docs.

At the very least, we should have a single slug solution within Keystone itself, right?

Vultraz:
The changeset patch levels should be "patch", not "major":
```
'@keystonejs/adapter-knex': patch
'@keystonejs/adapter-mongoose': patch
'create-keystone-app': patch
'@keystonejs/fields': patch
'@keystonejs/website': patch
```

JedWatson:
Given that we're seeing different slugs being generated with the new version (e.g default path for a database from the project name) I'm in favour of making this a major change. It has the potential to break users projects and that's pretty dangerous. 

Maybe not all the packages need to be major though? would have to review the implications of each usage in detail

.changeset/slow-bikes-wonder.md
```diff
@@ -0,0 +1,9 @@
+---
+'@keystonejs/adapter-knex': major
+'@keystonejs/adapter-mongoose': major
+'create-keystone-app': major
+'@keystonejs/fields': major
+'@keystonejs/website': major
```
MadeByMike, line 6:
```suggestion
'@keystonejs/adapter-knex': major
'@keystonejs/adapter-mongoose': major
'create-keystone-app': major
'@keystonejs/fields': major
```

.changeset/slow-bikes-wonder.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystonejs/adapter-knex': major
+'@keystonejs/adapter-mongoose': major
+'create-keystone-app': major
+'@keystonejs/fields': major
+---
+
+Updated @sindresorhus/slugify to fix a problem where it was producing unexpected output, eg. adding unexpected underscores: 'NAME1 Website' => 'nam_e1_website'. The slugify output for db name may be different with this change. For the above example, the output will now be 'name_1_website' for the same string.
```
MadeByMike, line 8:
```suggestion
Updated @sindresorhus/slugify to fix a problem where it was producing unexpected output, eg. adding unexpected underscores: 'NAME1 Website' => 'nam_e1_website'. The slugify output for db name may be different with this change. For the above example, the output will now be 'name_1_website' for the same string.

If your database name changes unexpectedly, add an environmental variable called `DATABASE_URL` with a full path to the database.
```

MadeByMike:
> The changeset patch levels should be "patch", not "major":
> 
> ```
> '@keystonejs/adapter-knex': patch
> '@keystonejs/adapter-mongoose': patch
> 'create-keystone-app': patch
> '@keystonejs/fields': patch
> '@keystonejs/website': patch
> ```

It's major because database URLs may change meaning it's a breaking change.

Vultraz:
Ah, I see. My bad 😬

.changeset/slow-bikes-wonder.md
```diff
@@ -0,0 +1,10 @@
...
+
+If your database name changes unexpectedly, add an environmental variable called `DATABASE_URL` with a full path to the database.
```
MadeByMike, line 10:
```suggestion
If your database name changes unexpectedly, add an environment variable called `DATABASE_URL` with a full path to the database.
```

MadeByMike:
> Ah, I see. My bad 😬

I totally did the same thing at first. "Bumping a dep? That's not breaking!"

PR #2627, opened 2020-04-01
Document missing argument in custom resolver functions

jordanoverbye:
There is a 5th argument passed into custom resolver functions that isn't documented. [See here](https://github.com/keystonejs/keystone/blob/master/packages/keystone/lib/providers/custom.js#L93)

### DIFF
--- a/.changeset/lazy-elephants-repeat.md
+++ b/.changeset/lazy-elephants-repeat.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/lazy-elephants-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Updated documenation for the last argument in custom resolver functions
```
timleslie, line 5:
```suggestion
Updated documentation for the last argument in custom resolver functions.
```

PR #2629, opened 2020-04-01
Docs/Fields: list each fields package and their corresponding fields in one place

Vultraz:
For easy reference.


MadeByMike:
This section is the API docs and Readme for `@keystone/fields` I think it might be a little odd to list other field packages here...

But I get why, any other thoughts on how we might surface this with out loosing the focus on the readme of the package documenting the API for THAT package?

Vultraz:
🤔 I was thinking this page served as more of a general field API overview, since it already discusses common field config options (not unique to the `fields` package), and the native type aliases (which are handled by the `List` class anyway). Most people are probably going to visit this first just to get a sense of the whole fields API, IMO, not for the `fields` package specifically.

Vultraz:
Another option would be to add individual pages for each package, then split some of the general field API stuff into its own page.

Vultraz:
@MadeByMike any more thoughts on this, or should I close it?

molomby:
I both share @MadeByMike's concern and really like having a broader index like this. At some point in the not-too-distant future I hope we'll need an even more complete list with links to 3rd party field type as well. Is there any good place in the website for an index like this?

I suggest we merge this content for now, with an eye to restructuring it down the track.

Vultraz:
Closing this since #3140 supersedes it.

PR #2631, opened 2020-04-01
Docs: Code Snippet Meta

jossmac:
Introduce code snippet meta. This moves the code and pre styles from the `<Content />` wrapper to the MDX renderer.

We take advantage of the `metastring` provided by `MDXProvider` from "@mdx-js/react", which takes arguments after the language declaration on a code block and passes them to the component as props e.g. (pretend they're back ticks)

```
'''json title=package.json allowCopy=false
{
  "scripts": {
    "start": "node server.js"
  }
}
'''
```

> **Note:** Meta properties are ignored in environments like GitHub and NPM so will not effect rendering.

Supported meta properties. Willing to change these to something less verbose, like `copy`/`lang`, but prefer explicit:

| Key          | Type      | Description                                                                    |
| :----------- | :-------- | :----------------------------------------------------------------------------- |
| title        | `string`  | Default `undefined`. Presents a title above the code snippet                   |
| allowCopy    | `boolean` | Default `true`. Display a button that allows the user to copy the code snippet |
| showLanguage | `boolean` | Default `true`. Display a prettified label of the snippet language             |

An example of the updated UI:

image

I'll need to come through in another PR and cleanup any weirdness that arises from this change. There's some snippets where displaying the language or copy button doesn't really make sense.

* * *

Some unrelated changes were made in this PR, and can be found under the commit messages:
- [fix bug with focus on heading anchors](https://github.com/keystonejs/keystone/pull/2631/commits/f781f70548d987cd7d53cf98aa9adb7e2abe3fb7)
- [filter out paragraphs that contain badges](https://github.com/keystonejs/keystone/pull/2631/commits/c80540e9cecf60bec7f2a2993bbfd30016c9ac03)

### DIFF
--- a/.changeset/poor-readers-care.md
+++ b/.changeset/poor-readers-care.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/website': minor
+---
+
+Introduce meta data to code blocks
--- a/.changeset/poor-readers-care.md
+++ b/.changeset/poor-readers-care.md
@@ -2,4 +2,4 @@
 '@keystonejs/website': minor
 ---
 
-Introduce meta data to code blocks
+Introduced meta data to code blocks.
--- a/website/package.json
+++ b/website/package.json
@@ -58,6 +58,7 @@
     "react-helmet": "^5.2.0",
     "rehype-slug": "^2.0.3",
     "underscore.string": "^3.3.5",
-    "unist-util-visit": "^1.4.0"
+    "unist-util-visit": "^1.4.0",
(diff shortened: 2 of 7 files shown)


.changeset/poor-readers-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/website': minor
+---
+
+Introduce meta data to code blocks
```
MadeByMike, line 5:
```suggestion
Introduced meta data to code blocks.
```

PR #2633, opened 2020-04-01
Fixed state updates on Tooltip when unmounted (fixes #2419)

### LINKED ISSUE #2419: Admin UI: state update on unmounted component
Been seeing this every so often recently. Haven't tracked down the exact cause yet.
image

Vultraz:
Converted `Tooltip` to a functional component and the problem went away. Also removed the `immediatelyHide` and `immediatelyShow` states in the process since they weren't being used (they weren't the cause of the problem, though).

### DIFF
--- a/packages/arch/packages/tooltip/src/index.js
+++ b/packages/arch/packages/tooltip/src/index.js
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { Component, createRef, Fragment } from 'react';
+import { Fragment, useState, useRef, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import flushable from 'flushable';
 import styled from '@emotion/styled';
@@ -24,7 +24,7 @@ const TooltipElement = styled.div({
   zIndex: 2,
 });
 
-let TooltipPositioner = props => {
+const TooltipPositioner = props => {
   return createPortal(
     <Popper
       referenceElement={props.targetNode}
@@ -48,7 +48,6 @@ let TooltipPositioner = props => {
 // ==============================
 
 const LISTENER_OPTIONS = { passive: true };
-const NOOP = () => {};
 
 let pendingHide;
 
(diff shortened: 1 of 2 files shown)


packages/arch/packages/tooltip/src/index.js
```diff
@@ -57,118 +56,122 @@ const showTooltip = (fn, defaultDelay) => {
...
-    }
-    if (this.props.hideOnMouseDown && this.ref.current) {
-      this.ref.current.addEventListener('mousedown', this.cancel, LISTENER_OPTIONS);
-    }
-    if (this.props.hideOnKeyDown) {
-      document.addEventListener('keydown', this.cancel, LISTENER_OPTIONS);
-    }
+    let cancelPendingSetState = () => {};
+
+    const cancel = () => {
```
irwin-r, line 98: It may be a bit more succinct/readable to pull these event handlers outside of the `useEffect` and limit the body of the `useEffect` to registering + removing the event listeners.

packages/arch/packages/tooltip/src/index.js
```diff
@@ -57,118 +56,122 @@ const showTooltip = (fn, defaultDelay) => {
...
-    if (!this.state.isVisible) {
-      return;
-    }
-    if (this.props.hideOnMouseDown && this.ref.current) {
-      this.ref.current.removeEventListener('mousedown', this.cancel, LISTENER_OPTIONS);
-    }
-    if (this.props.hideOnKeyDown) {
-      document.removeEventListener('keydown', this.cancel, LISTENER_OPTIONS);
-    }
+Tooltip.defaultProps = {
```
irwin-r, line 172: The React team is deprecating `defaultProps`:

https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md

It may be worthwhile pulling these in as default argument values.

MadeByMike: We've got this in a bunch of places... I agree but I think we could defer this to another PR for now.

packages/arch/packages/tooltip/src/index.js
Vultraz: I could move `cancelPendingSetState` out, but everything else uses props/state.

packages/arch/packages/tooltip/src/index.js
Vultraz: Yes, I'll handle that separately at some point.

packages/arch/packages/tooltip/src/index.js
irwin-r: Sorry, to clarify, I didn't mean outside of the component, but outside of the `useEffect` method.

I think `cancel`, `handleMouseEnter` and `handleMouseLeave` should all live above the `useEffect` method.

Vultraz: I think that makes things just a little messier since then `cancelPendingSetState` has to be managed with `useRef` as well. 🤔

PR #2640, opened 2020-04-02
Docs: improve search experience

jossmac:
- support focus on "/" press
- update styles

image

### DIFF
--- a/website/src/components/Search.js
+++ b/website/src/components/Search.js
@@ -29,6 +29,7 @@ export const Search = () => {
     let keyboardHandler = event => {
       if (allowKeyboardRequest(event)) {
         inputRef.current.focus();
+        // prevent the "/" character from making it into the input
         event.preventDefault();
       }
     };
@@ -137,7 +138,6 @@ const HiddenLabel = props => (
   />
 );
 
-// label for screen-readers
 const SearchIcon = () => (
   <svg
     fill="currentColor"
--- a/website/src/components/Search.js
+++ b/website/src/components/Search.js
@@ -4,15 +4,41 @@ import { jsx } from '@emotion/core';
 import { useEffect, useRef } from 'react';
 import { navigate } from 'gatsby';
 import { Input } from '@arch-ui/input';
+import { borderRadius, colors } from '@arch-ui/theme';
 
 import { addCallback } from '../utils/async-load-search';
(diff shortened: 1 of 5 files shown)


MadeByMike:
Nice

website/src/components/Search.js
```diff
@@ -4,15 +4,42 @@ import { jsx } from '@emotion/core';
...
 import { addCallback } from '../utils/async-load-search';
-import { algoliaStyles } from '../utils/algolia-styles';
 
 const searchId = 'algolia-doc-search';
 
+function allowKeyboardRequest(event) {
+  if (event.key !== '/') {
+    return false;
+  }
+  if (['INPUT', 'TEXTAREA'].includes(document.activeElement.nodeName)) {
```
jordanoverbye, line 17: Maybe this could just be

```
if (event.key !== '/' || ['INPUT', 'TEXTAREA'].includes(document.activeElement.nodeName)) {
  return false;
}
```

website/src/components/Search.js
```diff
@@ -36,15 +63,97 @@ export const Search = () => {
...
+  <div
+    css={{
+      position: 'relative',
+
+      '.algolia-autocomplete': {
+        width: '100%',
+      },
+      '.ds-dropdown-menu': {
+        boxShadow: `rgba(9, 30, 66, 0.31) 0px 0px 1px, rgba(9, 30, 66, 0.25) 0px 4px 8px -2px`,
+        borderRadius: borderRadius,
```
jordanoverbye, line 99: Maybe this could just be the property value shorthand:

```
{
  ....
  borderRadius,
  ....
}
```

website/src/components/Search.js
```diff
@@ -36,15 +63,97 @@ export const Search = () => {
...
+
+        // hide the arrow
+        '::before': {
+          display: 'none',
+        },
+
+        // remove border and adjust radius on inner dialog wrapper
+        '[class^=ds-dataset-]': {
+          border: 0,
+          borderRadius: borderRadius,
```
jordanoverbye, line 111: Same as above

website/src/components/Search.js
jossmac: It definitely could be, but I think this is clearer

PR #2644, opened 2020-04-02
Fix Relationship field error when passed null value

### LINKED ISSUE #2642: React crashes when Relationship field is passed null value
An item in User List A has a to-many relationship with an item in Task List B. That is to say that a User can have many tasks and a Task can also have many users.

However, if `access.read` on the Relationship field In Task List B is a function that doesn't give a User read access to that field (for example, to see who the users who have been given that Task) then the React throws the following error:

image

I think this because in the [Relationship field view](https://github.com/keystonejs/keystone/blob/master/packages/fields/src/types/Relationship/views/Field.js), <LinkToRelatedItems /> is not expecting a null value but an empty array. Simply adding `value = value || [];` on line 59 seems to fix the issue.

justintemps:
Fixes #2642 .

### DIFF
--- a/packages/fields/src/types/Relationship/views/Field.js
+++ b/packages/fields/src/types/Relationship/views/Field.js
@@ -62,6 +62,8 @@ function LinkToRelatedItems({ field, value }) {
   let isDisabled = false;
   let label;
   let link = `${adminPath}/${refList.path}`;
+  value = value || [];
+
   if (many) {
     label = 'View List of Related Items';
 


Vultraz:
I think it would be better to do this in the main `RelationshipField` component so that all child components get the fallback. Suggest doing
```js
value = [],
```
on line 164 and removing the fallbacks on line 207 and 219 (ie, just change them to `value.concat`)

PR #2647, opened 2020-04-03
Admin UI: Handle non-admin users better

zenflow:
This PR fixes two minor issues handling unauthorized/non-admin users:
- After signing in, unauthorized users are left in signin page.
  - Now after signing in, unauthorized users are redirected to the `/` path (the assumed public home)
- When trying to access a private page, unauthorized users are redirected to signin page, even if they're signed in already
  - Now when trying to access a private page, unauthorized users are redirected to signin page only if they're signed out, otherwise they see the following "Access Denied" page. The styles and layout are the same as SignoutPage & SigninPage.
image

### DIFF
--- a/.changeset/metal-ties-camp.md
+++ b/.changeset/metal-ties-camp.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': minor
+---
+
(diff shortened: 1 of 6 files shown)


zenflow:
See #2657

PR #2654, opened 2020-04-05
Docs: blockquote variants

jossmac:
Parse mdx blockquotes and apply variant styles:
- standard
- tip
- note
- important

image


JedWatson:
I really like this, as a direction.

My feedback, read this like a stream of consciousness as I reviewed the changes:

1. Not sure about the semantics of renaming those "hints" to "tips", doesn't feel like that makes the content clearer (ref. changes to `docs/api/access-control.md` and `docs/guides/graphql-philosophy.md`) -- kind of want to get a second opinion on this, maybe @timleslie?

2. Notes should maybe be a different color than tips? This got me thinking about "why don't we use green"

3. Colours generally convey semantic meaning, "important" being orange made me feel like it was a "warning". Maybe that's correct? depends on the content of the important I guess.

4. With that in mind, a red "danger" or "warning" could be really helpful to draw attention to really important notes, e.g calling out the consequences of the `dropDatabase` option

I think it's worth a discussion with a few people to land on how we should really design this feature for use but generally it's got a lot of upside 👍

jossmac:
> 1. Not sure about the semantics of renaming those "hints" to "tips"

These can probably be removed entirely. If it should be highlighted, make it a "note", otherwise just let it be content.

> 2. Notes should maybe be a different color than tips? This got me thinking about "why don't we use green"

See above

> 3. Colours generally convey semantic meaning, "important" being orange made me feel like it was a "warning". Maybe that's correct? depends on the content of the important I guess.

That is intentional. Wherever we use an "important" note should be to warn consumers about potentially unexpected (from their point-of-view) behaviour e.g.

image

4. With that in mind, a red "danger" or "warning" could be really helpful to draw attention to really important notes, e.g calling out the consequences of the dropDatabase option

See above

jesstelford:
magic

🚢

PR #2664, opened 2020-04-06
Added a `where: { key }` argument to List meta queries

### LINKED ISSUE #1279: Add where clause to List Meta Queries
See also _Field_ Meta Query where clause: #1281

For example;

```javascript
keystone.createList('Post', {
  fields: {
    title: { type: Text },
  },
});
keystone.createList('User', {
  fields: {
    name: { type: Text },
  },
});
```

```graphql
query {
  _ksListsMeta(where: { name: "Post" }) {
    name
  }
}
```

Only returns the matching list name

```javascript
{
  data: {
    _ksListsMeta: [{
      name: "Post"
    }]
  }
}
```

This would make querying for a particular list much nicer, for example to get the fields meta of one list: https://github.com/keystonejs/keystone-5/issues/1280

Vultraz:
Resolves #1279.


jesstelford:
Oooh, I like it!

Can we add a test too?

I'm wondering if `name` is the best option here? We refer to it as `listKey` elsewhere in the docs, etc. What do you think about calling it `listKey` or `key` or something like that?

Vultraz:
@jesstelford switched to `key` and added a test 👍

Issue #1280, opened 2019-06-05
Expose field meta info via the graphQL API


jesstelford:
Related: https://github.com/keystonejs/keystone-5/issues/354

For example;

```javascript
keystone.createList('Post', {
  fields: {
    title: { type: Text },
    status: {
      type: Select,
      defaultValue: 'draft',
      options: [
        { label: 'Draft', value: 'draft' },
        { label: 'Published', value: 'published' }
      ],
    },
  },
});
```

Then it gets added to the meta query:

```graphql
query {
  _ksListsMeta {
    name
    schema {
      fields {
        name
        type
        ...on FieldTypeSelect {
          options {
            label
            value
          }
        }
      }
    }
  }
}
```

Which would return something like:

```javascript
{
  data: {
    _ksListsMeta: [{
      name: "Post"
      schema: {
        fields: [{
          name: "title",
          type: "Text",
        }, {
          name: "status",
          type: "Select",
          options: [
            { label: 'Draft', value: 'draft' },
            { label: 'Published', value: 'published' }
          ],
        }]
      },
    }]
  }
}
```

NOTE: For the `Select` field in particular, this is distinct from https://github.com/keystonejs/keystone-5/issues/354 which is about getting the label of any one item, whereas this feature is for getting the total possible set of labels (and values).

Issue #354, opened 2018-09-21
Work out pattern for providing Select label to UI builders


timleslie:
The graphql API makes the `value` of a `Select` field available, but does not make the `label` available. The Admin UI obtains the `label` field over the `adminMeta` interface. We need to decide what is the correct way to make this available to project builders. We can perhaps do this over the graphql API, or possibly by making some `adminMeta`-like object available. Either way, we need to document the pattern.

jesstelford:
I really like the graphQL `type` option ☝️ Great solution!

MadeByMike:
@jesstelford how much of the `adminMeta` could be passed via the `graphQLAPI`? If the answer it all of it I definitely think an `adminMeta` GQL endpoint should be a priority issue.

If not, I don't think we should invest in changing this except to fix any bug. 

Is this related to #1768 ?

MadeByMike:
Work is in progress on this: https://github.com/keystonejs/keystone/pull/3056 as well as a recent overhaul of adminMeta. There are also other issues related to custom views for the select component. And adminUI hooks have been added since this issue... I think one way or another this can be solved now and I'd rather see a new issue if required that takes into account recent developments. Closing this one as stale.

PR #2673, opened 2020-04-07
Added `fields` list to ListSchema

### LINKED ISSUE #1281: Add where clause to Field Meta Queries
See also _List_ Meta Query where clause: #1279 

For example, the query in #1280 can become:

```javascript
keystone.createList('Post', {
  fields: {
    title: { type: Text },
    status: {
      type: Select,
      defaultValue: 'draft',
      options: [
        { label: 'Draft', value: 'draft' },
        { label: 'Published', value: 'published' }
      ],
    },
  },
});
```

```graphql
query {
  _ksListsMeta {
    name
    schema {
      fields(where: { name: "status" }) {
        name
        type
        ...on FieldTypeSelect {
          options {
            label
            value
          }
        }
      }
    }
  }
}
```

Which returns only the `status` field:

```javascript
{
  data: {
    _ksListsMeta: [{
      name: "Post"
      schema: {
        fields: [{
          name: "status",
          type: "Select",
          options: [
            { label: 'Draft', value: 'draft' },
            { label: 'Published', value: 'published' }
          ],
        }]
      },
    }]
  }
}
```

Vultraz:
@jesstelford 

Resolves #1281
Partially implements #1280 (only includes `name` and `type` keys right now)

Example:
```graphql
query {
  _ksListsMeta(where: { key: "User" }) {
    name
    schema {
      fields(where: { type: "Text" }) {
        name
        type
      }
    }
  }
}
```

Produces:
```js
{
  "data": {
    "_ksListsMeta": [
      {
        "name": "User",
        "schema": {
          "fields": [
            {
              "name": "name",
              "type": "Text"
            },
            {
              "name": "email",
              "type": "Text"
            }
          ]
        }
      }
    ]
  }
```

### DIFF
--- a/.changeset/thin-mirrors-refuse.md
+++ b/.changeset/thin-mirrors-refuse.md
@@ -3,4 +3,4 @@
 '@keystonejs/keystone': minor
 ---
 
-Added `fields` list to ListSchema. This optionally takes a `where: { type }` argument and returns all matching field names on a list along with their types.
+Added `fields` list to ListSchema. This optionally takes a `where: { type }` argument and returns metadata for all matching fields on list, including name, type, and user-provided config.
--- a/.changeset/thin-mirrors-refuse.md
+++ b/.changeset/thin-mirrors-refuse.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': minor
+---
+
+Added `fields` list to ListSchema. This optionally takes a `where: { type }` argument and returns all matching field names on a list along with their types.
--- a/.changeset/thin-mirrors-refuse.md
+++ b/.changeset/thin-mirrors-refuse.md
@@ -3,4 +3,4 @@
 '@keystonejs/keystone': minor
 ---
 
-Added `fields` list to ListSchema. This optionally takes a `where: { type }` argument and returns metadata for all matching fields on list, including name, type, and user-provided config.
+Added `fields` list to ListSchema. This optionally takes a `where: { type }` argument and returns all matching field names on a list along with their types.
--- a/.changeset/thin-mirrors-refuse.md
+++ b/.changeset/thin-mirrors-refuse.md
@@ -1,4 +1,5 @@
 ---
+'@keystonejs/api-tests': minor
 '@keystonejs/keystone': minor
 ---
 
--- a/api-tests/queries/meta.test.js
+++ b/api-tests/queries/meta.test.js
@@ -117,7 +117,6 @@ multiAdapterRunners().map(({ runner, adapterName }) =>
                 fields {
                   name
                   type
-                  config
                 }
                 relatedFields {
                   type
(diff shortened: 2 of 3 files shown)


Vultraz:
@jesstelford done 👍

JedWatson:
🎉 this is fantastic! Thanks @Vultraz

I'd love to see other meta available as well - given config is field-dependent, maybe it makes sense to include a `config` key which is of JSON type and could contain additional properties?

For what I'm aware of, it makes sense to generally include the same set of properties that `getAdminMeta` includes, except all the standard properties that the base `Implementation` returns could be hoisted up as actual GraphQL fields, where extensions (like `options` for select) could be in the object that the `config` object returns.

Vultraz:
@JedWatson added a `config` key ^ Example in use:
```js
      {
        "name": "FileTest",
        "schema": {
          "fields": [
            {
              "name": "name",
              "type": "Text",
              "config": {}
            },
            {
              "name": "file",
              "type": "File",
              "config": {
                "adapter": {
                  "src": "C:\\Users\\exodi\\Documents\\keystone-5\\demo-projects\\todo\\files",
                  "path": "/files"
                }
              }
            }
          ]
        }
      },
      {
        "name": "SelectTest",
        "schema": {
          "fields": [
            {
              "name": "options",
              "type": "Select",
              "config": {
                "options": [
                  "a",
                  "b",
                  "c",
                  "d",
                  "e",
                  "f",
                  "g",
                  "h",
                  "i",
                  "j",
                  "k",
                  "l",
                  "m",
                  "n",
                  "o"
                ]
              }
            }
          ]
        }
      },
```

Though, I'm not sure if having the file adapter src path exposed is good? 🤔

JedWatson:
> I'm not sure if having the file adapter src path exposed is good?

Yeah, you'll really want to use the `getAdminMeta` function rather than passing raw `config` - anything in there has already been cleaned and safe for sharing publicly (it's awkward naming and I think we'll address that in the near future but this is the field config that's serialised for front-end use in the Admin UI)

jesstelford:
Agreed we shouldn't be exposing the source directory or files.

@jedwatson the original intent in https://github.com/keystonejs/keystone/issues/1280 (and discussed in https://github.com/keystonejs/keystone/issues/354) is as you say; to provide enough information for the admin UI to function.

I'm not sure that exposing the entire config option is quite the right solution as it's a different concern. I think the exposed data should definitely be _derived_ from the config, but it might differ slightly (such as not exposing the source directories) 

Can I propose that we remove it from this PR so it can get merged without contention, then we can figure out (maybe in #354) how exactly we want to expose the rest of the info?

@Vultraz thanks for adding the test 👍

Vultraz:
Ok, reverted the `config` field for now.

jesstelford:
Sorry for the churn! Thanks for getting stuck into it super quick 🎉

jesstelford:
@jedwatson are you ok if we merge this as-is then add more info in a further PR?

JedWatson:
Yep! Let’s do it.

Thanks again @Vultraz

PR #2681, opened 2020-04-07
Added a Keystone meta provider

Vultraz:
Going off what @jesstelford said in #2673, I went looking at the various `getAdminMeta` getters to see what could be added to the meta APIs. Most of the stuff in list and field-level `getAdminMeta` could be added to `_ksListsMeta`/`_${List}Meta`, but there wasn't anything to handle the Keystone-level meta.

This adds a new provider that enables this query:
```gql
query {
  _ksMeta {
    projectName
  }
}
```
to get
```js
{
  "data": {
    "_ksMeta": {
      "projectName": "Keystone To-Do List"
    }
  }
}
```

Right now it only provides a `projectName` field since that's all `keystone.getAdminMeta` added (there called simply `name`, but `projectName` seems more descriptive). I gave it its own return type so we could easily add more stuff later.

Side note: I realize this overlaps somewhat with the App Version Provider... maybe it should be merged with this? 🤔 Then one could do, say:
```gql
query {
  _ksMeta {
    projectName
    appVersion
  }
}
```

### DIFF
--- a/.changeset/tasty-masks-count.md
+++ b/.changeset/tasty-masks-count.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': minor
+---
+
+Added a Keystone meta provider.
--- a/packages/keystone/lib/Keystone/index.js
+++ b/packages/keystone/lib/Keystone/index.js
@@ -32,7 +32,12 @@ const {
 } = require('./relationship-utils');
 const List = require('../List');
 const { DEFAULT_DIST_DIR } = require('../../constants');
-const { CustomProvider, ListAuthProvider, ListCRUDProvider } = require('../providers');
+const {
+  CustomProvider,
+  ListAuthProvider,
+  ListCRUDProvider,
+  KeystoneMetaProvider,
+} = require('../providers');
 
 module.exports = class Keystone {
   constructor({
@@ -86,6 +91,7 @@ module.exports = class Keystone {
         access: appVersion.access,
         schemaNames,
       }),
+      new KeystoneMetaProvider({ name: this.name }),
     ];
 
     if (adapters) {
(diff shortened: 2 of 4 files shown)


timleslie:
Hmm, I'm not sure about this, but I'm not prepared to rule it out just yet. The `projectName` thing is interesting. It used to be the case that `keystone.name` was used as a default DB name, but we've since removed that and I believe that the only place it's still used is as a title in the admin UI. I was actually thinking about removing it completely from the `keystone` object and pushing it onto the admin UI config. Making it available over the graphQL api is an interesting use case I hadn't thought of before. I'm trying to think of when someone might want/need to use this. Did you have an example use case in mind?

Vultraz:
Well, I don't really have a specific usecase in mind. When I opened this I was mostly looking to see what could be converted to use the GQL API instead of the `adminMeta` object to populate the Admin UI, which I gathered from @jesstelford was a long-term goal.

Side note: it does still look like `keystone.name` is being used to construct the default db name:

Mongoose adapter:
```js
    if (!uri) {
      const defaultDbName = slugify(name) || 'keystone';
      uri = `mongodb://localhost/${defaultDbName}`;
      logger.warn(`No MongoDB connection URI specified. Defaulting to '${uri}'`);
    }
```

Knex adapter:
```js
    if (!knexConnection) {
      const defaultDbName = slugify(name, { separator: '_' }) || 'keystone';
      knexConnection = `postgres://localhost/${defaultDbName}`;
      logger.warn(`No Knex connection URI specified. Defaulting to '${knexConnection}'`);
    }
```

I wouldn't be opposed to simply making it an Admin UI-only option. I think I might have misunderstood the whole "populate admin ui via gql" thing, or perhaps it needs some spec still?

Vultraz:
Some more thoughts. It's probably best to keep the static admin meta (name, apiPath, etc) in the JSON global and only use the GQL API for fetch list data. That would render this unnecessary 🤔

Vultraz:
Since `name` is now part of the Admin UI config, I'm closing this. It should probably stay in the static JSON global anyway, like mention above.

PR #2684, opened 2020-04-08
Update create-keystone-app and associated quick-start guide.

timleslie:
These changes will hopefully make it much easier for a user to get started, particularly if they are running postgres. In particular, the default connection string we previously used had a vanishingly small chance of ever working. We now give much more explicit instructions on how to find a connection string which will work and then allow the user to explicitly provide and test this connection string when setting up their project.

### DIFF
--- a/docs/quick-start/README.md
+++ b/docs/quick-start/README.md
@@ -7,26 +7,41 @@ slug: /quick-start/
 
 # Getting started
 
-This quick start guide will get you up and running in just a few minutes. Let's build a simple todo app with a fresh install of Keystone!
+Welcome to KeystoneJS!
+This quick start guide will get you up and running in just a few minutes.
+Let's build a simple Todo app with a fresh install of Keystone!
 
 ## Requirements
 
-Before we start, check that your computer meets the following requirements:
+Before we start, make sure that you have a basic development environment set up, with the following tools installed on your system.
 
 - [Node.js](https://nodejs.org/) >= 10.x: Node.js is a server platform which runs JavaScript.
+- [yarn](https://yarnpkg.com/) or [npm](https://docs.npmjs.com/cli/npm): yarn and npm are different package managers which can be used to install Keystone.
 
-And ONE of the following databases:
+```shell allowCopy=false showLanguage=false
+$ node --version
+v12.11.0
+$ npm --version
+6.9.0
+$ yarn --version
+1.17.3
+```
+
+## Database Setup
+
+You will also need to have a database for Keystone to store your application data in.
+You can use either `MongoDB` or `PostgreSQL`.
 
 - [MongoDB](https://www.mongodb.com/) >= 4.x: MongoDB is a powerful NoSQL document storage database.
 - [PostgreSQL](https://www.postgresql.org) >= 9.x: PostgreSQL is an open source relational database that uses the SQL language.
 
-Finally, make sure [your database is configured and running](/docs/quick-start/adapters.md).
+Follow the [database setup](/docs/quick-start/adapters.md) instructions to install and configure your databasem, and to find out what your `connection string` is.
 
-All set? Great, let's get started!
+> **Important:** You will need to make sure you have a valid `connection string` for your database in order to set up Keystone.
 
-## Installation
+## Installing Keystone
 
-To create a new Keystone application, run the following commands in your terminal:
+To create a new Keystone application, run the following command in your terminal:
 
 ```shell allowCopy=false showLanguage=false
 npm init keystone-app my-app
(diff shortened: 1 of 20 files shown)


docs/quick-start/README.md
```diff
@@ -37,15 +52,37 @@ yarn create keystone-app my-app
 You'll be prompted with a few questions:
 
 1. **What is your project name?** Pick any name for your project. You can change it later if you like.
-2. **Select a starter project.** Select the `Todo` application if you wish to follow this guide.
-3. **Select an adapter.** Choose `Mongoose` if you're running a MongoDB database and `Knex` if you're running a PostgreSQL one.
+2. **Select a starter project.** These are some preconfigured projects you can use as the base of your own application. Select the `Todo` application if you wish to follow the rest of the Keystone tutorials.
+3. **Select an database type.** Choose between `MongoDB` and `PostgreSQL`.
```
MadeByMike, line 56:
```suggestion
3. **Select a database type.** Choose between `MongoDB` and `PostgreSQL`.
```

MadeByMike, line 56:
```suggestion
3. **Select a database type.** Choose between `MongoDB` and `PostgreSQL`.
```

docs/quick-start/adapters.md
```diff
@@ -3,67 +3,149 @@ section: quick-start
...
 The simplest way to install MongoDB is using [Homebrew](https://brew.sh/).
 Refer the [official guide](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/) for more information.
 
-### Windows
+#### Windows
 
 Follow the [official guide](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-windows/) for installing MongoDB on Windows.
 
-### Other platforms
+#### Other platforms
```
MadeByMike, line 34: Total bikeshed, but I've not been using title case generally. I noticed every other title here is changed to title case except this one... I don't hold strong opinions but like consistency, not just here but with the rest of the docs

docs/quick-start/adapters.md
```diff
@@ -3,67 +3,149 @@ section: quick-start
...
 
 For Windows and other platforms see the [download instructions](https://www.postgresql.org/download/) on the [postgresql.org](https://postgresql.org) website.
 
 ### Setup
 
-By default the Knex adapter will attempt to connect to a PostgreSQL database as the current user.
+<!-- ### MongoDB -->
```
MadeByMike, line 96: Remove these comments

docs/quick-start/adapters.md
timleslie: From our style guide:

- Headings at all levels use [Title Case](https://en.wikipedia.org/wiki/Letter_case#Stylistic_or_specialised_usage)

I'll change this one to `Other Platforms` 👍

packages/create-keystone-app/example-projects/config.js
```diff
@@ -1,14 +1,20 @@
...
+
 const adapters = {
-  Mongoose: {
+  MongoDB: {
+    name: 'MongoDB',
     file: 'adapter-mongoose.js',
     dependencies: ['@keystonejs/adapter-mongoose'],
-    description: 'Connect to a Mongo database.',
+    description: 'Connect to a MongoDB database.',
+    defaultConfig: name => `mongodb://localhost/${slugify(name)}`,
```
MadeByMike, line 9: This has a different slugify method to the postgres example, is that consistent with how we slugify the project names for default connections strings?

I'm going to assume it is and that the reason is mongo names can have `-` but postgres can't?

Maybe not for this PR but we should totally just make that consistent and always `_`. Unfortunately that's breaking :(

docs/quick-start/adapters.md
Vultraz: Wait, I'm confused - #2592 changed all the headings to sentence case 🤔

packages/create-keystone-app/lib/copy-example-project.js
```diff
@@ -8,7 +8,7 @@ const createNewProjectFolder = newProjectFolder => {
   fs.mkdirpSync(newProjectFolder);
   const readDir = fs.readdirSync(newProjectFolder);
   if (readDir && readDir.length > 0) {
-    error(`The project directory ${newProjectFolder} is not empty`);
+    error(`The project directory "./${newProjectFolder}" is not empty`);
```
MadeByMike, line 11: There is an edge-case where if passed as an argument, it might not be a relative path. Leading to this error being wrong. Pretty edgy though.

packages/create-keystone-app/lib/install-project-dependencies.js
```diff
@@ -9,6 +9,7 @@ const installProjectDependencies = async () => {
   }
 
   console.log('Installing dependencies with yarn. This could take a few minutes.');
```
MadeByMike, line 11:
```suggestion
  console.log('Installing dependencies with yarn. This will take a few minutes.');
```

packages/create-keystone-app/example-projects/config.js
timleslie: Your have surmised correctly 👍

packages/create-keystone-app/lib/copy-example-project.js
timleslie: Yeah, not too fussed about that I don't think. Will accept a PR in the future if anyone cares enough :-)

PR #2685, opened 2020-04-08
Made certain functions error if called after prepare()

### LINKED ISSUE #2678: Please document that 'keystone.extendGraphQLSchema' needs to come BEFORE 'keystone.prepare'
# Feature request

## Is your feature request related to a problem? Please describe.

I was trying to use `keystone.extendGraphQLSchema` in my code to add a custom query and spent a great deal of time not understanding why the new query I'd added did not appear in the KS app's schema (i.e., it did not appear in the Playground) and threw an error when called.  Finally I tried putting it BEFORE my call to `keystone.prepare` and it worked fine.  Apparently if you put it after `keystone.prepare` it fails silently.

## Describe the solution you'd like

Just to document it, perhaps with an explanation of why one needs to follow the other.

Vultraz:
Instead of failing silently. Resolves #2678.

### DIFF
--- a/.changeset/hungry-panthers-tickle.md
+++ b/.changeset/hungry-panthers-tickle.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Made keystone.extendGraphQLSchema() error if called after prepare().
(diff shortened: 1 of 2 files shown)


Vultraz:
@timleslie added those two

PR #2696, opened 2020-04-09
Re-enable all login tests

timleslie:
c.f #2657


zenflow:
@timleslie Title should be "Re-enable _most_ login tests" but otherwise looks good. 

But why not just merge #2657, which gets that failing test passing, and enables *all* login tests? 😀 Do you have any thoughts on the PR? Would you support merging it?

timleslie:
@zenflow Yep, I really like #2657, but there's a few different things going on in there which I want to run by some other people to make sure everything is going in the right direction. I want to pull out this change separately just to better understand where our tests are currently failing so that it's more clear which parts of that PR are about fixing that specific test failure and which parts are about adding new features 👍

Issue #1684, opened 2019-09-20
Images not loading from custom field type in Admin UI


MadeByMike:
# Bug report

## Describe the bug

The Admin UI is returning the wrong mime-type for svg which breaks the star custom field type.

A clear and concise description of the bug.

## To Reproduce

Steps to reproduce the behaviour. Please provide code snippets or a repository:

1. Launch the test project in `test-projects/basic`
1. Create new content 
1. Compare with screenshots

## Expected behaviour

The image should render.
The content can be loaded but the network panel shows a `type` of document.
I think that is incorrect for SVG?

## Screenshots

![image](https://user-images.githubusercontent.com/1320567/65289850-cb5a2680-db8f-11e9-88da-5dd69ccc8cef.png)

![image](https://user-images.githubusercontent.com/1320567/65289860-d614bb80-db8f-11e9-9c07-9501cd80ebd7.png)

jesstelford:
🤔 That URL is on the root `/admin` path which implies it's part of the Admin UI bundle. Which then implies it's something to do with webpack not serving the file's mimetype correctly.

@MadeByMike do you see the same behaviour when doing `yarn build && yarn start` (ie; a "Production" build)?

MadeByMike:
Ahhh! Nice pick-up @jesstelford it works with a build and start. It's getting the correct mime-type. What I'm not sure about is why it was able to load the SVG source in the first place.

Is there a bug here or should we close this one?

MadeByMike:
@jesstelford I'm going to close this I think. Could you re-open if you think there is an issue here.

jesstelford:
It looks like there's still a bug in dev mode where it's not setting the right mime type.

Vultraz:
Is this related to #1823 (or vice-versa)?

Issue #1823, opened 2019-10-24
File upload thumbnails not displaying in AdminUI


Vultraz:
# Bug report

## Describe the bug

When uploading a file with the File field (I'm using the LocalFileAdapter), the thumbnail will show up in the Create dialog, but not when viewing the entry in the list (see below). I have both `src` and `path` properties of the file adapter set to the same place:

```js
const fileAdapter = new LocalFileAdapter({
  src: './files',
  path: './files',
});

keystone.createList('FileTest', {
  fields: {
    name: { type: Text },
    file: { type: File, adapter: fileAdapter },
  }
});
```
A look in Chrome seems to indicate it's trying to fetch the files from `localhost:3000/admin/files` instead of `localhost:3000/files` (where they actually are). Not sure how to fix this one.

## Screenshots
![image](https://user-images.githubusercontent.com/3558659/67446102-25e91580-f65b-11e9-8cb7-ef5e492c5303.png)

![image](https://user-images.githubusercontent.com/3558659/67446180-6fd1fb80-f65b-11e9-9215-de541c225787.png)

## System information

- OS: Windows 10 1903
- Browser: Chrome 78

MadeByMike:
@Vultraz what happens if you change the path to `/files`? `./files` would indicate a relative path on the web.

gautamsi:
I think the path should be `files` and `app-static` configured with same path and src.

Vultraz:
Ok, with the absolute path (`/files`) it *does* look in `localhost:3000/files` but it still returns a 404.
![image](https://user-images.githubusercontent.com/3558659/67448687-7a908e80-f663-11e9-83c7-bc1bdbc8f2b0.png)

(not sure if the `text/html` bit is relevant).

When trying to visit the path in the browser you get:
> Cannot GET /files/5db06df14660714c0c734571-bigthonker.jpg

And I have confirmed the file exists in the filesystem at that path.

gautamsi:
You also need the `@keystonejs/app-static` or express.static path configured

Vultraz:
Ok, this just gets weirder and weirder! When `src` is set to a *relative* path (i.e. `./files`), that folder is created at server toplevel (`/`). For me that would be `C:\Users\exodi\Documents\testapp2\files`.

But if it an absolute path (i.e. `/files`), the folder is created at OS root - `C:\files`! There's literally a toplevel folder in `C:\` with files from when I was testing with an absolute `src` path 😱

[...]

PR #2708, opened 2020-04-11
Some minor cleanup to the file field

Vultraz:
- Removed unused `route` and `directory` config options
- Added `originalFilename` to the admin meta query
- Explicitly added `originalFilename` and `encoding` to the Mongoose schema.


packages/fields/src/types/File/Implementation.js
```diff
@@ -4,19 +4,17 @@ import { KnexFieldAdapter } from '@keystonejs/adapter-knex';
...
   Types: { ObjectId },
 } = mongoose;
 
 export class File extends Implementation {
-  constructor(path, { directory, route, adapter }) {
+  constructor(path, { adapter }) {
     super(...arguments);
     this.graphQLOutputType = 'File';
-    this.directory = directory;
-    this.route = route;
```
MadeByMike, line 19: I think this should stay. It might be what's needed to fix #1684, and #1823. They are added to the Admin meta but not used in the field views. If we add the route to the image path in views it should allow users to configure a path for images. I'm not sure that directory is as useful.

Also as I write this I'm thinking now that maybe this info should be pulled off the `fileAdapter`.

Thoughts?

Vultraz: Yeah, I'd say it it should probably be pulled off the adapter. It avoid having to have things configured in two places.

PR #2715, opened 2020-04-12
Expanded `_ksListsMeta` query input to take a key list

Vultraz:
I was thinking more about the `where` input I added to `_ksListsMeta` in #2664, and it occurred to me that that change was perhaps not the most useful. If you already know the key you want - for example, `User` - you could just do a `_UsersMeta` query instead of `_ksListsMeta(where: { key: "User" })`.

This expands `key` to accept an array, so as to be more versitile:
```gql
query {
  _ksListsMeta(where: { key: ["User", "Todo"] }) {
    name
  }
}
```
gives
```js
{
  "data": {
    "_ksListsMeta": [
      {
        "name": "User"
      },
      {
        "name": "Todo"
      }
    ]
  }
}
```

Though, admittedly, one could still just do...
```gql
query {
  _UsersMeta {
    name
  }
  _TodosMeta {
    name
  }
}
```

So perhaps the `where` input still isn't that useful after all... 🤔 @jesstelford thoughts?

### DIFF
--- a/.changeset/silver-monkeys-shop.md
+++ b/.changeset/silver-monkeys-shop.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Expanded `_ksListsMeta` query input to take a key list.
--- a/packages/keystone/lib/providers/listCRUD.js
+++ b/packages/keystone/lib/providers/listCRUD.js
@@ -91,7 +91,7 @@ class ListCRUDProvider {
         count: Int
       }`,
       `input ${this.gqlNames.listsMetaInput} {
-        key: String
+        key: [String]
       }`,
       `input _ListSchemaFieldsInput {
         type: String
@@ -193,7 +193,7 @@ class ListCRUDProvider {
       // And the Keystone meta queries must always be available
       [this.gqlNames.listsMeta]: (_, { where: { key } = {} }, context) =>
         this.lists
-          .filter(list => list.access[schemaName].read && (!key || list.key === key))
+          .filter(list => list.access[schemaName].read && (!key || key.includes(list.key)))
           .map(list => list.listMeta(context)),
     };
   }


.changeset/silver-monkeys-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
```
MadeByMike, line 2:
```suggestion
'@keystonejs/keystone': major
```

This is a breaking change right?

Vultraz: No, since the single input still works.

Vultraz:
Added an aux filter param too:
```js
_ksListsMeta(where: { auxiliary: false })
```

Vultraz:
Async callback not invoked failure.

MadeByMike:
Hmmm I'm less sure we need this one @Vultraz, not sure we needed the where either... 🤔 

Isn't this exactly why GraphQL and not rest:

```
query {
  _UsersMeta {
    name
  }
  _TodosMeta {
    name
  }
}
```

Why not make GraphQL the API?

Vultraz:
Originally, I was just implementing #1279 as @jesstelford requested. Admittedly, there is some benefit to being able to just pass in an array of list keys without manually constructing the query, but I'm not sure I'll actually need to use that.

The `auxiliary` flag is definitely useful, though, since the admin UI doesn't need aux lists.

MadeByMike:
Yep `auxiliary` is good. I'd merge that right away. 

I want to get another opinion on adding the array. Writing a larger query is not too bad and you could write a helper to generate it if you needed it for huge amounts of lists.

Up to you if you want to merge `auxiliary` now or wait for more input on the array input. Other people have stronger opinions of the shape of this API than me.

Vultraz:
Split the `auxiliary` change into #3052

MadeByMike:
I think we arrived at not needing this. Closing.

PR #2719, opened 2020-04-12
Starter template: Clarify userOwnsItem return type

mjomble:
When studying the example code, at first I assumed the function should always return a boolean.
Conditionally returning an object looked like a bug or a leftover from an older version of the code.

After reading more of the docs, I realised this code is correct, but a comment like this would have been useful :)


timleslie:
Hi @mjomble, good point about this perhaps not being obvious to a new user. Could you perhaps update the inline comment to the docs which explain this form of access control? Thanks!

mjomble:
@timleslie done

PR #2724, opened 2020-04-13
Used functional components for email templates

Vultraz:
Wasn't able to test these directly, but nothing in the `express-react-views` seems to indicate this would break anything.


Vultraz:
Whaa?
>FAIL  api-tests/relationships/crud/many-to-many.test.js (8.618s)

PR #2731, opened 2020-04-13
Arch: cleaned up Control component implementation

Vultraz:
- Converted to functional component
- Removed `memoize-one` dependency (no longer needed since I switched to a clearer impl with `useMemo`.
- Removed instance ref being passed to `HiddenInput`. It wasn't being used for anything. Also removed the `focus()` and `blur()` members.
- Removed `innerRef` props from `CheckboxPrimitive` and `RadioPrimitive`. These weren't being used anywhere (there's one non-functional case I'll remove later (#2497)) and can be re-added with `forwardRef` if needed.
- Used proper SVG components for the icons instead of `dangerouslySetInnerHTML`.


Vultraz:
Can #2497 get merged with this?

PR #2734, opened 2020-04-14
[RFC] changed the WYSIWYG field to fetch from CDN

Vultraz:
Small diff but a big change for the WYSIWYG field. Prior to this, TinyMCE was loaded locally from the `tinymce` module. This removes that and instead delegates the responsibility to `tinymce-react` to fetch from CDN, allowing people to access pro features with their Tiny cloud API key. @ra-external especially was asking for this.

(BTW, You don't *need* an API key. The only difference is now you get a little banner telling you to sign up for Tiny Cloud.)

I honestly can't figure out a way to conditionally use the local import if the API key isn't provided. TinyMCE has this weird implementation where unless you're calling `init()` yourself (which we're no longer doing), you can do a side effect-only import just to set the global `window.tinymce` variable. The React integration first checks if that variable is set; if so, it uses it, if not, it fetches from CDN.

I couldn't find a way to properly instance that to the field. Basically, the first TinyMCE source you load up - local or CDN - will (as far as I can tell) be used for all field instances. It feels cleaner just to let the CDN handle everything; everything is always up-to-date, and we don't have to worry about version conflicts between local or remote versions...

Will let @ra-external elaborate on his usecase.


ra-external:
I would, as @Vultraz said, appreciate this.  We have a use-case where we need to upload images within TinyMCE.  Our original plan was to write a custom image upload handler in a custom WYSIWYG field based on the KS TinyMCE WYSIWYG field, which I did, but this proved very problematic -- it was returning somewhat strange file paths which we had to massage -- and in the end seemed to conflict with the AdminUI to the point where it stopped the AdminUI from uploading images.

So what we'd like to do as a fallback is upload our images to TinyMCE's cloud, but to do this we need to be able to pass the APIKey to TinyMCE, but according to @Vultraz this requires that TinyMCE be loaded from the CDN.  

Further, there will be many use cases where people wish to make use of other 'premium' TinyMCE features but won't be able to without being able to pass the TinyMCE APIKey.  For instance, we want to be able to user their easy multi-media insert plugin, which is a premium feature.  We need to be able to pass the APIKey to do this too.

Finally, if one uses the CDN, as @Vultraz said, it's always up to date and no one has to maintain the local TinyMCE library in Keystone.

(Of course if one day someone can figure out how to write custom image handlers for TinyMCE that one could pass on from within the Keystone environment, and which won't interfere with Keystone nor the AdminUI, that would be great too!)

JedWatson:
Thanks @Vultraz, assigning this to myself to take a closer look at 🙂

ra-external:
@Vultraz --  if one is using TinyMCE’s TinyDrive and wants to pass in a ‘JWT’ providing function (see https://www.tiny.cloud/docs/tinydrive/configuration/) in the `editorConfig` in a Keystone custom field based on your PR, the first option

`tinydrive_token_provider: '/jwt',
`
 doesn’t seem to work.  If I put this directly inside defaultOptions  in `Field.tsx` it works fine.  I traced it out and the correct value is being passed in overrideOptions.

The KeystoneJS WYSIWYG file, in addition to allowing the APIKey to be passed, should also allow the TinyDrive related JWT token provider path to be passed.

Vultraz:
Resolved conflicts.

jobmbay:
Hello,

Why does Wysiwyg is not deplayed after deployment ?

bladey:
Thanks for this PR @Vultraz.

**Keystone 5** has officially moved into active maintenance mode as we push towards the next major new version **Keystone Next**, you can find out more information about this transition [here](https://github.com/keystonejs/keystone/issues/4962).

In an effort to sustain the project going forward, we're cleaning up and closing old PRs such as this one.

However, you are welcome to create a PR under the [Keystone 5](https://github.com/keystonejs/keystone-5/pulls) repository.

Please let me know if you have any questions.

PR #2745, opened 2020-04-15
Allowed configuring default/maximumPageSize for all lists

Vultraz:
~It occurred to me that these options make a *lot* more sense when applied to the whole Admin UI, instead of per-list. I really doubt anyone is thinking, hey, we want list A to show 50 items at once while lists B and C have 10 items and this other one has 20. Max page size makes even less sense on a per-list basis (who would want to see a completely arbitrary maximum number of items depending on the given list?)~

~This leaves the sort options untouched (since those are indeed list-dependent) and moves the `defaultPageSize` and `maximumPageSize` config options to the Admin UI ctor.~

Update: this now just adds `defaultPageSize` and `maximumPageSize` config options that allows configuring the defaults for all lists. All handling of `adminConfig` options (such as stripping whitespace) is now handled by the Admin UI app.

### DIFF
--- a/.changeset/tasty-cooks-begin.md
+++ b/.changeset/tasty-cooks-begin.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': minor
+'@keystonejs/keystone': patch
+---
+
+Added defaultPageSize and maximumPageSize config options fto the Admin UI. These can be used to set defaults for all lists (previously, these defaults were 50 and 1000, respectively).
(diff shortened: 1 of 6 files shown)


Vultraz:
Rebased

gautamsi:
I like this but am sure someone would want to limit few list higher size rather than small limit lists.

Vultraz:
Hmmmm....

VinayaSathyanarayana:
We should have per List options... taking the global Admin UI values if not set at the List level.

Vultraz:
Ok, that's a good idea. Will do that.

Vultraz:
Ok, this is ready.

Vultraz:
@timleslie conflict-be-gone

Vultraz:
I'll fix the conflicts on this after #3056 is merged.

Vultraz:
@timleslie since that other one is stalled I just went and fixed the conflicts on this.

Issue #1220, opened 2019-05-23
Errors thrown from hooks are not caught


jesstelford:
For example, on a Field Type:

```javascript
class Foo extends File {
  async validateInput() {
    throw new Error('hi');
  }
}
```

### Expected behaviour

Fail the mutation and show the thrown error in the `errors` array

### Actual behaviour

Causes the server to hang when creating a new item.

Vultraz:
Tested with this:
```js
    validateInput: () => {
      console.log("HOOK TEST");
      throw new Error('Y U DO DIS');
    }
```
The error *does* show up as a GraphQL Error popup in the admin UI, but not the console. So, this is fixed, I guess?

PR #2767, opened 2020-04-17
Various improvements to validation hooks

Vultraz:
@ra-external pointed out `addFieldValidationError` doesn't actually do anything when called in list-level hooks. I checked, and this is indeed so; field-level hook errors are already handled by that point. Turns out list hooks get their own (undocumented) `addValidationError` function:

```js
    if (this.hooks[hookName]) {
      const listValidationErrors = [];
      await this.hooks[hookName]({
        ...args,
        addValidationError: (msg, _data = {}, internalData = {}) =>
          listValidationErrors.push({ msg, data: _data, internalData }),
      });
      if (listValidationErrors.length) {
        this._throwValidationFailure(listValidationErrors, operation, originalInput);
      }
    }
```

This changes `addFieldValidationError` to `addValidationError` for field hooks, while keeping the former around as an alias for compatibility. I also made `addFieldValidationError` `undefined` in list hooks since it wasn't actually doing anything before.

I also made it that any errors thrown in validation hooks get surfaced as `ValidationFailureError`s instead of just generic errors. Should address #1220 (I'm still not able to reproduce any hang, but this should ensure any thrown errors now have consistent type)

Finally, also made the field type and list hook validation run in parallel. This also makes both validation steps for a field execute with each other (as opposed to checking all types hooks, then all field-level hooks). Do let me know if I did this right 😬 

Also, `addValidationError` no longer returns the result of `push` (the number of errors).


packages/keystone/lib/List/index.js
```diff
@@ -1182,25 +1182,53 @@ module.exports = class List {
...
-    await this._mapToFields(fields, field => field[hookName](args));
-    await this._mapToFields(
-      fields.filter(field => field.hooks[hookName]),
-      field => field.hooks[hookName](args)
-    );
+    };
+
+    // Compatibility. Remove at some point
+    args.addFieldValidationError = (...passthrough) => {
+      console.warn('addFieldValidationError is deprecated. Use addValidationError instead');
```
MadeByMike, line 1193: 👍

Vultraz:
FTR, I just noticed the docs explicitly do claim that field type hooks and field hooks run in parallel. So... I guess this bring the actual behavior in line with the docs? 🤔

packages/keystone/lib/List/index.js
leonalicious: please update the documentation as well ~

Vultraz: I already replaced all mentions of `addFieldValidationError` with `addValidationError`.

Vultraz:
@timleslie anything more I should do with this?

timleslie:
Hoping to get to this one today or tomorrow. I'm trying to make at least some progress on every PR in the backlog this week, but this is one of the most complex ones, so I keep skipping it, sorry 😓 

You might want to take a quick look at #3002 , since it gives an example of using `addFieldValidationError`. I haven't worked out what the implications for this PR might be, but it's something to consider from an impact point of view.

Vultraz:
No hurry! 😃 

Here's a breakdown of the effects on each type of hook:
| Function                  | Field type hooks                   | Field hooks                        | List hooks                                |
| ------------------------- | ---------------------------------- | ---------------------------------- | ----------------------------------------- |
| `addFieldValidationError` | Still available, will give warning | Still available, will give warning | No longer available as was non-functional |
| `addValidationError`      | Newly available                    | Newly available                    | Already available but undocumented        |

So, for #3002, all you'd need to do is replace `addFieldValidationError` with `addValidationError` (same signatures) and you'd be good to go.

Vultraz:
Had another report on Slack of this being run into. Their use of `addFieldValidationError` at the list level wasn't working.

Vultraz:
Small tweak. Instead of making `addFieldValidationError` just `undefined` in list-level hooks, I made it throw an error telling people to use `addValidationError`. Field-level hooks still just get a console warning.

Vultraz:
@timleslie rebased for #3235

Vultraz:
Unrelated failure.

timleslie:
This PR has been superseded by #4652.

PR #2772, opened 2020-04-18
Set test-utils as a dev dependency

### LINKED ISSUE #2763: Not using Mongo, yet Keystone installs a Mongo binary
# Bug report

## Describe the bug

When using the latest Keystone packages, mongodb-memory-server is installed in node_modules, which then drags in a Mongo binary. This should not be the case as I'm using the Knex Adapter and Postgres.

## To Reproduce

Steps to reproduce the behaviour:

1. Set up a postgres app
2. npm i

## Expected behaviour

Mongo Binary not installed

## Screenshots

```
> mongodb-memory-server@6.5.2 postinstall [project]/node_modules/mongodb-memory-server
> node ./postinstall.js

mongodb-memory-server: checking MongoDB binaries cache...
mongodb-memory-server: binary path is [project]/node_modules/.cache/mongodb-memory-server/mongodb-binaries/4.0.14/mongod
```

## System information

- OS: macOS
- Browser (if applies) N/A

## Additional context

I think this comes from test utils, when dependencies on Mongo or Knex should happen at the adapter level.

Abogical:
Fixes #2763


.changeset/funny-avocados-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Set test-utils as development dependency.
```
MadeByMike, line 5:
```suggestion
Changed test-utils to a development dependency.
```

PR #2774, opened 2020-04-18
Switched to new @primer/octicons package and regenerated icon components

Vultraz:
Added all the new icons that appeared in the process.


packages/arch/packages/icons/src/build/octicons.js
```diff
@@ -29,7 +29,10 @@ Object.getOwnPropertyNames(octicons).forEach(octiconName => {
 
   iconsIndex.push({ iconName, componentName });
 
-  svgContents = svgContents.replace(/fill-rule="/g, 'fillRule="');
+  svgContents = svgContents.replace(/([fill|clip])-rule="/g, '$1Rule="');
+
+  // FIXME: we don't really want to do this here, but prettier >= 1.18 won't close empty tags anymore...
+  svgContents = svgContents.replace(/><\/path>/g, ' />');
```
Vultraz, line 35: Apparently Prettier no longer handles closing the tags. See https://github.com/prettier/prettier/issues/6452. Not sure if we should disable the rule.

MadeByMike: SVG will handle `<path/></path>` totally fine. It might not be pretty but it's fine.

Vultraz: Right, but there's currently an [`eslint` rule](https://github.com/keystonejs/keystone/blob/master/.eslintrc.js#L58) enabled that will cause the tests to fail if the closing tags are left.

Issue #2370, opened 2020-02-15
[AdminUI] defaultValue as function not working on Calendar Day


maximelebreton:
Hey there,

I wanted to put the **current day** as default value on the `CalendarDay` field.  

So I read the [fields config documentation](https://www.keystonejs.com/keystonejs/fields/#config): 
![image](https://user-images.githubusercontent.com/1072425/74587133-11a49c00-4fef-11ea-9346-4e187ab1b766.png)

And then [CalendarDay config documentation](https://www.keystonejs.com/keystonejs/fields/src/types/calendar-day/#config)

I tried to do this, but **it doesn't work**:

```js
fields: {
    date: {
      type: CalendarDay,
      defaultValue: () => Date.now().toISOString() // doesn't work
```

So to be sure, I tried this, and **it works**:

```js
fields: {
    date: {
      type: CalendarDay,
      defaultValue: "2020-01-01" // works
    }
```

So I tried this, and **it doesn't work**...
```js
fields: {
    date: {
      type: CalendarDay,
      defaultValue: () => "2020-01-01" // doesn't work
    }
```

So the **defaultValue** prop as **function** doesn't work...

Maybe it's related to all fields, I haven't tested it?

**Personal thought:**
Please switch to Typescript, we shouldn't meet anymore this kind of stuff in 2020 :)

2xAA:
This also seems to be the case with the `DateTime` type.

My schema:

```JavaScript
fields: {
  createdAt: {
    type: DateTime,
    defaultValue: () => new Date().toISOString()
  }
}
```

reubano:
dupe #2043

MadeByMike:
Yes, this is a duplicate and I've just updated the documentation so that this doesn't happen again. @maximelebreton I'm sorry for the confusion please for why this must be a string: #2043

Vultraz:
Actually, after #2347 the function default *should* work as long as `isRequired` is false.

Issue #2043, opened 2019-12-05
defaultValue function values possibly broken


Vultraz:
# Bug report

## Describe the bug

Ok, so the docs say fields can take a `defaultValue` key as a string or function. Strings work fine, but I can't seem to get functions to do anything at all.

I notice they aren't passed to the Admin UI, so it makes sense they wouldn't resolve and display when creating an item, but even if you simply save a field with no value the `defaultValue` function is never called.

Testing using this config in the TODO demo:
```js
keystone.createList('Todo', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    name: {
      type: Text,
      schemaDoc: 'This is the thing you need to do',
      isRequired: true,
      defaultValue: () => { return 'test text' },
    },
  },
});
```

`List._resolveDefaults` is where I would expect the default function to be resolved server-side, but `getDefaultValue` is only called for `id` since `name` has a `null` value already.

MadeByMike:
So I did a complete investigation of this and have concluded that it is not a bug.

The `getAdminMeta` function is executed at build time and generates a JSON blob that is passed to the AdminUI. I could resolve the function at build time but if that is possible it might as well be a string. This is how data is passed from Keystone configuration to the AdminUI.

This means that data can pretty much only be scalar data types and not functions or other complex types. There are a few exceptions to this including field views and pages. For these to work they need to be separate files and they are imported via an `importView` (which does a bunch of complex and magic stuff not worth mentioning here) and results in functions built and packaged for use in the front-end.

This is not a good API if you want to intersperse bits of configuration for the Admin UI amongst the list and field configurations.

```keystone.createList('Todo', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    name: {
      type: Text,
      schemaDoc: 'This is the thing you need to do',
      isRequired: true,
      defaultValue: importViews(`./path-to-some-function-with-default-export`),
    },
  },
});
```   
Not only that, it requires a bunch of work to implement each time we want to do it. That why I'm moving towards a single configuration file for the Admin UI via hooks:

```
const adminApp = new AdminUIApp({
  hooks: require.resolve('./admin-ui/')
});
```

I think hooks is the correct API in future for any customisation of the Admin UI that needs to occur at runtime. In the index file for hooks we export a configuration:

```
import CustomDashboard from './dashboard';

export default {
  pages: () => [
    {
      label: 'A new dashboard',
      path: '',
      component: CustomDashboard,
    }
  ],
}
```

[...]

PR #2778, opened 2020-04-19
Corrected documented type for defaultValue

mjomble:
Updated the type in the table to match the extended description below


Vultraz:
So, interestingly, the mention of `Function` was deliberately removed from the table since there were various issues with `Function` values of `defaultValue`... (#2043, #2370)... except my testing seemed to indicate they were fixed in at least *some* cases (for non-required fields, at least). So... the current behavior definitely needs some clarification.

MadeByMike:
The base field class implements a `getDefaultValue` method that looks like this:

```
  getDefaultValue({ context, originalInput, actions }) {
    if (typeof this.defaultValue !== 'undefined') {
      if (typeof this.defaultValue === 'function') {
        return this.defaultValue({ context, originalInput, actions });
      } else {
        return this.defaultValue;
      }
    }
    // By default, the default value is undefined
    return undefined;
  }
```

This can be changed in the implementation for individual field types, but none of our fields currently change it.

This means `defaultValue` can really be anything. (`String` \| `Function` \| `null`) is not really accurate as many fields don't expect a string.

packages/fields/README.md
```diff
@@ -55,7 +55,7 @@ Fields share some standard configuration options.
 | -------------- | ----------------------------------- | ----------- | --------------------------------------------------------------------------------------- |
 | `type`         | `FieldType`                         | (required)  |                                                                                         |
 | `schemaDoc`    | `String`                            | `false`     | A description for the field used in the AdminUI.                                        |
-| `defaultValue` | `Boolean`                           | `undefined` | A default value of the field.                                                           |
+| `defaultValue` | `String` \| `Function` \| `null`    | `undefined` | A default value of the field.                                                           |
```
MadeByMike, line 58:
```suggestion
| `defaultValue` | `Any` \| `Function` | `undefined` | A valid default value for the field type. Functions must return a valid value. Use `undefined` to set no default, and `null` to sets an empty default.                                                    |
```

PR #2779, opened 2020-04-19
Admin UI: added a current user display to the sidebar

Vultraz:
Small thing I thought would be helpful.

image

And you can click the name to go right to the list item.


packages/app-admin-ui/client/components/Nav/index.js
```diff
@@ -289,8 +292,70 @@ function PrimaryNavItems({
...
+
+const UserIcon = styled.div`
+  flex-shrink: 0;
+  width: 40px;
+  height: 40px;
+  border-radius: 50%;
+  background-color: ${colors.primary};
+  color: white;
+  display: flex;
+  flex-direction: column;
```
MadeByMike, line 313:
```suggestion
```
With only one child item column is not doing anything here

packages/app-admin-ui/client/components/Nav/index.js
```diff
@@ -289,8 +292,70 @@ function PrimaryNavItems({
...
+      user: authenticated${authListKey} {
+        id
+        name
+      }
+    }
+  `;
+
+  // FIXME: figure out how best to handle a case where the user changed their name from the Admin UI.
+  // Perhaps a subscription once those are implemented?
+  const { data: { user } = {}, loading } = useQuery(AUTHED_USER_QUERY);
```
MadeByMike, line 332: Happy to live with this but will ask if anyone knows a good way to clear the apollo cache

packages/app-admin-ui/client/components/Nav/index.js
Vultraz: I needed it to get the `PersonIcon` vertically centered.

Vultraz: Ah, wait, I see what you mean. Removed.

Vultraz:
Hmm... now it's failing because it says the user the info box is fetching has no ID... 🤔

Vultraz:
Annnnd we're green! 🎉

PR #2780, opened 2020-04-19
Document the file field crash issue on node 13

Vultraz:
People keep running into #2101. The `resolutions` fix works when put in the top-level `package.json` (#2601), but I couldn't get it working in the `app-graphql` package file. Might as well at least document this until `apollo-server` either bumps their deps or v3 comes out.


MadeByMike:
I'd support this. Should we link to the Github issues as well? cc/ @timleslie

Vultraz:
Done.

PR #2786, opened 2020-04-20
Migrated to react-popper 2

Vultraz:
Refactored to use the new hooks-based API. The only thing I couldn't test was the Content field alignment tooltip since I couldn't get the Add Block button to show up 🤔 

Also removed the `react-popper` and `react-popper-tooltip` deps from the fields package since those were unused there.


jossmac:
> [...]couldn't test the Content field alignment tooltip since I couldn't get the Add Block button to show up

@MadeByMike perhaps you could point @Vultraz in the right direction?

PR #2794, opened 2020-04-21
Custom fields demos

MadeByMike:
Adds a new demo to help people creating custom fields.

Add an updated star field and a multi-check field.

image


demo-projects/custom-fields/fields/MultiCheck/views/Controller.js
```diff
@@ -0,0 +1,24 @@
+const FieldController = require('@keystonejs/fields/Controller').default;
```
mitchellhamilton, line 1: Why are you using CommonJS here?

MadeByMike: Because it borked out when I didn't 😆

mitchellhamilton: That's very strange. The Controller should be able to be ESM

pahaz:
https://github.com/keystonejs/keystone/issues/2851 related to this PR

PR #2798, opened 2020-04-22
Admin UI: restructured sidebar to made dev actions more noticeable

### LINKED ISSUE #1190: Make GraphQL Playground link more obvious in AdminUI
Since we only display the GraphQL Playground in dev mode, and it's a highly useful feature of the admin UI, we should surface it in a more obvious place for app developers to find.

Mockup:

Screen Shot 2019-05-20 at 11 11 02 am

Vultraz:
Resolves #1190 based on @jesstelford 's mockup. Open to suggestions as to the design, but I think it looks pretty good!

image


lucatk:
Quick question (also, the redesign looks really good I think!):
Currently, the sidebar does not have any background color, resulting in overlapping elements when List content ends up being wider than my browser window.

A quick fix would be adding a background color to the sidebar. I determined it would belong here:
```diff
<PrimaryNav
    onMouseEnter={handleMouseEnter}
    onMouseLeave={handleMouseLeave}
-   style={makeResizeStyles('width')}
+   style={{
+       ...makeResizeStyles('width'),
+       backgroundColor: colors.page,
+   }}
 >
```
I haven't worked out how to contribute with all the monorepo stuff and this being just a single line of code deemed it too much work to figure it out right now. Therefore, this PR seemed like a good place for that change (if the PR is going to be merged any time soon, otherwise a hotfix would be appreciated)

timleslie:
@Vultraz Are you able to follow up on the suggestion from @lucatk ? Let me know, I'll hold off merging this until you get back to me just so we don't lose the info.

Vultraz:
@timleslie done 👍

Vultraz:
How come Netlify is on the fritz today? 🤔

PR #2807, opened 2020-04-23
Fix bug querying one-to-one relationships with mongoose

timleslie:


packages/mongo-join-builder/lib/join-builder.js
```diff
@@ -57,7 +57,7 @@ function relationshipPipeline(relationship) {
   const { cardinality, columnNames } = rel;
   const extraPipeline = pipelineBuilder(relationship);
   const extraField = `${uniqueField}_all`;
-  if (cardinality === '1:N' || cardinality === 'N:1') {
+  if (cardinality !== 'N:N') {
```
jesstelford, line 60: If not `1:N`, `N:1`, or `N:N`, what _is_ the cardinality here?

timleslie: `1:1`

jesstelford: Of course. 🤦‍♂️

PR #2808, opened 2020-04-23
Fixed list-level `adminDoc` not actually doing anything

### LINKED ISSUE #494: IDEA: Add basic docs info to List definition which can be surfaced in the Admin-UI
It would be kinda of nice to be able to do something like this:

```js
keystone.createList('StudyTypes', {
    fields: {
        name: { type: Text },
    },
    docs: `Each study type defines an entire category of studies.
    Adding a new study type requires adding the appropriate UI components to MAP to support displaying the new study type.`
})
```

And then have that `docs` field content made available to the user within the admin UI. This is just a rough idea, and I'm sure there are lots of different ways we could make this happen. I'd be curious to think what other people think.

Vultraz:
Resolves #494.

#2768 added documentation for list-level `adminDoc`, but it turns out that was never actually implemented. `schemaDoc` was, and both keys were implemented for *fields*, but `adminDoc` was never added to lists. This rectifies that.

View in the Admin UI:
image


packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -97,8 +97,6 @@ export function ListLayout(props) {
   const Render = ({ children }) => children();
   return (
     <main>
-      <div ref={measureElementRef} />
```
Vultraz, line 100: Stray div, wasn't doing anything.

PR #2827, opened 2020-04-25
Admin UI: Refactored handling of list state

### LINKED ISSUE #2302: Admin UI makes duplicate API requests
# Bug report

## Describe the bug

The Admin UI makes at least 2 requests to the API when showing a list page.

In the worst case, this becomes a repeating request if the list is sorted by a field which will not return a consistent sort order.

## To Reproduce

This can be reproduced in a standard starter app "Starter (Users + Authentication)" created with `yarn create keystone-app`

1. Create and start the app

Screen Shot 2020-01-30 at 10 59 14 am

2. Sign in to the new app
3. Open the browsers Network panel 
4. Navigate to `http://localhost:3000/admin/users`
5. In the network panel you will see 2 identical api requests

Screen Shot 2020-01-30 at 11 01 51 am

Screen Shot 2020-01-30 at 11 02 10 am

## Expected behaviour

Expected only one api request.

## System information

- OS: Mac
- Node v12.6.0
- Browser: reproduced on Chrome, Firefox and Safari for Mac
- Package Versions:
```
    "@keystonejs/adapter-mongoose": "^5.1.3",
    "@keystonejs/app-admin-ui": "^5.3.0",
    "@keystonejs/app-graphql": "^5.0.1",
    "@keystonejs/auth-password": "^5.0.1",
    "@keystonejs/fields": "^6.0.0",
    "@keystonejs/keystone": "^5.3.0",
```

## Additional context

The more problematic case is harder to reproduce ... but if the list is sorted by a field which will not give a consistent sort order, the request will repeat indefinitely. This appears to only occur for requests which take more than 1 or 2 seconds to resolve.

In my use case, we have a table with ~166000 items (using Postgres). One of the fields is a text field. When I set a `contains` filter for that field and have the sort order set to a Relationship field we see infinite api requests about 1 second apart. ie. as each request completes, a new one begins.

### LINKED ISSUE #2613: Search bar in UI Admin List view doesn't work
# Bug report

## Describe the bug

The Search component in the UI Admin List View is unresponsive. When user attempts to interact with it, React throws the following error:

```
Warning: Cannot update a component (`Search`) while rendering a different component (`Search`).
To locate the bad setState() call inside `Search`, follow the stack trace as described in
https://fb.me/setstate-in-render
```

## System information

This happens on MacOS Catalina Chrome. Current Keystone installs look like this

```
"@keystonejs/adapter-mongoose": "^5.2.1",
"@keystonejs/app-admin-ui": "^5.9.2",
"@keystonejs/app-graphql": "^5.1.2",
"@keystonejs/auth-password": "^5.1.2",
"@keystonejs/fields": "^7.0.0",
"@keystonejs/keystone": "^6.0.1",
```

Vultraz:
Fixes #2613
Fixes #2302

Turns out both of those bugs were caused by the same underlying issue. The client was running multiple `useQuery`s in different places to fetch list data.

This expands the `ListProvider`'s functionality, making it the sole component responsible for calling `useQuery`. ~It also stores a copy of the current decoded URL state.~

These data hooks were removed:
- `useList`
- ~`useListUrlState`~
- `useListQuery`
- `useListItems`

None of the data hooks take a `listKey` argument anymore. Instead, they use the current list as given by the provider. A few components that also took a `listKey` prop just to pass it to the data hooks have also had that removed.


JedWatson:
@Vultraz looks good! just reviewed, would you mind checking on the failing cypress test before I approve it though?

Vultraz:
Ok, those failures are odd... everything seems to work fine here... Is there some way to actually view the screenshots Cypress spits out?

JedWatson:
You should be able to see them in the **artifacts** tab in the CircleCI UI here: https://app.circleci.com/pipelines/github/keystonejs/keystone/3848/workflows/661dde8f-f7ae-4c51-9349-a52078c9ee19/jobs/19147/artifacts

It does look (from the screenshots) like there are issues refreshing the list after a change has been made (or at least, cypress isn't waiting for the reload maybe?)

e.g https://19147-128193054-gh.circle-artifacts.com/0/tmp/screenshots/add-edit-delete-data.js/Deleting%20data%20--%20Deleting%20data%20to%20Posts%20%28failed%29.png

Vultraz:
Ok, so I can repro the delete issue locally. Investigating.

Vultraz:
Turns out the delete issue wasn't in my new code. T'was an issue my new code exposed 😬

gautamsi:
I like this, this also means refactor to all some of my pending PR :)

Vultraz:
Ok, the editing tests work now with the addition of a `wait`. Not entirely sure why that became necessary. Trying to figure out the last one.

Vultraz:
Huh! Ok, we're green. I just added delays to the tests 😬 Whether that's a good fix or not, not entirely sure.

Vultraz:
EDIT: after giving it a day to think, I decided to restore the `useListUrlState` since it's not really part of the state of the list, but rater the list influences the URL state.

MadeByMike:
Finally got around to testing this locally. LGTM :)

Issue #2836, opened 2020-04-26
Can't build AdminUI properly


lucatk:
# Bug report

## Describe the bug

I'm trying to contribute to the project. I wanted to try out some changes on the admin-ui.
I've read the contributing guidelines.

## To Reproduce

Steps to reproduce the behaviour. Please provide code snippets or a repository:

1. git clone project
2. Ran `yarn`
3. Ran `yarn` in packages/app-admin-ui
4. Made changes to app-admin-ui
5. Ran `yarn build`
6. Substituted @keystonejs/app-admin-ui package in my project for my local version (using `yarn add PATH_TO_LOCAL_CLONE/packages/app-admin-ui`
7. Ran yarn start:dev, navigated to admin UI in browser
8. See error

## Expected behaviour

The admin UI to work.

## Screenshots

![](https://snipboard.io/hjxVa2.jpg)

## System information

- OS: macOS Catalina
- Browser (if applies): Google Chrome 80.0.3987.163

## Additional context

The build and KeystoneJS startup seems to work like normal (no errors)

Vultraz:
The way I've been contributing is via running one of the demo projects. They're symlinked to the packages in `/packages`, so you can clone the monorepo and test out your changes that way.

1. Clone repo and create your branch.
2. Make changes directly to `packages/app-adminu-i/
3. In root `package.json`, change the `demo` script to `"demo": "yarn --cwd ./demo-projects",` (needed because of #2632 )
4. Run `yarn dev` to launch the default demo (Todo List) to test your changes, or `yarn demo $demoName dev` to pick another.

gautamsi:
@Vultraz you can make us of your own projects using a `projects` folder. see https://github.com/keystonejs/keystone/pull/1832

Vultraz:
....huh! That I did not know. 😱

MadeByMike:
@lucatk this is likely because of the workspace configuration. There are ways around it but a better way would be to edit the admin-ui package directly on your own branch. Copy it somewhere if you'd like to reference the code. Also hit my up on the slack https://community.keystonejs.com/ if you'd like some help

PR #2842, opened 2020-04-27
Include projects folder in the repo while ignoring sub folders

gautamsi:
improvement over #1832.

This adds `projects` folder in repo with readme and (git) ignores any other files/folders in projects folder. It is already included in workspace but not physically present

would be useful as many did not know it is there at their disposable and they have to use [symlinked and other hacks to get it working. ](https://github.com/keystonejs/keystone/issues/2836#issuecomment-619652717)  @Vultraz 

> Someone using this will need to make sure they are not polluting yarn.lock due to additional packages being used in those projects.

Don't think a changeset is needed but can add that if required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,5 +109,4 @@ tags
 
 # End of https://www.gitignore.io/api/vim
 
-projects/
 temp/
--- a/projects/.gitignore
+++ b/projects/.gitignore
@@ -1,0 +1,3 @@
+*
+!.gitignore
+!README.md
--- a/projects/README.md
+++ b/projects/README.md
@@ -1,0 +1,5 @@
+# projects folder
+
+You can add your projects in this folder to work with keystone monorepo, contribute to Keystone but still wants to have separate test projects.
+
+This folder is not added to git except the .gitignore and README.md file.


Vultraz:
For the record, I've been directly working with the Todo demo since I've *only* been developing KS, not a project of my own, but this should be helpful once I get to actually doing that. 👍

And about the symlinks, BTW, I didn't set those up myself. KS handles that internally.

MadeByMike:
I added this to the main `.gitignore`months ago: `demo-projects/test-project` for this purpose.

I copied the todo and renamed it `test-project`, I'm a bit worries now that we are going to have multiple ways of doing this and #1832. It really up to individual contributors where and how they want to develop and if we are going to offer something like this I say we offer one method.

I like it in the demo-project because it's automatically set-up as a workspace... but I don't have strong preferences.

If we think a readme is required, let work out where the main project dev folder should be and add it there.

I'm less in favour of adding more folders to the root.

gautamsi:
`projects` is also in set up in workspace.

MadeByMike:
> `projects` is already added in workspace config as part of earlier PR..

#1832 only adds the folder to `.gitignore` which doesn't actually create a folder. This PR actually adds a folder to the root of the workspace.

I'm not convinced that we should be signaling to contributors that "this is where we want you to develop projects in the mono-repo". AND if we do want to do that, I'd like to discuss if the demo-projects folder is better as it is already set-up with yarn workspaces and I think encourages people to think in terms of demo-projects, and encouraging the use the other demo-projects for development. 

But that said, I'm not sure we should do either. It's really an individual preference that I don't think Keystone should have a strong opinion about.

gautamsi:
I don't have any say either. anyone searching would know what are their option when they land here :)

PR #2843, opened 2020-04-28
Updated cloudinary to 1.21.0

Vultraz:
Fixes a whole bunch of the "Accessing non-existent property [...] of module exports inside circular dependency" warnings on startup when using Node 14.


Vultraz:
Netlify still unhappy... :/

PR #2853, opened 2020-04-29
Upgrade dev dependencies

timleslie:


Vultraz:
BTW, `mongodb` 3.5.7 was just released. It fixes more of those "Accessing non-existent property of module exports inside circular dependency" warnings

PR #2864, opened 2020-04-30
Fix for incorrect Content-Type headers returned for Admin UI bundles

### LINKED ISSUE #2741: Keystone Admin UI serves incorrect MimeType
# Bug report

## Describe the bug

Currently, bundles returned for the Admin UI (eg. `[/admin/js/main.b84baf6b30d5f4898129.bundle.js](http://localhost:3000/admin/js/main.b84baf6b30d5f4898129.bundle.js)`) are returned to the browser with a `content-type: */*` header. This occurs in dev and production modes.

Usually this doesn't cause a problem; I believe this is because browsers have `strict MIME type checking` disabled by default. If a nginx config on the server specifically turns this checking on:
```
    # disallow circumventing declared MIME types
    add_header  X-Content-Type-Options nosniff;
```

It fails to load.

So, the combination of the hardened nginx config and Keystone's sloppy headers causes an error.

It refuses to execute the script because its MIME type ('*/*') is not executable, and strict MIME type checking is enabled.

molomby:
Fixes #2741.

When the Express `res.format()` function matches a request (to a non-default function) it replaces the responses `Content-Type` header with the key it matched on. This causes our `*/*` format handler (which is basically a noop) to return requests with header of `Content-Type: */*`. This in turn causes problems if a `X-Content-Type-Options: nosniff` header is also set on the request (ie. somewhere else in the Express middleware, by a reverse proxy, etc.).

We can fix this by resetting the `Content-Type` header with a mime-type derived from the requested files extension.

Unfortunately, despite being quite redundant, we can't just remove the `*/*` handler. Doing so causes requests with an `Accept: */*` header or no `Accept` at all to fall though to the `text/html` handler, rather than the default.

The `default` handler itself doesn't require this additional logic as Express has already given the response a correct `Content-Type` header and doesn't override it when the `default` format handler is used. If the `text/html` handler is hit, the overridden `Content-Type` is correctly set to `text/html`.

### DIFF
--- a/packages/app-admin-ui/index.js
+++ b/packages/app-admin-ui/index.js
@@ -160,7 +160,13 @@ class AdminUIApp {
         default: () => {
           next();
         },
+        // Without this, request with an 'Accept: */*' header get picked up by
+        // the 'text/html' handler rather than the default
         '*/*': () => {
+          // We need to reset the res 'Content-Type' otherwise it gets replaced by the format we've matched on: '*/*'.
+          // Returning a wildcard mimetype causes problems if a 'X-Content-Type-Options: nosniff' header is also set.
+          // See.. https://github.com/keystonejs/keystone/issues/2741
+          res.type(path.extname(req.url));
           next();
         },
         // For page loads, we want to redirect back to signin page
--- a/packages/app-admin-ui/index.js
+++ b/packages/app-admin-ui/index.js
@@ -166,7 +166,8 @@ class AdminUIApp {
           // We need to reset the res 'Content-Type' otherwise it gets replaced by the format we've matched on: '*/*'.
           // Returning a wildcard mimetype causes problems if a 'X-Content-Type-Options: nosniff' header is also set.
           // See.. https://github.com/keystonejs/keystone/issues/2741
-          res.type(path.extname(req.url));
+          const extension = path.extname(req.url);
+          if (extension) res.type(extension);
           next();
         },
         // For page loads, we want to redirect back to signin page
--- a/.changeset/lazy-clocks-try.md
+++ b/.changeset/lazy-clocks-try.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Static assets loaded for the Admin UI (eg. JS bundles) now have correct Content-Type response headers


packages/app-admin-ui/index.js
```diff
@@ -160,7 +160,13 @@ class AdminUIApp {
         default: () => {
           next();
         },
+        // Without this, request with an 'Accept: */*' header get picked up by
+        // the 'text/html' handler rather than the default
         '*/*': () => {
+          // We need to reset the res 'Content-Type' otherwise it gets replaced by the format we've matched on: '*/*'.
+          // Returning a wildcard mimetype causes problems if a 'X-Content-Type-Options: nosniff' header is also set.
+          // See.. https://github.com/keystonejs/keystone/issues/2741
+          res.type(path.extname(req.url));
```
jesstelford, line 169: What happens when `req.url` = `/foo/bar`?

Wouldn't we want the content type to be `text/html` then? Or maybe `*/*` to let the client decide?

dominikwilkowski: Don't we control 100% of the requests? In the adminUI... which so far I can only see `/admin/js/*`

jesstelford: Maybe something like this?
```suggestion
          const extension = path.extname(req.url);
          res.type(extension ? extension : '*/*');
```

jesstelford: @dominikwilkowski This express app handles _everything_ under `/admin`, which could include custom pages. Not to mention the pages such as `/admin/todos` which doesn't have an extension.

molomby: Great point. So...

1. If I request `/admin/blerg` from the browser (address bar), the browser (Chrome in this case) adds a sensible `Accept` header of..

```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
```

Which works fine as it matches the `text/html` resolver (returns with `Content-Type: text/html; charset=utf-8`).

2. If I request `/admin/blerg` _without_ an `Accept` header, it hits the `*/*` format resolve and the `Content-Type` is returned as `application/octet-stream`. Presumably this is some kind of default that's applied since we're passing `res.type()` and empty string.

3. If I request with an unrecognised `Accept` header (`food/pasta`) it hits the `default` format resolver, returning with `Content-Type: text/html; charset=UTF-8` 
which is correct for the content -- the main html page for the admin UI.

This all seems to be true regardless of the depth of the path, trailing slashes, etc.

Point 2 is probably wrong. The content being returned is actual html; the incorrect mime type here is our fault -- I've patched to only override if a file extension is present, meaning it will return `*/*` as before.

Note this still isn't totally correct. If we removed the `res.format()` block entirely, these requests would return `text/html`. But as you say, at least the browser can decide (and it's still better than we have now).

PR #2855, opened 2020-04-29
Consolidated #2427 and #2497 (functional components)

Vultraz:
Consolidated the two for easier review. The latter had some stuff (moving route props) that's now covered in #2827, and I figured out what to do with the nonexistent `innerRef` in #2731. Should be good to go.

PR #2870, opened 2020-05-01
[WIP] Explicit fields loader

timleslie:
This PR makes explicit exactly which values are passed through to the `fields-views-loader` function. It also removes `hooks` and `pages` from the `adminMeta` passed to the client, since they were never actually used there.

(Builds on top of #2869, which should be merged before this is considered).

### DIFF
--- a/packages/app-admin-ui/index.js
+++ b/packages/app-admin-ui/index.js
@@ -47,21 +47,6 @@ class AdminUIApp {
     };
   }
 
-  getAdminMeta() {
-    return {
-      adminPath: this.adminPath,
-      pages: this.pages,
-      hooks: this.hooks,
-      ...this.routes,
-      ...(this.authStrategy
-        ? {
-            authStrategy: this.authStrategy.getAdminMeta(),
-          }
-        : {}),
-      ...this._adminMeta,
-    };
-  }
-
   isAccessAllowed(req) {
     if (!this.authStrategy) {
       return true;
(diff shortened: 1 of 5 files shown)


gautamsi:
Do not remove the hooks and pages from `getAdminUIMeta` they are in tern fed to `AdminMetaProvider`.

I also see that you have factored this out into `adminViews` along with `adminMeta`, solution would be to move this `adminViews` into `AdminMetaProvider` or create another context with `AdminViewsProvider` and wrap admin-ui client in that.

I like the listViews which would be helpful, finally heading in right direction.

packages/app-admin-ui/index.js
```diff
@@ -120,16 +122,14 @@ class AdminUIApp {
...
-    const { adminPath, apiPath, graphiqlPath, pages, hooks } = this;
+    const { adminPath, apiPath, graphiqlPath } = this;
     const { signinPath, signoutPath } = this.routes;
     const { lists, name } = keystone.getAdminMeta({ schemaName: this._schemaName });
     const authStrategy = this.authStrategy ? this.authStrategy.getAdminMeta() : undefined;
     return {
       adminPath,
       apiPath,
       graphiqlPath,
-      pages,
```
gautamsi: this is also part of `AdminMetaProvider` which I use in several PR as well. would be helpful in customising admin-ui.

packages/app-admin-ui/index.js
```diff
@@ -281,6 +292,7 @@ class AdminUIApp {
     const secureCompiler = webpack(
       getWebpackConfig({
         adminMeta,
+        adminViews,
```
gautamsi, line 295: I hope this end up in `AdminMetaProvider` or a new `AdminViewsProvider`

timleslie:
This is all still really WIP, don't take anything here as gospel just yet until I have a chance to get to the bottom of where all these things go (it's all very non-obvious right now, which is making it difficult to add anything new). Nothing that's actually used on the client side will be removed once it's complete.

gautamsi:
I usually see things in WIP and next morning I wake up and it is merged and published (timezone to blame), so putting my thoughts early.

Vultraz:
I was also working on something to do with the adminMeta! :P I noticed that not only is it accessible from the provider with `useAdminMeta`, it's also saved in every Admin UI `List` object AND in every list's field controller. I'm trying to figure out how best to clean up some of that duplication.

packages/app-admin-ui/index.js
Vultraz: I think that's the old `pages` config. People should be using the new `pages` hook instead of the direct config option, shouldn't they?

gautamsi: yeah, the hooks should be enough.

Vultraz: Indeed. Might also be worth adding a deprecation/removal notice for `pages` since technically its functionality was never removed until now.

Vultraz:
@timleslie BTW, these can all be converted to `useAdminMeta` once #2871 and #2855 get merged (they convert the remainder of these - except `AddFilterPopup` - to functional components so we can use the hook):
image

PR #2871, opened 2020-05-01
Admin UI: Improvements to Update workflow

Vultraz:
- Converted `UpdateManyModal` to a functional component. Dropped `UpdateManyModalWithMutation` in the process.
- Cleaned up `ListManage` a bit
- Surface update errors in a toast. Same behavior as the Create modal
- Ensured an error wouldn't dismiss the Update popout
- Fixed an error making it impossible to deselect the final selected field

### DIFF
--- a/.changeset/chilly-adults-search.md
+++ b/.changeset/chilly-adults-search.md
@@ -2,4 +2,4 @@
 '@keystonejs/app-admin-ui': patch
 ---
 
-Improved Update workflow.
+Refactored `UpdateManyModal` component.
--- a/.changeset/chilly-adults-search.md
+++ b/.changeset/chilly-adults-search.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Improved Update workflow.
--- a/packages/app-admin-ui/client/components/Nav/index.js
+++ b/packages/app-admin-ui/client/components/Nav/index.js
@@ -383,7 +383,7 @@ const ActionItems = ({ mouseIsOverNav }) => {
       ...(ENABLE_DEV_FEATURES
         ? [
             {
-              label: 'GraphiQL Playground',
+              label: 'GraphQL Playground',
               to: graphiqlPath,
               icon: TerminalIcon,
               target: '_blank',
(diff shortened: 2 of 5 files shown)


packages/app-admin-ui/client/components/Nav/index.js
```diff
@@ -383,7 +383,7 @@ const ActionItems = ({ mouseIsOverNav }) => {
       ...(ENABLE_DEV_FEATURES
         ? [
             {
-              label: 'GraphiQL Playground',
+              label: 'GraphQL Playground',
```
Vultraz, line 386: Unrelated, I just happened to realize that GraphiQL =/= GraphQL Playground.

MadeByMike: ¯\_(ツ)_/¯ OK

packages/app-admin-ui/client/pages/List/Management.js
```diff
@@ -17,74 +17,75 @@ export const ManageToolbar = styled.div(({ isVisible }) => ({
...
-const SelectedCount = styled.div({
-  color: colors.N40,
-  marginRight: gridSize,
-});
 
-export default function ListManage(props) {
-  const { onDeleteMany, onUpdateMany, selectedItems } = props;
-  const [deleteModalIsVisible, setDeleteModal] = useState(false);
-  const [updateModalIsVisible, setUpdateModal] = useState(false);
+const SelectedCount = styled.div`
```
MadeByMike, line 21: Why did we change to template functions here? We use object notation throughout most of the codebase. I'd rather be consistent.

.changeset/chilly-adults-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Improved Update workflow.
```
MadeByMike, line 5:
```suggestion
Refactored `UpdateManyModal` component.
```

MadeByMike:
I still like the idea of keeping the `updateItem` function separate from the UI. The real problem is we need some more basic UI components here rather than this large mess. 

I'm inclined to move forward with this for now

packages/app-admin-ui/client/pages/List/Management.js
Vultraz: There's no consistent usage of `styled` object notation vs template literals. Lot of both. I've been gradually switching to the latter as I've come across it. Haven't touched the function-based ones since I only yesterday realized that they can return a template literal too 😬

PR #2874, opened 2020-05-02
Admin UI: explicit list config

Vultraz:
The client-side `List` objects were storing a *ton* of duplicated and unnecessary data.

Here's what one of the Admin UI `List` objects looked like before:
image

Notice how not only was the *entire* `adminMeta` object duplicated in every list, but the `config` object was being saved despite being `assigned`ed to the `List` object. This meant that the entire config object was being saved twice, *and* it also meant completely unnecessary keys like `views` were being saved too (these are used by the field view loader and aren't needed here).

This removes the copy of `adminMeta` (it wasn't used anywhere) and the `config` copy + assignment in favor of explicitly picking out the pieces of `config` we want to save. Note this does mean the raw field config is no longer accessible at `list.config.fields`, but it's still available via the field controller in `list.fields`.

Lot cleaner!

### DIFF
--- a/.changeset/curly-llamas-compare.md
+++ b/.changeset/curly-llamas-compare.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Cleaned up duplicated and/or unnecessary list config data.
--- a/packages/app-admin-ui/client/classes/List.js
+++ b/packages/app-admin-ui/client/classes/List.js
@@ -7,14 +7,22 @@ export const gqlCountQueries = lists => gql`{
 }`;
 
 export default class List {
-  constructor(config, adminMeta, views) {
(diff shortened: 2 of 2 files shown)


MadeByMike:
I think this is ok. @timleslie has been looking at some stuff parallel to this so he might want to get eyes on this one as well.

jesstelford:
Possibly related: https://github.com/keystonejs/keystone/issues/1039

timleslie:
Yes, love this, very much in line with the PRs I've got lined up to make all the `adminMeta` more explicit and more constrained so that we can get a really tight grip on this API surface area. I'm still doing a bit of fine tooth comb work to understand all the implications of this change and be 100% confident nothing will break, but I'm very hopeful to give this a ✅ today.

PR #2887, opened 2020-05-05
Use the getRefList method in more places

timleslie:
All part of my effort to have fewer references to the borg `adminMeta` object :-)

### DIFF
--- a/packages/fields/src/types/Relationship/views/Field.js
+++ b/packages/fields/src/types/Relationship/views/Field.js
@@ -56,12 +56,12 @@ function SetAsCurrentUser({ listKey, value, onAddUser, many }) {
 }
 
 function LinkToRelatedItems({ field, value }) {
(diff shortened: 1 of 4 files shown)


packages/fields/src/types/Relationship/views/Field.js
```diff
@@ -56,12 +56,12 @@ function SetAsCurrentUser({ listKey, value, onAddUser, many }) {
 }
 
 function LinkToRelatedItems({ field, value }) {
-  const { many, ref } = field.config;
-  const { adminPath, getListByKey } = field.adminMeta;
-  const refList = getListByKey(ref);
+  const { many } = field.config;
+  const { adminPath } = field.adminMeta;
+  const { path } = field.getRefList();
```
MadeByMike, line 61: This is so much nicer

PR #2888, opened 2020-05-05
[WIP] Pass readView through to field controllers

timleslie:
More removal of reliance on the full `adminMeta` object 👍

### DIFF
--- a/packages/fields/src/types/OEmbed/views/blocks/o-embed.js
+++ b/packages/fields/src/types/OEmbed/views/blocks/o-embed.js
@@ -21,7 +21,7 @@ const Embed = ({ url, oembedData }) => {
 
   if (options.previewComponent) {
     // The adapter should implement this option
-    const [Preview] = options.adminMeta.readViews([options.previewComponent]);
+    const [Preview] = options.readViews([options.previewComponent]);
     return <Preview url={url} options={options} />;
   } else {
     // This is a fallback so we can at least try to render _something_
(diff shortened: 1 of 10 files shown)


Vultraz:
Idea, actually. Make `initCellView`, `initFieldView`, and `initFilterView` return their result and then change all the cases of `field.readViews([field.views.Field])`, etc to just `field.initFieldView`, etc. They seem to do the same thing.

packages/fields/src/types/OEmbed/views/blocks/o-embed.js
```diff
@@ -21,7 +21,7 @@ const Embed = ({ url, oembedData }) => {
 
   if (options.previewComponent) {
     // The adapter should implement this option
-    const [Preview] = options.adminMeta.readViews([options.previewComponent]);
+    const [Preview] = options.readViews([options.previewComponent]);
```
Vultraz, line 24: I think this is wrong. The blocks are passed their own copy of `adminMeta` here: https://github.com/keystonejs/keystone/blob/master/packages/field-content/src/views/Controller.js#L57 . You'll need to explicitly pass in `readViews`.

Once #2890 is merged the `adminMeta` copy in `options` can be removed entirely, I think.

MadeByMike: Tested #2890 👍

PR #2890, opened 2020-05-05
Refactored out use of fetch in Unsplash block

Vultraz:
Note this is currently **untested** (I don't have a Content + Unsplash setup).

@timleslie this removes that pesky stray use of `adminMeta.apiPath`.


timleslie:
Very nice, it's like you read my mind. I saw this code yesterday and was very much wondering why it was using `fetch`.

PR #2892, opened 2020-05-05
Explicitly unpack List config items from adminMeta

timleslie:


packages/app-admin-ui/client/providers/AdminMeta.js
```diff
@@ -63,11 +62,17 @@ export const AdminMetaProvider = ({ children }) => {
...
+    ([key, { access, adminConfig, adminDoc, fields, gqlNames, label, path, plural, singular }]) => {
+      const list = new List(
+        { access, adminConfig, adminDoc, fields, gqlNames, key, label, path, plural, singular },
+        adminMeta,
+        views[key]
+      );
+      listsByKey[key] = list;
+      listsByPath[list.path] = list;
+    }
+  );
```
MadeByMike, line 75: Not sure I follow why this expansion. In this specific case I found it harder to reason about. ...not that I mind much.

timleslie: The idea here is to make it explicit about what things we're taking out of the `adminMeta` object and where we're sending them. Once all of this is complete I want to know precisely how we're using the `adminMeta` object. The reason this is so important is that it's particularly wacky the way this thing gets passed across to the client via webpack, so it helps to have it all laid out what we're doing with it, rather than having the whole object being shipped off to other parts of the code base.

PR #2897, opened 2020-05-05
Explicit FieldController config, elevated some keys

Vultraz:
This adds `isOrderable`, `isRequired`, and `adminDoc` as top-level `FieldController` properties (previously they were only available in `config` despite other keys like `isPrimaryKey` being explicitly set. It also removes the duplicate keys between `field` and `field.config`.

This leaves `field.config` containing only type-specific config options (see below, for example) plus any options the user passed through with field-level `adminConfig`.

image


Vultraz:
Forgot to commit it 🤦

Issue #2867, opened 2020-04-30
Webpack can't resolve imported css files in custom pages.


LiamAttClarke:
# Bug report

## Describe the bug

Importing a css file in a custom page no longer works with the latest release (`@keystonejs/app-admin-ui v5.12.0`).

## To Reproduce

1. Create a [custom page](https://www.keystonejs.com/keystonejs/app-admin-ui/#pages).
2. Import a local css file.
Example:
```jsx
import './CustomPage.css';

export default () => <div className="some-class"></div>;
```
3. Start Keystone and open in the browser.

## Expected behaviour

The css file should be resolved by Webpack as it did in previous versions of `app-admin-ui`.

## Screenshots

![Annotation 2020-04-30 172950](https://user-images.githubusercontent.com/9361948/80762282-37c7db80-8b0a-11ea-9fd6-40522de696a0.jpg)

## System information

- OS: Windows

## Additional context

@Nikodermus Suggested this might be caused by his recent change (https://github.com/keystonejs/keystone/pull/2860). Perhaps this implicitly excludes local css file resolution?

gautamsi:
this should be reverted and investigated on why the css from `node_modules` not included when there is no `exclude` option specified.

MadeByMike:
Resolved in #2860

PR #2860, opened 2020-04-29
fix(packages/keystonejs getwebpackconfig.js): include node_modules stylesheets in webpack build

Nikodermus:
Related to #2859

Issue #2859, opened 2020-04-29
node_module css stylesheet not being loaded in production


Nikodermus:
# Bug report
Stylesheets loaded as imports inside a custom page, are not bundled in a production build.

## Describe the bug

In a custom page, I have a CSS import in the way of
```javascript
import '@secret/venue-map-component/dist/VenueMap.css';
```

This works in development mode, but the stylesheet is missing in the production build

## Expected behaviour

node_modules stylesheet should be added in production build as with local css stylesheets.

## Screenshots
Dev
![Screen Shot 2020-04-29 at 3 35 35 PM](https://user-images.githubusercontent.com/7080621/80647751-9fabf280-8a34-11ea-914c-79c24d8bdcd1.png)
Prod
![Screen Shot 2020-04-29 at 3 35 04 PM](https://user-images.githubusercontent.com/7080621/80648729-48a71d00-8a36-11ea-96aa-6f1455a78f3a.png)

## System information

- macOS 10.14 (18A391)
- Browser: All

Nikodermus:
Fixed in #2860

MadeByMike:
Resolved: #2860

PR #2904, opened 2020-05-05
[App-admin-ui] Revert change in getWebpackConfig.js

Nikodermus:
I tried to fix this issue #2859 with the PR #2860

But ended up causing #2867 

I'll do a more in depth investigation and send a more proper PR for that, I'd probably need some help on how to test the changes for the production build

cc @LiamAttClarke @gautamsi


LiamAttClarke:
@Nikodermus Thanks for following up on this!

PR #2906, opened 2020-05-06
Pass named config through to Controller

timleslie:
Supersedes #2888

### DIFF
--- a/packages/fields/src/Controller.js
+++ b/packages/fields/src/Controller.js
@@ -14,7 +14,7 @@ export default class FieldController {
       defaultValue,
       ...config
     },
-    adminMeta,
+    { readViews, preloadViews, getListByKey, adminPath, authStrategy },
     views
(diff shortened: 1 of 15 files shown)


packages/fields/src/Controller.js
```diff
@@ -3,19 +3,23 @@ import isEqual from 'lodash.isequal';
...
-    adminMeta,
+    { readViews, preloadViews, getListByKey, adminPath, authStrategy },
     views
   ) {
     this.config = config;
-    this.label = label;
-    this.path = path;
-    this.type = type;
-    this.maybeAccess = access;
+    this.label = config.label;
```
Vultraz, line 10: Shouldn't pulling from `config` here (already destructured).

timleslie: Yep, merge error, fixed now :+1:

PR #2916, opened 2020-05-07
Cleaned up handling of defaultValue in field controllers

Vultraz:
- `defaultValue` is no longer saved in `field.config`.
- Cleaned up field-specific default handling
- Removed a redundant check (`defaultValue` will always be `undefined` if not specified in the passed-in config, so no need for an extra condition)

### DIFF
--- a/.changeset/quick-melons-occur.md
+++ b/.changeset/quick-melons-occur.md
@@ -1,6 +1,6 @@
 ---
 '@keystonejs/fields': patch
+'@keystonejs/field-content': patch
 ---
 
 Cleaned up handling of defaultValue in field controllers.
-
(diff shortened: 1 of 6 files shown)


.changeset/quick-melons-occur.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/fields': patch
```
timleslie, line 2:
```suggestion
'@keystonejs/field-content': patch
```

timleslie: Boo, I'm bad at github, that's meant to be adding `field-content`, not replacing `fields` 🤷‍♂️

Vultraz: Done.

Vultraz:
`field-content` in the changeset.

PR #2919, opened 2020-05-07
Made handling of field-specific config options more consistent

Vultraz:
Brings these in line with the rest of the fields.


packages/fields/src/types/Virtual/Implementation.js
```diff
@@ -4,27 +4,33 @@ import { KnexFieldAdapter } from '@keystonejs/adapter-knex';
...
-    return { [`${this.path}`]: this.config.resolver };
+    return { [`${this.path}`]: this.resolver };
   }
   gqlQueryInputFields() {
     return [];
   }
   extendAdminMeta(meta) {
     return {
       ...meta,
-      isOrderable: false,
```
Vultraz, line 26: No need to set `isOrderable` here, `false` is the default in the base.

Issue #2565, opened 2020-03-21
Date picker for CalendarDay or DateTimein AdminUI?


ra-external:
I've seen this referenced in other issues (for instance https://github.com/keystonejs/keystone/issues/2366), but I don't know if there's been any movement on this.  It's a fairly large inconvience at our end, since the client isn't happy having to enter date strings 'manually'.  Is there any hope of a change in the near future?  The issues seemed to indicate that a fairly simple workaround was possible, using the native date picker.

MadeByMike:
This is a duplicate of #2185, and #2366.

Totally understand that this is an important issue and it's high on our priorities.  Please consider contributing to https://github.com/keystonejs/keystone/pull/2922 where we've broken down the work required into a number of smaller tasks.

Issue #2366, opened 2020-02-14
[AdminUI] CalendarDay: Date picker doesn't work...


maximelebreton:
Hello keystone Team,
and thanks for your amazing work.

It's a real pleasure to discover this framework, I'm very excited about it!

I have this strange problem: With the CalendarDay field (and the other date fields), there is **no Date Picker**...
Only a simple text input (So I suppose this is not normal).
And no errors in console...

Any ideas?

![todo](https://user-images.githubusercontent.com/1072425/74569585-06178d80-4f7b-11ea-8fc1-d3ca56896a87.gif)

Versions:
```
    "@keystonejs/adapter-mongoose": "^5.1.3",
    "@keystonejs/app-admin-ui": "^5.3.0",
    "@keystonejs/app-graphql": "^5.0.1",
    "@keystonejs/app-static": "^5.0.0",
    "@keystonejs/fields": "^6.0.0",
    "@keystonejs/keystone": "^5.3.0",
```

Thanks!

Vultraz:
That's expected. There's no built-in date picker for CalendarDay yet.

maximelebreton:
Oh!
And any plans to implement it?
Or a way to customize it by myself?
It's very problematic for the project Im working on now.

Thanks for your help

maximelebreton:
Just adding `type="date"` on the `input` tag would do the job, no?

![date](https://user-images.githubusercontent.com/1072425/74576253-564a1c00-4f8a-11ea-9a7e-8ad54f718e2a.gif)

maximelebreton:
https://github.com/keystonejs/keystone/pull/2367

Vultraz:
I was working on a date picker impl last year (#1965) but I didn't finish 😬

MadeByMike:
This is a duplicate of #2185.

This is an important issue and we are working on it. Please consider contributing to https://github.com/keystonejs/keystone/pull/2922 where we've broken down the work required into a number of smaller tasks.

Issue #1136, opened 2019-05-14
Default values as function


dcousens:
When defining default values it would be useful to allow a functions in addition to static values. For example:

### Current:
``` js
dob: {
    type: CalendarDay,
    format: 'Do MMMM YYYY',
    yearRangeFrom: 1901,
    yearRangeTo: getYear(new Date()), // Gets the date when Keystone is started
},
```

### Suggested:

``` js
dob: {
    type: CalendarDay,
    format: 'Do MMMM YYYY',
    yearRangeFrom: 1901,
    yearRangeTo: () => getYear(new Date()), // Get the date when a user creates a new item
},
```

MadeByMike:
We need to do a better job of describing what this is less technical way. "Default value initialization as lambdas" is not very welcoming for community contributions. @dcousens would you mind having a go at describing what this enables for developers\users?

dcousens:
> Default value initialization functions

MadeByMike:
Let me update it then :p

bladey:
**Keystone 5** has officially moved into active maintenance mode as we push towards the next major new version **Keystone Next**, you can find out more information about this transition [here](https://github.com/keystonejs/keystone/issues/4962).

In an effort to sustain the project going forward, we're cleaning up and closing old issues such as this one. If you feel this issue is still relevant for **Keystone Next**, please let us know.

PR #2922, opened 2020-05-08
[WIP] Updates to the CalendarDay field

### LINKED ISSUE #2748: Docs don't specify the supported date and time formatting tokens for DateTime and CalendarDay
From the DateTime docs:
> **format**
Defines the format of the string that the component generates. For example, `MM/DD/YYYY h:mm A`.

A link could be added, pointing to the table of tokens of the date utility library used for formatting.

timleslie:
This PR is a bit of a mixed bag to make general improvements to the `CalandarDay` field type.

I'll probably break it into multiple smaller PRs as it comes together and we find answers to some open questions.

The docs for this field type need a bit of work and there are a few open questions around the API which need to be clarified.

Things to consider:
 * `format` only applies to the adminUI. Do we want to have this somehow apply to the graphQL API as well? ( #2702, 
 * `yearRangeFrom` and `yearRangeTo` only apply to the adminUI. Do we want to apply validation at the graphQL layer as well? This would probably required date from/to rather than year from/to. Probably would want to be a half-open range I guess [from, to).
 * Is the `Field` view component, based on `TextDayPicker` doing what we want? Do we want to be able to configure it more? How easy is it for someone to use a different component here if they want?
 * The `Filter` component uses ISO8601 formatting, rather than the `format` config value used in the `Field` component. Is this correct?

Need to make sure we fix #2748, #2565, #2366

Should also consider #1136 

Should look into #215 and see if there's anything useful left in that issue that we still need to address. Joss makes a good comment in here: https://github.com/keystonejs/keystone/issues/215#issuecomment-589484850 that we should probably wait for `react-day-picker` v8 before launching into any serious UI work.

Loads of interesting related work in #1965.
 * We should definitely upgrade to date-fns `2.x`. This will be a breaking change, as they changed their format string handling completely.
 * An example `{ from, to }` API for date range restriction.

### DIFF
--- a/demo-projects/todo/index.js
+++ b/demo-projects/todo/index.js
@@ -1,9 +1,20 @@
 const { Keystone } = require('@keystonejs/keystone');
 const { MongooseAdapter } = require('@keystonejs/adapter-mongoose');
-const { Text } = require('@keystonejs/fields');
+const { Text, CalendarDay, Virtual } = require('@keystonejs/fields');
 const { GraphQLApp } = require('@keystonejs/app-graphql');
 const { AdminUIApp } = require('@keystonejs/app-admin-ui');
 const { StaticApp } = require('@keystonejs/app-static');
+const { format, parseISO } = require('date-fns');
+
+class _CalendarDayImpl extends CalendarDay.implementation {
+  gqlOutputFieldResolvers() {
+    return {
+      [`${this.path}`]: item =>
+        item[this.path] && format(parseISO(item[this.path]), this.config.gqlFormat),
+    };
+  }
+}
+const _CalendarDay = { ...CalendarDay, implementation: _CalendarDayImpl };
 
 const keystone = new Keystone({
   adapter: new MongooseAdapter({ mongoUri: 'mongodb://localhost/todo' }),
@@ -13,6 +24,17 @@ keystone.createList('Todo', {
   schemaDoc: 'A list of things which need to be done',
   fields: {
     name: { type: Text, schemaDoc: 'This is the thing you need to do', isRequired: true },
+    date: { type: CalendarDay, adminDoc: 'For testing only' },
+    niceDate: {
+      type: Virtual,
+      resolver: item => item.date && format(parseISO(item.date), 'do MMMM, yyyy'),
+    },
+    nicerDate: {
+      type: Virtual,
+      resolver: (item, { formatAs = 'do MMMM, yyyy' }) =>
+        item.date && format(parseISO(item.date), formatAs),
+      param: 'formatAs: String',
+    },
   },
 });
 
--- a/demo-projects/todo/index.js
+++ b/demo-projects/todo/index.js
@@ -1,20 +1,9 @@
 const { Keystone } = require('@keystonejs/keystone');
 const { MongooseAdapter } = require('@keystonejs/adapter-mongoose');
-const { Text, CalendarDay, Virtual } = require('@keystonejs/fields');
+const { Text } = require('@keystonejs/fields');
 const { GraphQLApp } = require('@keystonejs/app-graphql');
 const { AdminUIApp } = require('@keystonejs/app-admin-ui');
 const { StaticApp } = require('@keystonejs/app-static');
-const { format, parseISO } = require('date-fns');
-
-class _CalendarDayImpl extends CalendarDay.implementation {
-  gqlOutputFieldResolvers() {
-    return {
-      [`${this.path}`]: item =>
-        item[this.path] && format(parseISO(item[this.path]), this.config.gqlFormat),
-    };
-  }
-}
-const _CalendarDay = { ...CalendarDay, implementation: _CalendarDayImpl };
 
 const keystone = new Keystone({
   adapter: new MongooseAdapter({ mongoUri: 'mongodb://localhost/todo' }),
@@ -24,17 +13,6 @@ keystone.createList('Todo', {
   schemaDoc: 'A list of things which need to be done',
   fields: {
     name: { type: Text, schemaDoc: 'This is the thing you need to do', isRequired: true },
-    date: { type: CalendarDay, adminDoc: 'For testing only' },
-    niceDate: {
-      type: Virtual,
-      resolver: item => item.date && format(parseISO(item.date), 'do MMMM, yyyy'),
-    },
-    nicerDate: {
-      type: Virtual,
-      resolver: (item, { formatAs = 'do MMMM, yyyy' }) =>
-        item.date && format(parseISO(item.date), formatAs),
-      param: 'formatAs: String',
-    },
   },
 });
 
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -438,6 +438,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "matheuschimelli",
+      "name": "Matheus Chimelli",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/10470074?v=4",
+      "profile": "https://github.com/matheuschimelli",
+      "contributions": [
+        "doc"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/.changeset/big-llamas-sparkle.md
+++ b/.changeset/big-llamas-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Refactored hook management into a separate module.
--- a/.changeset/famous-pumpkins-raise.md
+++ b/.changeset/famous-pumpkins-raise.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Fixed a bug where `existingItem` had the wrong value in hooks during an `updateManyMutation`.
--- a/.changeset/light-llamas-wonder.md
+++ b/.changeset/light-llamas-wonder.md
@@ -1,0 +1,11 @@
+---
+'@keystonejs/demo-project-meetup': patch
+---
+
+Updated usage of Apollo based on Next.js [example](https://github.com/vercel/next.js/blob/canary/examples/with-apollo). 
+
+The **key changes** are: 
+- Less boilerplate code for setting/initializing Apollo Client
+- Update home, events and about page to opt in for SSG
+- Removed MyApp.getIntialProps as it will stop the [Auto Static optimization](https://nextjs.org/docs/api-reference/data-fetching/getInitialProps#caveats). 
+- Handle the authentication in client-side only.
--- a/.changeset/lucky-jokes-warn.md
+++ b/.changeset/lucky-jokes-warn.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Factored out utility `List` functions into a separate module.
--- a/.changeset/pretty-radios-obey.md
+++ b/.changeset/pretty-radios-obey.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Updated internal access control functions to directly accept access control definition.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -275,11 +275,8 @@ We also build commonjs builds to run in node (for testing with jest or etc.) and
 Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
 
 <!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
-
 <!-- prettier-ignore-start -->
-
 <!-- markdownlint-disable -->
-
 <table>
   <tr>
     <td align="center"><a href="http://www.thinkmill.com.au"><img src="https://avatars3.githubusercontent.com/u/872310?v=4" width="80px;" alt=""/><br /><sub><b>Jed Watson</b></sub></a><br /><a href="https://github.com/keystonejs/keystone/commits?author=JedWatson" title="Code">💻</a></td>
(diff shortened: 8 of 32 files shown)


Vultraz:
FTR, I'm not sure in hindsight that the Material day picker was a great idea. It kinda clashed theme-wise with Arch, and it runs the risk of making everything look too much like generic Android.

Regarding `date-fns` 2.x, based on my work in #1965 the change wasn't massive. The biggest change is mostly stuff like `yyyy` -> `YYYY` and (IIRC) there is a compatibility layer one can use. I've been meaning to give the update a second pass at some point.

Definitely 👍 on date range replacing year range.

timleslie:
## Proposal 1

Add two new config options, `{ dateFrom, dateTo }` to replace `{ yearRangeFrom, yearRangeTo }`. 
 * Each is optional and takes a string in ISO8601 format (`YYYY-MM-DD`).
 * Validation would be applied at the the graphQL API level.
 * The admin UI could optionally use these values to perform client side validation and/or restrict the input fields range.
 * The option value would be validated as a valid ISO8601 string when the system started and throw an error if invalid.
 * If both options were set, they would be validated that `dateFrom <= dateTo` when the system started and throw an error if invalid.

This proposal would replace the `{ yearRangeFrom, yearRangeTo }` which a) didn't support days/months and b) only applied client side validation in the admin UI. The graphQL API could still be used to set any value. It's hard to imagine a use case where you would want validation in the admin UI which wasn't also enforced by the graphQL API.

timleslie:
## Proposal 2

Upgrade `date-fns` to version `2.x`.

This will be a breaking change to this field type (and the other date/time fields) as the formatting rules of that library have changed. These changes make for a much more complete formatting API, which we should adopt.

timleslie:
## Proposal 3

Update the `TextDayPicker` component to use [`react-date-picker`](https://github.com/gpbl/react-day-picker) version [`8.x`](https://github.com/gpbl/react-day-picker/issues/942), which is currently in `alpha`, but appears to be under active development at the moment.

Vultraz:
Ahhh, that `react-day-picker` looks nice! Much cleaner than some of the other options on offer (`react-date-picker`, `react-datepicker`, etc).

timleslie:
## Proposal 4

Add a `format` argument to graphQL queries as per https://github.com/keystonejs/keystone/pull/2702#issuecomment-623945286

This would take a `String`, validate it as a valid format string, and cause the date to be formatted accordingly when resolved at the field level.

The default would continue to be the ISO860 format (`YYYY-MM-DD`). I do not think that we should support changing the default, as I want to maintain the symmetry between create/update and reading. I also do not think we should attempt to support alternate formats for create/update.

This would merely be a convenience to users querying the API to prevent them having to include explicit formatting code in their clients.

timleslie:
## Proposal 5

Rename the current `format` argument to `adminFormat` and update the documentation to indicate that this format is only applied in the list view of the admin UI. It is currently not particularly obvious that this is the case, and users might think that this format is applied at either the storage of graphQL layer.

gautamsi:
Should this be moved to `adminConfig` for field? Can still call it format then

gautamsi:
### another improvement 
 also remove use of `moment` from `@arch-ui/day-picker` or unify the library being used. another alternative is to use `dayjs` which claims momentjs compatibility with 90% reduced size.

MadeByMike:
@timleslie I like all these proposals (except for 4). I don't think we need to provide different options for incoming or out going dates in graphQL. 

If we say graphQL queries accept and provide ISO860 format, I think this is much clearer and easier for users to work with. It's easy to implement a format options on the client to make sure this is in the format you want you want and it simplifies our work as well as potentially saving people from a mess of handling multiple different in one application.

molomby:
My brain dump...

## Re: Proposal 1 - From/To Validation Range

> Add two new config options, { dateFrom, dateTo } to replace { yearRangeFrom, yearRangeTo }.

Yep, this all sound great. Some ideas:

* We could still support years by assuming `from` is from the start and `to` is till the end. Eg. `{ dateFrom: '2000', dateTo: '2007' }` would be interpreted as `{ dateFrom: '2000-01-01', dateTo: '2007-12-31' }`. Supporting this would enable us to be mostly backward compatible with the current `yearRangeFrom`/`yearRangeTo` options (by aliasing them).
* `CalendarDay` values (like these options) are weird pseudo-dates. When I'm working with values like this I distinguish them from actual dates by referring to the former as "days" (as in the name of this type). As such, I don't love having "date" in the name of these options. It would also be nice to call out in the option name that these are used for validation. Maybe something like `fromDay, toDay` or `minValid, maxValid`?

I'm not invested in either of these ^^ ideas, just throwing them in the mix.

## Re: Proposal 4 - Server-Side Formatting

> `format` only applies to the adminUI. Do we want to have this somehow apply to the graphQL API as well??

> Add a format argument to graphQL queries as per #2702 (comment)

To me, applying formatting to a date is about getting it ready for human consumption. This doesn't seem like something that should be part of a data API like GraphQL.

One of the issues is how consistent we want `CalendarDay` fields to be with `DateTime` fields, so let me segue into the `DateTime` type for a bit first...

If you're formatting a a UTC datetime for someone and _not_ accounting for time zones and localisation, you're doing it wrong. `2020-03-09T04:31:11z` in `d mmmm YYYY` isn't "9 February 2020"; if you're in Ecuador it's "8 Febrero 2020". Supporting this server-side requires a function that takes a mask, locale and either a time zone identifier or an offset.

And to what benefit? There are a few use case I can think of where this would be helpful:

* Bundle size -- You're building a site for mobile and trying to keep it really light, server-side date formatting can save you from downloading and parsing [20](https://bundlephobia.com/result?p=date-fns@2.13.0) to [95](https://bundlephobia.com/result?p=moment-timezone@0.5.28) Kb of compressed JS. This is a pretty legitimate use case.
* Data size -- You're building an export that includes `createAt` and `modifiedAt` `DateTime` fields but you actually only need the date (`YYMMDD`). Stripping the excess data on the server saves you 36 Kb for every 1000 records. Pretty minor but worth mentioning.

Am I'm missing others?

My question is, **would we ever want to build out server-side formatting for the `DateTime` type, complex enough to actually work well, just to solve these limited use cases?**

Then, if the answer is "no" (and I suspect it is) then **would be want add this functionality to _just_ the `CalendarDay` field?**

Again, I suspect the answer is "no". It's not quite as complex as datetimes as we can ignore time zones but..

* We should really still solve for localisation
* It makes the `CalendarDay` type conspicuously inconsistent `DateTime`
* There doesn't seem to be much benefit (unless there's some use case I'm missing)
* Formatting on the server remains architecturally questionable

TL:DR; Supporting this well and constantly will be a lot of work for little gain. Besides, formatting data for human consumption is the browsers job.

(Ps. If we _did_ support this, don't miss the naming discussing in #2702.)

## Re: Proposal 5 - Admin Format

Yep, this is useful as config for the Admin UI but, as as @gautamsi notes, it should probably be in `adminConfig`. Though, personally, I prefer the clarity of `formatAs` for the name.

## Re: The Rest

Proposals 2 and 3 sound good to me.

Vultraz:
I propose we use the standard `interval` terminology. Something like `interval: { start, end }` would fit perfectly with `date-fns`'s own API: https://date-fns.org/v2.13.0/docs/isWithinInterval

timleslie:
> We could still support years by assuming from is from the start and to is till the end. Eg. { dateFrom: '2000', dateTo: '2007' } would be interpreted as { dateFrom: '2000-01-01', dateTo: '2007-12-31' }. Supporting this would enable us to be mostly backward compatible with the current yearRangeFrom/yearRangeTo options (by aliasing them).

I'm not such a fan of this. It makes our API more complex and it introduces an element of magic. If I stumble across someone else's config and it says `{ dateFrom: '2000', dateTo: '2007' }` then it's probably not going to be obvious to me that this goes from the `2000-01-01` to `2007-12-31` and not some other particular choice of days for each year. I'd rather that we have exactly one explicit way to encode this info. I think the maintenance/docs/ongoing confusion cost outweighs the backwards compatibility benefit.

jesstelford:
I'm +1 for a `format` input param:

✅ The bundle size issue is quite bad. If I can remove libs like `moment` or `date-fns`, etc, from the client, that is a huge win
✅ It'd be an optional parameter, so it's only additive
✅ I would never expect the date to be formatted for any timezone than the one it was stored as (we store the offset info already afaik, so we have everything we need). If the user wants the same moment in time represented in their local timezone, then that has to be done client side because there's no possible way for the server to know where they are.
✅ When I'm querying the GraphQL API, I'm looking at the Playground and what options are available there. if I see a `format` option, I will use it. But if it's hidden somewhere in the docs and I have to add it myself to my Keystone config, I'm unlikely to use it.

gautamsi:
@jesstelford `dayjs` seems to have small footprint and moment compatible (I have not personally tested extensively)

jesstelford:
In support of a `filter` input parameter on the GraphQL API... Here's a fun one I _just ran into_. I get this error from React when doing SSR:

> Warning: Text content did not match. Server: "December 2, 2020." Client: "2 December 2020."

This is generated after running an ISO8086 datestring through this function:

```
new Date(time).toLocaleDateString(undefined, {
    day: 'numeric',
    month: 'long',
    year: 'numeric',
  })
```

The warning arises because it is run on both server-side and client side. Even _on the same computer_, there are differences between how the browser formats the date, and how node formats it.

If I could normalise that so it is only ever run in the server (ie; the GraphQL resolver), I would not get such warnings.

Note that this isn't only a console warning, it also causes a flash of wrong content (ie; it shows the node.js version on page load, but after the first client render, it shows the browser version). That's bad.

Vultraz:
> The bundle size issue is quite bad. If I can remove libs like moment or date-fns, etc, from the client, that is a huge win

That was actually what gave me the idea for #2702. The date formatting functionality is already there, it's just a matter of exposing it. Otherwise, you end up with the end user having to add a date lib dependency just to format potentially just one date string 😬

timleslie:
### Re: prop 4

I think the bundle size argument is a particularly compelling argument for this prop. An important question to ask is: can the user get server-side-formatted dates using some other mechanism? I put together the following example using the `Virtual` field type:

```
keystone.createList('Todo', {
  schemaDoc: 'A list of things which need to be done',
  fields: {
    name: { type: Text, schemaDoc: 'This is the thing you need to do', isRequired: true },
    date: { type: CalendarDay, adminDoc: 'For testing only' },
    niceDate: {
      type: Virtual,
      resolver: item => item.date && format(parseISO(item.date), 'do MMMM, yyyy'),
    },
    nicerDate: {
      type: Virtual,
      resolver: (item, { formatAs = 'do MMMM, yyyy' }) =>
        item.date && format(parseISO(item.date), formatAs),
      param: 'formatAs: String',
    },
  },
});
```

Screen Shot 2020-05-11 at 4 37 44 pm

Caveat: The `param` config option for `Virtual` fields doesn't actually exist yet, I added it as a quick hack in my local setup, but I think I'll ship it as a separate PR, because there's no reason not to have it.

I think this pattern is a simple enough mechanism for devs to implement (I'd happily write up a guide on how to do this for the docs) and it allows them the flexibility to make their implementation as simple/complex as they need/want, allows them to pick their formatting library of choice, etc, etc. I think overall having over these decisions to the dev is a better solution than us trying to make a one size fits all API.

molomby:
Re: server-side formatting, cont.

Bundle size is important, but remember, for this to actually be a win all date times used in the app (or on a page?) need to be loaded through GraphQL. If you're doing anything else with dates other than loading them from Keystone, you're going to need in-browser formatting capacity. 

All that's going to happen is, people will start building an app using the server-side `format` option, then, after a few months, will need to format a date in the browser and end up pulling in a date library anyway. They'll have half their dates being formatted on the server and half in the browser. We'll shepherd them into a situation that's worse than if they'd just kept formatting concerns on the client (where they belong!) from the start.

Plus, we're talking about [17.5 Kb](https://bundlephobia.com/result?p=date-fns@2.13.0) for the whole of `date-fns@2.13.0` and, if you _really_ want to squeeze your bundle you [don't even need the whole thing](https://date-fns.org/docs/I18n).

> It might seem complicated to require and pass locales as options, but unlike Moment.js which bloats your build with all the locales by default date-fns forces developer to manually require locales when needed.

So, given..

* How rare these "smaller bundle" circumstances are likely to be in practice
* How much work is involved to get server-side formatting [right](https://github.com/keystonejs/keystone/pull/2922#issuecomment-626108025), tested, documented, etc.
* How it kind of blurs presentation logic into the data API
* How it shepherd solution devs down what's likely to be a technical dead-end
* And how small the bundle size "win" would actually be..

Is it really worth considering?

The saving grace here is that, yes, it is "purely additional" at an API level and, yes, we _can_ actually solve the entire problem well if we choose to. Formatting like this will be convenient for some small number of apps and, maybe more importantly, it demo's well. It's good for the "first sprint" (I guess).

So yeah, I think it's a bad idea but at having it isn't going to create serious problems.

Vultraz:
RE removing moment and luxon, there's [`date-fns-tz`](https://github.com/marnusw/date-fns-tz) which is compatible with date-fns 2 and includes some helpers for timezone-related functionality. Might be able to replace those two and use `date-fns[-tz]` everywhere.

MadeByMike:
Did a little spike on adding functions to dayTo/From ... had to revert because of course you can't pass functions across the webpack barrier. This kinda sucks because the most common `dayTo` value is probably "TODAY". Easy enough to have a function set this at build time, but this needs to be enforced at runtime and shared with the AdminUI.

# Proposal 6: Solve this

demo-projects/todo/index.js
```diff
@@ -14,6 +25,17 @@ keystone.createList('Todo', {
   schemaDoc: 'A list of things which need to be done',
   fields: {
     name: { type: Text, schemaDoc: 'This is the thing you need to do', isRequired: true },
+    date: { type: CalendarDay, adminDoc: 'For testing only' },
```
MadeByMike, line 28: Will remove this

MadeByMike:
Ok I've gone through all this work and I'm pretty sure this PR has run it's course...

Proposals 1-4 have been implemented.

No graphQL formatting yet but I like Tim's suggestions of using the virtual field.

Re: 5. - Admin format is contentious in terms of how it should be implemented and definitely not essential, not to be solved here.

Re: 6 - Also not to be solved here.

PR #2925, opened 2020-05-08
website: Apply hash link offset to H4s

jordanoverbye:
Fixes an issue where h4 headings will be anchored to the very top of the page and be hidden by the header

https://www.keystonejs.com/keystonejs/fields/src/types/calendar-day/#format


website/src/components/markdown/Heading.js
```diff
@@ -4,7 +4,15 @@ import { jsx } from '@emotion/core';
 import { colors, gridSize } from '@arch-ui/theme';
 import { LinkIcon } from '@arch-ui/icons';
 
-import { mq } from '../../utils/media';
```
jordanoverbye, line 7: We're not using any responsive stuff here so this can be removed

website/src/components/markdown/Heading.js
```diff
@@ -4,7 +4,15 @@ import { jsx } from '@emotion/core';
...
+// offset header height for hash links
+const hashLinkOffset = {
+  '::before': {
+    content: '" "',
+    height: 'calc(32px + 60px)',
```
jossmac, line 11: Not sure why this is using `calc`, I must've had some dynamic number in there at one point. Happy to leave as is for now.

PR #2930, opened 2020-05-09
Fixed many usability issues with the CalendarDay field

Vultraz:
Went down a rabbit hole with this one 😬 I think I probably caused most of these with #2424.

- Fixed field not functioning as a proper controlled component in the Create popout.
- Fixed field initially displaying "Invalid Date" before defaulting to 1970-01-01.
- Filter input no longer defaults to the current date. This was bugged; submitting the form with no changes would match nothing.
- Filter input now falls back to no value when given an invalid date. Previously, it was falling back to 1970-01-01.
- Fixed filter input not initially displaying the current value when in edit mode (it was displaying the current date).
- Fixed filter input not being initially focused.
- Fixed filter input not being submitted properly if focus wasn't lost first.

I also dropped the initial and on-invalid-input current-day default for the filter. Most filters don't match the default, and it's not really likely someone would want to use the current date anyway. It just meant having to clear the field first. As for falling back to the current day on an invalid value (which was bugged anyway!), that also felt completely arbitrary. Now it falls back to empty input (`""`).

### DIFF
--- a/.changeset/sour-taxis-hunt.md
+++ b/.changeset/sour-taxis-hunt.md
@@ -1,0 +1,13 @@
+---
+'@arch-ui/day-picker': patch
+'@keystonejs/fields': patch
+---
+
+Fixed many usability issues with the CalendarDay field:
+- Fixed field not functioning as a proper controlled component in the Create popout.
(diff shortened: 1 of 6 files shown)


Vultraz:
Includes a chrono-node version update.

PR #2260, opened 2020-01-17
add dependsOn option in field which is same as v4

gautamsi:
related to #1355 
closes #2231 

this being v4 feature `dependsOn` which is useful for showing or hiding the component in admin ui based on other field values. 

example: 
dependsOn

following config used in basic test project

```js
keystone.createList('User', {
  fields: {
    name: { type: Text },
    email: { type: Text, isUnique: true, dependsOn: { name: 'email' } },
    dob: {
      type: CalendarDay,
      format: 'Do MMMM YYYY',
      yearRangeFrom: 1901,
      yearRangeTo: getYear(new Date()),
    },
    lastOnline: {
      type: DateTime,
      format: 'MM/DD/YYYY h:mm A',
      yearRangeFrom: 2013,
    },
    password: { type: Password },
    isAdmin: { type: Checkbox },
    favouritePosts: { type: Relationship, ref: 'Post', many: true, dependsOn: { isAdmin: true } },
    company: {
      type: Select,
      options: [
        { label: 'Thinkmill', value: 'thinkmill' },
        { label: 'Atlassian', value: 'atlassian' },
        { label: 'Thomas Walker Gelato', value: 'gelato' },
        { label: 'Cete, or Seat, or Attend ¯\\_(ツ)_/¯', value: 'cete' },
      ],
    },
    attachment: {
      type: File,
      adapter: fileAdapter,
      dependsOn: { $or: [{ company: ['cete', 'thinkmill'] }] },
    },
    color: { type: Color },
    website: { type: Url },
    ...(embedAdapter && { profile: { type: OEmbed, adapter: embedAdapter } }),
    ...(cloudinaryAdapter && { avatar: { type: CloudinaryImage, adapter: cloudinaryAdapter } }),
    ...(unsplash.accessKey && { favouriteImage: { type: Unsplash, ...unsplash } }),
  },
  labelResolver: item => `${item.name} <${item.email}>`,
  hooks: {
    validateInput: async ({ resolvedData, addValidationError }) => {
      if (resolvedData.name === 'Homer') {
        addValidationError('Sorry, no Homers allowed', { a: 1 }, { b: 2 });
      }
    },
  },
});

keystone.createList('Post', {
  fields: {
    name: { type: Text },
    slug: { type: Text },
    status: {
      type: Select,
      defaultValue: 'draft',
      options: [
        { label: 'Draft', value: 'draft' },
        { label: 'Published', value: 'published' },
      ],
    },
    author: {
      type: Relationship,
      ref: 'User',
[...]

PR #2939, opened 2020-05-11
 Admin UI: fixed no-access indicators not showing up in list table

Vultraz:
~~The list query doesn't include an `error` argument at all. It looks something like this:~~
```js
{
      
      allTodos(first: 20 orderBy: "name_ASC") {
        id _label_
        name updatedAt
      }
      _allTodosMeta(search: "") { count }
    }
```

~~Plus, I cannot actually find any other place where errors are part of the query data result instead of in `query.errors`. This looks like it's meant to gray out specific items from the list table you don't have access to, but in my testing you simply just don't see those items at all.~~

### DIFF
--- a/.changeset/popular-mayflies-doubt.md
+++ b/.changeset/popular-mayflies-doubt.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Fixed no-access indicators not showing up in list table.
--- a/packages/app-admin-ui/client/components/ListTable.js
+++ b/packages/app-admin-ui/client/components/ListTable.js
@@ -216,7 +216,7 @@ class ListRow extends Component {
           if (itemErrors[path] instanceof Error && itemErrors[path].name === 'AccessDeniedError') {
             return (
               <BodyCell key={path}>
-                <ShieldIcon title={itemErrors[path].message} css={{ color: colors.N10 }} />
+                <ShieldIcon title={itemErrors[path].message} css={{ color: colors.N20 }} />
                 <A11yText>{itemErrors[path].message}</A11yText>
               </BodyCell>
             );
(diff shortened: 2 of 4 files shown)


jesstelford:
There are two ways to restrict access to a field:

```
keystone.createList('Things', {
  fields: {
    name: { type: Text, access: false },
    title: { type: Text, access: () => false },
  }
});
```

In the first form (for `name`), the field should simple disappear from the Admin UI.

In the second form (for `title`), should field should still be visible, but greyed out in the Admin UI? Or should it also disappear completely?

If it's to disappear completely, this change seems reasonable. Otherwise, we might need to keep those errors around to catch the case where access control is determined dynamically based on the function passed to `access`.

Vultraz:
That's field access, though. The code in question here was part of the list query and applies to the list table.

Fields to which you do not have access are indeed already not editable as you say (though they're not actually grayed out/disabled, you just can't - there's #2258 for that)

gautamsi:
there is two way to use this functional access thing. 
1. use GraphQL to provide access control information to admin ui, I see there is already plan in - https://github.com/keystonejs/keystone/pull/2679#issuecomment-610717619, caveat
   - it would not help evaluate access based on current item, only authedItem would be accessible.
   - can be very complex to evaluate the current Item or related item in the access control function. 
2. use UI hooks api to provide admin-ui way to evaluate access control. caveats:
   - can be tedious as you have to create access control separately for admin-ui.
3. use `expression-match` or similar lib to use templating to determine access control in backend as well as server. I did that with `dependsOn` PR https://github.com/keystonejs/keystone/pull/2260

Vultraz:
Update: turns out I just needed to actually grab the errors from the right source and format them properly and everything worked again!
image

PR #2947, opened 2020-05-12
Add support for arguments on Virtual fields

timleslie:


jesstelford:
How're we testing the Virtual field?

timleslie:
You're right, it's currently pretty light on for tests, I'll add some as part of this PR 👍

PR #2956, opened 2020-05-13
Added a password field cell view

Vultraz:
Previously, you just got an empty column. This is more useful, especially since it gives admins an at-a-glance look at unsecured accounts without having to explicitly filter for Not Set.
image


Vultraz:
@timleslie done

PR #2958, opened 2020-05-13
Improved/clarified at-tracking date display format

Vultraz:
Makes this more readable and removes any ambiguity of m/d/y vs d/m/y.

Before:
image

After:
image

Am open to tweaking the format if you guys think it should be something different.


VinayaSathyanarayana:
How does this work, when different users are in different time zones? Do we show timestamps converted to user's local timezone or UTC?

Vultraz:
That's a good question, and I'm not entirely sure. Internally, the plugin uses the `DateTime` field, not `DateTimeUtc`, so I *think* it's local time for the person who created it. Could definitely use some clarification, though.

VinayaSathyanarayana:
Let's use UTC to store the timestamp. Display to be based on user timezone. The timezone of the user can be shown along with the user id field

PR #2961, opened 2020-05-13
Admin UI: used consistent formatting for emotion styled components

Vultraz:
@MadeByMike you mentioned we should have consistent formatting.

- Makes all styled components use template literals.
- Components that used the function argument now use the `css` tagged literal to return their result. This ensures they also get css syntax highlighting if an extension such as [`vscode-styled-components`](https://marketplace.visualstudio.com/items?itemName=jpoissonnier.vscode-styled-components) is installed.
- `px` units were added to raw measurements when necessary (been doing this all along as I converted stuff).
- The `TableRow` component was removed. It was never used with an `isActive` prop, and its only other rule (`> td { background-color: null }`) was invalid. Fixing it caused it to override the selected-row style.
- `BodyCell` now uses `none` instead of `null` for its `background-color` fallback (latter was invalid).

Stuff in `components.js` is excluded since I'm doing some additional work there.


MadeByMike:
I think a lot of people here prefer object notation but I'm not fussed as long as it's consistent. happy to approve. I'll give others a short amount to time to object ;)

JedWatson:
Sorry to jump in here with the opposite direction but we have a strong preference for styles to be expressed as object, and the result of functions, as opposed to the styled-components `styled` syntax.

The reasons for this are that we intend to introduce theming to the design system in the not-too-distant future as part of the improvements to Admin UI customisation, and it'll be based on style functions.

I noticed the other day that we've had a lot of "styled" components pop up in the codebase and things were getting inconsistent, so my apologies for not picking this up and making it clearer sooner.

Having said all that, this isn't a big diff and so if it's the easiest way to land on something consistent for now, I'm not minded to block it and we can deal with switching it all to the other format when there's a clearer reason to prefer one syntax over the other in the design system.

But all things being otherwise equal, I'd prefer we express (for example)

```js
const HeaderCell = styled.th(
  ({ isSelected, isSortable }) => css`
    background-color: white;
    box-shadow: 0 2px 0 ${alpha(colors.text, 0.1)};
    box-sizing: border-box;
    color: ${isSelected ? colors.text : colors.N40};
    cursor: ${isSortable ? 'pointer' : 'auto'};
    display: table-cell;
    font-weight: normal;
    padding: ${gridSize}px;
    :hover {
      color: ${isSortable && !isSelected ? colors.N60 : null};
    }
  `
);
```

as

```js
const HeaderCell = ({ isSelected, isSortable, ...props }) => (
  <th css={{
    backgroundColor: 'white',
    boxShadow: `0 2px 0 ${alpha(colors.text, 0.1)}`,
    boxSizing: 'border-box',
    color: isSelected ? colors.text : colors.N40,
    cursor: isSortable ? 'pointer' : 'auto',
    display: 'table-cell',
    fontWeight: 'normal',
    padding: gridSize,
    ':hover' {
      color: isSortable && !isSelected ? colors.N60 : undefined,
    }
  }} {...props} >
);
```

... because, among other things, we'll be using hooks shortly to pull theme from context for things like the `colors` object; like this:

```js
const HeaderCell = ({ isSelected, isSortable, ...props }) => {
  const { colors, gridSize } = useTheme();
  return (<th css={{
    backgroundColor: 'white',
    boxShadow: `0 2px 0 ${alpha(colors.text, 0.1)}`,
    boxSizing: 'border-box',
    color: isSelected ? colors.text : colors.N40,
    cursor: isSortable ? 'pointer' : 'auto',
    display: 'table-cell',
    fontWeight: 'normal',
    padding: gridSize,
    ':hover' {
      color: isSortable && !isSelected ? colors.N60 : undefined,
    }
  }} {...props} >);
};
```

Vultraz:
Ahh, my bad! 😬 I can go back and convert the `styled` stuff back to object syntax. Or would you prefer I fully convert all `styled` components to use the `css` prop instead?

timleslie:
Just followed up with @JedWatson who is busy in meetings today and his reply was:

```
Jed Watson  11:17
tl;dr
> Or would you prefer I fully convert all styled components to use the css prop instead?
ideally, yes, use the css prop
```

If that's not too much effort, @Vultraz? 🙏 

Might be an opportunity to break this into more than one PR to get some of the simpler stuff through quicker than the big styled component work?

Vultraz:
Ok, will get on it, and I'll split the `TableRow` change out. 👍

Issue #2965, opened 2020-05-14
@keystone/file-adapter's S3Adapter incorrectly requires constructor parameters


intmainvoid:
# Bug report

## Describe the bug

Creating an instance of `S3Adapter` (from `@keystone/file-adapters`) without providing accessKeyId, secretAccessKey and region throws an error and terminates Keystone's Node.js instance. The error is thrown with the following message:

> S3Adapter requires accessKeyId, secretAccessKey, region, bucket, folder.

## To Reproduce

Create an instance of the `S3Adapter` as follows:
``` Javascript
const fileAdapter = new S3Adapter({
	// accessKeyId // not provided
	// secretAccessKey: // not provided
	// region: // not provided
	bucket: 'website-submission-attachments',
	folder: 'test',
	s3Options: {
		apiVersion: '2006-03-01',
	},
	uploadParams: ({ filename, id, mimetype, encoding }: UploadParams) => {
		return {
			Metadata: {
				keystone_id: String(id),
			},
		};
	},
});
```

## Expected behaviour

The `S3Adapter` connects to the S3 bucket regardless of the presence of the 3 missing parameters and that no attempt is made by the `S3Adapter` to pass the the three missing parameters on to the `AWS.S3` constructor params.

## Screenshots

N/A

## System information

macOS

## Additional context

`@keystone/file-adapter`'s `S3Adapter` requires - among others - the following parameters in its constructor which are then passed to the AWS SDK constructor: `AWS.S3`:

```
    accessKeyId,
    secretAccessKey,
    region,
```

There are many common use cases where creating an S3 instance without these parameters are valid - and in fact - `accessKeyId` and `secretAccessKey` are [deprecated](https://github.com/aws/aws-sdk-js/blob/master/lib/config.d.ts#L197) in the AWS SDK.

## Proposed solution

Remove the hard requirement for these parameters to be provided and instead simply allow the user to provide them as required in the `s3Options` object.

MadeByMike:
Fixed in #2967

PR #2967, opened 2020-05-14
file-adapters - Removed unnecessary required fields for S3Adapter

intmainvoid:
This PR implements the proposed solution for issue https://github.com/keystonejs/keystone/issues/2965 where an error was thrown if alternative AWS SDK authentication methods were used


Vultraz:
You'll need to add a changeset describing this change too.

.changeset/long-bags-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/file-adapters': minor
```
MadeByMike, line 2:
```suggestion
'@keystonejs/file-adapters': major
```

Any one using this will need to change how they provided: `accessKeyIdm` `secretAccessKey` and `region`, making this a breaking change.

We'll also need to inform people of this in the changeset with specific details on how to update if using those options.

packages/file-adapters/lib/s3.js
```diff
@@ -3,28 +3,13 @@ const AWS = require('aws-sdk');
...
-    this.s3 = new AWS.S3({
-      accessKeyId,
-      secretAccessKey,
-      region,
-      ...s3Options,
-    });
+    this.s3 = new AWS.S3(s3Options);
     this.bucket = bucket;
-    this.folder = folder;
+    this.folder = folder || '';
```
MadeByMike, line 12: Let's move this to a default in the constructor.

```
constructor({ bucket, folder='', ...
```

packages/file-adapters/lib/s3.js
```diff
@@ -3,28 +3,13 @@ const AWS = require('aws-sdk');
...
     if (getFilename) {
```
MadeByMike, line 13: What happens here if getFilename is not provided?

packages/file-adapters/lib/s3.js
```diff
@@ -3,28 +3,13 @@ const AWS = require('aws-sdk');
...
       this.getFilename = getFilename;
```
MadeByMike, line 14: What happens on save (line 27) if getFilename is not provided?

packages/file-adapters/lib/s3.js
intmainvoid: All done.

.changeset/long-bags-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/file-adapters': major
+---
+
+Removed unnecessary required fields for S3Adapter
```
MadeByMike, line 5: Move `accessKeyId`, `secretAccessKey` and `region` to `s3Options` and renamed `secretAccessKey` to `s3`.

Previously:

```
const fileAdapter = new S3Adapter({
  bucket: 'bucket-name',
  accessKeyId: 'ACCESS_KEY_ID',
  secretAccessKey: 'SECRET_ACCESS_KEY',
  region: 'us-west-2',
});
```

now:

```
const fileAdapter = new S3Adapter({
  bucket: 'bucket-name',
  s3Options: {
    accessKeyId: 'ACCESS_KEY_ID',
    s3: 'SECRET_ACCESS_KEY',
    region: 'us-west-2',
  }
});
```

intmainvoid: A couple of things:

1/ The renaming of that property from `secretAccessKey` to `s3` was an error in the docs. Fix forthcoming.

2/ Not sure about the accuracy of your description. These weren't "moved" anywhere. It's just that we stopped accepting them on the constructor. It was always possible (as it is after this change) to supply these properties via the `s3Options` parameter.

timleslie: I don't think we need to language lawyer the changeset too much, I think the examples do must of the heavy lifting to clarify exactly what has changed and what changes a developer would need to make in order to upgrade to the new package.

If we can make that update (or something sufficiently similar) then I'm happy to give this a ✅ and get it shipped 🚀

intmainvoid: Cool. I've made all the changes I think need to be done. Anything else specific we're waiting for?

timleslie: Just need to update the changeset to include the info from Mike's comment then we're good to go.

intmainvoid: Yep. I think [this commit](https://github.com/keystonejs/keystone/pull/2967/commits/68eda7b3960f2bef224d85d037f1c9e067b622a0) takes care of that. Have I missed something?

timleslie: Yep, the changeset file (`.changeset/long-bags-itch.md`) needs to contain the text that Mike suggested in this comment thread (or something similar). This is the text that will get put into the CHANGELOG file, and is the thing that developers will refer to when upgrading this major package version, so it's important that it clearly explains the before and after of this version change so that they can successfully this package.

PR #2976, opened 2020-05-15
Admin UI: big cleanup of query construction

Vultraz:
And now, my big weekend PR.

This refactors the Admin UI's list and item query construction. Instead of using complicated and very hard-to-read string concatenation, this makes use of query variables wherever possible.

- All queries are now named (prep for another change I have coming)!
- Changes the item query to a single, static query (since we're always fetching all fields) taking an `id` argument.
- List queries now use variables for all input (`search`, `where`, `skip`, `first,` and `sortBy`). The new `getListQuery` function only takes `fields` to control which fields we're querying since that can't be done via variables. Even better, using variables makes it clear exactly what arguments are accepted and their types.
- Since `where` filters are just passed in via variables now, `getFilterGraphQL` doesn't need to handle messy strings anymore and can just return plain objects!
- `gqlCountQueries` was removed and replaced with a helper in `Home/index.js`.

### DIFF
--- a/.changeset/few-bears-eat.md
+++ b/.changeset/few-bears-eat.md
@@ -1,8 +1,8 @@
 ---
-'@keystonejs/app-admin-ui': minor
-'@keystonejs/field-content': patch
-'@keystonejs/fields': patch
-'@keystonejs/fields-mongoid': patch
+'@keystonejs/app-admin-ui': major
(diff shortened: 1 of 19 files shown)


Vultraz:
@timleslie done 👍

PR #2978, opened 2020-05-17
Admin UI: refactored how query refetches are handled

### LINKED ISSUE #2795: Relationship fields created during editing of one list item do not immediately show up when editing another
I hesitated to raise this since I don't have a reproducible example, but it's happened to two members on my dev team.  

We have an `Article` list which has a `Topic` Relationship field:

```
		
topic: {
				label: 'Assign topics',
				type: Relationship,
				ref: 'Topic',
				many: true
			},

```

The issue is that if we _create_ a `Topic` in the middle of editing an `Article`-- that is, we click on the `+` symbol in the `Topic` field in the AdminUI and a new 'create Topic window' opens, where we fill in a name and save 

image

That `Topic` (in the case above, 'Something') does not appear in the pulldown list of `Topic` when one saves the first `Article` one is editing, then opens another to edit it.  

The Relationship field (`Topic`) only appears in the pulldown list when you either do a page refresh, or go to the list itself in the AdminUI.

For what it's worth, we are using the custom Admin UI, the 'old school' way (I wasn't able to get the newer way to work for me):

```
import Dashboard from './pages/dashboard';
import About from './pages/about';

export default {
	pages: () => [
		/*   {
      label: 'A new dashboard',
      path: '',
      component: Dashboard,
    },*/
		{
			label: 'About this project',
			path: '/about',
			component: About
		},
		{
			label: 'Blog',
			children: [
				{ listKey: 'Article' },
				{ label: 'Proposed Articles', listKey: 'Proposal' },
				{ label: 'Authors', listKey: 'Author' },
				{ label: 'Images', listKey: 'Image' }
			]
		},
		{
			label: 'People',
			children: ['User']
		}
	]
};


```

Vultraz:
Relies on #2976. Fixes #2795 

Instead of manually using `refech`, use the nice `refetchQueries` functionality Apollo provides. This is what the named queries in #2976 was for.

Relevant commit: https://github.com/keystonejs/keystone/commit/e4e372ec3cacc072b8cd43ce0bf8ba1c4fa214d0

### DIFF
--- a/.changeset/pink-radios-join.md
+++ b/.changeset/pink-radios-join.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Refactored how query refetches are handled.
--- a/packages/app-admin-ui/client/components/CreateItemModal.js
+++ b/packages/app-admin-ui/client/components/CreateItemModal.js
@@ -54,6 +54,7 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate }) => {
   const [createItem, { loading }] = useMutation(list.createMutation, {
     errorPolicy: 'all',
     onError: error => handleCreateUpdateMutationError({ error, addToast }),
+    refetchQueries: ['getList', 'RelationshipSelect'],
   });
 
   const { fields } = list;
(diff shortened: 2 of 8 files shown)


Vultraz:
Rebased.

PR #2987, opened 2020-05-18
Admin UI: simplified ItemLink handling

Vultraz:
🧹 🧹 🧹 

- Removed `adminPath`  and `itemLink` props from `ListTable`. The latter wasn't even expected to have a value.
- Removed `link` prop from `ListRow` (this was passed the value of `itemLink` from `ListTable`)
- Converted `ItemLink` to a proper component taking `path` and `item` props. These match the data expected by the old `itemLink` function, except `path` is now populated with `list.fullPath`.
- Removed the now-unnecessary `LinkComponent` component.

Also renames the `onChange` prop in `ListTable` to `onDelete` since it was incorrectly named (previously, it was passed a delete callback).


gautamsi:
I used `itemLink` prop to create a link to my custom page instead of default edit page.

PR #2992, opened 2020-05-19
Admin UI: minor sort usability fixes

Vultraz:
- Fixed list view sort indicator pointing the wrong way. Previously, it would unintuitively point down for ascending down and vice-versa for descending.
- Fixed Sort Select popup losing focus when pressing Alt.

FTR, arrow-up-for-ascending is consistent with how Windows Explorer does things.

There are also a couple more issues with the Sort Select popup, but I'll need to do some more extensive work to resolve those.


timleslie:
> FTR, arrow-up-for-ascending is consistent with how Windows Explorer does things.

I can also confirm that this is true in OSX Finder.

PR #3001, opened 2020-05-20
(WIP) Extend MySql support

rmalvarado1805:
This adds partial MySQL support as a series of shims/mixins, following previous efforts by @LiamAttClarke on #2577.
The goal is to provide MySQL support without limiting the current capabilities of the project's main database engine (PostgreSQL).


timleslie:
Closing this for now as MySQL support in the `knex` adapter is not something we intend to support. MySQL support may be introduced in the `Prisma` adapter in the future.

PR #3011, opened 2020-05-21
List plugin improvement

gautamsi:
ref: #2838 
This is split from #2838 and works on single feature of adding additional parameter to list plugins.

additional parameter is `{listKey, keystone}` name of the list helps in setting up further if want to build dynamic queries or logging. `keystone` parameter further help add feature in plugin which may need keystone instance. for example some list author can create additional list or meta list. 

( I can think of a list history plugin which can create a list by name of `${listkey}History` and use beforeChange/afterChange hooks to copy changed content to history list). 

[best viewed whitespace off](https://github.com/keystonejs/keystone/pull/3011/files?diff=unified&w=1)

### DIFF
--- a/docs/api/create-list.md
+++ b/docs/api/create-list.md
@@ -286,9 +286,8 @@ Changes the path in the Admin UI. Updating `plural` and `singular` values will n
 
 ### `plugins`
 
-An array of functions that modify config values. Plugin functions receive a config object and can modify or extend this. They should return a valid list config. starting Keystone version 10.0.0, plugins receives second parameter with `{ listKey, keystone}`. Where listKey is the name of list and keystone instance.
+An array of functions that modify config values. Plugin functions receive `(config, { listKey, keystone })`, where `config` is the a list config object, `listKey` is the name of the list, and `keystone` is the keystone object. They should return a valid list config. Plugin functions are executed in the order provided in the list, with the output config of one being passed as input to the next. The output of the final plugin is used to construct the `List` instance.
 
-> This keystone instance is not fully functional when plugins are initialised. Features like `executeQuery` is ready only after `keystone.connect` is run. You can use keystone instance from within hooks and access control methods.
 
 ```javascript
 const setupUserList = ({ fields, ...config }) => {
--- a/docs/api/create-list.md
+++ b/docs/api/create-list.md
@@ -286,7 +286,9 @@ Changes the path in the Admin UI. Updating `plural` and `singular` values will n
 
 ### `plugins`
 
-An array of functions that modify config values. Plugin functions receive a config object and can modify or extend this. They should return a valid list config.
+An array of functions that modify config values. Plugin functions receive a config object and can modify or extend this. They should return a valid list config. starting Keystone version 10.0.0, plugins receives second parameter with `{ listKey, keystone}`. Where listKey is the name of list and keystone instance.
+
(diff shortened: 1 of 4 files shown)


packages/utils/src/index.js
```diff
@@ -110,6 +110,9 @@ export const zipObj = obj =>
 // compose([f, g, h])(o) = h(g(f(o)))
 export const compose = fns => o => fns.reduce((acc, fn) => fn(acc), o);
 
+// composePlugins([f, g, h])(o, e) = h(g(f(o, e), e), e)
+export const composePlugins = fns => (o, e) => fns.reduce((acc, fn) => fn(acc, e), o);
```
timleslie, line 114: I think we'd be better off having this function defined inline in `Keystone/index.js`, as that's the only place it's ever really going to be used, as opposed to the more general `compose` function. In principle (although not quite always in practice) we want to keep the `utils` package generic enough that it could be used by other projects that aren't keystone, and this function is just a bit too keystone specific. Putting the function inline also means we only need to have a `minor` version bump on the keystone package, since it's a backwards compatible change.

docs/api/create-list.md
```diff
@@ -286,7 +286,9 @@ Changes the path in the Admin UI. Updating `plural` and `singular` values will n
 
 ### `plugins`
 
-An array of functions that modify config values. Plugin functions receive a config object and can modify or extend this. They should return a valid list config.
+An array of functions that modify config values. Plugin functions receive a config object and can modify or extend this. They should return a valid list config. starting Keystone version 10.0.0, plugins receives second parameter with `{ listKey, keystone}`. Where listKey is the name of list and keystone instance.
```
timleslie, line 289:
```suggestion
An array of functions that modify config values. Plugin functions receive `(config, { listKey, keystone })`, where `config` is the a list config object, `listKey` is the name of the list, and `keystone` is the keystone object. They should return a valid list config. Plugin functions are executed in the order provided in the list, with the output config of one being passed as input to the next. The output of the final plugin is used to construct the `List` instance.
```

docs/api/create-list.md
```diff
@@ -286,7 +286,9 @@ Changes the path in the Admin UI. Updating `plural` and `singular` values will n
...
+
+> This keystone instance is not fully functional when plugins are initialised. Features like `executeQuery` is ready only after `keystone.connect` is run. You can use keystone instance from within hooks and access control methods.
```
timleslie, line 291:
```suggestion
```

I think we can probably leave this block out, I don't think it introduces more confusion than it solves.

packages/utils/src/index.js
gautamsi: done

gautamsi:
@timleslie should be all good

packages/keystone/lib/Keystone/index.js
```diff
@@ -34,6 +33,9 @@ const List = require('../List');
 const { DEFAULT_DIST_DIR } = require('../../constants');
 const { CustomProvider, ListAuthProvider, ListCRUDProvider } = require('../providers');
 
+// composePlugins([f, g, h])(o, e) = h(g(f(o, e), e), e)
+export const composePlugins = fns => (o, e) => fns.reduce((acc, fn) => fn(acc, e), o);
```
timleslie, line 37: Can we move this down to line 310? It's only used in that one specific place and it's much easier to understand the code when you can see it all in one place. Also, there's no need to be exporting this function.

Issue #1639, opened 2019-09-16
Singleton list types (single item list)


MadeByMike:
A number of people in the community have expressed a desire to be able to configure a global 'options' page. This would work like a Keystone list however there is only a single item.

Currently settings like this are achieved with custom pages. What is wanted is the ability to use the fields API to generate these pages, along with access control and GraphQL endpoints.

I'm suggesting the following API that closely matches the `createList` API:

 ```JavaScript
Keystone.createSingle('Settings', {
  schemaDoc,
  hooks,
  access,
  adapterConfig,
  path,
  plugins,
  fields,
  itemQueryName
})
```
All of the above would work the same as for lists. 

The following list config would not be required on options pages:

`labelField, labelResolver, listQueryName, singular, plural`

I thought potentially you might want a different itemQueryName than what appears on the settings page? However since you would not be required to change the value of a setttings page name due to an ambiguous plural, I don't think we need a 'singular` option? If we do it should just be label?

I only see the need for one query which would match the `itemQueryName`. i.e:

```
settings {
  fieldName
}
```

For mutations we should only need update. No `id` should be required. e.g:

```
updateSettings(data: { siteName: "My Site" }) {}
```

We keep the `data` key for consistency.

timleslie:
I would prefer not to add to the keystone API for this unless we determine that it's not possible to do using pure config and the existing API. From the POV of setting up the list with graphQL, I think everything described here can be achieved with hooks and access control. The admin UI stuff might require some additional config, but we should probably build out the graphQL side of things first and make sure we understand the use case before building in highly specific UI.

MadeByMike:
I get there is not much difference between options and lists in the back-end, however there are big differences between how they are represented in the AdminUI and the required `graphQL` API. I think I am still pro this being a 'special case'.

At the end of the day there are 2 goals for me here: 
 1. First is to make it easy to set-up options pages with the correct GraphQL end-points, hooks, access control etc... so that it effectively looks and smells like a single item. Currently a new user could sink more than a day on setting this up, as it touches multiple advanced concepts.
 1. The other important thing is that the end result looks professional enough to hand over to non-technical users. I.e. Proper treatment of 'Settings' in the admin UI.

Even after our discussion I still feel `.createSingle` or `.createOptions` is worth the cost. You can take a look at any professional CMS and you will find this feature.

gautamsi:
[...]

PR #3013, opened 2020-05-21
Singleton List refactor

gautamsi:
closes #2838 
ref #1639 

~built on top of ~#3012~  #3011, see actual diff for this PR - https://github.com/keystonejs/keystone/pull/3013/files/d16188bfe8eeaec03b0fd23b6c1e9f741034494c..85ba50307242c50eaa84840e6ea2f1d8aded6e05?diff=unified&w=1~

This is refactor from #2838. additional argument to list plugins is moved to #3011 and disabling ui based on access control in Item Detail page footer is also moved to #3012

~this PR adds TWO way of handling singleton features:~

~. if you set `{ adminConfig: { singleton: true }, access: { delete: false } }` (due to change in UI in this PR) this can give you plugin free UI  version of singleton list. List view will redirect to first item in the list and you have no way to create a new item or delete the existing item.~
  ~1.1 if you also set `create` access to false and seed the list (create item on `keystone.connect`) then you do not need any plugin. you are already covered.~

~2. above setting (except 1.1) is set automatically by `singleton` plugin. It also adds ability to create first item then you can not create more item, attempting this would give error in GraphQL. seeding is optional.~

UI portion of changes are split and will be in future PR. 

how to use: 
```js
const { singleton } = require('@keystonejs/list-plugins');

keystone.createList('SettingsList', {
  fields: {
    name: { type: Text },
    slug: { type: Slug, from: 'name' },
    field2: { type: Text },
    field3: { type: Text },
    field4: { type: Text },
    field5: { type: Text },
  },
  plugins: [singleton()],
});
```

@timleslie: should we set `create = false` and enforce seeding to make use of this plugin (if you do not see, it is unusable). Or we leave it as recommended approach for single item list.


packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -278,6 +279,10 @@ const ListPage = props => {
     );
   }
 
+  if (list.adminConfig.singleton && itemCount > 0) {
+    history.push(`${adminPath}/${list.path}/${items[0].id}`);
```
Vultraz, line 283: `list.fullPath` is preferred now.

packages/app-admin-ui/client/pages/Item/ItemTitle.js
```diff
@@ -29,17 +29,21 @@ export const ItemTitle = memo(function ItemTitle({ titleText }) {
...
-            variant="subtle"
-            icon={ChevronLeftIcon}
-            to={list.getFullPersistentPath()}
-            css={{ marginLeft: -12 }}
-          >
-            {list.label}
-          </IconButton>
-          <Search list={list} />
-        </div>
+        {!list.adminConfig.singleton ? (
```
gautamsi, line 32: I am sure we can handle this differently as this will make plugin dependent on UI and not generic feature of UI being changed by plugin.

I need some suggestion on what that alternate way would be for 
* hiding search field and add new button

packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -247,6 +246,12 @@ const ListPage = props => {
     }
   }, []);
 
+  useEffect(() => {
+    if (list.adminConfig.singleton && itemCount > 0) {
```
gautamsi, line 250: again, should this be a different naming to redirect to first item? should we do that as generic feature?

packages/list-plugins/README.md
```diff
@@ -103,3 +103,20 @@ const { createdBy, updatedBy } = require('@keystonejs/list-plugins');
...
+
+## Usage
+
+```js
+const { singleton } = require('@keystonejs/list-plugins');
+
+keystone.createList('ListWithPlugin', {
+  fields: {...},
+  plugins: [
+    singleton(),
```
gautamsi, line 119: This plugin does not need argument. We have this pattern to call the plugin when adding to list, if there is no argument this nesting of function call can be avoided. should this be a pattern or dependent on plugin whether to add as `[plugin]` or as `[plugin()]`.

gautamsi:
@timleslie removed UI changes as requested

PR #3029, opened 2020-05-23
Admin UI: cleaned up react-router usage to match latest recommendations

Vultraz:
As of `react-router` 5.1 it's [recommended](https://reacttraining.com/blog/react-router-v5-1/) to use `children` + hooks instead of `render` or `component`. 5.1 also included a [fix](https://github.com/ReactTraining/react-router/commit/966565956a1634006f0a924ce7078990036737a4) to `children` so they no longer render when the route doesn't match.

These changes should make it easier for me to upgrade to v6 once it comes out.

- Uses `children` for routes instead of `render`. I decided to stick with prop form since this is going to be renamed `element` in router v6 with children reserved for nested routes.
- Moves the no-lists-defined page into its own file to be consistent with the other error/info pages.
- Utilized `useParams` for route params

Minor code formatting:
- `HomePage`: moved the `DocTitle` down into `<main>` and dropped the enclosing `Fragment`
- `ListNotFoundPage`: simplified string construction

### DIFF
--- a/.changeset/wild-mirrors-leave.md
+++ b/.changeset/wild-mirrors-leave.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Cleaned up react-router usage to match latest recommendations.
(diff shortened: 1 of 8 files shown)


Vultraz:
Blah, why does this "Async callback was not invoked within the 5000ms" test failure keep happening :(

timleslie:
Yeah, I dunno, but it's a right pain in the butt :-( If you or someone else has some time to get to the bottom of it that'd be great.

PR #3044, opened 2020-05-26
[RFC] Singleton demo

MadeByMike:
I'd like to work on this more with @gautamsi, @timleslie and the rest of the community.

This demo aims to showcase the list plugins we've created so far as well as how to create one.

In doing so it highlights some of the shortcomings of both Keystone and the current plugins. In particular I'd like to use this example to explore what more we can do to support the concept of a Singleton list type in Keystone.

We should aim to solve the following problems with this demo:

1. The graphQL queries should really be different. Specifically, it makes no sense to use a query like `allSingletons` then pluck the first item, or to provide a `where: { id: "" }` clause, when there is only one item.
2. Additionally there are a bunch of unneeded `gqlNames` on singleton lists. The core architecture should make it easy to extend these.
3. Singleton lists should accept any list key as a name regardless of the plural
4. An initial "item" for singletons should be created automatically
5. We should remove or present singleton items differently on dashboard and navigation
6. We should have no item list page for singletons.
7. The item page and surrounding UI should also be different for singleton items

Each of the above tasks should if possible be looked at as a single PR. As these improvements are added to Keystone we can go back to this RFC and improve until the point that we think it makes good demo.

**About the demo:**

To run checkout branch and go to: `/demo-projects/plugins`

I chose role based access control as I think it's something that a lot of people would like to be able to configure within the AdminUI similar to WordPress or other CMSs with more "content focused" workflow.

There are probably some other challenges this demo highlights, in relation the list plugin architecture, or keystone itself. For example I needed something like a factory function pattern to create plugins that the user then has to add to lists, this is ok, but could the API it be better? Anything else this demo surfaces please try to think of a minimal PR we could spin-off for Keystone add it to the discussion here, we can build it, then come back and improve this demo.

I've converted the issue to a discussion: https://github.com/keystonejs/keystone/discussions/1639 there are few ideas there that might still be relevant.


MadeByMike:
I'd also welcome mock-ups of what the community thinks the dashboard and item page should look like for singletons. Rough sketches are fine.

gautamsi:
I am looking into this, access control may not be best example but it is ok to start with. Role based access control may need more complex setup to be effecting in combination with access config.

more suitable example would be to have theme colour, copyright url, whether signup enabled or not, email verification enforcement etc.

MadeByMike:
> I am looking into this, access control may not be best example but it is ok to start with. Role based access control may need more complex setup to be effecting in combination with access config.
> 
> more suitable example would be to have theme colour, copyright url, whether signup enabled or not, email verification enforcement etc.

I chose this specifically because I want to surface those problems

PR #3048, opened 2020-05-26
Enable list header actions hook

MadeByMike:
@gautamsi I had a little git slip and I had to revert and re-open this one. We're still looking at moving forward with it. Forgive me for the confusion.


packages/app-admin-ui/client/pages/List/CreateItem.js
```diff
@@ -0,0 +1,28 @@
...
+import { PlusIcon } from '@arch-ui/icons';
+
+import { useList } from '../../providers/List';
+
+const CreateItem = () => {
+  const {
+    list: { access },
+    openCreateItemModal,
+  } = useList();
+  if (!access.create) return null;
```
gautamsi, line 14: We re taking care of access control here instead of List->index

packages/app-admin-ui/client/pages/List/index.js
```diff
@@ -70,7 +68,7 @@ export function ListLayout(props) {
         <HeaderInset>
           <FlexGroup align="center" justify="space-between">
             <PageTitle>{list.plural}</PageTitle>
-            {list.access.create ? listHeaderActions ? listHeaderActions() : <CreateItem /> : null}
+            {listHeaderActions ? listHeaderActions() : <CreateItem />}
```
gautamsi, line 71: Do not want to remove all the header items which dev may have created, list create access check is moved to `<CreateItem/>` which will not render itself if there is no access for create.

we can leave this to hook author to decide whether they need action for create access or not.

MadeByMike: Good call to move it to create item.

.changeset/real-pots-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': minor
+---
+
+enable use of `listHeaderActions` hook
```
MadeByMike, line 5:
```suggestion
Enabled the use of `listHeaderActions` hook in the Admin UI.

Usage: 

new AdminUIApp({
  hooks: require.resolve('./admin-ui/'),
});

The index file in the `admin-ui` directory should export a hooks which will be packaged for use in the Admin UI during the Keystone build:

import { CreateItem } '@keystonejs/admin-ui/components/'
export default {
  // re-implement the default create item button + custom text
  listHeaderActions: () => (<div><CreateItem /><p>Hello world</p></div>), 
};
```

PR #3051, opened 2020-05-27
Make simulated password self documenting

mikehazell:
I was reading some code and found `password1234` hard coded in the auth provider and it gave me a small hart attack.

Thought this might save someone else going on the same heart attack inducing goose chase.

### DIFF
--- a/.changeset/fluffy-baboons-relax.md
+++ b/.changeset/fluffy-baboons-relax.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/auth-password': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/fluffy-baboons-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/auth-password': patch
+---
+
+Made simulated password self-documenting
```
MadeByMike, line 5:
```suggestion
Made simulated password self-documenting.
```

PR #3056, opened 2020-05-27
Admin UI: fetch most list metadata via query

Vultraz:
~Relies on #3055.~

This covers most *list* metadata except for `adminConfig` (which I have an idea for once #2745 is merged) and `fields`, which will be handled separately.

Lists to which you do not have `read` access will no longer show up in the sidebar, and attempting to navigate to them will show a not-found error rather than a no-access error. Let me know if this is desirable.

Also adds a `schema.sortType` field to the meta query. *Might* move it if it feels out of place.

I also needed to include two extra fixes to get everything working. These have their own changeset:
- Don't refetch queries when signing in/out (new behavior still clears the cache, just doesn't do a refetch). Fetching new authed data when you're signing out didn't make much sense and was causing errors trying to re-render the `MasterListProvider`. I also  `await` ed the `clearCache` calls to ensure they finish before proceeding. 
- Fixed the Sign Out nav link acting like an external link and causing a full page refresh.


packages/app-admin-ui/client/providers/MasterList.js
```diff
@@ -0,0 +1,176 @@
...
+    partialRefetch: true,
+  });
+
+  // Don't render the rest of the app until this is done fetching everything
+  if (loading) {
+    return <FetchingListsPage />;
+  }
+
+  //
+  // We've consciously made a design choice that the `read` permission on a
```
Vultraz, line 75: This block was copied from `keystone.getAdminMeta`.

packages/keystone/lib/ListTypes/list.js
```diff
@@ -326,7 +326,6 @@ module.exports = class List {
   }
 
   getAdminMeta({ schemaName }) {
-    const schemaAccess = this.access[schemaName];
```
Vultraz, line 329: Most of `list.getAdminMeta` has been removed.

test-projects/access-control/cypress/integration/list/admin-ui.js
```diff
@@ -79,13 +79,13 @@ describe('Access Control Lists > Admin UI', () => {
         const prettyName = prettyListName(name);
         const slug = listSlug(name);
 
-        // Still navigable
-        cy.get('body nav').should('contain', prettyName);
+        // Not navigable
+        cy.get('body nav').should('not.contain', prettyName);
```
Vultraz, line 83: Non-readable lists no longer show up in the sidebar.

test-projects/access-control/cypress/integration/list/admin-ui.js
```diff
@@ -116,13 +116,13 @@ describe('Access Control Lists > Admin UI', () => {
         const prettyName = prettyListName(name);
         const slug = listSlug(name);
 
-        // Still navigable
-        cy.get('body nav').should('contain', prettyName);
+        // Not navigable
```
Vultraz, line 119: Non-readable lists no longer show up in the sidebar.

test-projects/access-control/cypress/integration/list/admin-ui.js
```diff
@@ -159,13 +159,13 @@ describe('Access Control Lists > Admin UI', () => {
           const prettyName = prettyListName(name);
           const slug = listSlug(name);
 
-          // Still navigable
-          cy.get('body nav').should('contain', prettyName);
+          // Not navigable
```
Vultraz, line 162: I think this test was backwards. Its description said "does **not** show items when readable & not admin" but it was checking that the sidebar did contain it. Changed to not contain.

Vultraz:
Green!

packages/app-admin-ui/client/pages/Home/index.js
```diff
@@ -51,33 +52,11 @@ const Homepage = () => {
...
-      return (
-        <PageError>
-          <p>{error.message}</p>
-        </PageError>
-      );
-    }
-
-    allowedLists = allowedLists.filter(
-      list => deniedQueries.indexOf(list.gqlNames.listQueryMetaName) === -1
+    return (
```
Vultraz, line 56: We're no longer using Access Denied Errors to filter dashboard cards to show, so I figured it's better to just show any error.

packages/app-admin-ui/client/providers/MasterList.js
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+    }
+  }
+`;
+
+/**
+ * Handles metadata on all available lists.
+ */
+export const MasterListProvider = ({ children }) => {
```
gautamsi, line 64: should we use a better name for this? `ListMetaProvider`? and `useListmeta` may be?

Vultraz: I like. 👍

.changeset/twelve-ducks-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/app-admin-ui': patch
```
gautamsi, line 2: you can merge changesets I guess.
manually editing is fine.

Vultraz: Well, I wanted these to appear as patch changes.

packages/keystone/lib/ListTypes/list.js
gautamsi: this `list.getAdminmeta` may be useful to wrap few things from new provider, can it be?

Vultraz: This is server-side. What were you thinking of putting here?

.changeset/twelve-ducks-mix.md
gautamsi: you need to split PR if you want separate patch changes independent of this change, having minor and extra patch change, what value it has

packages/keystone/lib/ListTypes/list.js
gautamsi: ah, I missed, there is another list class in client side as well.

I guess you did not reuse the getAdminMeta to populate the resulting gql response.

Vultraz: Indeed. That is handled by a combination of `list.listMeta` and the CRUD provider.

Vultraz:
Split out the signin/out fixes.

MadeByMike:
I updated this one to give it another test drive and I do like the meta provider. But I'm less convinced about using a query to fetch the list and spoke with @JedWatson about it as well because he's been putting a lot of thought into how the next generation of the AdminUI might look.

The short sound bite from that conversation is: "I think we want a less dynamic runtime, not more".

The other concern is I tested it on a rather large schema and there is some delay we could possibly avoid.

Sadly, I think we should decline this PR in favour of the codegen example Jed started. We'll get a PR up shortly and see if we can get your input.

PR #3066, opened 2020-05-29
Formatted CRUD provider typedefs, moved to own file (no functional changes)

Vultraz:
Used [single-line comments](https://www.apollographql.com/docs/apollo-server/schema/schema/#documentation-strings) where applicable and ran each block through prettier with a `gql` literal attached. Also moved them to a getter helper in a separate file since they're getting really long.

### DIFF
--- a/packages/keystone/lib/providers/listCRUD.js
+++ b/packages/keystone/lib/providers/listCRUD.js
@@ -15,7 +15,7 @@ class ListCRUDProvider {
   getTypes({ schemaName }) {
     return unique([
       ...flatten(this.lists.map(list => list.getGqlTypes({ schemaName }))),
-      ..._getListCRUDTypes(this.gqlNames)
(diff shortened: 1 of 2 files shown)


packages/keystone/lib/providers/listCRUD.js
```diff
@@ -1,6 +1,8 @@
 const { GraphQLJSON } = require('graphql-type-json');
 const { flatten, objMerge, unique } = require('@keystonejs/utils');
 
+const { _getListCRUDTypes } = require('./listCRUDTypes');
```
JedWatson, line 4: This feels like a weird place for a dangling underscore. Is there a convention we're using that I missed? I've seen us use `_method` to informally denote "private" API on classes, but it feels weird when used as an export/import

Vultraz: Nope, not a convention, just a random choice on my part. Removed 👍

PR #3068, opened 2020-05-30
Improved image-not-loaded File field display

Vultraz:
Utilizes [`react-image`](https://github.com/mbrevda/react-image).

Before:
image

After:
image

This also means there's now a loading indicator while the image is loading.


Vultraz:
Asynncc calllbacckk

PR #3069, opened 2020-05-30
Don't parse the GraphQL schema twice on load

Vultraz:
Apollo already accepts the AST output of `gql`. No need to parse the schema both in ks and then again in Apollo.

Honestly, I might make `dumpSchema` just return a string so we can avoid that boilerplate in the tests. 😐


timleslie:
> Honestly, I might make `dumpSchema` just return a string so we can avoid that boilerplate in the tests. 😐

Yeah, this is a slightly awkward corner of the keystone object API. `getTypeDefs` is good when you're Apollo, and `dumpSchema` is good if you want to export your schema to a file, but we don't have anything in between for when you want your schema returned as just a string.

This PR should really be a `major` change, since we're changing the return type of `getTypeDefs`. If we're going to do that then I think we should really resolve the whole API all at once.

I think we could easily have `dumpSchema` return `schema`. If we also put in a check for `if (file)` and default the value of `schemaName` to `"public"` then we could have a backwards compatible API that would let you do:

```
console.log(keystone.dumpSchema())
```

and have it do something meaningful.

If we wanted to go with a backwards incompatible change then I'd propose the same behaviour but just change the methods args to be an object, e.g. `dumpSchema({ file, schemaName = 'public' })`.

Thoughts?

.changeset/four-trains-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Don't parse the GraphQL schema twice.
```
MadeByMike, line 5: Prevented parsing theGraphQL schema twice. Internal refactor only.

MadeByMike, line 5:
```suggestion
Prevented parsing theGraphQL schema twice. Internal refactor only.
```

Vultraz:
Idea I had last night want to make `dumpSchema` just return a string, then add a `--dumpSchemaTo` (or similarly named) argument to the CLI that handles the write-to-file bit. Right now, the only place `dumpSchema` is actually mentioned is the [performance monitoring docs](https://www.keystonejs.com/guides/performance#push-schema-to-apollo-engine). They're pretty outdated at this point (working on that), but the method they recommend to use `dumpSchema` is awkward.

timleslie:
I think I'd be in favour of your suggestion of making `dumpSchema` simply return a string. It could then be: `  dumpSchema(schemaName = 'public') {...`, and people could then do whatever they wanted with that string, whether it's writing it to a file, dumping it to the console, etc, etc.

I don't think there's any value in adding a CLI option, it just feels like API cruft to me.

Vultraz:
@timleslie ok, done

Vultraz:
BTW, does this really need to be a major patch? getTypeDefs is an internal function and dumpSchema is basically undocumented. I don't think it make sense to do a major version bump any time any API changes.

.changeset/four-trains-turn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystonejs/keystone': major
+---
+
+Prevented parsing the GraphQL schema twice.
+- `keystone.getTypeDefs` now returns the parsed GraphQL AST instead of the raw SDL.
+- `keystone.dumpSchema` now returns a the GraphQL schema as a string instead of writing it to file. Additionally, its first `file` argument was removed and now only takes a the schema name, which defaults to `public`.
```
timleslie, line 7:
```suggestion
- `keystone.dumpSchema` now returns the GraphQL schema as a string instead of writing it to file. Additionally, its first `file` argument was removed and now only takes a the schema name, which defaults to `public`.
```

PR #3071, opened 2020-05-31
Updated missing Docs for ui hooks

gautamsi:
some minimal docs for ui hooks
also fixed example in one of the merged changeset.

### DIFF
--- a/packages/app-admin-ui/README.md
+++ b/packages/app-admin-ui/README.md
@@ -86,8 +86,7 @@ Header components on the Item Details page can be replaced using this hook. Ths 
 
 > This must return a React component.
 
-```javascript
-// ./admin-ui/index.js
+```javascript title=/admin-ui/index.js
 import { ItemId, AddNewItem } '@keystonejs/admin-ui/components/'
 export default {
   // re-implement the default AddNewItem and ItemId button + custom text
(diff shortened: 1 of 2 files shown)


gautamsi:
I have to add document for #2483 adding which now will cause the conflict. once either is merged, I will fix the docs.

packages/app-admin-ui/README.md
```diff
@@ -72,14 +72,59 @@ export default {
...
+```
+
+#### `listHeaderActions`
+
+Header components on the List page can be replaced using this hook. This replaces components on list page for all Lists.
+
+> This must return a React component.
+
+```javascript
+// ./admin-ui/index.js
```
Vultraz, line 105: Should use the `title=./admin-ui/index.js` syntax on the line above.

gautamsi: i did that but what advantage it has?

Vultraz: image

gautamsi: thought so but the rendering does not work. does leading `.` matter in file name?

gautamsi: or is it thing for rendering in keystonejs.com?

Vultraz: Perhaps it's the leading `.` 🤔

gautamsi: I see that it is rendering on docs website

PR #3080, opened 2020-06-01
Updated Next.js dependency to 9.4.4

Vultraz:
Looks like 9.4 has some nice stuff: https://nextjs.org/blog/next-9-4

@MadeByMike I have no idea how you got partial commit attribution on this. I definitely did something wonky locally... 😂


gautamsi:
was updating dependency enough? I had some issue when I tried to update. I needed that for `NODE_OPTIONS=--inspect` to work which is a known issue with current version in keystone.

Vultraz:
I was able to run the meetup demo. What issues did you see?

gautamsi:
I dont remember now, there was some complain about babel i believe

Issue #2501, opened 2020-03-10
Docs homepage design tweaks


jesstelford:
See: https://github.com/keystonejs/keystone/pull/2475

@jordanoverbye did an excellent overhaul of the docs homepage which does a much better job of giving new users a jumping off point.

I think the wording on the homepage could be tweaked to be less generic and also more compelling for our target audience:

<img width="574" alt="Screen Shot 2020-03-10 at 10 26 11 am" src="https://user-images.githubusercontent.com/612020/76268622-be0e3100-62c2-11ea-9023-771dadfbf1c1.png">

There are a couple of things we could improve:

- Remove the "Wall of text" effect of the large header + long paragraph
- Aim for a "6 word" version which conveys as much meaning as possible
- Remove vague phrases such as "platform", "extensible", and "applications"
- Remove some of the buzz words
- Ensure the heading and the paragraph compliment eachother rather than restating
- Target our audience better. We're not for "All NodeJS application developers", we're for "Frontend devs who need a backend"

Random brainstormed ideas:

---

> **Headless GraphQL API framework for Node**

> Configure your data model in JavaScript, and KeystoneJS will generate a powerful GraphQL API + Admin UI.

---

> **For frontend devs who need a GraphQL backend**

> Configured in plain old JavaScript, KeytoneJS will auto generate a powerful GraphQL API + Admin UI in just minutes.

---

> **BYOF`*` GraphQL CMS for Frontend Devs**

> KeystoneJS auto generates your GraphQL API + Admin UI from plain old JavaScript so you can get back to building your product sooner. As your product matures, KeystoneJS is right there for you with multiple extensibility options and a strong community of passionate developers just like you.

> _`*`BYOF: Bring Your Own Frontend_

Vultraz:
Out of the three options I like the last best.

jahlherapath:
I like option one straight to the point and you know exactly what it is.

PR #3094, opened 2020-06-03
Update homepage copy

MadeByMike:
Closes: #2501


website/src/components/homepage/SectionHero.js
```diff
@@ -13,11 +13,10 @@ const SectionHero = () => (
     <Grid>
       <div>
         <Heading>
-          A scalable, extensible and open-source platform to build NodeJS applications
+          Headless CMS &amp; GraphQL API for Node
         </Heading>
         <Description>
-          KeystoneJS comes with first-class GraphQL support, a highly extensible architecture, and a
-          wonderful Admin UI
+          Configure your data model in JavaScript, and KeystoneJS will generate a powerful GraphQL API and CMS.
```
jesstelford, line 19: Just after this paragraph, the phrase `schema` is used instead of `data model`. Might be best to keep it consistent?
Screen Shot 2020-06-03 at 11 30 12 am

```suggestion
          Configure your schema in JavaScript, and KeystoneJS will generate a powerful GraphQL API and CMS.
```

jordanoverbye:
I think what you've said is good, but it's a shame theres a widow when the container is at full-width (1440px and above). Not the end of the world, but would be nice to fix

Screen Shot 2020-06-03 at 1 33 28 pm

jesstelford:
@jordanoverbye Fixed in https://github.com/keystonejs/keystone/pull/3099 👍

PR #3095, opened 2020-06-03
Provide all resolver info to custom access control functions

timleslie:


packages/keystone/lib/Keystone/index.js
```diff
@@ -659,6 +664,7 @@ module.exports = class Keystone {
     this.executeQuery = this._buildQueryHelper(
       this.getGraphQlContext({
         skipAccessControl: true,
+        // This is fucked!
```
jesstelford, line 667: 🤣

MadeByMike: Documentation is important

timleslie: Nothing to see here 😓

packages/keystone/lib/Keystone/index.js
```diff
@@ -157,10 +157,15 @@ module.exports = class Keystone {
       // We do it within the request callback so we can resolve it based on the
       // request info ( like who's logged in right now, etc)
       getCustomAccessControlForUser = memoize(
-        async access => {
+        async (item, args, context, info, access, gqlName) => {
```
jesstelford, line 160: Is this a major breaking change? Is `getCustomAccessControlForUser` considered part of our external API?

jesstelford, line 160: For consistency with how we call other methods internally, should this use object notation?

```suggestion
        async ({ item, args, context, info, access, gqlName }) => {
```

timleslie: Gosh I hope not. If someone's digging that deep then... 🤷 I think it's safe to call this an internal function, even if it is technically publicly available.

timleslie: Probably, but this keeps it locally consistent with the other internal functions in`getGraphQlContext`. I'd rather tackle this question as a separate PR, or else we'll end up changing everything.

PR #3112, opened 2020-06-04
Admin UI: converted ConnectivityListener component to a custom hook

Vultraz:
And cleaned it up a bit. Also removed it from the public app since editing isn't available there.


Vultraz:
Fixed conflicts.

PR #3120, opened 2020-06-05
adapter-knex: self-referential patch

pahaz:
At the moment it looks like no self-referential relationships are set to NULL.

image

I've tested it on my code. I really have 4 self-referential relationships. And the current logic is not detecting these relations.

# my Test model example #
```
keystone.createList('Test', {
    fields: {
        ...
        related: {
            type: Relationship,
            ref: 'Test',
            many: true,
        },
        self: {
            type: Relationship,
            ref: 'Test',
            many: false,
        },
    },
})
...
```
image
image

As you can see there is no `right` object here. Just `left` for all self-referential relations.


Vultraz:
Could you add a changeset?

PR #3137, opened 2020-06-10
Add "prop drawer"

MadeByMike:
image

I'm sure there is a simpler way to capture all selection changes and re-render the <PropsDrawer/> currently combining useEffect and some KeyUp events on the editor. Slate docs seem to have no standard way. Not sure why the BlockButtons always seem to re-render with selection changes.

### DIFF
--- a/packages/fields/package.json
+++ b/packages/fields/package.json
@@ -66,9 +66,9 @@
     "react-dom": "^16.13.1",
     "react-select": "^3.1.0",
     "react-toast-notifications": "^2.3.0",
-    "slate": "^0.47.4",
+    "slate": "^0.58.1",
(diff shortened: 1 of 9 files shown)


packages/fields/package.json
```diff
@@ -66,9 +66,9 @@
     "react-dom": "^16.13.1",
     "react-select": "^3.1.0",
     "react-toast-notifications": "^2.3.0",
-    "slate": "^0.47.4",
+    "slate": "^0.58.1",
```
JedWatson, line 69: FYI I've specifically left this out of my commits for now, just to hilight the problem merging in a PR that updates slate without allowing different versions between packages

Vultraz: Does this supersede #2510 ?

PR #3141, opened 2020-06-10
Adding URL formulation rules to the style guide

molomby:
Codifying the url formulation rules [we agreed back in #2041](https://github.com/keystonejs/keystone/issues/2041#issuecomment-566802733) by adding them to the style guide.

Now that we've figured out the lingering issue with NPM packages (#3139) there should be nothing stopping this from this from becoming the One True Way.

### DIFF
--- a/.changeset/real-plums-remain.md
+++ b/.changeset/real-plums-remain.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Adding URL formulation rules to the style guide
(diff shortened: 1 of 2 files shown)


.changeset/real-plums-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Adding URL formulation rules to the style guide
```
MadeByMike, line 5:
```suggestion
Adding URL formulation rules to the style guide.
```

PR #3146, opened 2020-06-11
Migrated to @primer/octicons-react v10

Vultraz:
The new [Octicons 10](https://github.com/primer/octicons/releases/tag/v10.0.0) no longer requires the awkward `<Octicons>` wrapper, so we can use the `@primer/octicons-react` package as a drop-in replacement for `@arch-ui/icons` without the maintenance overhead of manually creating icons from`@primer/octicons`.

I did a test, everything seems to be working fine. There are a couple icons we might want to change, and might want to use the new (?) `size` prop, but those can be done separately.


Vultraz:
Fixed conflicts.

gautamsi:
does this mean we are deprecating `@arch-ui/icons`?

Vultraz:
I support that.

PR #3147, opened 2020-06-12
[WIP] Access control query helper

gautamsi:
I have seen lot of request and query regarding how you can have gql query inside access control. 

seems like it was not easy. 
this is an effort to add this feature and guidelines as if there is lot of gql query it may slow down the app, there are lot of call to access control functions.

@timleslie do you think we should do this? or have some other pattern. I see there are hacks available to do it (wrap the schema in function accepting keyston instance and hold it there when returning the list config)

### DIFF
--- a/packages/keystone/lib/ListTypes/list.js
+++ b/packages/keystone/lib/ListTypes/list.js
@@ -437,7 +437,8 @@ module.exports = class List {
           data,
           existingItem,
           operation,
-          { gqlName, itemId: id, ...extraInternalData }
+          { gqlName, itemId: id, ...extraInternalData },
+          context
         );
         if (!access) {
           restrictedFields.push(field.path);
@@ -455,7 +456,7 @@ module.exports = class List {
     const access = await context.getListAccessControlForUser(this.key, originalInput, operation, {
       gqlName,
       ...extraInternalData,
-    });
+    }, context);
     if (!access) {
       graphqlLogger.debug(
         { operation, access, gqlName, ...extraInternalData },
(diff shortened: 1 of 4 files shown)


packages/keystone/lib/ListTypes/list.js
```diff
@@ -1483,11 +1484,11 @@ module.exports = class List {
       // NOTE: These could return a Boolean or a JSON object (if using the
       // declarative syntax)
       getAccess: () => ({
-        getCreate: () => context.getListAccessControlForUser(this.key, undefined, 'create'),
-        getRead: () => context.getListAccessControlForUser(this.key, undefined, 'read'),
-        getUpdate: () => context.getListAccessControlForUser(this.key, undefined, 'update'),
-        getDelete: () => context.getListAccessControlForUser(this.key, undefined, 'delete'),
-        getAuth: () => context.getAuthAccessControlForUser(this.key),
+        getCreate: () => context.getListAccessControlForUser(this.key, undefined, 'create', undefined, context),
```
gautamsi, line 1487: this is tricky part. In Future we may be using list meta queries to determine the access to a list and item, this way if you decide to not run this query for each item then the feature will still remain incomplete. Some users have sought out that they want to only allow deletion if the relationship item has some value. 

top level list access call may not be accurate.

Top level access check using gql should only be done for `create` access. 

should there be item specific meta when you use read query?
```json
{ 
  data: {
    allUsers: [
      {
         name: "user1",
         // other fields ....
        ___meta: {
           access: { update: false, delete: true },
         },
      },
    ]
}
      

```

timleslie:
Hey @gautamsi, thanks for bringing this up. I've also seen a lot of questions in this area, and I think it's something we can solve reasonably easily.

As well as having helpers in access control, we also need them in custom queries/mutations, hooks, virtual fields, and other server-side scripts. I'd like to come up with a general solution that covers all these bases in a consistent manner. I don't think it should be too hard, it'll just take a couple of tweaks on what we already have available. I'll write up a discussion around this today to outline the surface area that such a feature would need to support and examples of how it would be used in different contexts 👍

gautamsi:
one I have suggested and that worked for a dev is to wrap the field config in function

in schema
```js
exports.Post = keystone => ({
  fields: {
    title: { type: Text },
    //  other fields
  },
  hooks: {
    afterDelete: ({ existingItem }) => {
      if (existingItem.image) {
        fileAdapter.delete(existingItem.image);
      }
    },
  },
  access: {
    delete: async ({ authentication: { item, listKey }, existingItem }) => {
      const allu = await keystone.executeQuery(`{allUsers { name email isAdmin id }}`);
      console.log(JSON.stringify(allu,null, 2));
      return false;
    }
  }
});
```
use schema as `keystone.createList('Post', Post(keystone));`

timleslie:
Yep, that pattern of making the `keystone` object directly available to the access control methods is definitely the best work around at the moment, and is how I solve the issue in my projects for now 👍

timleslie:
I've written up my ideas in #3150 . I think it should adequately cover the use case required by this PR 👍

gautamsi:
I see that, this PR will be obsolete then :)

gautamsi:
closed in #3154 
any place you see `context` (GraphQL context) object, it has the features now to help query. more documentation updates will be done soon.

PR #3151, opened 2020-06-15
[RFC] POC of a new server-side graphql API

timleslie:
This is a POC implementation to go with the discussion in #3150. This code is purely to give a concrete thing to use in the discussion, this PR will never get merged. If #3150 is agreed on it will be implemented in a series of separate PRs.


timleslie:
This feature has now been completely implemented 🎉

PR #3152, opened 2020-06-15
[WIP] Field EditorJs

gautamsi:
Another attempt to have a block based editor which is based on https://editorjs.io/

**Why another block based editor field?**
I think this one is more simple and have some features without the complexity, we even have some npm package to convert most of the existing blocks to render them.

This renders something like this with default:

editorjs

I could not cove all the features, try creating a field `body: { type: EditorJs },` after importing it as `const { EditorJs } = require('@keystonejs/fields-editorjs');` and see yourself. 

more demo is on the website https://editorjs.io/

There is one caveat with this type of field, onChange for every text entry is expensive. There should be a hook (no UI hook) in admin UI which would call Field method to check if there is pending changes instead of tracking internal chnaged values based on onChange event.

TODO:
- [ ] Tools should be selectable
- [ ] have to enable tools specific options
- [ ] enable custom tools via hooks or some other way
- [ ] enable image upload

Some of the TODO will be done if we agree to add it in this repo.


Vultraz:
This looks a *lot* simpler than the Slate field, even with my Slate 0.5x work. I like 👍

gautamsi:
@timleslie @MadeByMike any feedback on this, worth completing or keep this a contrib package?

MadeByMike:
@gautamsi really sorry for taking so long to look to look at this. In terms of Keystone priorities a new WYSIWYG field has not been high on the list recently.

Although this is undeniably simpler and we don't love the existing slate document field I'd find it hare justify introducing a third WYSIWYG/blocks editor into core fields package at this time. @JedWatson in particular has done a few exploration branch to investigate how we might update the existing document field and what we've learnt is there a few other things that should be in place before we start. Namely better UI primitives that we'd like to integrate into the blocks.

@JedWatson would it be fair to say that you think the new version of Slate is the right platform to build a new version of a document field? I'd encourage you to have a look at this code - it is simple, and it does work with serialised JSON. At this stage it doesn't provide any custom blocks that show how to extend beyond standard WYSIWYG fields.

@JedWatson after looking at this it would be good for you to make a clear all on what the future of the blocks editor looks like. If it's not EditorJS, should we support another editor?

jesstelford:
Looking great @gautamsi!

This could live in a 3rd party npm package while we sort out what is/isn't in core, then we could link to it from the docs. I wouldn't want to block such awesome work from getting out there and being used as long as you're happy to contribute maintaining it as a seperate package for now?

gautamsi:
EditoJS also has plugin system which can be used to create custom blocks.

Currently Cloudinary, OEmbed and Unsplash fields adds block feature for Content field. EditorJS has Image upload feature as well as link validation which generate OEmbed like block, a backend implementation in PHP is available in their repo which I have to port to nodejs code aligned to keystone. I believe writing a plugin for unsplash would be easy and can be based on existing OEmbed like Link feature they have.

gautamsi:
> This could live in a 3rd party npm package while we sort out what is/isn't in core, then we could link to it from the docs. I wouldn't want to block such awesome work from getting out there and being used as long as you're happy to contribute maintaining it as a seperate package for now?

Sure, I am up for that. I do need little help in setting that up with build system and publishing workflow to npm. It is easier to just copy them in project and import the file, how do I do that with external lib, something which I have not figured out yet.

if you want to help, I will invite after copying the files in `keystonejs-contrib` org soon.

MadeByMike:
@gautamsi let me know when you've copied that over and I'll see if I can help (or get someone to help) with NPM publishing workflow.

gautamsi:
@MadeByMike nevermind, I was able to setup this with `build-field-types` and published "as is" (this PR WIP) for testing with versiopn `0.1.0` https://www.npmjs.com/package/@keystonejs-contrib/fields-editorjs

PR #3154, opened 2020-06-16
Improved server-side GraphQL support

timleslie:
This adds the key functionality for #3150. Once this is live we can look at deprecating/removing the other bits of server-side graphQL support which we no longer need.

### DIFF
--- a/docs/discussions/server-side-graphql.md
+++ b/docs/discussions/server-side-graphql.md
@@ -1,0 +1,118 @@
+<!--[meta]
+section: discussions
+title: Server-side GraphQL Operations
+[meta]-->
+
+# Server-side GraphQL Operations
+
+Keystone provides a GraphQL API which we generally think of as being an API that is accessed over HTTP from a client talking to a server.
+There are many uses cases however where we want to execute GraphQL operations directly on the server.
+These can be grouped into two main categories:
+
+As part of a resolver, e.g.
+
+- Hooks
+- Access control
+- Virtual field resolver
+- Custom field type resolvers
+- Custom query/mutation resolvers
+
+As part of some other process, e.g.
+
+- Migrations
+- Data seeding
+- Responding to a non-GraphQL HTTP request
+- Testing
+- Anything else a developer could possibly imagine
+
+Keystone provides an API which give you complete control over how you execute server-side GraphQL operations in each of these situations.
+
+## Background
+
+There are four main things that go into executing a server-side GraphQL operation:
+
+- A GraphQL schema
+- The query string
+- Variable values
+- A `context` object
+
+When executing queries over HTTP we only have control over the queries and variables.
+On the server side we can also control the schema and the `context`.
+
+In most Keystone systems there will only be a single schema, but it is possible to configure multiple schemas (although this is not well documented).
+
+The [`context` object](https://www.apollographql.com/docs/apollo-server/data/resolvers/#the-context-argument) is a general purpose object which is made available to all resolvers. When performing HTTP queries this object is built by Apollo server based on the `request` object. When directly performing server-side operations we need to be able to create our own `context` object.
+
+## Within Resolvers
+
+While Keystone generates the basic CRUD resolvers for you, there are lots of ways you can write custom code which is executed during the resolver phase of a query.
+These include:
+
+- Hooks
+- Access control
+- Virtual field resolver
+- Custom field type resolvers
+- Custom query/mutation resolvers
+
+The [signature of a resolver function](https://www.apollographql.com/docs/apollo-server/data/resolvers/#resolver-arguments) is `(parent, args, context, info)`.
+The `context` argument is an object which is provided to all resolvers and can contain and state or functions which should be shared across all resolvers.
+The `context` object is also passed through to all hooks and access control functions.
+
+As a developer you might want to execute some GraphQL operations as part of your custom code.
+This can be done using the function `context.executeGraphQL()`, which is added to the `context` object by Keystone.
+This async function takes a `query` and `variables` as arguments and returns `{ data, errors }` as a response.
+
+```
+const { data, errors } = await context.executeGraphQL({ query: '{...}', variables: { ... } });
+```
+
+### Custom context
+
+The function `context.executeGraphQL` will pass `context` through to the resolvers used in the operation.
+There are some instances where you may wish to use a different context to perform the operation.
+For example, you might wish to execute your operation with all access-control turned off, or authenticated as a different user.
+
+This can be achieved using the function `context.createContext({ schemaName, authentication, skipAccessControl })`, which will create a new `context` object based on the supplied arguments.
+
+- `schemaName`: This can be used to select the named schema if you have multiple schemas within your Keystone system.
+- `authentication`: This is an object of the shape `{ item: { id }, listKey: '...' }` which is passed through to access control functions. This can be used to emulate executing the operations as a different user.
+- `skipAccessControl`: This can be used to bypass all access control.
+
+By default each of these values will take the value of the original `context` object.
+You can use this functions like this:
+
+```
+const { data, errors } = await context.executeGraphQL({
+  context: context.createContext({ ... }),
+  query: '{...}',
+  variables: { ... }
+  });
+```
+
+## General purpose operations
+
+There are situations where you might want to execute a GraphQL operations outside of the context of a resolver function.
+These include:
+
+- Migrations
+- Data seeding
+- Responding to a non-GraphQL HTTP request
+- Testing
+- Anything else a developer could possibly imagine
+
+In these cases there is no `context` object for you to borrow, but there is an API to create one for use.
+
+The methods `keystone.createContext()` and `keystone.executeGraphQL()` behave almost exactly the same as the functions `context.createContext()` and `context.executeGraphQL()` respectively. There a few key differences
+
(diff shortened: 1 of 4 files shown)


docs/discussions/server-side-graphql.md
```diff
@@ -0,0 +1,123 @@
...
+While Keystone generates the basic CRUD resolvers for you, there are lots of ways you can write custom code which is executed during the resolver phase of a query.
+These include:
+
+- Hooks
+- Access control
+- Virtual field resolver
+- Custom field type resolvers
+- Custom query/mutation resolvers
+
+As a developer you might want to execute some GraphQL operations as part of your custom code.
```
MadeByMike, line 61: You'd get a little confused at this point if you don't know how a context object is provided to a resolved function I think we need to describe or show this before this point.

timleslie: Hmm, I wonder if we should just link out to the apollo docs at that point. If you're writing a custom resolver and you don't know about `context` then... 🤷 Maybe an simple example of using it in a hook and within access control, since those aren't full blown resolvers, but they have happen to have access to the context object, since they're called as part of the parent resolver.

docs/discussions/server-side-graphql.md
```diff
@@ -0,0 +1,123 @@
...
+
+- Migrations
+- Data seeding
+- Responding to a non-GraphQL HTTP request
+- Testing
+- Anything else a developer could possibly imagine
+
+In these cases there is no `context` object for you to borrow, but there is an API to create one for use.
+
+The methods `keystone.createContext()` and `keystone.executeGraphQL()` behave almost exactly the same as the functions `context.createContext()` and `context.executeGraphQL()`. There a few key differences
```
MadeByMike, line 110: I think you meant:
 
```suggestion
The method `keystone.executeQuery()` behaves almost exactly the same as the functions `context.createContext()` and `context.executeGraphQL()` with a few key differences:
```

timleslie: No, I meant what I had originally, I'm comparing each `keystone.foo` to the corresponding `context.foo`.

docs/discussions/server-side-graphql.md
timleslie: Or maybe we flip the order of this paragraph with the following one, where `context` objects are explained. I had something like that originally but then flipped it to this, but maybe that made it worse. 🤔

MadeByMike: This is also available in our hooks and stuff too right? It's less clear at that point that this is an Apollo context object.

MadeByMike: I've approved. Just this one general comment that dosen't need to block it.

packages/keystone/lib/Keystone/index.js
```diff
@@ -228,20 +228,50 @@ module.exports = class Keystone {
...
-      ...this._sessionManager.getContext(req),
-      ...this._getAccessControlContext({
+      ...this.createContext({
         schemaName,
         authentication: { item: req.user, listKey: req.authedListKey },
         skipAccessControl,
       }),
-      totalResults: 0,
-      maxTotalResults: this.queryLimits.maxTotalResults,
+      ...this._sessionManager.getContext(req),
```
gautamsi, line 274: `...this._sessionManager.getContext(req),` is this going to override `skipAccessControl` or anything else?

timleslie: Nope, `this._sessionManager.getContext(req)` returns `{ startAuthedSession, endAuthedSession, authedItem, authedListKey }`, however those last two values are redundant (we already compute them in `createContext`) and I'll probably remove them at some point.

gautamsi: 👍

packages/keystone/lib/Keystone/index.js
```diff
@@ -228,20 +228,50 @@ module.exports = class Keystone {
...
+      skipAccessControl = orig.skipAccessControl,
+    }) => this.createContext({ schemaName, authentication, skipAccessControl });
+    context.executeGraphQL = ({ context = context, query, variables }) =>
+      this.executeGraphQL({ context, query, variables });
+    return context;
+  }
+
+  executeGraphQL({ context, query, variables }) {
+    if (!context) {
+      context = this.createContext();
```
lklots, line 253: I think you want  ```context = this.createContext({});```

timleslie: Ooh, good catch, I think you're right!

lklots: @timleslie starting a new graphql project and running ` npx keystone create-tables` results in 
```
TypeError: Cannot read property 'schemaName' of undefined
    at Keystone.createContext (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/lib/Keystone/index.js:231:19)
    at Keystone.executeGraphQL (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/lib/Keystone/index.js:255:22)
    at Object.module.exports [as onConnect] (/Users/levyklots/my-app/initial-data.js:10:22)
    at Keystone.connect (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/lib/Keystone/index.js:540:33)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async createTables (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/bin/commands/create-tables.js:11:3)
```

timleslie: Yep, that'll do it, let me ship a fix for this now. Thanks for the bug report 🙏

timleslie: @lklots I've published `@keystonejs/keystone@11.1.3` which fixes this issue. Thanks again for the bug report!

lklots: @timleslie there is another error if you run `npx keystone create-tables` in a brand new adminui project below. I'll start debugging and let you know but wanted you to be aware


```
✖  
Error: No executable schema named 'public' is available. Have you setup '@keystonejs/app-graphql'?
    at Keystone.executeGraphQL (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/lib/Keystone/index.js:259:13)
    at Object.module.exports [as onConnect] (/Users/levyklots/my-app/initial-data.js:10:22)
    at Keystone.connect (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/lib/Keystone/index.js:540:33)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async createTables (/Users/levyklots/my-app/node_modules/@keystonejs/keystone/bin/commands/create-tables.js:11:3)
```

timleslie: 😓 This is also definitely a bug, but it's a little more confusing. In particular, I'm wondering how the code that we had *before* the change to `executeGraphQL` ever worked! We shouldn't be trying to initialise data via server side graphQL operations in the `onConnect` phase, because at this point we haven't actually set up our graphQL schemas yet! That happens at a later phase of the system set up.

I think that perhaps that it *didn't* work in the past, but it was failing silently, rather than throwing an error, which is even worse!

Let me dig into this and try and work out what's going on and how to fix it.

lklots: @timleslie that's for sure what's happening since the errors are being gobbled up when running `yarn dev` since it is not reporting failures on `npx keystone create-tables`

```
yarn dev
yarn run v1.22.4
$ cross-env NODE_ENV=development DISABLE_LOGGING=true keystone dev
ℹ Command: keystone dev
✔ Validated project entry file ./index.js
✔ Keystone server listening on port 3000
No cookieSecret value was provided. Please generate a secure value and add it to your app. Until this is done, a random cookieSecret will be generated each time Keystone is started. This will cause sessions to be reset between restarts. See [https://www.keystonejs.com/keystonejs/keystone/#config] for details.
✔ Initialised Keystone instance
✔ Connected to database
⠋ Verifying tables

Database is empty. Initialise tables with: 'npx keystone create-tables'
See: 🔗 https://www.keystonejs.com/quick-start/#installing-keystone


```

timleslie: I've been able to reproduce the error you're getting, which is a good start.

timleslie: I think the bug is here:

```
  onConnect: process.env.CREATE_TABLES !== 'true' && initialiseData,
```

When we run `create-tables` we connect directly to the database, without ever bothering to setup graphQL servers (since all we're doing is executing some database commands).

As such, our start app signals to only run `initialiseData` when we're *not* running `create-tables`.

If you run:

```
yarn create-tables
```

or

```
npm run create-tables
```

then it will run the script:

```
    "create-tables": "CREATE_TABLES=true keystone create-tables"
```
from `package.json`, which sets up the env var correctly, and you should be able to successfully create your tables.

The *real* culprit here is the bad instructions on the quick-start guide page which say:

```
yarn keystone create-tables
# or
npx keystone create-tables
```

This will work fine for projects which don't have anything in their `onConnect` function, but happen to be bad advice in the case of the `starter` project, which requires that env-var.

I will look into resolving this, either by updating the docs on the quick-start page, or finding a better mechanism that plays nicely with the existing instructions.

lklots: @timleslie you're awesome thank you!

timleslie: Problem solved in https://github.com/keystonejs/keystone/pull/3196

PR #3173, opened 2020-06-17
Remove _buildQueryHelper and _executeOperation

timleslie:


gautamsi:
We should give time after deprecation notice.

PR #3174, opened 2020-06-17
Replace getGraphQlContext with createHTTPContext

timleslie:


gautamsi:
Is this aligned with removal of executeQuery? Should this be staged for later?

timleslie:
> Is this aligned with removal of executeQuery? Should this be staged for later?

Yes, that's why this PR is still in `[WIP]`.

packages/keystone/lib/Keystone/index.js
```diff
@@ -263,17 +263,16 @@ module.exports = class Keystone {
...
-  // It is also used for generating the `keystone.query` method
-  getGraphQlContext({ schemaName, req = {}, skipAccessControl = false } = {}) {
+  createHTTPContext({ schemaName, req }) {
+    // The GraphQL App uses this method to build up the context required for each incoming query.
     return {
       ...this.createContext({
         schemaName,
         authentication: { item: req.user, listKey: req.authedListKey },
-        skipAccessControl,
+        skipAccessControl: false,
```
gautamsi, line 272: is this going to break use when want to skip access control

MadeByMike: That's a default value so if you create a context with `skipAccessControl: true`, no prob.

packages/keystone/lib/Keystone/index.js
```diff
@@ -234,17 +234,16 @@ module.exports = class Keystone {
     return execute(schema, query, null, context, variables);
   }
 
-  // The GraphQL App uses this method to build up the context required for each
-  // incoming query.
-  // It is also used for generating the `keystone.query` method
-  getGraphQlContext({ schemaName, req = {}, skipAccessControl = false } = {}) {
+  createHTTPContext({ schemaName, req }) {
```
jordanoverbye, line 237: Do we need `req = {}` here?

timleslie: No, `req` is a required argument for this use case, which is specifically for HTTP requests.

PR #3178, opened 2020-06-17
Arch: redesigned Pagination component

Vultraz:
Old:
image

New:
image

I also did away with the Expand button. It's a real pain to expand the thing every time to go to a new page.


gautamsi:
I wish we could always show `1` for first page then `...` and then other page number and last page number. essentially you have first and last page number always visible.
`< 1 ...2 3 4 ... 11 >`

VinayaSathyanarayana:
I agree with view of @gautamsi - that we should show First ..... Range, Current, Range .... Last

PR #3183, opened 2020-06-17
Improve utils docs

Vultraz:
Have had this sitting around locally for a couple months. Wanted to add more, but I might as well add what I have.

### DIFF
--- a/packages/utils/src/index.js
+++ b/packages/utils/src/index.js
@@ -82,20 +82,40 @@ export const filterValues = (obj, predicate) =>
     {}
   );
 
-// [x, y, z] => { x[keyedBy]: mapFn(x), ... }
-// [{ name: 'a', animal: 'cat' },
-//  { name: 'b', animal: 'dog' },
-//  { name: 'c', animal: 'cat' },
-//  { name: 'd', animal: 'dog' }]
-// arraytoObject(obj, 'name', o => o.animal) =>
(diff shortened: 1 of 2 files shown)


packages/utils/src/index.js
```diff
@@ -201,18 +222,22 @@ export const captureSuspensePromises = executors => {
...
  * Compares two version strings or number arrays in the major.minor.patch format.
- * Returns true if comp if each element of comp is greater than than base.
+ * @param {Array<Number>|String} comp The version to compare.
+ * @param {Array<Number>|String} base The version against which to compare.
+ * @returns True if each element of comp is greater than or equal base.
  */
 export const versionGreaterOrEqualTo = (comp, base) => {
   const parseVersion = input => {
-    if (typeof input === 'object') {
+    if (input instanceof Array) {
```
Vultraz, line 240: Tiny random change. When I originally wrote this function I didn't know about `instanceof` and I had intended this check to be for arrays.

pahaz: It's better to use `Array.isArray()`. instanceof with Arrays produce some problems ...

Vultraz: Do Javascript’s quirks never end ☹️

pahaz: I can create PR. Because it's may produce problems for me.

PR #3185, opened 2020-06-17
Reexported gql from apollo server via app-graphql

Vultraz:
I pulled it from Apollo Server instead of `graphql-tag` since the former [exports it as of v2](https://www.apollographql.com/docs/apollo-server/migration-two-dot/#the-gql-tag) and it's encouraged to use it.

### DIFF
--- a/.changeset/gorgeous-toes-stare.md
+++ b/.changeset/gorgeous-toes-stare.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-graphql': minor
+---
+
+Reexported `gql` template literal from Apollo Server.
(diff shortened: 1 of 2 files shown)


Vultraz:
Which parent package did you mean?

timleslie:
> Which parent package did you mean?

`'apollo-server-express'`

As a dev, surely it's better for me to explicitly write 

```
const { gql } = require('apollo-server-express');
```

rather than 

```
const { gql } = require('@keystonejs/app-graphql');
```

Vultraz:
Hmmmm.... Yes, that would indeed work just as well.

Vultraz:
The only downside to importing from the `apollo-server-express` package is later down the road, it will be a major breaking change for end users once Apollo Server v3 is out, since the package changes to `@apollo/server`.

timleslie:
I'm going to close this for now.

PR #3199, opened 2020-06-22
Support GraphQL AST input in executeGraphQL

Vultraz:
As says. Still supports string input.


timleslie:
> Make sense, @timleslie , thoughts?

Yep, all good 👍

Issue #3208, opened 2020-06-28
Deleting value in Location field leads to 'Unable to render view'


petrkrizek:
# Bug report

## Describe the bug

See the title.
Error in console: 
Uncaught TypeError: Cannot destructure property 'googlePlaceID' of 'value' as it is null.

## To Reproduce

1. Create a Location field
2. Select some address
3. Delete address
4. Unable to render view

## Expected behaviour
To see map again

## The problem
In the file 'packages/fields/src/types/Location/views/Field.js' on line 11 there is destructuring of 'value' that is null. Assigning default value does not help, since null is not undefined.
[https://github.com/keystonejs/keystone/blob/master/packages/fields/src/types/Location/views/Field.js#L11](https://github.com/keystonejs/keystone/blob/master/packages/fields/src/types/Location/views/Field.js#L11)

## The fix
Move the destructuring out from fn argument and add empty object 'fallback' like this:

`const {
    googlePlaceID,
    formattedAddress,
    lat,
    lng
  } = (value || {})`

PR #3209, opened 2020-06-28
fix: fix Location field 'unable to render' after deleting the field value

petrkrizek:
Closes: #3208

### DIFF
--- a/packages/fields/src/types/Location/views/Field.js
+++ b/packages/fields/src/types/Location/views/Field.js
@@ -20,7 +20,7 @@ const LocationField = ({
     formattedAddress,
     lat,
     lng
-  } = (value || {})
+  } = { ...value }
   const htmlID = `ks-input-${field.path}`;
   const autocompleteService = new google.maps.places.AutocompleteService();
   const geocoder = new google.maps.Geocoder();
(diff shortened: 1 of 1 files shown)


packages/fields/src/types/Location/views/Field.js
```diff
@@ -8,13 +8,19 @@ import { Map, Marker, GoogleApiWrapper } from 'google-maps-react';
...
   google,
   renderContext,
   isDisabled,
 }) => {
+  const {
+    googlePlaceID,
+    formattedAddress,
+    lat,
+    lng
+  } = (value || {})
```
MadeByMike, line 23:
```suggestion
  } = { ...value }
```

packages/fields/src/types/Location/views/Field.js
```diff
@@ -8,13 +8,19 @@ import { Map, Marker, GoogleApiWrapper } from 'google-maps-react';
 
 const LocationField = ({
   field,
-  value: { googlePlaceID, formattedAddress, lat, lng } = {},
+  value = {},
```
MadeByMike, line 11:
```suggestion
  value,
```

Default no longer needed.

packages/fields/src/types/Location/views/Field.js
MadeByMike: Spreading null or undefined will result in an empty object

MadeByMike:
If we can add a changeset this is good to go

MadeByMike:
Updated in #3242

PR #3222, opened 2020-07-02
Remove support for multiple database adapters in a single Keystone system

timleslie:
One of Keystone's perhaps lesser used features is the ability to connect to more than one database at a time. This feature was added very early in development in anticipation that this might be a feature that we needed. In practice, none of the systems we have built have ever come close to requiring this feature. I'm not aware of any systems using it in the wild.

I would like to remove support for this feature, as it adds complexity to the code, surface area to the API, and potentially confusion to developers trying to get a solid mental model of a Keystone system.

I'm putting up this PR as an RFC to see if there is any vocal support for keeping this feature. Ideally I'd love to hear a practical example of this being used in a production context.

### DIFF
--- a/.changeset/smart-deers-trade.md
+++ b/.changeset/smart-deers-trade.md
@@ -1,0 +1,13 @@
+---
+'@keystonejs/keystone': major
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
+'@keystonejs/demo-project-meetup': patch
+'@keystonejs/cypress-project-access-control': patch
+'@keystonejs/cypress-project-basic': patch
+'@keystonejs/cypress-project-client-validation': patch
+'@keystonejs/cypress-project-login': patch
+'@keystonejs/cypress-project-social-login': patch
+---
+
+Removed support for multiple database adapters in a single `Keystone` system. The `adapters` and `defaultAdapter` config options were removed from the `Keystone()` constructor.
--- a/docs/guides/prisma.md
+++ b/docs/guides/prisma.md
@@ -96,7 +96,7 @@ keystone.extendGraphQLSchema({
     {
       schema: 'allComplete: Boolean',
       resolver: async () => {
-        const { prisma } = keystone.adapters.PrismaAdapter;
+        const { prisma } = keystone.adapter;
         const unfinished = await prisma.todo.count({ where: { isComplete: { equals: false } } });
         return unfinished === 0;
       },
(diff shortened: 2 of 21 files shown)


jesstelford:
Agreed on the concept 👍

molomby:
> In practice, none of the systems we have built have ever come close to requiring this feature.

I suspect this is due to the lack of db adapters. For example...

* If we had a Stripe DB adapter we would have used this functionality
* With A JSON DB adapter you could use static data along side your DB content or use it as the basis of an import process
* Imagine you have a separate process that runs each hour and dumps a set of CSVs into a directory. Your KS could define a list for this data, now it's available via GraphQL to your front end.

None of this is _really_ an argument to keep the functionality in though since none of these adapters are likely to be added anytime soon. Even if they were, _it might not make sense to solve these with a single "keystone" instance_.

@JedWatson and I were talking along these lines a few weeks ago. Trying to figure out where this "multiple" boundary should sit. Ie. if you have multiple data sources you want to work with within a project should that be multiple data adapters? multiple keystone objects? multiple apps? etc.

There's a related conversation around the requirement for with multiple graphQL schemas and/or endpoints. Are these just multiple "keystones" as well?

These are pretty far reaching/important decisions. I think there's a reasonable argument for keeping each "keystone" pretty simple then scaling those simple objects our for more complex solutions. It would let us simplify some things I'm sure.

It's probably worth evaluating this from the perspective of specific examples. Ie. picking some of the above examples (or probably better ones) and sketching out how it would work in practice. I'm guessing that's the only way to get a real handle on the implications of an approach.

molomby:
Some question for you @timleslie..

* Is it practical to image we'll ever support relationships that span database adapters?
* whjat

If the answer is "no", what's the benefit in having

VinayaSathyanarayana:
Would vote to keep multiple adapters. Let us not drop this feature.

timleslie:
@VinayaSathyanarayana 

> Would vote to keep multiple adapters. Let us not drop this feature.

Is this a feature you are currently using? Could you give some more details on why you think we should keep this feature?

VinayaSathyanarayana:
We have use cases where there are multiple data sources from which the data is fetched.

timleslie:
> We have use cases where there are multiple data sources from which the data is fetched.

That's really interesting! Could you give a code snippet of how you set up your `Keystone` object? Do you call `createList` with different adapters, or do you only use the other data sources for custom queries/mutations? Is there a particular reason for choosing to store the data across multiple sources, rather than consolidating in a single location.

This info is really important for our decision making, as if we choose to remove the multiple-adapter feature then we want to make sure there are appropriate alternatives in places for developers like yourself who might be using the feature.

gautamsi:
only place I see this could be used (not that I used it) is when you have in memory adapter or redis adapter. I see use with auth item caching (for authed item relationships query in access control) but that caching can be used differently and not necessarily with multiple adapters.

molomby:
Without details it's hard to be sure but I strongly suspect the use cases people have alluded to here don't require the Keystone concept of an adapter. Just having "multiple data sources from which the data is fetched" doesn't require multiple adapters. Similarly for caching authed items data.

I think this functionality can definitely go.

molomby (approved):
Yep, nice.

PR #3223, opened 2020-07-02
Remove default database connection string

timleslie:
If a dev does not supply a connection string for their database then Keystone will go looking at certain environment variables, and failing that will just make something up based on the value of `name` passed to the `Keystone()` constructor.

Defaulting to a connection based on the name of the project is not particularly elegant, and does not set up devs for success. It is important that they explicitly state where the database is and have control over this. There's no good reason keystone should be guessing.

The `create-keystone-app` package now ensures that the dev has set up this connection string, so very few users should be relying on this "feature", and those who are can very easily update their project to have an explicit connection string.

### DIFF
--- a/packages/adapter-knex/README.md
+++ b/packages/adapter-knex/README.md
@@ -47,17 +47,9 @@ _**Default:**_
 ```javascript
 {
   client: 'postgres',
-  connection: '<DEFAULT_CONNECTION_URL>'
+  connection: 'process.env.CONNECT_TO || process.env.DATABASE_URL || process.env.KNEX_URI'
 }
 ```
-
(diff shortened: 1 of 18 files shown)


packages/adapter-knex/README.md
```diff
@@ -47,18 +47,10 @@ _**Default:**_
 ```javascript
 {
   client: 'postgres',
-  connection: '<DEFAULT_CONNECTION_URL>'
+  connection: 'process.env.CONNECT_TO || process.env.DATABASE_URL || process.env.KNEX_URI'
```
jesstelford, line 50:
```suggestion
  connection: process.env.CONNECT_TO || process.env.DATABASE_URL || process.env.KNEX_URI
```

Issue #3004, opened 2020-05-20
Fields with Underscores Cause Runtime Error when Using Postgres/Knex


hew:
# Bug report

## Describe the bug

If you create a List with a field that has an underscore using Postgres/Knex, Keystone errors out. 

## To Reproduce

1. Start a new keystone project
2. Choose postgres/Knex
3. Create a list with a field that has an underscore (ie: `user_id`)
4. Try to view that list

```
 "@keystonejs/adapter-knex": "^10.0.0",
 "@keystonejs/app-admin-ui": "^6.0.0",
```

## Expected behaviour

Keystone should not crash when viewing a list that has an underscore.

## Screenshots

![image](https://user-images.githubusercontent.com/3103241/82475068-c6cc7180-9a80-11ea-9bf6-e9316559dc20.png)

## System information

- OS: macOS
- All browsers

## Additional context

Fields with underscores seem to be missing at the graphQL filtering level.

hew:
Reproduced here. Just needs a database with a `player` table and a field with an underscore.

https://github.com/hew/keystone-postgres-error-reproduced.

hew:
Issue is caused by [this line].(https://github.com/keystonejs/keystone/blob/63e00d805f3653863002befdaeda74c711f36f6b/packages/adapter-knex/lib/adapter-knex.js#L659) and [this line](https://github.com/keystonejs/keystone/blob/63e00d805f3653863002befdaeda74c711f36f6b/packages/adapter-knex/lib/adapter-knex.js#L652)

I'll see about pushing a fix tomorrow at work.

MadeByMike:
Oh nice pick up! That should be fixed.

BikramjeetSingh:
Is this already fixed? If not, can I pick this up?

hew:
@BikramjeetSingh You can pick it up. Thanks.

BikramjeetSingh:
Okay, I am looking into this issue.

BikramjeetSingh:
I have created a PR for this bug [here](https://github.com/keystonejs/keystone/pull/3225). Could you please take a look? @hew @MadeByMike

BikramjeetSingh:
My PR has been merged, so this issue can be marked as resolved and closed.

hew:
Nice work. Closing this issue. Thanks @BikramjeetSingh!

PR #3225, opened 2020-07-02
Fix bug where fields that contain underscores cause runtime error in Knex Adapter

BikramjeetSingh:
This is a PR to fix the bug reported in [issue #3004](https://github.com/keystonejs/keystone/issues/3004): if a list is created with a field whose name contains one or more underscores (_), a runtime error is thrown by the Knex adapter.


.changeset/dirty-points-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/adapter-knex': minor
```
timleslie, line 2:
```suggestion
'@keystonejs/adapter-knex': patch
```

PR #3227, opened 2020-07-03
Moved `name` config option from Keystone to Admin UI

Vultraz:
As says, since it's no longer used by the KS core and only passed to the Admin UI.


timleslie:
Haha yep, this is exactly where I was going with the changes down in the adapter code.

I'm just pondering the timing of this change, and whether we want to bundle it with this big batch of `major` changes, or wait for one more cycle. Let me have a think about it. I'm 💯 in favour of the change in principle 👍

Vultraz:
I think it makes sense to include with the others. It logically goes with the change to db names.

packages/app-admin-ui/index.js
```diff
@@ -11,6 +11,7 @@ const getWebpackConfig = require('./server/getWebpackConfig');
 
 class AdminUIApp {
   constructor({
+    name,
```
timleslie, line 14: Can we default this to: `name = 'Keystone'`?

Vultraz:
@timleslie done

timleslie:
> OK, I'm board with this.

Oops, *on board*, definitely not bored! 😅

PR #3234, opened 2020-07-07
Factor out list utils

timleslie:
Internal refactor, no functional changes.


packages/keystone/lib/ListTypes/utils.js
```diff
@@ -0,0 +1,95 @@
+const { upcase } = require('@keystonejs/utils');
+const { logger } = require('@keystonejs/logger');
+
+const keystoneLogger = logger('keystone');
+
+const preventInvalidUnderscorePrefix = str => str.replace(/^__/, '_');
```
MadeByMike, line 6: This is only used once and very simple - we could just inline it - but I won't hold this up for that

PR #3235, opened 2020-07-07
Factor out hooks handling code

timleslie:
Internal refactor, no functional changes.

### DIFF
--- a/packages/keystone/lib/ListTypes/hooks.js
+++ b/packages/keystone/lib/ListTypes/hooks.js
@@ -1,0 +1,166 @@
+const { omitBy, arrayToObject } = require('@keystonejs/utils');
+const { mapToFields } = require('./utils');
+const { ValidationFailureError } = require('./graphqlErrors');
+
+class HookManager {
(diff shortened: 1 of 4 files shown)


packages/keystone/lib/ListTypes/hooks.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  async _validateHook({ args, fields, operation, hookName }) {
+    const { originalInput } = args;
+    const fieldValidationErrors = [];
+    // FIXME: Can we do this in a way where we simply return validation errors instead?
+    args.addFieldValidationError = (msg, _data = {}, internalData = {}) =>
+      fieldValidationErrors.push({ msg, data: _data, internalData });
+    await mapToFields(fields, field => field[hookName](args));
+    await mapToFields(
```
singhArmani, line 112: Do we need two `await` calls here?

timleslie: Yep, we want to ensure that the hooks defined by the field type itself are executed before any user defined hooks for the same field.

PR #3236, opened 2020-07-07
[WIP] Apply cache hints to mutation resolvers.

timleslie:
This change passes the `info` object down into `_itemsQuery` in mutations to allow `cacheHints` to be utilised in these cases.


timleslie:
After thinking about this some more I'm no longer convinced that this is a good change. Thanks Jess for the call out. Will close this PR without merging.

PR #3240, opened 2020-07-08
Add new email field type

jordanoverbye:
This PR adds a new field type `Email`. This is essentially the same as the `Text` field type, but with some built-in email validation.

### Screenshots

Show a warning when creating an item with an invalid email address. The tooltip is native to the browser and happens because we've added `type="email` to the input.

Screen Shot 2020-07-08 at 11 13 23 am

Show a warning when trying to update an item with an invalid email address.

Screen Shot 2020-07-08 at 11 09 10 am

### GraphQL errors

Using this list

```js
keystone.createList('User', {
  fields: {
    email: { type: Email, isRequired: true },
  },
});
```

```
mutation {
  createUser(data: { email: "__invalid__email" }) {
    id
    email
  }
}


// Returns

{
  "errors": [
    {
      "message": "You attempted to perform an invalid mutation",
      "name": "ValidationFailureError",
      "time_thrown": "2020-07-09T01:21:24.941Z",
      "data": {
        "messages": [
          "Invalid email"
        ],
        "errors": [
          {
            "value": "__invalid__email"
          }
        ],
        "listKey": "User",
        "operation": "create"
      },
      "path": [
        "createUser"
      ],
      "uid": "ckce3v2ll0001creff2e019iu"
    }
  ],
  "data": {
    "createUser": null
  }
}
```

### DIFF
--- a/.changeset/calm-pans-leave.md
+++ b/.changeset/calm-pans-leave.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Add `React.useCallback` and `React.useMem` inside of Text field view
--- a/.changeset/calm-pans-leave.md
+++ b/.changeset/calm-pans-leave.md
@@ -2,4 +2,4 @@
 '@keystonejs/fields': patch
 ---
 
-Add `React.useCallback` and `React.useMem` inside of Text field view
+Add `React.useCallback` and `React.useMemo` inside of Text field view
--- a/packages/fields/src/types/Email/Implementation.js
+++ b/packages/fields/src/types/Email/Implementation.js
@@ -15,6 +15,7 @@ export class Email extends Text {
   }
 
   isValidEmail(email) {
-    return /\S+@\S+\.\S+/.test(email);
+    // https://www.w3.org/TR/2012/WD-html-markup-20120329/input.email.html#input.email.attrs.value.single
+    return /^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/.test(email);
   }
 }
--- a/packages/fields/src/types/Email/Implementation.js
+++ b/packages/fields/src/types/Email/Implementation.js
@@ -1,0 +1,20 @@
+import { Text } from '../Text/Implementation';
+
+export class Email extends Text {
+  validateInput({ resolvedData, addFieldValidationError }) {
+    const fieldValue = resolvedData[this.path];
+
+    if (!this.isRequired && !fieldValue) {
+      // If the field isn't required and we don't have a value we can skip any validation checks
+      return;
+    }
+
+    if (!this.isValidEmail(fieldValue)) {
+      addFieldValidationError(`Invalid email`, { value: fieldValue });
+    }
+  }
+
+  isValidEmail(email) {
+    return /\S+@\S+\.\S+/.test(email);
+  }
+}
--- a/packages/fields/src/types/Text/views/Field.js
+++ b/packages/fields/src/types/Text/views/Field.js
@@ -1,9 +1,12 @@
 /** @jsx jsx */
 
 import { jsx } from '@emotion/core';
+import { useCallback, useMemo } from 'react';
 
 import { FieldContainer, FieldLabel, FieldDescription, FieldInput } from '@arch-ui/fields';
 import { Input } from '@arch-ui/input';
+
+const isAccessDeniedError = error => error instanceof Error && error.name === 'AccessDeniedError';
 
 const TextField = ({
   onChange,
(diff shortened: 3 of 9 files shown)


packages/fields/src/types/Text/views/Field.js
```diff
@@ -1,21 +1,33 @@
...
+import { useCallback, useMemo } from 'react';
 
 import { FieldContainer, FieldLabel, FieldDescription, FieldInput } from '@arch-ui/fields';
 import { Input } from '@arch-ui/input';
 
-const TextField = ({ onChange, autoFocus, field, errors, value = '', isDisabled }) => {
-  const handleChange = event => {
-    onChange(event.target.value);
-  };
+const isAccessDeniedError = error => error instanceof Error && error.name === 'AccessDeniedError';
```
jordanoverbye, line 9: The changes here are mainly just moving things around / using `useCallback` and `useMemo`.

packages/fields/src/types/Email/Implementation.js
```diff
@@ -0,0 +1,20 @@
...
+      return;
+    }
+
+    if (!this.isValidEmail(fieldValue)) {
+      addFieldValidationError(`Invalid email`, { value: fieldValue });
+    }
+  }
+
+  isValidEmail(email) {
+    return /\S+@\S+\.\S+/.test(email);
```
jesstelford, line 18: I'm fairly certain it's valid for an email to _not_ have a period: `hi@io` if the TLD supports it.

jordanoverbye: Good catch @jesstelford! I didn't know that was a thing. 

I've just updated that regex to be the same as what browsers use for input `type=”email”`: `https://www.w3.org/TR/2012/WD-html-markup-20120329/input.email.html#input.email.attrs.value.single`

.changeset/calm-pans-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Add `React.useCallback` and `React.useMem` inside of Text field view
```
Vultraz, line 5: Typo.

gautamsi:
My thought would be to add validation features to Text field with certain in built validation features.

Vultraz:
I was also thinking that, but how would the admin UI views be handled?

gautamsi:
@Vultraz this PR is mostly same as Text Field just adding a validation of email in controller and setting type on input field. 
practically it can be a flag in Text field as `validate=email` (or any suitable thing) which triggers email validation inside the Text controller instead of brand new field type

packages/fields/src/types/Email/Implementation.js
MadeByMike: This is also valid: `very.”(),:;<>[]”.VERY.”very@\\ "very”.unusual@strange.example.com`

MadeByMike:
I'm of 2 minds about whether we need this or just a better way to specify adminUI interfaces.

packages/fields/src/types/Email/Implementation.js
molomby: Email addresses are super weird and very hard to validate. Pretty sure you can have almost anything in the user portion since it's up to the receiving server to interpret it. I believe you also don't need a dot in the host portion as you can refer to hosts on a local network, eg. `me@localhost` is valid (if I'm the kind of person to run a personal mail server on my own machine where I'm also hosting Keystone I guess and, for some reason, emailing myself.. 🤷🏻‍♂️).

Glad we're delegating all this to the W3C suggested regex!

JedWatson:
@jordanoverbye this looks good to me, but a quick question about the graphql errors you included as an example above - it demonstrates that the missing `name` field is required where I think you meant to demo the invalid value in the email field not passing validation?

jordanoverbye:
> @jordanoverbye this looks good to me, but a quick question about the graphql errors you included as an example above - it demonstrates that the missing `name` field is required where I think you meant to demo the invalid value in the email field not passing validation?

@JedWatson in the demo I setup for this, I created a list with `name` and `email` field. I probably should have excluded the `name` field which would have simplified those errors. I've just updated the PR description which does that 👍

VinayaSathyanarayana:
I support the need for Email Field.  

Some suggestions:- 
We should support (directly or through hooks) ability to validate email domain names and handle common typing errors

Please look at
* [Mailcheck](https://github.com/mailcheck/mailcheck) 
When your user types in "user@gmil.con", Mailcheck will suggest "user@gmail.com".

* [Email Domain validator](https://www.npmjs.com/package/email-domain-validator)

packages/fields/src/types/Text/views/Field.js
```diff
@@ -26,7 +38,7 @@ const TextField = ({ onChange, autoFocus, field, errors, value = '', isDisabled
           autoComplete="off"
```
irwin-r, line 38: Probably a dumb call-out on my behalf;  but should we not enable `autoComplete` for email addresses?

MadeByMike:
Sadly @jordanoverbye I think we should close this PR :( essentially this is a text field. We need to make custom interfaces easier but should not add to core field types every time there is a variation of text.

If @molomby suggestions were to be considered and the issues resolved then this would warrant a new field type.

molomby:
I'm definitely going to come back to this; the requirement comes up so often, it's worth solving.

PR #3241, opened 2020-07-08
Fix existingItem in updateManyMutation hooks

### LINKED ISSUE #3239: Incorrect exitistingItem value passed to hooks in updateManyMutation
# Bug report
Seems like the updateManyMutation which takes the input in the form of `[{id, data}, {id, data}]` randomizes the sequence of update item with corresponding item id, I got a very upset customer where the prices were mixing up and some actual sold items caused loss :( (I know there is no warranty, just expressing my frustration)

## Describe the bug

When I try to update multiple records with `updateManyMutation`, It is not providing `existingItem` corresponding to item trying to update. You do not face any issue when there are no hooks or the hooks do not update `resolveData`, but when you change any field in `resolveInput` or compare the value for `updatedItem.id` and `existingItem.id` in `afterChange` hook, you will see difference data.

## To Reproduce

clone blog demo

1. Add additional field to list, say "postId" in Post list.
2. add `resolveInput` hook to copy value of `id` field to `postId` field
3. add `afterChange` hook and log value of `existingItem.id` and ``updatedItem.id`
4. have at least 4-5 Posts
5. try to update multiple Post (min 3) and set field like `status` to something else currently set ( you can use "Update" button on items selection or use GraphQL method for `UpdateManyPosts` mutation)
6. notice the difference in `id` of the post and id value in `postId` field
7. also notice the console log in `afterChange` hook

try to update few time if it does not happen at once or select 6-7 posts in update.

update: see diff, what I had to do to repro https://github.com/keystonejs/keystone/compare/master...gautamsi:error-in-many-mutation?expand=1

## Expected behaviour

There should not be any issues like that, you must always have `existingItem.id === updatedItem.id`

## Screenshots

N/A

## System information

- OS: Windows/Mac
- Browser (if applies) [e.g. chrome, safari]

## Additional context

I traced this issue to following few lines

where call to `zipObj` util function is unreliable 
https://github.com/keystonejs/keystone/blob/5ad84ccd8d008188e293629e90a4d7e7fde55333/packages/keystone/lib/ListTypes/list.js#L837

I will also create a PR

timleslie:
Fixes #3239

### DIFF
--- a/packages/keystone/lib/ListTypes/list.js
+++ b/packages/keystone/lib/ListTypes/list.js
@@ -10,6 +10,7 @@ const {
   flatten,
   zipObj,
   createLazyDeferred,
+  arrayToObject,
 } = require('@keystonejs/utils');
 const { parseListAccess } = require('@keystonejs/access-control');
 const { logger } = require('@keystonejs/logger');
@@ -833,9 +834,10 @@ module.exports = class List {
     const access = await this.checkListAccess(context, data, operation, extraData);
 
     const existingItems = await this.getAccessControlledItems(ids, access);
+    const existingItemsById = arrayToObject(existingItems, 'id');
 
     const itemsToUpdate = zipObj({
-      existingItem: existingItems,
+      existingItem: ids.map(id => existingItemsById[id]),
       id: ids, // itemId is taken from here in checkFieldAccess
       data: data.map(d => d.data),
     });
--- a/.changeset/famous-pumpkins-raise.md
+++ b/.changeset/famous-pumpkins-raise.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Fixed a bug where `existingItem` had the wrong value in hooks during an `updateManyMutation`.


packages/keystone/lib/ListTypes/list.js
```diff
@@ -833,9 +834,10 @@ module.exports = class List {
     const access = await this.checkListAccess(context, data, operation, extraData);
 
     const existingItems = await this.getAccessControlledItems(ids, access);
+    const existingItemsById = arrayToObject(existingItems, 'id');
 
     const itemsToUpdate = zipObj({
-      existingItem: existingItems,
+      existingItem: ids.map(id => existingItemsById[id]),
```
timleslie, line 840: There's nothing to guarantee the order of items in `existingItems`, so once we have them we need to map them back into the correct order before handing them off to the hooks.

jesstelford:
Need a test showing the failure & therefore the fix?

timleslie:
> Need a test showing the failure & therefore the fix?

In theory yes, in practice it's hard to trigger the failure condition, because it relies on arbitrary behaviour of the adapter which we don't have explicit control over. I've validated the fix by inline logging during runs of the queries which exercise this code path.

MadeByMike:
@timleslie approved based on code review, then noticed CI failing?

packages/keystone/lib/ListTypes/list.js
Vultraz: Perhaps worth adding this as a code comment?

gautamsi: one doubt, does `getAccessControlledItems` filters out items we have no access? then `existingItem` will be undefined, and there will be undesired update to item if someone is not checking existingItem be undefined. right?

PR #3242, opened 2020-07-09
Fix location field bug

MadeByMike:
Fix a bug with the location field where it is unable to render in admin ui after deleting the value.

### DIFF
--- a/.changeset/quiet-bees-rescue.md
+++ b/.changeset/quiet-bees-rescue.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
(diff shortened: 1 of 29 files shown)


.changeset/quiet-bees-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Fix Location field unable to render in admin ui after deleting the value.
```
MadeByMike, line 5:
```suggestion
Fix a bug with the location field where it is unable to render in admin ui after deleting the value.
```

PR #3243, opened 2020-07-09
Refactor createItems

singhArmani:
[WIP]: Refactoring `createItem` API for CRUD operations
    
This branch tracks the work in progress for refactoring how we use `keystone.createItems` API. The existing API, allows the end-user to seed the data while setting up the keystone app. Doing so, the API bypasses the GraphQL layer and insert the data directly into the database.
    
This behaviour deviates from how we are using the executeGraphQl method, which relies on the existing API we get from `schema => ({api, AdminUI})`.
    
So, this branch is an effort to provide an API that follows the single API standard which is less ambiguous and provide a good developer experience.


test-projects/basic/server.js
```diff
@@ -53,21 +53,31 @@ function dropAllDatabases(adapters) {
...
-    ['User', 'PostCategory', 'ReadOnlyList'].map(listName =>
-      createItems({
-        keystone,
-        listName,
-        items: initialData[listName],
-        returnFields: listName === 'User' ? 'id, email' : 'id, name',
-      })
-    )
-  );
+  const { createUsers: users } = await createItems({
```
singhArmani, line 61: `createItems` uses `runChunkedMutation` which returns an Array of resolved promises. So, the line should be 
```js
const [{createUsers: users}] = await createItems(...);
```

MadeByMike:
Cypress is passing for me locally :(

demo-projects/blog/index.js
```diff
@@ -18,7 +19,11 @@ const keystone = new Keystone({
     const users = await keystone.lists.User.adapter.findAll();
     if (!users.length) {
       const initialData = require('./initialData');
-      await keystone.createItems(initialData);
+      await createItems({
+        keystone,
+        listName: 'User',
+        items: initialData.User.map(x => ({ data: x })),
```
timleslie, line 25: Let's update the data in the `initialData` file to have the `{ data: ... }` key so that we can keep this line of code a bit cleaner.

packages/test-utils/lib/test-utils.js
```diff
@@ -15,11 +15,14 @@ async function setupServer({
   createLists = () => {},
   keystoneOptions,
   graphqlOptions = {},
+  dbName,
```
timleslie, line 18: Do we still need this `dbName` support? If not, let's revert the relevant changes.

test-projects/access-control/index.js
```diff
@@ -144,7 +158,19 @@ listAccessVariations.forEach(createListWithDeclarativeAccess);
 module.exports = {
   keystone,
   apps: [
-    new GraphQLApp(),
-    new AdminUIApp({ name: projectName, adminPath: '/admin', authStrategy }),
+    new SchemaRouterApp({
+      routerFn: () => 'testing',
+      apps: {
+        seeding: createGraphQLApp('seeding'),
+        testing: createGraphQLApp('testing'),
```
timleslie, line 165:
```suggestion
        seeding: new GraphQLApp({ schemaName: 'seeding' }),
        testing: new GraphQLApp({ schemaName: 'testing' }),
```
No need for the extra function definition here.

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,179 @@
...
+
+    return accum;
+  }, []);
+};
+
+const runQuery = async ({
+  query,
+  variables,
+  keystone,
+  schemaName = 'public',
```
MadeByMike, line 21:
```suggestion
  schemaName = 'internal',
```

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+
+  if (verbose) {
+    console.log(`Finished paginated query`);
+    console.log(`${allItems.length} items found`);
+  }
+
+  return allItems;
+};
+
+const runChunkedMutation = async ({ gqlName, pageSize = 500, items, ...rest }) => {
```
MadeByMike, line 75: Let's name this paginate query and move the contents of `chunkedArray` inline here as it's the only place it's called

MadeByMike: Alternatively we could just call this `runMutation`

packages/test-utils/lib/test-utils.js
```diff
@@ -18,7 +18,7 @@ async function setupServer({
   const Adapter = { mongoose: MongooseAdapter, knex: KnexAdapter }[adapterName];
 
   const argGenerator = {
-    mongoose: getMongoMemoryServerConfig,
+    mongoose: async () => await getMongoMemoryServerConfig(),
```
MadeByMike, line 21: If we decided we don't need the DBName we can run this without async?

packages/orm/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@keystonejs/orm",
```
MadeByMike, line 2: Let's decide on a name for this package

packages/orm/lib/orm.js
MadeByMike: Hah just realised we have `runPaginatedQuery` so `runPaginatedMutation` makes sense

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+      accum[chunkIndex] = [];
+    }
+
+    accum[chunkIndex].push(item);
+
+    return accum;
+  }, []);
+};
+
+const runQuery = async ({
```
MadeByMike, line 17: I feel like this should be an internal method (not exported) since it is essentially just a proxy for `executeGraphQL` if someone needs to call this directly they probably don't need to be using this package

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+
+    allItems.push(...latestResult);
+
+    skip += pageSize;
+  } while (latestResult.length);
+
+  if (verbose) {
+    console.log(`Finished paginated query`);
+    console.log(`${allItems.length} items found`);
+  }
```
MadeByMike, line 76:
```suggestion
```

packages/orm/lib/orm.js
timleslie: This is useful if you want to run custom queries or mutations, so for consistency should remain part of the public API.

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+  ...rest
+}) => {
+  let skip = 0;
+  let latestResult;
+  const allItems = [];
+
+  if (verbose) {
+    console.log(`Started paginated query`);
+    console.log({ pageSize });
+  }
```
MadeByMike, line 53:
```suggestion
```

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+
+  do {
+    if (verbose) console.log({ skip });
```
MadeByMike, line 56:
```suggestion
```

packages/test-utils/lib/test-utils.js
timleslie: yep, this change can be reverted.

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+  let skip = 0;
+  let latestResult;
+  const allItems = [];
+
+  if (verbose) {
+    console.log(`Started paginated query`);
+    console.log({ pageSize });
+  }
+
+  do {
```
MadeByMike, line 55: Is a do...while the best way to do this? It's probably fine but have you considered GOTO? :trollface:

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+    variables,
+  });
+  if (errors) throw errors[0];
+  return data;
+};
+
+const runPaginatedQuery = async ({
+  pageSize = 1,
+  variables = {},
+  verbose = false,
```
MadeByMike, line 43:
```suggestion
```

packages/orm/lib/orm.js
```diff
@@ -0,0 +1,235 @@
...
+  return runQuery({ keystone, query, variables: { item }, schemaName, extraContext });
+};
+
+const createItems = ({
+  keystone,
+  listName,
+  items,
+  returnFields = `id`,
+  schemaName,
+  extraContext,
```
MadeByMike, line 129: Can we get rid of extraContext now?

api-tests/access-control/authed.test.js
```diff
@@ -43,7 +45,17 @@ multiAdapterRunners().map(({ before, after, adapterName }) =>
           }),
         {}
       );
-      items = await keystone.createItems(initialData);
+
+      for (const [listName, _items] of Object.entries(initialData)) {
+        const newItems = await createItems({
+          keystone,
+          listName,
+          items: _items.map(x => ({ data: x })),
```
MadeByMike, line 53: This sucks! `createItems` should completely abstract this `data: {}` BS - this confuses so many people.

packages/test-utils/lib/test-utils.js
MadeByMike, line 21:
```suggestion
    mongoose: getMongoMemoryServerConfig,
```

api-tests/access-control/authed.test.js
timleslie: It sucks, but the right solution is to fix out GraphQL. Let's fix the root of the problem, rather than giving people two subtley different APIs to deal with.

MadeByMike: Cross posting my points from slack:

In this context it kinda is a different API already. Items is an array and why would people think this should be in the shape of `[{data:{ }]`

Yes, we should fix the queries, but create items is intended to hide the query syntax no?

packages/server-side-graphql-client/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@keystonejs/server-side-graphql-client",
+  "description": "An ORM library for Keystone systems.",
```
timleslie, line 3: Need to update this line.

MadeByMike, line 3:
```suggestion
  "description": "A library for running server-side graphQL queries and mutations against in Keystone.",
```

packages/server-side-graphql-client/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "engines": {
+    "node": ">=10.0.0"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.8.4",
+    "p-is-promise": "^3.0.0",
+    "p-lazy": "^3.0.0",
+    "p-reflect": "^2.1.0",
+    "semver": "^5.7.1"
+  },
```
timleslie, line 18: I don't think any of these dependencies are actually used.

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,371 @@
...
+  const user = await createItem({
+    keystone,
+    listName: 'User',
+    item: userInput,
+    returnFields: `name, email`,
+  });
+  console.log(user); // { name: 'keystone user', email: 'keystone@test.com'}
+};
+
+addUser({ name: 'keystone user', email: 'keystone@test.com' });
```
timleslie, line 88: Can we simplify this example (and the ones below) be removing the intermediate function `userInput`? Ideally we'd have the minimal example demonstrating the behaviour of the functions.

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,371 @@
...
+[Shared Config Options](#shared-config-options) apply to this function.
+
+| Properties | Type                             | Default    | Description                                                                                    |
+| ---------- | -------------------------------- | ---------- | ---------------------------------------------------------------------------------------------- |
+| `items`    | GraphQL `[listName]sCreateInput` | (required) | The array of objects to be inserted.                                                           |
+| `pageSize` | `Number`                         | 500        | The create mutation batch size. This is useful when you have large set of data to be inserted. |
+
+### getItem
+
+[Shared Config Options](#shared-config-options) apply to this function.
```
timleslie, line 145:
```suggestion
```

packages/server-side-graphql-client/lib/server-side-graphql-client.js
```diff
@@ -0,0 +1,269 @@
...
+      runQuery({ query, variables: { items: chunk }, keystone, schemaName, extraContext })
+    )
+  );
+
+  /*
+   * The result is of the format: [{createUsers: [{id: '123', name: 'aman'}]}, {createUsers: [{id: '456', name: 'mike'}]}].
+   * We need to combine all objects into one array keyed by the `createUsers`, such that, the output is: [{id: '123', name: 'aman'}, {id: '456', name: 'Mike'}]
+   */
+
+  return result.map(item => item[gqlName]).flat(1);
```
timleslie, line 54: `.flat()` isn't supported in Node v10, which is our supported version.

packages/server-side-graphql-client/tests/index.test.js
```diff
@@ -0,0 +1,225 @@
...
+    data: {
+      name: 'test',
+      age: 30,
+    },
+  },
+  { data: { name: 'test2', age: 40 } },
+];
+const schemaName = 'testing';
+
+const seedDb = ({ keystone, ...rest }) =>
```
timleslie, line 26: We never use `...rest` so it should be removed.

packages/server-side-graphql-client/tests/index.test.js
```diff
@@ -0,0 +1,225 @@
...
+          expect(allItems).toEqual(testData.map(x => x.data));
+        })
+      );
+    });
+    describe('udpate', () => {
+      test(
+        'Should update single item',
+        runner(setupKeystone, async ({ keystone }) => {
+          // Seed the db with single item
+          const seedItems = await seedDb({ keystone, item: testData[0].data });
```
timleslie, line 99:
```suggestion
          const seedItems = await seedDb({ keystone });
```

packages/server-side-graphql-client/README.md
MadeByMike: I did this because the previous examples had:

```
// inside an asyc function
await ....
```

I agree that an asyc example is much neater to read than a promise, so I think it makes sense to show a tiny little function that demonstrates it is async as well as giving a bit of a hint about the types of functions users might want to create.

packages/server-side-graphql-client/lib/server-side-graphql-client.js
MadeByMike: Oh ok
```suggestion
  return [].concat(result.map(item => item[gqlName]));
```

MadeByMike: Like this?

packages/server-side-graphql-client/README.md
MadeByMike: Why? It accepts the shared config items.

packages/server-side-graphql-client/lib/server-side-graphql-client.js
timleslie: Yep, except you need to spread the args in.

```suggestion
return [].concat(...result.map(item => item[gqlName]));
```

packages/server-side-graphql-client/tests/index.test.js
MadeByMike: I use it! What do you want to name it? options?

packages/server-side-graphql-client/README.md
timleslie: It's in the wrong spot of the file.

packages/server-side-graphql-client/tests/index.test.js
MadeByMike: I guess this test works ok without that, I'd assumed Aman had done this for a reason - but if not we can remove that `...rest`

timleslie: Well this code doesn't actually do anything. Passing `item` into `createItems` doesn't ever get used 🤷‍♂️

packages/server-side-graphql-client/README.md
MadeByMike: Good point ;)

packages/server-side-graphql-client/README.md
timleslie: Yeah, I think this is tricky. As is, these examples aren't cut and pastable, we're in a bit of a no-man's-land between cut-and-pastable. I think perhaps those kind of examples belong in a more learning oriented resource, like a data-seeding users guide, rather than the information-oriented API docs. We don't have to get this right straight away, let's just flag it as something to think about and loop back on in the future.

packages/server-side-graphql-client/tests/index.test.js
MadeByMike: Correct, should be `items` but I still think we can just remove -I'll do it and see if CI blows up :)

packages/server-side-graphql-client/README.md
MadeByMike: Agree, as it is I think these are better structure than the previous examples -but we can probably do more in future

packages/server-side-graphql-client/package.json
MadeByMike, line 18:
```suggestion
```

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,370 @@
...
+
+- `createItem`
+- `createItems`
+- `updateItem`
+- `updateItems`
+- `deleteItem`
+- `deleteItems`
+- `runCustomQuery`
+
+> NOTE: All functions accepts an `config` object as an argument, and return a `Promise`.
```
timleslie, line 48:
```suggestion
> NOTE: All functions accept a config object as an argument, and return a `Promise`.
```

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,370 @@
...
+
+### Shared Config Options
+
+The following config options are common to all server-side graphQL functions.
+
+| Properties     | Type       | Default    | Description                                                                                                  |
+| -------------- | ---------- | ---------- | ------------------------------------------------------------------------------------------------------------ |
+| `keystone`     | `Keystone` | (required) | Keystone instance.                                                                                           |
+| `listName`     | `String`   | (required) | Keystone list name name.                                                                                     |
```
timleslie, line 57:
```suggestion
| `listName`     | `String`   | (required) | Keystone list nam.                                                                                     |
```

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,370 @@
...
+| `returnFields` | `String`   | `id`       | A graphQL fragment of fields to return. Must match the graphQL return type.                                  |
+| `schemaName`   | `String`   | `public`   | Name of your GraphQL API schema. To override the access-control mechanism, provide `internal` as schemaName. |
```
timleslie, line 59:
```suggestion
| `schemaName`   | `String`   | `public`   | Name of your GraphQL schema. To override the access-control mechanism, provide `internal` as schemaName. |
```

packages/server-side-graphql-client/tests/index.test.js
```diff
@@ -0,0 +1,225 @@
...
+];
+const schemaName = 'testing';
+
+const seedDb = ({ keystone }) =>
+  createItems({
+    keystone,
+    listName: 'Test',
+    items: testData,
+    schemaName,
+    ...rest,
```
singhArmani, line 32:
```suggestion
```

packages/server-side-graphql-client/tests/index.test.js
```diff
@@ -0,0 +1,225 @@
...
+
+          expect(allItems).toEqual(testData.map(x => x.data));
+        })
+      );
+    });
+    describe('udpate', () => {
+      test(
+        'Should update single item',
+        runner(setupKeystone, async ({ keystone }) => {
+          // Seed the db with single item
```
singhArmani, line 98:
```suggestion
          // Seed the db
```

packages/server-side-graphql-client/tests/index.test.js
```diff
@@ -0,0 +1,225 @@
...
+            schemaName,
+          });
+
+          expect(allItems).toEqual([]);
+        })
+      );
+    });
+    describe('getItems', () => {
+      test(
+        'Should get all items when no where caluse',
```
singhArmani, line 191:
```suggestion
        'Should get all items when no where clause',
```

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,370 @@
...
+});
+
+const updateUsers = async (updateUser) => {
+  const users = await updateItems({
+    keystone,
+    listName: 'User',
+    items: [
+      {id: '123', data: {name: 'newName1'},
+      {id: '456', data: {name: 'newName2'}
+    ],
```
singhArmani, line 278:
```suggestion
    items: updateUsers
```

packages/server-side-graphql-client/README.md
```diff
@@ -0,0 +1,370 @@
...
+const { updateItems } =  require('@keystonejs/server-side-graphql-client')
+
+keystone.createList('User', {
+  fields: {
+    name: { type: Text },
+    email: { type: String },
+  },
+});
+
+const updateUsers = async (updateUser) => {
```
singhArmani, line 271:
```suggestion
const updateUsers = async (updateUsers) => {
```

timleslie:
This branch has served its purpose. See #3300 #3301 #3302 #3303 for final implementation.

PR #3254, opened 2020-07-16
Combine the result of runChunkedMutation into an object. 

singhArmani:
**Combine result of `runChunkedMutation` query**. 
Initial implementation of `runChunkedMutation` array was returning the result in the format of nested array. 
```js
  // The result is of the format: 
  const result = [{createUsers: [{id: '123', name: 'aman'}]}, {createUsers: [{id: '456', name: 'mike'}]}]


```
It was making hard to extract the actual data. This refactoring combines the result of chunk promise arrays into a single object with key based of `createManyInputName`. 
```js
  // Combine all objects into one array keyed by the `createUsers`, such that, the output is: 
  const newResult =  {createUsers: [{id: '123', name: 'aman'}, {id: '456', name: 'Mike'}]}
```
Also, it aligns to the data structure returned by the graphql query 😃.


.changeset/sweet-books-sparkle.md
```diff
@@ -0,0 +1,21 @@
+---
```
timleslie, line 1: No need for a changeset in this PR, as it's just merging into another branch.

packages/orm/lib/orm.js
```diff
@@ -66,13 +66,22 @@ const runPaginatedQuery = async ({ pageSize = 1, variables = {}, verbose = false
   return allItems;
 };
 
-const runChunkedMutation = async ({ pageSize = 500, items, ...rest }) => {
+const runChunkedMutation = async ({ listName, pageSize = 500, items, ...rest }) => {
   const chunks = chunkArray(items, pageSize);
 
+  const { createManyMutationName } = rest.keystone.lists[listName].gqlNames;
```
timleslie, line 72: Pulling out `createManyMutationName` here will only be the right thing to do if the chunked mutation is a `create` operation. We should really be passing in the `gqlName` from whoever calls `runChunkedMutation`.

singhArmani: Ah, that's a very good point. Thanks for bringing it up. @timleslie

MadeByMike: This is a great idea - createItems should pass that 👍

PR #3257, opened 2020-07-19
Update README.md

zamkevich:
I understand that you can name your const as a class argument (adapter ed.) and it will work, but code like this is hard to read for beginners.


packages/fields/src/types/CloudinaryImage/README.md
```diff
@@ -19,7 +19,7 @@ const { CloudinaryImage } = require('@keystonejs/fields');
 
 const keystone = new Keystone({...});
 
-const adapter = new CloudinaryAdapter({
+const fileAdapter = new CloudinaryAdapter({
```
gautamsi, line 22: should be cloudinaryAdapter

zamkevich: Ok

zamkevich:
I returned back to **filletAdapter** name for several reasons:

- Consistency: we use this name everywhere for any type of adapter;
- Actually it's don't make sense to name adapter so definitely. If user have more than one adapter he will figure it out himself. It is important for us to show a basic example without misleading the user.
- It's got confused CloudinaryAdapter != cloudinaryAdapter.

MadeByMike:
`filletAdapter` sounds perfect 😂 Thank for the contribution.

PR #3265, opened 2020-07-20
Call jest.setTimeout from jest/setup.js

timleslie:
Out package.json contains the following:

```
  "jest": {
    "setupFilesAfterEnv": [
      "./jest/setup.js"
    ],
    ...
  }
```

which means we can use this file to consistently call `jest.setTimeout()` fo all our tests.

https://jestjs.io/docs/en/configuration#setupfilesafterenv-array

### DIFF
--- a/api-tests/access-control/schema.test.js
+++ b/api-tests/access-control/schema.test.js
@@ -10,10 +10,6 @@ const {
   getFieldName,
 } = require('./utils');
 
-// `mongodb-memory-server` downloads a binary on first run in CI, which can take
-// a while, so we bump up the timeout here.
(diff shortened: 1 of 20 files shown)


api-tests/access-control/schema.test.js
```diff
@@ -10,10 +10,6 @@ const {
   getFieldName,
 } = require('./utils');
 
-// `mongodb-memory-server` downloads a binary on first run in CI, which can take
-// a while, so we bump up the timeout here.
-jest.setTimeout(60000);
```
MadeByMike, line 15: :(

PR #3272, opened 2020-07-21
Admin UI: stop passing unnecessary props around

Vultraz:
Removed
- Removed `onDelete` prop from `ListTable` and `ListRow` (was called in `ListRow::handleDelete`, but no callback was ever given)
- Removed unnecessary wrapping `Fragment` from `ListManage`

List view components
- Stop passing `items`, `itemCount`, and `queryError` props to `ListLayout` (now fetched via hooks)
- Stop passing `fields`, `items`, `queryErrors`, `list`, `onSelectChange`, `selectedItems`, `currentPage`, `filters`, and `search` to `ListTable` (now fetched via hooks)
- Stop passing `list`, `fields`, and `onSelectChange` to `ListRow` (now fetched via hooks)
- Stop passing `handleSortChange` to `SortLink` (now fetched via hook)

Also stop passing `list` to the following (now fetched via hook)
- `DeleteItemModal`
- `DeleteManyItemsModal`
- `UpdateManyItemsModal`
- `NoResults`
- `Search`
- `ItemDetails`
- `ItemTitle` (this one was unused)
- `ActiveFilters`

### DIFF
--- a/.changeset/chilled-mails-build.md
+++ b/.changeset/chilled-mails-build.md
@@ -2,4 +2,4 @@
 '@keystonejs/app-admin-ui': patch
 ---
 
-Stop passing unnecessary props around through ListLayout, ListTable, ListRow, and SortLink.
+Stopped passing unnecessary props around, especially ones which can be accessed via hooks.
--- a/.changeset/chilled-mails-build.md
+++ b/.changeset/chilled-mails-build.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/app-admin-ui': patch
+---
+
+Stop passing unnecessary props around through ListLayout, ListTable, ListRow, and SortLink.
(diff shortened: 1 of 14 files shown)


molomby:
Isn't this technically a breaking change, since projects could be depending on these (redundant) variables, or are all these variables internal only?

Vultraz:
I suppose, yes. Many of these are exported.

MadeByMike:
Unfortunately I think this is a breaking change and not necessarily one we want to make. I'm very pro using the hooks but for items exported from `packages/app-admin-ui/components/index.js` we want people to be able to pass their own props to these and in some cases I think a deliberate decision was made to use the props rather than pluck it from context.

The following files look safe to change:

```
packages/app-admin-ui/client/pages/List/Management.js 
packages/app-admin-ui/client/pages/Item/index.js 
packages/app-admin-ui/client/components/NoResults.js
```

PR #3299, opened 2020-07-28
Fix isUnique for DateTime on MongooseAdapter

timleslie:
* Correctly apply unique schema config for `DateTime` field on mongoose.
* Provides a higher level flag to indicate whether a field type supports uniqueness or not.
* Adds tests for uniqueness to core fields.
* Update `isRequired` tests to be more similar to `isUnique` tests.

### DIFF
--- a/packages/fields/src/Implementation.js
+++ b/packages/fields/src/Implementation.js
@@ -30,6 +30,12 @@ class Field {
     this.getListByKey = getListByKey;
     this.listKey = listKey;
     this.label = label || inflection.humanize(inflection.underscore(path));
+    if (this.config.isUnique && !this._supportsUnique) {
+      throw new Error(
+        `isUnique is not a supported option for field type ${this.constructor.name} (${this.path})`
+      );
+    }
+
     this.adapter = listAdapter.newFieldAdapter(
       fieldAdapterClass,
       this.constructor.name,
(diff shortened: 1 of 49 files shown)


packages/fields/src/Implementation.js
```diff
@@ -51,6 +57,12 @@ class Field {
     });
   }
 
+  // By default we assume that fields support unique.
+  // Fields should override this method if not.
+  get _supportsUnique() {
+    return true;
```
JedWatson, line 63: I feel reasonably strongly that this isn't a safe default.

When creating a new field type, I shouldn't need to know to opt-out of behaviour that I'm not implementing.

This is more likely to be relevant for compound field types (e.g images, locations, embeds) etc than simple field types (text, numbers, etc) but given we have both, and in one case (creating a compound field) it's going to break by default, I'd prefer to flip the default

timleslie: Yep, I can get on board with this.

timleslie: Although that makes this a `major` breaking change.

timleslie:
Good point. I did it this way initially as I was thinking about ways to integrate this flag into the way the tests are executed. In the end I didn't end up using this mechanism and so I think you're right, it makes more sense for this to be a simple property set in the constructor. I'll make that change 👍

PR #3304, opened 2020-07-28
Rename  field-content to fields-content

MadeByMike:
Renamed `@keystonejs/field-content` to `@keystonejs/fields-content`.


MadeByMike:
I made a decision to keep the version number despite the rename. Since there were no functional change to the package I felt this makes sense.

.changeset/smart-baboons-shave.md
```diff
@@ -0,0 +1,9 @@
+---
+'@keystonejs/fields-content': major
```
MadeByMike, line 2:
```suggestion
'@keystonejs/field-content': major
'@keystonejs/fields-content': major
```

.changeset/smart-baboons-shave.md
```diff
@@ -0,0 +1,10 @@
+---
+'@keystonejs/field-content': major
```
mitchellhamilton, line 2: This can't be here given the package no longer exists
```suggestion
```

timleslie: I think perhaps what we should do is to keep the original package, perform a `major` bump, and have that final package simply throw an error which points the dev towards the new package.

Once that final package has been published we can remove the package from the repo.

Thoughts?

mitchellhamilton: Yep, that sounds good.

PR #3305, opened 2020-07-28
Move the location to fields-location-google

MadeByMike:
Moved the `Location` field from `@keystonejs/fields` to `@keystonejs/fields-location-google`.


packages/fields/package.json
```diff
@@ -46,7 +46,7 @@
     "cuid": "^2.1.8",
     "date-fns": "^2.14.0",
     "dumb-passwords": "^0.2.1",
-    "google-maps-react": "^2.0.2",
+    "google-maps-react": "^2.0.6",
```
timleslie, line 49: This can be removed from the `fields` package.

mitchellhamilton:
This is done code-wise though @JedWatson suggested that this should be named `fields-location-google`  in https://github.com/keystonejs/keystone/pull/3280#pullrequestreview-456233019. Is that what we want? What should the name of the field type be? `LocationGoogle`?

timleslie:
> This is done code-wise though @JedWatson suggested that this should be named `fields-location-google` in [#3280 (review)](https://github.com/keystonejs/keystone/pull/3280#pullrequestreview-456233019). Is that what we want? What should the name of the field type be? `LocationGoogle`?

I'm +1 on `fields-location-google` and `LocationGoogle`. Before reading the code I had assumed that this would be a simple lat/long geo location, not a google location specific thing, so specificity in naming would clear that up, while leaving the door open for other Location fields in the future.

PR #3306, opened 2020-07-29
Add blog section to website

MadeByMike:
For discussion.

If we plan to ship this as a blog it's probably ok. But the second half moved into my musing about current field types in a very specific way and needs to be generalised for docs.

Full of typos - will fix later :) 

Adds a blog section to the website:

image

I need to work on the template for blog pages... I think they should look different:

image


gautamsi:
I want to add references to contrib field and other packages,  what is best section for this. I guess new pr is needed for that.

MadeByMike:
Yeah I can do that. Let me work out where this is going to live and then I can add that in an appropriate way.

molomby:
I'd strongly suggest we put the date in the URL of the posts and probably the physical file name too (ie. `docs/blog/2020-08-03/field-types.md` or something). As we touched on today, one of the main distinguishing points between posts and the other docs is their point-in-time-ness. It's important that the age of a post be obvious. 

(It's also important we get the URLs correct up front so we don't break links later on)

Noviny (approved):

packages/create-keystone-app/example-projects/starter/initial-data.js
```diff
@@ -5,7 +5,7 @@ module.exports = async keystone => {
   // Count existing users
   const {
     data: {
-      _allUsersMeta: { count },
+      _allUsersMeta: { count = 0 },
```
timleslie, line 8: Does this change need to be in this PR? If so, we need a changeset for it.

timleslie (changes requested):

PR #3308, opened 2020-07-30
Move fields-datetime-utc into fields

MadeByMike:
Move `@keystonejs/fields-datetime-utc` into `@keystonejs/fields`


packages/fields/src/types/DateTimeUtc/README.md
```diff
@@ -13,12 +13,10 @@ Produces JS `Date` objects and ISO 8601 strings.
 
 Unlike the core `DateTime` field type only the UTC value is stored.
```
timleslie, line 14:
```suggestion
Unlike the `DateTime` field type only the UTC value is stored.
```

packages/fields-datetime-utc/CHANGELOG.md
```diff
@@ -1,267 +0,0 @@
-# @keystonejs/fields-datetime-utc
```
timleslie, line 1: Should we keep the CHANGELOG file around until after the release so that a final version including the deprecation notice can be published?

PR #3312, opened 2020-07-30
Use keystone.executeGraphQL in place of authedGraphqlRequest

timleslie:


timleslie:
> This looks good to me 👍 . If we are not using `authedGraphqlRequest` anywhere, we can remove it safely out of the `test-utils` package.

Yep, I'd like to do that at some point in the not entirely distant future, but there are a few other things in the `test-utils` package which should disappear at the same time. Other soon to land PRs will make them a bit more clear hopefully 👍

Issue #2822, opened 2020-04-24
AdminUI doesn't confirm on 'cancel in item create


ra-external:
When one is creating a new item, for example an 'Article', if one clicks on the 'create' button top right 

![image](https://user-images.githubusercontent.com/60699805/80214133-94bf1f80-863a-11ea-8772-6746ff8b72ba.png)

and one fills out the item's fields and then clicks 'cancel' on the bottom

![image](https://user-images.githubusercontent.com/60699805/80214200-b1f3ee00-863a-11ea-9cc3-25b08aa60c6f.png)

one does not receive a confirmation asking if one really wants to abandon editing.  This causes all entered data to be lost.

What should happen is that you get a confirmation 'dialog' such as one gets when one deletes an item 

![image](https://user-images.githubusercontent.com/60699805/80214276-d9e35180-863a-11ea-9268-d9761d320985.png)

In the meantime, is there any way I can implement the desired behavior?

gautamsi:
when you hit create again, you should have the date filled in already

ra-external:
Ok, that's good to know, but it's not very intuitive -- I would think a standard pop-up warning would be, well, standard...

MadeByMike:
Yep, seems reasonable. This is probably a good first issue if you'd like to have a crack at it @ra-external you are getting fairly familiar with the Admin UI. If not I'd say it's going to be a lower priority compared with other things on our list for a while.

ra-external:
@MadeByMike  -- Thanks, when I can come up for air from the project I'm working on I'll take a look at it.

PR #3317, opened 2020-07-30
Alert user before closing the createItemModal form

singhArmani:
The form data inside the `createItemModal` was getting lost when user
hit the `cancel` button. This could lead to a poor UX.

This fix will identifies if form data has changed from its initial
values before closing the modal. If yes, then it alerts the user for
confirmation. Otherwise, will close the modal.
admin

Closes: #2822

### DIFF
--- a/packages/app-admin-ui/client/components/CreateItemModal.js
+++ b/packages/app-admin-ui/client/components/CreateItemModal.js
@@ -16,6 +16,7 @@ import { useToasts } from 'react-toast-notifications';
 
 import { Button, LoadingButton } from '@arch-ui/button';
 import Drawer from '@arch-ui/drawer';
+import Confirm from '@arch-ui/confirm';
 import {
   arrayToObject,
   captureSuspensePromises,
@@ -50,6 +51,7 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
   const { list, closeCreateItemModal, isCreateItemModalOpen } = useList();
 
   const [item, setItem] = useState(list.getInitialItemData({ prefill: prefillData }));
+  const [isConfirmOpen, setConfirmOpen] = useState(false);
   const [validationErrors, setValidationErrors] = useState({});
   const [validationWarnings, setValidationWarnings] = useState({});
 
@@ -129,14 +131,39 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
     }
   });
 
-  const _onClose = () => {
-    if (loading) return;
+  // Identifies if the user has changed the initial data in the form.
+  const hasFormDataChanged = () => {
+    const data = arrayToObject(creatable, 'path');
+    let hasChanged = false;
+    const initialData = list.getInitialItemData({ prefill: prefillData });
+    const initialValues = getValues(data, initialData);
+    const currentValues = getValues(data, item);
+    for (const path of Object.keys(currentValues)) {
+      if (data[path].hasChanged(initialValues, currentValues)) {
+        hasChanged = true;
+        break;
+      }
+    }
+    return hasChanged;
+  };
+
+  const _createItemModalClose = () => {
     closeCreateItemModal();
     setItem(list.getInitialItemData({}));
-    const data = arrayToObject(creatable, 'path', field => field.serialize(item));
     if (onClose) {
+      const data = arrayToObject(creatable, 'path', field => field.serialize(item));
       onClose(data);
     }
+  };
+
+  const _onClose = () => {
+    if (loading) return;
+    if (hasFormDataChanged()) {
+      // Ask for user confirmation before canceling.
+      setConfirmOpen(true);
+      return;
+    }
+    _createItemModalClose();
   };
 
   const _onKeyDown = event => {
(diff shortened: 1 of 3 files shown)


packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -244,10 +271,36 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
               ));
             }}
           </Render>
+          <ConfirmModal
+            isOpen={isConfirmOpen}
+            onYes={() => {
```
jordanoverbye, line 276: I'd call this `onConfirm` and `onCancel`

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -129,16 +131,41 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
...
     closeCreateItemModal();
     setItem(list.getInitialItemData({}));
-    const data = arrayToObject(creatable, 'path', field => field.serialize(item));
     if (onClose) {
+      const data = arrayToObject(creatable, 'path', field => field.serialize(item));
       onClose(data);
     }
   };
 
+  const _onClose = () => {
```
jordanoverbye, line 159: Can you make this use `useCallback`

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -129,16 +131,41 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
...
+    for (const path of Object.keys(currentValues)) {
+      if (data[path].hasChanged(initialValues, currentValues)) {
+        hasChanged = true;
+        break;
+      }
+    }
+    return hasChanged;
+  };
+
+  const _createItemModalClose = () => {
```
jordanoverbye, line 150: Can you make this use `useCallback`

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -129,16 +131,41 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
     }
   });
 
-  const _onClose = () => {
-    if (loading) return;
+  // Identifies if the user has changed the initial data in the form.
+  const hasFormDataChanged = () => {
```
jordanoverbye, line 135: Can you make this use `useCallback`

packages/app-admin-ui/client/components/CreateItemModal.js
jordanoverbye: I'd also move these inline functions to callbacks too.

```
 <ConfirmModal
  isOpen={isConfirmOpen}
  onConfirm={onConfirm}
  onCancel={onCancel}
/>
```

MadeByMike:
Looks pretty good. I do agree about changing `onYes` to `onConfirm`. 

I also wondering if the has data changed logic could be more simple? Normally I'd just set an `isDirty` flag when someone interacts with the form elements. It's not perfect but avoids owning more logic. -- What you've done looks ok. This is just a question.

Make what changes you think necessary and I'll ✅

packages/app-admin-ui/client/components/CreateItemModal.js
singhArmani: This may not require useCallback decorator as it's not being passed as a prop.

singhArmani:
> This is looking good! I just have some comments around variable names and callbacks.

I have refactored the variables' names 👍 . But I am not sure I agree with wrapping all the functions with `useCallback` hook. `useCallback` does have some cost associated with it. And IMO seems pre-optimization.

MadeByMike:
> I am not sure I agree with wrapping all the functions with `useCallback` hook.  

I care little about the pre-optimisation cost or value of the hook (in this specific case) - all I care about here is how readable the code is and I think there's little difference either way here. Not going to hold out for that change.

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -147,6 +174,17 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
...
+  const _onConfirm = () => {
+    setConfirmOpen(false);
+    _createItemModalClose();
+  };
+
+  // Handler when user hits the cancel on confirm modal
+  const _onCancel = () => {
+    setConfirmOpen(false);
+  };
+
```
MadeByMike, line 187:
```suggestion
```

packages/app-admin-ui/client/components/CreateItemModal.js
```diff
@@ -244,10 +282,29 @@ const CreateItemModal = ({ prefillData = {}, onClose, onCreate, viewOnSave }) =>
               ));
             }}
           </Render>
+          <ConfirmModal isOpen={isConfirmOpen} onConfirm={_onConfirm} onCancel={_onCancel} />
```
MadeByMike, line 285:
```suggestion
          <ConfirmModal 
            isOpen={isConfirmOpen} 
            onConfirm={() => {
              setConfirmOpen(false);
              _createItemModalClose();
            }} 
            onCancel={() => {
              setConfirmOpen(false);
            }} 
          />
```

MadeByMike: You might need to run prettier after this.

packages/app-admin-ui/client/components/CreateItemModal.js
MadeByMike: Yeah that was exactly my comment :)

MadeByMike: What's the value in moving an inline function unless it's used more than once or is huge? 

My preference is usually to keep the code as small as possible until it hinders readability.

singhArmani:
> Normally I'd just set an `isDirty` flag when someone interacts with the form elements. It's not perfect but avoids owning more logic. -- What you've done looks ok. This is just a question.

I completely agree that the logic is verbose, and should have a boolean flag to check the form state. But implementing `isDirty` may be tricky too IMO. It's not only that if the user has interacted (touched),  but also if the form values have changed from its previous values.

PR #3323, opened 2020-08-01
Plugins: fix for tracking update bug & minor tracking improve

zamkevich:
- Fixed bug in byTracking plugin where updatedBy field was not updated.

**Minor update**
Now in both of tracking plugins "updated" fields are updating only after event "update ", and remain empty for an event "create".

**Why?** 
This is logical and consistent:
- If the record has just been created, then it will not have update marks
- If the record is updated by anonymous the same thing
- We know that if there are no update marks, this means that the record has not yet been updated and we can filter the records.


zamkevich:
I thought for a while and looks like now I understand why you update Updated By/At fields by create event 😅

**Useful flow**
Using only createdBy fields for track changes like a "touch" trigger. In this option, a first iteration will be untracked if createdBy fields will be set only by "update" event.

And we have two options: first logical and consistent (read my [first comment](#issue-460963079)), second - situative. 
Both of them have a right to exist.

Please, tell me if I wrong 😬

**Possible solution**
We can add modal behavior to plugins: 
- If used default setup updated fields will be updated only by "create" event.
- if used only updated fields - they will be updated on both events (create & update).

**Code** 
updatedBy
```
…
    if (
      // if it's just created - we reset updatedBy field
      operation === 'create' &&
      // modal behavior - reset only if used both type of fields
      created &&
      updated
    ) {
      resolvedData[updatedByField] = null;
    }
…
```

createdBy
```
…
 if (operation === 'create') {
      // create mode
      if (created) {
        resolvedData[createdAtField] = dateNow;
      }
      // modal behavior - set If only update field is used
      if (!created && updated) {
        resolvedData[createdAtField] = dateNow;
      }
    }
…
```

So, what do you think?

MadeByMike:
Cross posting my comments from Slack:

Ok, I had quick look at your PR over the weekend and I have one thing I need to resolve in my head first.
I'm pretty sure the "standard" behaviour for most content management systems with `[updated/created][by/at]` fields is that create also sets the updated field. Another way of thinking about this is that create is a special kind of update whereas update is not a create.

I know this is different to what you were expecting so I am going to go an re-confirm what the typical behaviour is, however I do recall having this discussion when this was first built and I believe some similar "research" was done.


There is one other change there that I need to investigate:

```
if (
  ...
  // nothing is changed we keep existing item as well
  Object.keys(originalInput).length !== 0
) {
   ...
}
```

Just because the input was empty we cannot assume that an update has not happened. defaultValues and hooks are both applied after the originalInput is resolved. For that reason I think we've taken the approach that if an update mutation was successful (even with no change) the record is considered updated.

To me this make sense as you might want to record any attempt to update records in the system. I need to think about this more.

Step 1 I think is to write some unit test :slightly_smiling_face:

MadeByMike:
Sadly I could not push the tests and my fix in #3332 to your branch.

My fix doesn't account for your suggestion of modifying the update fields explicitly only on update. That is a breaking change (and not one I think we should make) however I might be convinced to accept a PR that allows this behaviour when the user opts in via config on the plugin.

zamkevich:
I understand all your points about standard behavior for updated fields. And your points to update updated fields by every touch. And I agree that may be standard behavior for developers.

My arguments were for the UX side. By UX and your explanation, it's must be three fields:
- **Create** → set when item is creating.
- **Update** → updated every time when some data is updated (not created).
- **Touch** → updated by any actions with item.

It's my thoughts as an interface designer based on how, and for what the fields are used 🤗

PR #3332, opened 2020-08-03
Add tests and fix bug in byTracking plugin

MadeByMike:
Add tests and fix bug in tracking plugin.


MadeByMike:
See my comment about the AuthRelationship field - I think we should remove it from here. It will make testing easier and will remove complexity. we rely on it ONLY to set the value since we provide custom access control. Sadly it's only in the way here and I regret that we added it.

MadeByMike:
Closes #3332

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -25,9 +25,12 @@ const _byTracking = ({ created = true, updated = true }) => ({
   }
 
   const newResolveInput = ({ resolvedData, operation, originalInput, context }) => {
+    // if no data received from the mutation, skip the update
+    if (Object.keys(originalInput).length === 0) {
+      return resolvedData;
+    }
+
```
MadeByMike, line 32:
```suggestion
```

packages/list-plugins/lib/tracking/byTracking.js
```diff
@@ -26,8 +26,6 @@ const _byTracking = ({ created = true, updated = true }) => ({
 
   const newResolveInput = ({ resolvedData, operation, originalInput, context }) => {
```
timleslie, line 27:
```suggestion
  const newResolveInput = ({ resolvedData, operation, context }) => {
```

MadeByMike:
Removed the following from atTracking for consistency:

```
      // if no data received from the mutation, skip the update
      if (Object.keys(originalInput).length === 0) {
        return resolvedData;
      }
```

To document reasoning. It's valid that a create or update may cause a side effect that system admins want to track. This is not about "has the data been input into the form?" as much as "has the create/update been triggered, regardless of changes?"

gautamsi:
that makes sense, we can live with that for now and in later PR we can make this configurable.

PR #3345, opened 2020-08-05
Add `sortBy`, `first`, and `skip` in getItems

singhArmani:
Enhance the `getItems` functions by adding:
- `sortBy`
- `first`
- `skip`

This will allow users to achieve some advance use cases.

### DIFF
--- a/.changeset/violet-points-sparkle.md
+++ b/.changeset/violet-points-sparkle.md
@@ -1,0 +1,12 @@
+---
+'@keystonejs/server-side-graphql-client': minor
+---
+
+Enhanced the `getItems` functionality by adding: 
+- `sortBy`
+- `first`
+- `skip`
+
+These variable are part of the public interface, and allow users to achieve some advance use cases like sorting and skipping.
+
+
--- a/packages/server-side-graphql-client/lib/server-side-graphql-client.js
+++ b/packages/server-side-graphql-client/lib/server-side-graphql-client.js
@@ -85,23 +85,38 @@ const getItems = async ({
   keystone,
   listKey,
   where = {},
+  sortBy,
+  first,
+  skip,
   pageSize = 500,
   returnFields = `id`,
   context,
 }) => {
-  const { listQueryName, whereInputName } = keystone.lists[listKey].gqlNames;
-  const query = `query ($first: Int!, $skip: Int!, $where: ${whereInputName}) { ${listQueryName}(first: $first, skip: $skip, where: $where) { ${returnFields} }  }`;
+  const { listQueryName, whereInputName, listSortName } = keystone.lists[listKey].gqlNames;
 
-  let skip = 0;
+  let query;
+  if (sortBy) {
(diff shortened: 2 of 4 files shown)


packages/server-side-graphql-client/lib/server-side-graphql-client.js
```diff
@@ -84,31 +84,39 @@ const getItem = async ({ keystone, listKey, itemId, returnFields = `id`, context
 const getItems = async ({
   keystone,
   listKey,
+  first,
+  skip,
   where = {},
   pageSize = 500,
   returnFields = `id`,
+  sortBy = 'id_ASC',
```
timleslie, line 92: We don't want a default for this. By default it should have no sorting applied.

packages/server-side-graphql-client/lib/server-side-graphql-client.js
```diff
@@ -84,31 +84,39 @@ const getItem = async ({ keystone, listKey, itemId, returnFields = `id`, context
...
+  let _first = first || pageSize;
   const allItems = [];
 
   do {
     const response = await runQuery({
       keystone,
       query,
       context,
-      variables: { first: pageSize, skip, where },
+      variables: { first: _first, skip: _skip, sortBy, where },
```
timleslie, line 108: I don't think this works as expected.

If I ask for `first: 100` with `pageSize: 10` then each query should be pulling `10` items at a time until we have collected a total of 100 items.

The current implementation will completely override all pagination if `first` is supplied.

.changeset/violet-points-sparkle.md
```diff
@@ -0,0 +1,13 @@
+---
+'@keystonejs/server-side-graphql-client': minor
+---
+
+Enhanced the `getItems` functionality by adding: 
+- `sortBy`
+- `first`
+- `skip`
```
timleslie, line 8: We want to add this info to the `README` file as well.

packages/server-side-graphql-client/lib/server-side-graphql-client.js
timleslie: We also need to be careful not to overfetch. If I have `first: 101, pageSize: 10` we need to be careful not to pull down `110` items.

singhArmani: Oh, nice catch. Thank you.

.changeset/violet-points-sparkle.md
```diff
@@ -0,0 +1,13 @@
...
+
+These variable are part of the public interface, and allow users to achieve some advance use cases like sorting and skipping.
+
+**Note**: We are still keeping the internal pagination behaviour as it is, but now prioritising `first` over the `pageSize`.
```
timleslie, line 12: I think we can remove this note, nothing in the described changes implies that pagination changed, so it's more confusing to call it out than to not.

packages/server-side-graphql-client/lib/server-side-graphql-client.js
timleslie, line 92: Let's order the parameters `where, orderBy, first, skip`.

PR #3347, opened 2020-08-05
Update initial-data.md

gpersable:
Added createItems init


MadeByMike:
Yep that make sense - thanks! @all-contributors please add @gpersable for docs

PR #3361, opened 2020-08-10
Fix fields-cloudinary-image views not being compiled

mitchellhamilton:
Found this while I was completely removing build-field-types, looks like I missed changing to entrypoints from build-field-types' importView even though I changed it to be built with Preconstruct instead of build-field-types.

Note that this currently means that fields-cloudinary-image is broken on npm right now so this should probably be released quickly

### DIFF
--- a/.changeset/violet-pandas-burn.md
+++ b/.changeset/violet-pandas-burn.md
@@ -2,4 +2,4 @@
 '@keystonejs/fields-cloudinary-image': patch
 ---
 
-Fix field type views not being compiled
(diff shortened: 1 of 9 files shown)


.changeset/violet-pandas-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields-cloudinary-image': patch
+---
+
+Fix field type views not being compiled
```
MadeByMike, line 5:
```suggestion
Fixed ` fields-cloudinary-image` views not being compiled.
```

PR #3365, opened 2020-08-10
Use pnpm instead of Yarn

mitchellhamilton:
Why pnpm?
- [It's faster than Yarn in most cases](https://github.com/pnpm/benchmarks-of-javascript-package-managers)(v1, using Yarn v2 would be unpalatable in my opinion because of the compatibility problems of PnP)
- Because of the node_modules structure it creates, it's possible to have conflicting versions of some package and the conflicting versions can be shared across different usages rather than the package being duplicated(This is important if we want to have different versions of packages so that we don't run into problems around duplicated copies of the same package at a given version)
- It doesn't let packages access all the dependencies of every other package because of the node_modules structure
- I've been using it on quite a few projects recently and have found it to be more reliable/I've basically never had to `rm -rf node_modules`

(This isn't ready rn, I'm just opening this so CI runs)


mitchellhamilton:
It seems like Cypress or a dependency of it has an incorrect implementation of Node's module resolution algorithm and it's probably not worth digging deep into it right now so I'm gonna close this. I might have another look at this in the future though.

PR #3374, opened 2020-08-11
Clean up Slug unique value generation

timleslie:
Replaces a somewhat confusing `do...while` loop with a more standard `for` loop, and makes use of `server-side-graphql-client` to simplify executing a query inside the hook.

### DIFF
--- a/packages/fields/package.json
+++ b/packages/fields/package.json
@@ -36,6 +36,7 @@
     "@keystonejs/adapter-knex": "^11.0.1",
     "@keystonejs/adapter-mongoose": "^9.0.1",
     "@keystonejs/app-admin-ui": "^7.3.0",
+    "@keystonejs/server-side-graphql-client": "^1.1.0",
     "@keystonejs/utils": "^5.4.2",
     "@primer/octicons-react": "^10.0.0",
     "@sindresorhus/slugify": "^0.11.0",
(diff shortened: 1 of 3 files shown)


packages/fields/package.json
```diff
@@ -36,6 +36,7 @@
     "@keystonejs/adapter-knex": "^11.0.1",
     "@keystonejs/adapter-mongoose": "^9.0.1",
     "@keystonejs/app-admin-ui": "^7.3.0",
+    "@keystonejs/server-side-graphql-client": "^1.1.0",
```
gautamsi, line 39: IMO the fields package should have minimum level of dependency, this can be achieved using context as done in earlier.

timleslie: The `server-side-grpahql-client` is an extremely lightweight package, especially when compared to the other dependencies of `fields`, so adding this dependency has a negligible impact.

packages/fields/src/types/Slug/Implementation.js
```diff
@@ -199,72 +200,43 @@ export class SlugImplementation extends Text {
...
-      const { data, errors } = await context.executeGraphQL({
-        // Access Control may filter out some results, so we wouldn't be
-        // retreiving an accurate list of all existing items. Because we add the
-        // unique constraint to the field, the database will throw an error if
-        // we miss a match and try to insert anyway.
-        context: context.createContext({ skipAccessControl: true }),
-        query: queryString,
-        variables: {
+      try {
+        const result = await getItems({
```
gautamsi, line 213: IMO we can use some helper in utils to help simplify queries if needed. (I see the query is dumped in favour of getItems)

timleslie: `getItem` *is* the query helper, there's no need to re-invent that wheel. This is the exact use case that `getItem` was designed to support, e.g. running queries from within a hook.

PR #3298, opened 2020-07-27
Add a Prisma adapter

timleslie:
[Prisma 2](https://www.prisma.io/blog/announcing-prisma-2-n0v98rzc8br1/) has recently been released. It provides a powerful javascript library for interfacing with databases. The Prisma team have recently received [significant funding](https://www.prisma.io/blog/prisma-raises-series-a-saks1zr7kip6/) to continue development of their library.

This PR introduces support for using Prisma as a database adapter within Keystone. 

We are still reviewing Prisma for feature completeness and correctness to ensure that all of Keystone's features will work as expected.

We are also investigating the Prisma migration framework in order to provide a better DX for developers who need to perform migrations on their database as their Keystone system evolves.

Note: This PR is still WIP, particularly in regard to how, where and when we are generating the Prisma client.

See also: https://github.com/prisma/prisma/issues/3265

#### Relationships

| Type | Create | Read | Count | Update | Delete | Notes |
| :---  | :---: | :---: | :---: | :---: | :---: | :--- |
| One-to-many one-sided | ✅  | ✅  | ✅  | ✅  | ✅  |
| One-to-many two-sided | ✅  | ✅  | ✅  | ✅  | ✅  | ~~https://github.com/prisma/prisma/issues/3141~~ |
| Many-to-many one-sided | ✅  | ✅  | ✅  | ✅  | ✅  |
| Many-to-many two-sided  | ✅  | ✅  | ✅  | ✅  | ✅  |
| One-to-one   | ✅  | ✅  | ☑️   | ✅  | ✅  |
| **Self-referential**
| One-to-many one-sided | ✅  | ✅  | ✅  | ✅  | ✅  |
| One-to-many two-sided | ✅ | ✅ | ✅  | ✅ | ✅ | ~~https://github.com/prisma/prisma/issues/3078~~ |
| Many-to-many one-sided | ✅  | ✅  | ✅  | ✅  | ✅  |
| Many-to-many two-sided | ✅  | ✅  | ✅  | ✅  | ✅  |
| One-to-one  | ✅ | ✅ | ☑️   | ✅  | ✅  | ~~https://github.com/prisma/prisma/issues/3153~~ |

#### Fields

| Name | Prisma Type | Core | Filter | Required | Unique | Notes |
| :--- | :---: | :---: | :---: | :---: | :---: | :--- |
| **Core**
| `CalendarDay` | `DateTime` | ✅ | ✅ | ✅  | ✅  |
[...]

PR #3375, opened 2020-08-11
Add filter tests to URL field type

singhArmani:


.changeset/slow-bananas-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Add `Filter` tests for the `Url` field type
```
MadeByMike, line 5:
```suggestion
Added `Filter` tests for the `Url` field type.
```

singhArmani:
> We will want to test the case-insensitive filter options as well.

Yeah, I was thinking that as it's not recommended to have small letter URL, is it relevant to those scenarios? 🤔 @timleslie

timleslie:
> Why are testing this since it uses the `Text` implementation?

This is about black-box testing the API to ensure that it behaves in the way we expect, independent of what we know about the implementation details. This is important for verifying the correctness of the prisma adapter in #3298 .

PR #3382, opened 2020-08-12
small docs tweak

Noviny:
This broke when I was spinning some things up and took a little while to debug what the problem was.

### DIFF
--- a/docs/guides/intro-to-graphql.md
+++ b/docs/guides/intro-to-graphql.md
@@ -194,15 +194,15 @@ mutation AddUser($name: String!) {
   }
 }`;
 
-fetch('/admin/api', {
+const data = await fetch('/admin/api', {
   method: 'POST',
   headers: {
     'Content-Type': 'application/json',
   },
-  body: {
+  body: JSON.stringify({
     query: ADD_USER,
     variables: { name: 'Mike' },
-  },
+  }),
 }).then(result => result.json());
 ```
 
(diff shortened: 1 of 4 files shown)


docs/guides/intro-to-graphql.md
```diff
@@ -170,15 +170,17 @@ query GetUsers {
...
-fetch('/admin/api', {
-  method: 'POST',
+fetch("/admin/api", {
+  method: "POST",
   headers: {
-    'Content-Type': 'application/json',
+    "Content-Type": "application/json",
   },
-  body: {
+  body: JSON.stringify({
```
MadeByMike, line 178: Do we need to stringify a string?

docs/guides/intro-to-graphql.md
```diff
@@ -170,15 +170,17 @@ query GetUsers {
...
     query: GET_ALL_USERS,
-  },
-}).then(result => result.json());
+  }),
+})
+  .then((result) => result.json())
+  .then((r) => updateData(r));
```
MadeByMike, line 183: Is `updateData` and example of how to do something with the result? If so, maybe we should just assign it to something? i.e?

`const result = fetch(...)`

docs/guides/intro-to-graphql.md
singhArmani: We are stringifying an object here: `{query: ...}`.

PR #3388, opened 2020-08-13
Cleanup test-utils module

singhArmani:
- Removed some of the redundant functions like `findOne`, `findById` from `_keystoneRunner` function. Using utils from `server-side-graphql-client` package. 
- Removed `authedGraphqlRequest` and `graphqlRequest` as they are redundant now.

PR #3390, opened 2020-08-13
Cleanup test-utils module

singhArmani:
We were using some crud helper functions, for example `findById`, `findOne`,
and `update`, inside `_keystoneRunner` function. But now with the addition
of ssgc package we don't need them anymore.


singhArmani:
Consolidating this change into #3388 with `major` change.

PR #3392, opened 2020-08-13
Update matchFilter to use getItems

timleslie:
This PR removes `matchFilter` and instead uses direct calls to `getItems` from `server-side-graphql-client`. It removes the need to do client side sorting and removes an unnecessary level of abstraction which makes these `test-fixtures` harder to understand.

### DIFF
--- a/.changeset/small-cups-protect.md
+++ b/.changeset/small-cups-protect.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Updated `match` function in tests to use `getItems` function from `server-side-graphql-client`.
--- a/packages/fields/src/types/CalendarDay/test-fixtures.js
+++ b/packages/fields/src/types/CalendarDay/test-fixtures.js
@@ -1,4 +1,4 @@
-import { matchFilter } from '@keystonejs/test-utils';
+import { getItems } from '@keystonejs/server-side-graphql-client';
 import Text from '../Text';
 import CalendarDay from './';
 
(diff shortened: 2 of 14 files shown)


packages/test-utils/lib/test-utils.js
```diff
@@ -6,7 +6,6 @@ const { Keystone } = require('@keystonejs/keystone');
 const { GraphQLApp } = require('@keystonejs/app-graphql');
 const { KnexAdapter } = require('@keystonejs/adapter-knex');
 const { MongooseAdapter } = require('@keystonejs/adapter-mongoose');
-const { runCustomQuery } = require('@keystonejs/server-side-graphql-client');
```
timleslie, line 9: I believe this means we can remove the package dependency on `server-side-graphql-client`

.changeset/small-cups-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': minor
+---
+
+Updated `match` function to use `getItems` function from `server-side-graphql-client`.
```
timleslie, line 5: The `fields` package doesn't need a `major` change, as it's only update its tests. We do however need a `major` changeset on `test-utils`, as we're removing a function from its public API.

singhArmani: Yeah, I was kind of thinking on the same line. But wasn't sure.

.changeset/small-cups-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields': patch
+---
+
+Updated `match` function to use `getItems` function from `server-side-graphql-client`.
```
timleslie, line 5:
```suggestion
Updated `match` function in tests to use `getItems` function from `server-side-graphql-client`.
```

packages/test-utils/index.js
```diff
@@ -4,7 +4,6 @@ const {
   graphqlRequest,
   authedGraphqlRequest,
   networkedGraphqlRequest,
-  matchFilter,
```
timleslie, line 7: Need to revert these changes.

Issue #3389, opened 2020-08-13
Invalid schema generated when a related list has `access: false`


timleslie:
When a list has a relationship to another list with `access: false`, keystone generates an invalid graphQL schema.

```js
keystone.createList('User', {
  fields: {
    name: { type: Text },
    post: { type: Relationship, ref: 'Post', many: false },
  },
  access: true,
});

keystone.createList('Post', {
  fields: {
    name: { type: Text },
  },
  access: false,
});
```

```
Error: Unknown type "PostWhereUniqueInput". Did you mean "UserWhereUniqueInput", "PostRelateToOneInput", "UserWhereInput", "UserUpdateInput", or "UsersUpdateInput"?
```

While this is a fairly obscure access control configuration, it's something that can come up in systems which support multiple different schemas.

The solution involves checking for `refList.access[schemaName]` in the `Relationship` methods `getGqlAuxTypes`, `gqlUpdateInputFields`, and `gqlCreateInputFields`. Unfortunately the latter 2 of these are `getter` functions, so we can't pass in `schemaName` as required. Fixing the issue will require us to change these `getter` functions and pass in `{ schemaName }`.

PR #3394, opened 2020-08-14
Convert field getters into methods

timleslie:
This change is required to solve #3389, which will be solved on a separate PR once this is in.

~~This PR is currently blocked on #3391, which is moving around a Mock class which will need to be updated once #3391 lands.~~

### DIFF
--- a/packages/fields-auto-increment/src/Implementation.js
+++ b/packages/fields-auto-increment/src/Implementation.js
@@ -40,10 +40,10 @@ export class AutoIncrementImplementation extends Implementation {
       ...this.inInputFields(this.gqlType),
     ];
   }
-  get gqlUpdateInputFields() {
+  gqlUpdateInputFields() {
     return [`${this.path}: ${this.gqlType}`];
   }
-  get gqlCreateInputFields() {
+  gqlCreateInputFields() {
     return [`${this.path}: ${this.gqlType}`];
   }
 }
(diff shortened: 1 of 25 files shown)


packages/fields-auto-increment/src/Implementation.js
```diff
@@ -40,10 +40,10 @@ export class AutoIncrementImplementation extends Implementation {
       ...this.inInputFields(this.gqlType),
     ];
   }
-  get gqlUpdateInputFields() {
+  gqlUpdateInputFields() {
```
singhArmani, line 43: This function doesn't accept any param as mentioned in the changeset. Should it be: 
```js
 gqlUpdateInputFields({schemaName})
```
Same for `gqlCreateInputFields`.

packages/keystone/lib/ListTypes/list.js
```diff
@@ -1013,7 +1013,9 @@ module.exports = class List {
     if (schemaAccess.update && updateFields.length) {
       types.push(`
         input ${this.gqlNames.updateInputName} {
-          ${flatten(updateFields.map(field => field.gqlUpdateInputFields)).join('\n')}
+          ${flatten(updateFields.map(field => field.gqlUpdateInputFields({ schemaName }))).join(
```
singhArmani, line 1016: We need to update the method definition to accept `{schemaName}` object here.

packages/fields-auto-increment/src/Implementation.js
timleslie: It accepts the parameter, it just happens to ignore them. We don't specify unused arguments in our method definitions, so what we have here is correct.

A future PR will make use of the `{ schemaName }` argument specifically in the `Relationship` field type.

PR #3408, opened 2020-08-17
Setup TypeScript things

mitchellhamilton:
This is just setting up tsc and linting for TypeScript files.

No changesets (edit: for most packages)in this PR even though there are changes in packages because this stuff doesn't really need to be released.


packages/arch/packages/hooks/src/index.ts
```diff
@@ -1,15 +1,17 @@
 import ResizeObserver from 'resize-observer-polyfill';
```
mitchellhamilton, line 1: I just changed this file to TS because it seemed like the smallest thing I could test things on.

gautamsi:
I have tried this in past #1454. Finally seeing this land is good thing.

PR #3410, opened 2020-08-17
Add filter & Crud tests for Markdown field

singhArmani:


packages/fields-markdown/src/test-fixtures.js
```diff
@@ -1,7 +1,391 @@
...
+    withKeystone(({ keystone }) =>
+      match(keystone, { content_not_starts_with_i: '**This' }, [
+        { name: 'b', content: 'This is a [link](https://keystonejs.com)' },
+        { name: 'c', content: '~~Strike through~~' },
+        { name: 'e', content: null },
+      ])
+    )
+  );
+
+  test(
```
timleslie, line 267: Do these have `_i` variations?

singhArmani: @timleslie : Added `_i` variations too.

PR #3415, opened 2020-08-18
Add filter and CRUD tests for AutoIncrement field

singhArmani:
- Add a filter to accommodate `adapter` specific field API tests.
- Update `fields.test.js` to insert items in an order as in array returned by `initItems()` function of the module.

### DIFF
--- a/.changeset/few-icons-ring.md
+++ b/.changeset/few-icons-ring.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/api-tests': patch
+'@keystonejs/fields-auto-increment': patch
+---
+
+Added filter and CRUD tests for `AutoIncrement` field type.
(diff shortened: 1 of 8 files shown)


api-tests/fields.test.js
```diff
@@ -3,56 +3,81 @@ const path = require('path');
...
+      } else {
+        test.todo('Filtering - tests missing');
+      }
+    });
   });
+};
+
+describe('Fields', () => {
+  const testModules = globby.sync(
+    [`packages/**/src/**/test-fixtures.js`, `!packages/fields-auto-increment/src/test-fixtures.js`],
```
timleslie, line 53: I don't like having package-specific code in this general file. Can we find a way to implement this logic in such a way that it becomes part of the `test-fixtures.js` configuration?

timleslie (approved):
👍

PR #3420, opened 2020-08-19
Add listKey argument to hooks

timleslie:
Writing generic plugins is difficult if you can't access the list name, so this PR makes the `listKey` available to all hook functions. This is useful for plugins like logging which need to be applied to many lists and want to know the name of the list.


molomby (approved):
Hard to argue with that

PR #3421, opened 2020-08-19
Remove .md files from the prettier checks, as remark lints these files

timleslie:
We format markdown files with `remark`, which disagrees with `prettier` on the formatting rules to apply.

This PR allows `yarn format` to generate code which passes `yarn lint:prettier`.


mitchellhamilton (approved):
To unblock you, sure, let's do this.

PR #3425, opened 2020-08-20
Add a logging plugin

timleslie:


gautamsi:
this inspired me to write a plugin system RFC  for keystone instance, check #3430

PR #3433, opened 2020-08-21
Add filter and CRUD test for File field type

singhArmani:
NOTE: 
`skipRequiredTest` is still set to `true`. This is because in our update test inside `required-tests.js` file, making the mutation like this: 
```js
              keystoneTestWrapper(async ({ keystone }) => {
                const { data: data0, errors: errors0 } = await keystone.executeGraphQL({
                  query: `
                  mutation {
                    createTest(data: { name: "test entry", testField: ${mod.exampleValue} } ) { id }
                  }`,
                });
```
This is throwing an error of our `File` type object. 
On the contrary doing  the following makes other, for example Uuid, MongoId` fails because of casting/type issue: 
```js
              keystoneTestWrapper(async ({ keystone }) => {
                const { data: data0, errors: errors0 } = await keystone.executeGraphQL({
                  query: `
                  mutation($data: TestCreateInput) {
                    createTest(data: $data ) { id }
                  }`,
                  variables: {
                    data: {
                      name: 'test entry',
                      testField: mod.exampleValue,
                    },
                  },
                });
```

We need to revisit this` required-test.js` file and fix these accordingly in a separate PR. @timleslie


packages/fields/src/types/File/test-fixtures.js
```diff
@@ -1,7 +1,140 @@
...
+export const subfieldName = 'originalFilename';
+
+// Grab all the image files from the directory
+const testFiles = globby.sync(`packages/**/src/**/File/test-files/**`, { absolute: true });
+const directory = './files';
+
+// Ensures 'directory' exist
+if (!fs.existsSync(directory)) {
+  fs.mkdirSync(directory);
+}
```
timleslie, line 24: Let's put this into a `beforeAll` function to match the `afterAll` function

packages/fields/src/types/File/test-fixtures.js
```diff
@@ -1,7 +1,140 @@
...
+
+const prepareFile = filePath => {
+  const filename = path.basename(filePath);
+  const mimetype = mime.getType(filePath);
+  const createReadStream = () => fs.createReadStream(filePath);
+  const encoding = 'utf-8';
+  const upload = new Upload();
+  upload.resolve({ createReadStream, filename, mimetype, encoding });
```
timleslie, line 32:
```suggestion
  const upload = new Upload();
  upload.resolve({
    createReadStream: () => fs.createReadStream(filePath),
    filename: path.basename(filePath),
    mimetype: mime.getType(filePath),
    encoding: 'utf-8',
  });
```

timleslie (changes requested):

timleslie (approved):
This looks great now!

PR #3435, opened 2020-08-24
Update required test for Field API

singhArmani:
Referring to this issue [here](https://github.com/keystonejs/keystone/pull/3433#issue-471428574) here, `create` mutation now uses data argument as query variables. 
For this reason, we need to update all the example values in the `test-fixtures.js` files.


timleslie (approved):
Looks good!

PR #3462, opened 2020-08-25
Add tests for `OEmbed` field

singhArmani:
**NOTE**: As we are using github secrets, we need to look solving the issue when PR is created from a forked repository. For more info please check this [link](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow).

### DIFF
--- a/.changeset/tender-eyes-joke.md
+++ b/.changeset/tender-eyes-joke.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/api-tests': patch
+'@keystonejs/fields-oembed': patch
+---
+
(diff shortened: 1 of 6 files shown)


timleslie (changes requested):
Let's remove the attempted support for `_contains` and other string search operations.

singhArmani:
> Let's remove the attempted support for `_contains` and other string search operations.

This PR #3501 removes the unsupported filters.

timleslie (approved):
This all looks good 👍

PR #3463, opened 2020-08-25
Fix many-to-one-one queries

timleslie:


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -869,10 +872,13 @@ class QueryBuilder {
             );
           }
 
+          const subSubQuery = listAdapter.parentAdapter.knex.raw(
+            `SELECT "${selectCol}" FROM (${subQuery}) AS unused_alias`
```
MadeByMike, line 876: Why do you need an alias?

timleslie: https://www.postgresql.org/docs/12/sql-select.html

> A sub-SELECT can appear in the FROM clause. This acts as though its output were created as a temporary table for the duration of this single SELECT command. Note that the sub-SELECT must be surrounded by parentheses, and an alias must be provided for it. A VALUES command can also be used here.

MadeByMike (approved):

PR #3467, opened 2020-08-25
Add API tests for CloudinaryImage field type

singhArmani:
- Added filter, isRequired tests for `CloudinaryImage` field type. 
- Used github secrets in git workflow ('jest')
- Cleanup cloudinary resources after tests are completed to avoid exceeding the upload limit.


.changeset/tiny-rockets-lick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/api-tests': patch
+'@keystonejs/fields-cloudinary-image': patch
+---
+
+Add `filter`, `isRequired` and CRUD tests for `CloudinaryImage` field type.
```
timleslie, line 6:
```suggestion
Added `filter`, `isRequired` and CRUD tests for `CloudinaryImage` field type.
```

packages/keystone/package.json
```diff
@@ -31,7 +31,7 @@
     "express-pino-logger": "^4.0.0",
     "falsey": "^1.0.0",
     "fs-extra": "^7.0.0",
-    "globby": "^11.0.0",
+    "globby": "^11.0.1",
```
timleslie, line 34: The package version bumps here should result in a changeset for all the affected packages. In general it's probably best to leave and version bumps out of a feature PR like this unless they're actually required.

packages/fields-cloudinary-image/src/test-fixtures.js
```diff
@@ -1,6 +1,247 @@
...
+          expect(data.image.originalFilename).toBe(items[0].image.originalFilename);
+        })
+      )
+    );
+  });
+};
+
+// This function will run after all the tests are completed.
+// We use it to cleanup the resources (e.g Cloudinary images) which are no longer required.
+export const afterAll = () => cloudinary.v2.api.delete_resources_by_prefix('cloudinary-test');
```
timleslie, line 247: Let's move this code up to the top section of the file. We want to keep all the module config in a single location.

packages/fields-cloudinary-image/src/test-fixtures.js
```diff
@@ -1,6 +1,247 @@
...
+
+export const getTestFields = () => {
+  return {
+    name: { type: String },
+    image: { type: CloudinaryImage, adapter: fileAdapter },
+  };
+};
+
+export const initItems = () => {
+  let items = testFiles.map((p, i) => ({ name: `file${i}`, image: prepareFile(p) }));
```
timleslie, line 59:
```suggestion
  const items = testFiles.map((p, i) => ({ name: `file${i}`, image: prepareFile(p) }));
```

packages/fields-cloudinary-image/src/test-fixtures.js
```diff
@@ -1,6 +1,247 @@
+const path = require('path');
+require('dotenv').config({ path: path.resolve(__dirname, '../.env') });
+
+import { createItem, getItem, getItems, updateItem } from '@keystonejs/server-side-graphql-client';
+import { CloudinaryAdapter } from '@keystonejs/file-adapters';
+
+import cloudinary from 'cloudinary';
+
+cloudinary.v2.config({
```
timleslie, line 9: This config should happen after all `import` statements.

packages/fields-cloudinary-image/src/test-fixtures.js
```diff
@@ -1,6 +1,247 @@
+const path = require('path');
+require('dotenv').config({ path: path.resolve(__dirname, '../.env') });
```
timleslie, line 2: What's the purpose of this line?

timleslie (changes requested):
A few comments and questions below.

packages/fields-cloudinary-image/src/test-fixtures.js
singhArmani: @timleslie : This is to grab all the env variables defined in  `.env` file. 
Additionally, `{path: ...}` was required to identify the nested `.env` file location as node process was not picking up if we leave the config option object empty.

.github/workflows/tests.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - name: Unit tests
         run: yarn jest --ci --maxWorkers=1
         env:
+          CLOUDINARY_CLOUD_NAME: keystone-cloudinary
+          CLOUDINARY_KEY: 758252673115372
+          CLOUDINARY_SECRET: ${{ secrets.CLOUDINARY_SECRET }}
```
mitchellhamilton, line 59: Hmmm, we're going to need to conditionally run the tests or something given the secret won't be available on PRs from forks.

singhArmani: @mitchellhamilton: That's a valid point. 👍

singhArmani:
Fixing this mess in another branch.

PR #3473, opened 2020-08-25
Add filter and CRUD test for Unsplash field

singhArmani:
As we are using github secrets, we need to look solving the issue when PR is created from a forked repository. For more info please check this [link](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow).

### DIFF
--- a/packages/fields-unsplash/src/Implementation.js
+++ b/packages/fields-unsplash/src/Implementation.js
@@ -261,8 +261,8 @@ const CommonUnsplashInterface = superclass =>
   class extends superclass {
     getQueryConditions(dbPath) {
       return {
-        ...this.equalityConditions(dbPath, ({ unsplashId }) => unsplashId),
-        ...this.inConditions(dbPath, ({ unsplashId }) => unsplashId),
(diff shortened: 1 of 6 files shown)


packages/fields-unsplash/src/Implementation.js
```diff
@@ -265,9 +265,9 @@ const CommonUnsplashInterface = superclass =>
   class extends superclass {
     getQueryConditions(dbPath) {
       return {
-        ...this.equalityConditions(dbPath, ({ unsplashId }) => unsplashId),
-        ...this.stringConditions(dbPath, ({ unsplashId }) => unsplashId),
-        ...this.inConditions(dbPath, ({ unsplashId }) => unsplashId),
+        ...this.equalityConditions(dbPath, val => (val ? val.unsplashId : null)),
+        ...this.stringConditions(dbPath, val => (val ? val.unsplashId : null)),
+        ...this.inConditions(dbPath, val => (val ? val.unsplashId : null)),
```
timleslie, line 270: Nice catch 🚀

timleslie (approved):
Looks good 👍

PR #3476, opened 2020-08-26
Add filter, CRUD tests for LocationGoogle field

singhArmani:


packages/fields-location-google/src/test-fixtures.js
```diff
@@ -1,9 +1,216 @@
...
+            },
+            returnFields: 'name venue { googlePlaceID }',
+          });
+          expect(data).not.toBe(null);
+          expect(data.name).toBe('New location');
+          expect(data.venue.googlePlaceID).toBe(items[0].venue.googlePlaceID);
+        })
+      )
+    );
+  });
```
timleslie, line 215: Should our CRUD tests on this (and other fields) include a `delete` operation?

timleslie (changes requested):

packages/fields-location-google/src/test-fixtures.js
singhArmani: Hmm. That's a good point. I think we missed having `delete` test. I shall go ahead and add those too. 👍

timleslie (approved):
👍

PR #3481, opened 2020-08-26
updating major versions of dependencies

Noviny:
NB: lots of these are sindresorhus packages where the breaking change is to only support node>=10 - since we do this anyway, that's an easy upgrade.


Noviny:
Tests caught a problem I created! 🎉 

This PR is ready. This covers off all the 'low hanging fruit' updates. Post-this is going through categories of packages (jest, babel, gatsby etc) and upgrading those, as well as updating/discussing some packages that have more complicated upgrade paths.

.changeset/bright-gifts-return.md
```diff
@@ -0,0 +1,28 @@
...
+'@keystonejs/keystone': patch
+'@keystonejs/utils': patch
+'@keystonejs/cypress-project-access-control': patch
+'@keystonejs/cypress-project-basic': patch
+'@keystonejs/cypress-project-client-validation': patch
+'@keystonejs/cypress-project-login': patch
+'@keystonejs/website': patch
+---
+
+Update dependencies through a major version - this shouldn't require change by consumers
```
timleslie, line 28:
```suggestion
Updated dependencies through a major version - this shouldn't require change by consumers.
```

Out rule for changesets is "past tense, full sentences" 👍

timleslie (approved):

PR #3532, opened 2020-08-31
Consolidate field filter tests

timleslie:
This PR consolidates all the filter tests into a single module which captures the various different patterns which fields can support. This makes it more clear which fields support value equality vs null equality, for example. It also makes our tests more comprehensive. Most of our fields did not have tests for `null` equality, for example.

Adding these tests exposed bugs in the `Decimal`, `MongoId`, and `Uuid` types.


singhArmani (approved):
This all looks great to me 🎉

PR #3555, opened 2020-09-01
update website ia

Noviny:


Noviny:
Things to check out:
All navGroups (but not subgroups) are a link, and have a page which lists things related to that navGroup.
blog code mostly merged back into what it branched out of.
The `/documentation` page is just things I decided to do - needs review
I don't like the home pages for the sections but they're not... the worst...

Thing we need to figure out before merging (why this is still WIP) - the page for `/api` is just completely unacceptable

Noviny:
Changes that this PR represents:

1. The keystone landing page has a link for `docs` and a link for `blog`
2. There is a `/documentation` page in keystone - this outlines our philosophy of docs, and links to the different website sections
3. Each top level topic in the sidebar has a 'home page' - for everything other than 'API', this is currently mostly a duplicated version of the blog page (with slight specificities filed off) - for API it's a unique thing
4. In the sidebar, sections which you are not in are truncated, with a `see more` link - this only occurs if there would be hidden items.

Noviny:
The home pages are not ideal - but I want to be able to ship this and move on, rather than blocking it. The changes to the sidebar necessitate the home pages existing.

Noviny:
This also recombines 'create pages' and 'create blog post pages'

timleslie (approved):
This all looks much better. Let's ship and iterate on anything else that needs changing 👍

loklaan:
Hey there! If you have the brainspace @Noviny I was hoping to get an understanding of what the following change is achieving:

> Changes that this PR represents:
> 
> ...
> **4. In the sidebar, sections which you are not in are truncated, with a `see more` link - this only occurs if there would be hidden items.**

As someone who regularly traverses the docs in a nonlinear way, this has introduced delay in navigation as well as a fair bit of cognitive load.

eg. If I'm looking for a page:
1. I read through truncated items quickly for the page
2. If the page I'm looking for isn't it isn't one of those, I'll usually click through a few "See more..." items
3. If I don't find it quickly, I'll forget where I've been, as the page loads reset the other "See more..."s as well as the scroll positions
4. Then I try and use the page's Search :)
5. Forgot what I was doing in code, before looking through docs for a small reference _(not related to docs, just a point of reference for cognitive load)_

Ah sorry this comes off really complainy!

I'm hoping that if these changes were open to being iterated on, we could the sidebar back to being untruncated, or re-imagine a reduction of items in the side bar further (looking around, [some](https://reactjs.org/docs/getting-started.html) [other](https://nextjs.org/docs/getting-started) library/framework documentation sidebars take the approach of Accordions where the group titles are just for clicking not navigating).

Noviny:
Hi @loklaan - super happy to have iteration on here. The big problems we were trying to solve were:

1) too many items hiding important things
2) `API` got pushed below the fold

Happy if you want to revert the change to collapsing for now (though I'd like to leave the blog post doing it) while we work on accordianing (or add accordianing yourself)

The original reason not to do accordianing was that it has accessibility issues, which tbh I'm not sure how to solve.

loklaan:
Thanks for getting back Ben, and congrats to the team on the blog, btw 🎉

I'll open a PR as I get around to looking at this 👍

PR #3565, opened 2020-09-02
Add more tests for 1:1 create operations

timleslie:


docs/blog/field-types.md
```diff
@@ -44,17 +44,17 @@ Although technically similar to the `Text` field, conceptually, `Markdown` repre
 This completes the basic criteria of how we decide on new field types over config options or project specific customsation:
 
 1. Is data in the Database or GraphQL a different type?
-1. Is it conceptually a different field type for users?
-1. Is the UI incompatible with other values using the same type?
+2. Is it conceptually a different field type for users?
```
MadeByMike, line 47: 1. 
1.
1. 

Is the correct format for ordered lists in markdown. It's done this way so you never need to change the numbers when you reorder.

timleslie: You'll have to take it up with `yarn format`, I'm just doing whatever it does 🤷‍♂️

timleslie: (Which is to say I agree with you, but if we're going to go with that pattern, we need to update our tooling to make it the same everywhere)

MadeByMike (approved):

docs/blog/field-types.md
MadeByMike: It's it's prettier no drama

PR #3577, opened 2020-09-03
Allow field tests to iterate through different configs

timleslie:
In particular, this will let us apply all of our tests to the different types of `Select` field that exist.

### DIFF
--- a/packages/fields-location-google/src/test-fixtures.js
+++ b/packages/fields-location-google/src/test-fixtures.js
@@ -10,13 +10,11 @@ export const name = 'Location';
 export { LocationGoogle as type };
 export const supportsUnique = false;
 export const skipRequiredTest = false;
-export const exampleValue = 'ChIJOza7MD-uEmsRrf4t12uji6Y';
-export const exampleValue2 = 'ChIJ_9gDOjmuEmsRB7WRXm_Y6o8';
+export const exampleValue = () => 'ChIJOza7MD-uEmsRrf4t12uji6Y';
+export const exampleValue2 = () => 'ChIJ_9gDOjmuEmsRB7WRXm_Y6o8';
 export const fieldName = 'venue';
 export const subfieldName = 'googlePlaceID';
-export const fieldConfig = {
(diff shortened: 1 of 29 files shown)


packages/fields-location-google/src/test-fixtures.js
```diff
@@ -10,13 +10,11 @@ export const name = 'Location';
...
-export const exampleValue = 'ChIJOza7MD-uEmsRrf4t12uji6Y';
-export const exampleValue2 = 'ChIJ_9gDOjmuEmsRB7WRXm_Y6o8';
+export const exampleValue = () => 'ChIJOza7MD-uEmsRrf4t12uji6Y';
+export const exampleValue2 = () => 'ChIJ_9gDOjmuEmsRB7WRXm_Y6o8';
 export const fieldName = 'venue';
 export const subfieldName = 'googlePlaceID';
-export const fieldConfig = {
-  googleMapsKey: process.env.GOOGLE_API_KEY,
-};
+export const fieldConfig = () => ({ googleMapsKey: process.env.GOOGLE_API_KEY });
```
singhArmani, line 17: should we have default key value for google map too? @timleslie

tests/api-tests/fields/filter.test.js
```diff
@@ -8,7 +8,7 @@ const testModules = globby.sync(`packages/**/src/**/test-fixtures.js`, {
 });
 testModules.push(path.resolve('packages/fields/tests/test-fixtures.js'));
 
-multiAdapterRunners().map(({ runner, adapterName }) =>
+multiAdapterRunners('knex').map(({ runner, adapterName }) =>
```
singhArmani, line 11: do we need this only for `knex`?

singhArmani (changes requested):

packages/fields-location-google/src/test-fixtures.js
timleslie: It's not required to get tests to run, so no need for it.

singhArmani (approved):
This looks good now 👍

PR #3595, opened 2020-09-04
Document field

singhArmani:
**TODO**:

- [x] Write API tests
- [x] Fix rendering issue in the Admin UI


packages/fields-document/CHANGELOG.md
```diff
@@ -0,0 +1,277 @@
+# @keystonejs/fields-markdown
```
Noviny, line 1: Just a note - this file should probably be deleted as it's not relevant to this fields-document

Noviny:
createPost view doesn't render:

image

singhArmani:
> createPost view doesn't render:
> 
> image

@Noviny : I am writing Document Field API tests, and hopefully, I could fix it then.

packages/fields-document/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "dependencies": {
+    "@arch-ui/fields": "^3.0.1",
+    "@arch-ui/icons": "^0.1.1",
+    "@arch-ui/input": "^0.1.9",
+    "@arch-ui/theme": "^0.0.11",
+    "@arch-ui/tooltip": "^0.1.12",
+    "@arch-ui/typography": "^0.0.18",
+    "@babel/runtime": "^7.11.2",
+    "@emotion/core": "^10.0.35",
+    "@keystonejs/build-field-types": "^5.2.7",
```
timleslie, line 19:
```suggestion
```

This package no longer exists.

packages/fields-document/src/test-fixtures.js
```diff
@@ -0,0 +1,85 @@
...
+
+export const supportedFilters = [
+  'null_equality',
+  'equality',
+  'equality_case_insensitive',
+  'in_empty_null',
+  'in_value',
+];
+
+// Document field type doesn't support `start_with` and `ends_with`
```
timleslie, line 52: Hmm, I wonder if we want to be supporting the other string operations if we don't support these two? I'd like to have a discussion about this before we ship this.

timleslie (changes requested):

examples/content/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@keystonejs/demo-project-content",
+  "description": "An example KeystoneJS project showcasing Document Field type.",
+  "private": true,
+  "version": "5.1.8",
```
timleslie, line 5:
```suggestion
  "version": "0.0.0",
```

packages/fields-document/src/test-fixtures.js
singhArmani: This is because a document field value, being a string though, needs to have certain schema [interface](https://docs.slatejs.org/concepts/01-interfaces). 
```ts
interface Element {
  children: Node[]
}
```
And in our GraphQL filtering, we may be only interested in the text content which is: 
```ts
interface Text {
  text: string
}
```
So, given a unique schema,`contains` provides a value while filtering whereas `starts_` and `ends_` variants won't make much of a sense.

JedWatson:
Closing this because we've brought it forward in a different branch, and will open a new PR shortly

PR #3600, opened 2020-09-05
fix vulnerabilities with npx npm-check-updates -u and npm audit fix

arthurtalkgoal:


arthurtalkgoal:
We cannot pass the sontaype nodejs scan, and found that keystone created a lot of high vulnerabilities. This fix solve almost all the vulnerabilities.

timleslie:
Hi @arthurtalkgoal, unfortunately we can't just wholesale update our dependencies, as the major breaking changes may require changes to the keystone code, as well as possibly causing breaking changes to keystone itself.

If you have specific security concerns about a dependent package which is exploitable via keystone then please open a separate issue indicating the dependency chain and details of the vulnerability, then we can focus on making the changes required for that specific dependency chain.

PR #3601, opened 2020-09-06
correct SSG deleteItems parameter doc

gautamsi:
there is no `itemId` field, it is `items` filed in input

doc changes, no changeset needed.


timleslie (approved):
Nice catch 👍

PR #3616, opened 2020-09-07
Fix bug with count null queries on 1:1 rels

timleslie:
Adds tests for, and fixes a bug with, queries of the form `_allFooMeta(where: { bar_is_null: true }) { count }`. The result is a nicer general pattern for performing `count` queries, which simplifies `_addWheres` and `_addJoins` by not needing to worry about the `meta` flag any more.


singhArmani (approved):
Looks good to me 👍 . Well done 🎉

PR #3625, opened 2020-09-09
Update ID field API test

singhArmani:
**ID** field is a scalar type (unique identifier) and requires id to be
re-fetched from the server to perform filtering tests.

For this reason, this commit when applied will:
- Add `skipCommonFilterTest` flag to stop running the `common` filter
  tests for ID field
- Remove the `skipCrudTest` flag from **filter** tests, which was filtering it out before running tests.


timleslie (approved):
👍

PR #3628, opened 2020-09-10
upgrade manypkg and add scripts as hints

Noviny:
This is to help you with your Monday release. Both commands are documented but happy to chat about them if you want.


timleslie (approved):
👍

PR #3629, opened 2020-09-10
Remove native type mapping

timleslie:
This functionality allowed devs to writing fields like `name: { type: String }` and have this automagically map from `String` to `Text`. This gives us an API with two different ways to spell the same thing, with no real usage benefits. This PR removes the mapping, and forces devs to explicitly name the field type they want. This will remove ambiguity and increase consistency in schema definition. This is a `major` change, but shouldn't impact too many users, and the fix is a trivial replacement.

### DIFF
--- a/.changeset/giant-brooms-drop.md
+++ b/.changeset/giant-brooms-drop.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': major
+---
+
+Removed the automatic mapping of native types to keystone field types when defining fields. Keystone will no longer convert `{ type: String }` to `{ type: Text }`, `{ type: Number }` to `{ type: Float }`, or `{ type: Boolean }` to `{ type: Checkbox }`.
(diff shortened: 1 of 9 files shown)


mitchellhamilton (commented):
Yesssssssssss!!!!!

How about we throw a nice error telling people to use the appropriate field type though? (unless we think basically no one will have been using it?)

timleslie:
> How about we throw a nice error telling people to use the appropriate field type though?

If a user supplies an invalid `type` it gets caught just below where we removed this mapping, and they'll receive:

```
      if (typeof adapters === 'undefined' || Object.entries(adapters).length === 0) {
        throw `The type given for the '${this.key}.${fieldKey}' field doesn't define any adapters.`;
      }
```

We could finesse this error message a bit, but I don't think there's much to be achieved there 🤷

mitchellhamilton (approved):

PR #3630, opened 2020-09-10
Deprecate @keystonejs/logging

timleslie:
This module provides a single line of code, which does not warrant an entire package of its own. This PR performs a `major` version bump with a deprecation notice in the changeset and adds the deprecation notice to the README so that any users of  the package are aware of its status. Once these are published to `npm` we can consolidate the one line of code into the core `keystone` package where needed and remove this package from the monorepo.


mitchellhamilton:
I don't understand the point of this being separate to actually moving the stuff(similarly for mongo-join-builder). Why not move the stuff and change this package so it throws an error when someone requires it like we did for the field moving stuff?

timleslie:
> I don't understand the point of this being separate to actually moving the stuff(similarly for mongo-join-builder). Why not move the stuff and change this package so it throws an error when someone requires it like we did for the field moving stuff?

My thinking is that I want the bumped package to be a valid, usable package, for those people who might want to continue using it. If the current package works for someone, then it would be a (very mild) PITA to have an outdated package that you couldn't upgrade, because the next version only throws errors.

mitchellhamilton (approved):

PR #3638, opened 2020-09-11
Allow running tests for all available adapters

singhArmani:
At the moment, we are running tests for `knex` only.
When applied, this commit, will make sure to cover all available
adapters.


timleslie (approved):
Good catch!

PR #3652, opened 2020-09-14
Noviny/fix 3431

Noviny:
Maybe the `todo-knex` app should be its own PR

Also, would love to discuss both the solution, and whether/how this should be tested.


packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -680,7 +680,10 @@ class QueryBuilder {
   }
 
   _getQueryConditionByPath(listAdapter, path, tableAlias) {
-    const dbPath = path.split('_', 1);
+    const dbPath = path
+      .split('_')
+      .filter(a => a)
+      .join('_');
```
Noviny, line 686: The real goal here is to filter out `_somethingMeta` to be `somethingMeta`- not sure if this fix is too arcane and too cute.

packages/adapter-knex/lib/adapter-knex.js
```diff
@@ -680,8 +680,13 @@ class QueryBuilder {
   }
 
   _getQueryConditionByPath(listAdapter, path, tableAlias) {
-    const dbPath = path.split('_', 1);
-    const fieldAdapter = listAdapter.fieldAdaptersByPath[dbPath];
+    let fieldAdapter = listAdapter.fieldAdaptersByPath[path];
+
+    if (!fieldAdapter) {
+      let dbPath = path.split('_').slice(0, -1).join('_');
```
timleslie, line 686:
```suggestion
      const dbPath = path.split('_').slice(0, -1).join('_');
```

Noviny: Is it a lint rule that we use `let` unless it is redefined? If yes, can we enforce this as a lint rule?

timleslie: Not a rule at the moment, but definitely one I'd like, as that's what we aim for in practice 👍

Noviny:
@timleslie I thoroughly don't understand all the test failures but am jumping off this problem for tonight. If you know secrets that will accelerate me when I get back to this, lmk

.changeset/swift-items-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/adapter-knex': patch
+---
+
+Fix issue 3341, allowing fieldnames to include `_`, so that field names with an underscore in them can be filtered on
```
timleslie, line 5:
```suggestion
Fixed issue 3341, allowing fieldnames to include `_`, so that field names with an underscore in them can be filtered on.
```

timleslie (approved):
Looks good 👍

PR #3658, opened 2020-09-14
Noviny/renovate gatsby monorepo

Noviny:
This adds extra complexity and tech debt :( it is the way forward with gatsby.

@timleslie I couldn't remember if I was allowed to push to the renovate branch or not


Noviny:
Currently got this to "works on my machine"

😡

timleslie:
Let's leave sleeping dogs lie and live with the pinned gatsby versions until we have a more detailed plan of attack for the website.

PR #3660, opened 2020-09-14
incremental blog improvements

Noviny:
Things that were done here:

1. Author appears again on the home page
2. Title is the same in all locations
3. Author and date appear on the blog view.


timleslie (approved):
Looks good 👍

PR #3662, opened 2020-09-14
Removed performance monitoring guide

Vultraz:
This page is very out of date and the [Apollo Studio docs](https://www.apollographql.com/docs/studio/getting-started/) are much more comprehensive.

- Apollo Engine was renamed to Graph Manager and now to Apollo Studio.
- `ENGINE_API_KEY` is now `APOLLO_KEY`.
- The `privateVariables` Apollo Server config option it mentions is will be deprecated in server 3.
- It refers to a long-removed `WebServer` object. Nowhere else is this present in the docs anymore.
- You no longer need to manually upload your schema since there's a new `reportSchema` config option as of server 2.15 (we use 2.17), so half of this page is moot anyway.

It doesn't seem worth keeping around, especially with Apollo's own docs out there.


timleslie (approved):
Yeah, this is definitely out of date and providing negative value to anyone who happens to find it. Let's get rid of it 👍

PR #3663, opened 2020-09-14
Also enabled schema tracing if APOLLO_KEY is set

Vultraz:
I'm not actually sure if the `tracing` option is needed after 2.15 (the docs aren't clear), but since `APOLLO_KEY` is meant to be used instead of `ENGINE_API_KEY` as of 2.13, it makes sense that anything the latter toggled should also be toggled by the former. I also removed the explicit setting of `engine.apiKey` to keep the code cleaner. Apollo Server grabs the key from either env variable if set and handles the appropriate deprecation message.

Additionally, while you're now supposed to set `engine.reportSchema` to `true` (as of 2.15), I figured it better to have the user set that manually themselves in the GraphQL App config instead of having Keystone magically do it behind the scenes.


timleslie (approved):
Nice change 👍

PR #3664, opened 2020-09-14
Website: made Show More links stand out a bit

Vultraz:
As says


Noviny (approved):
🙏 this looks good - not sure about the test failures but happy to merge when that's stable

PR #3675, opened 2020-09-15
Utilized Octicons size prop (fixes minor icon size issues)

Vultraz:
This fixes a few minor visual issues, notable the user icon being too small and the info using the smaller, thicker icon rather than the thinner, larger one:

Before:
image

After:
image


timleslie (approved):
Looks good 👍

PR #3684, opened 2020-09-17
Make GraphQL id field non-nullable on output types

mitchellhamilton:
The `id` field should never be null so let's make that clear in the GraphQL schema


.changeset/neat-impalas-drive.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/adapter-knex': major
+'@keystonejs/adapter-mongoose': major
```
timleslie, line 3: Do these packages need bumps?

mitchellhamilton: Yes because they depend on `@keystonejs/fields` for the default id fields so the GraphQL schema they generate has changed. (they wouldn't if we didn't want this to be a major but because we want to, these need to have bumps)

timleslie (approved):
Looks good 👍

molomby:
This should actually hit field types like string, int, etc. too right? They can be `isPrimaryKey = true` too

molomby (changes requested):
This should actually hit field types like string, int, etc. too right? (eg. all "simple" field). They can be `isPrimaryKey = true` too. Kinda weird if we default nullability based on `isPrimaryKey` for some field types and not others.

mitchellhamilton:
@molomby given that none of the fields outside of MongoId, Uuid and AutoIncrement currently use the ID type when they're the primary key and we don't have any tests for using them as primary keys AFAIK, how about we not tie using other "simple" fields as ids with this PR?

molomby (approved):
Yeah sure, didn't meant to block. Would like to come back and support `text` at least but that's a problem for another day.

molomby:
(For my own future reference.. https://thinkmill.slack.com/archives/C0107LF95QX/p1600326707421600)

PR #3686, opened 2020-09-17
New blog post: 2020 September: Roadmap update

JedWatson:
Adding the blog post for our September 2020 Roadmap update 🎉


Noviny (approved):

JedWatson:
Note: I've added sort order to blog posts as a workaround for now, since we can't just sort by date or the posts end up in the wrong order (and this is a global setting across the site

PR #3696, opened 2020-09-18
Disable cloudinary API tests

timleslie:


mitchellhamilton (approved):

.changeset/tough-boxes-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/fields-cloudinary-image': patch
+---
+
+Disabled APi tests.
```
mitchellhamilton, line 5: IMO this shouldn't be here since consumers don't care about this.

PR #3716, opened 2020-09-21
Adding links to the new interfaces issues & PRs

JedWatson:
We have this tag now, so I've linked to them from the blog post to make following along easier

(Also corrected the tag itself, since we didn't end up hyphenating it)


timleslie (approved):
👍

PR #3730, opened 2020-09-21
New interfaces: Updates to Auth Package

JedWatson:
No major changes; mainly did this while I was testing and reviewing the new `auth` package.

I separated out the templates from the `createAuth` function so it's easier to see the logical flow, and added comments for each function/config being created, because there's a lot going on. It's all a bit easier to parse now.

I also made some small tweaks to the example project and admin ui templates for consistency.


packages-next/auth/src/index.ts
```diff
@@ -4,14 +4,41 @@ import {
...
+
+import { getExtendGraphQLSchema } from './getExtendGraphQLSchema';
 import { initFirstItemSchemaExtension } from './initFirstItemSchemaExtension';
+import { AuthConfig, Auth, ResolvedAuthGqlNames } from './types';
 
+import { signinTemplate } from './templates/signin';
+import { initTemplate } from './templates/init';
+
+/**
+ * createAuth function
```
mitchellhamilton, line 17: I don't get this bit(and the similar ones in the other comments)? If someone hovers over this, they'll see that it's a function named createAuth

JedWatson: tbh I was just trying to create some visual order to the file, when I got there it was really hard to see where the different things were. happy to do this any other way (whether that's better comments, or just using since `/*` so they don't show up as TS docs)

PR #3739, opened 2020-09-22
Add support for Prisma to create-keystone-app

timleslie:
To test this you can run the script locally with:

```sh
 node packages/create-keystone-app/bin/cli.js
```

### DIFF
--- a/.changeset/flat-pans-greet.md
+++ b/.changeset/flat-pans-greet.md
@@ -1,0 +1,5 @@
+---
+'create-keystone-app': minor
+---
+
+Added support for creating projects using the Prisma adapter.
--- a/docs/guides/prisma.md
+++ b/docs/guides/prisma.md
@@ -41,38 +41,14 @@ yarn create keystone-app my-app
 ```
 
 - Call your project `my-app`
-- Select `PostgreSQL` as your database type.
+- Select `Prisma (Experimental)` as your database type.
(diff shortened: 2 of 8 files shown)


JedWatson:
@timleslie just picking default values (more or less) without a database created leads to a hard error. Can we improve this?

(related; I was of the impression Prisma migrate can create the database if/when it needs to, assuming the server (i.e localhost, default port and auth) is available?)

image

Any meaningful fix here is possibly/probably out of scope for this PR, but throwing an error and crashing the process from the default choice seems like a poor experience for first users. We could catch the error, point them at the docs, and offer to continue anyway saying "You'll need to create your database after this" and that would be much nicer.

JedWatson:
Related to that last comment, when I skip the "test database" step, and follow the instructions it errors out without helping me know what next step to take to resolve the problem (spoiler: follow the docs to create a database)

I think we should do better here; we can _suggest_ a command that will probably do what they want; or at minimum point at the guide.

If I was trying Keystone out for the first time here with less knowledge, the current experience would just be a roadblock I'd walk away from.

image

timleslie:
> @timleslie just picking default values (more or less) without a database created leads to a hard error. Can we improve this?

I think in this case we should remove the default database string. At the moment there is no valid default database string which will work. We should structure the instructions such that they know that they need this value right up front.

JedWatson (approved):

PR #3744, opened 2020-09-22
Test AutoIncrement as both an ID and Int graphQL type

timleslie:
We need to make sure that this field behaves correctly when used as both an `ID` and `Int` field in graphQL.

### DIFF
--- a/packages/fields-auto-increment/src/test-fixtures.js
+++ b/packages/fields-auto-increment/src/test-fixtures.js
@@ -4,8 +4,9 @@ import { AutoIncrement } from './index';
 
 export const name = 'AutoIncrement';
 export const type = AutoIncrement;
-export const exampleValue = () => 35;
-export const exampleValue2 = () => 36;
+export const testMatrix = ['ID', 'Int'];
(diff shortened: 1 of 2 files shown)


mitchellhamilton (approved):

packages/fields-auto-increment/src/test-fixtures.js
```diff
@@ -34,19 +38,32 @@ export const initItems = () => {
...
-export const storedValues = () => [
-  { name: 'product1', orderNumber: 1 },
-  { name: 'product2', orderNumber: 2 },
-  { name: 'product3', orderNumber: 3 },
-  { name: 'product4', orderNumber: 4 },
-  { name: 'product5', orderNumber: 5 },
-  { name: 'product6', orderNumber: 6 },
-  { name: 'product7', orderNumber: 7 },
-];
+export const storedValues = matrixValue =>
```
mitchellhamilton, line 41: This feels like a strange name IMO since from reading the test, it's not actually the stored value, it's the value returned from graphql?

PR #3746, opened 2020-09-22
New Interfaces: Implementing signout button in Navigation

JedWatson:


packages-next/admin-ui/src/components/Navigation.tsx
```diff
@@ -64,6 +71,21 @@ const NavItem: FunctionComponent<NavItemProps> = ({ href, children }) => {
   );
 };
 
+const SignoutButton: FunctionComponent = () => {
+  const [endSession, { loading, data }] = useMutation(END_SESSION);
+  useEffect(() => {
+    if (data?.endSession) {
+      top.location.reload();
```
mitchellhamilton, line 78: Why top instead of just window? (non-blocking comment just asking because I feel like a future person would go "I should leave it using top bc that was probably intentional")

Also, I'd probs await endSession() and do the reload after that bc I feel like it's easier to reason about but also it doesn't matter so do whatever you feel like

packages-next/admin-ui/src/components/Navigation.tsx
```diff
@@ -64,6 +71,21 @@ const NavItem: FunctionComponent<NavItemProps> = ({ href, children }) => {
...
+    }
+  }, [data]);
+
+  return (
+    <Button size="small" isLoading={loading} onPress={() => endSession()}>
```
mitchellhamilton, line 83: Not about this PR but can we pls rename onPress to onClick? Calling it onPress just makes things confusing. (if it was actually a special thing, sure let's call it onPress but it's not)

mitchellhamilton (approved):

packages-next/admin-ui/src/components/Navigation.tsx
mitchellhamilton: actually, from seeing #3745, all we need to do is remove the onPress prop.

PR #3751, opened 2020-09-22
Make GraphQL _QueryMeta non-nullable on output types

jordanoverbye:
Related to #3684

In [this method](https://github.com/keystonejs/keystone/blob/master/packages/keystone/lib/providers/listCRUD.js#L45) it looks like Keystone always returns an object with a `count` field, so we should be specifying `_QueryMeta` as non-nullable.

It also looks like [looks like](https://github.com/keystonejs/keystone/blob/06be15471406ed09b0ed79b353500b37d019dc02/packages/keystone/lib/ListTypes/list.js#L500)  that the `count` field always returns an `Int`, so we can also set this field as as non-nullable.

### DIFF
--- a/.changeset/mighty-spoons-worry.md
+++ b/.changeset/mighty-spoons-worry.md
@@ -1,6 +1,4 @@
 ---
-'@keystonejs/adapter-knex': major
-'@keystonejs/adapter-mongoose': major
 '@keystonejs/fields': major
 '@keystonejs/keystone': major
 ---
--- a/.changeset/mighty-spoons-worry.md
+++ b/.changeset/mighty-spoons-worry.md
@@ -1,0 +1,6 @@
+---
+'@keystonejs/fields': major
+'@keystonejs/keystone': major
+---
+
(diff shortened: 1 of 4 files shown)


mitchellhamilton (approved):
👍  from me but I would wait for @timleslie's thoughts given it's a thing we can't really go back on(or rather, we can't go back on without causing problems)

.changeset/mighty-spoons-worry.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystonejs/adapter-knex': major
+'@keystonejs/adapter-mongoose': major
```
mitchellhamilton, line 3:
```suggestion
```
These aren't necessary here, they're in the PR making the id field non-nullable for a specific reason that isn't relevant here

timleslie:
I'm 👍 on this in principle. I think perhaps we should wait for this and #3684 so that we can take a bit of time and review whether there are other parts of our generated graphQL that we want to modify. I'm not sure anyone has capacity in their schedule to do this justice before the end of this week so maybe we can make that a goal for next week.

PR #3754, opened 2020-09-23
Small tweaks to the new Navigation component

JedWatson:
Nothing major, makes use of the new `<Link />` component @mitchellhamilton added recently, and neatens up the authenticate item status


packages-next/admin-ui/src/components/Navigation.tsx
```diff
@@ -86,6 +83,26 @@ const SignoutButton: FunctionComponent = () => {
   );
 };
 
+const AuthenticatedItem: FunctionComponent<{
+  item: { id: string; label: string };
+}> = ({ item }) => {
```
mitchellhamilton, line 88: non-blocking: pls never use the `FunctionComponent` type

JedWatson: 👍 Noted

I copied this from existing usage, will do a blitz on them

PR #3759, opened 2020-09-23
Styling the new List UI

JedWatson:
A long way from complete, but makes the new List UI look more like the current List UI

### DIFF
--- a/packages-next/admin-ui/src/pages/ListPage.tsx
+++ b/packages-next/admin-ui/src/pages/ListPage.tsx
@@ -1,8 +1,9 @@
 /* @jsx jsx */
 
 import { useQuery, gql } from '../apollo';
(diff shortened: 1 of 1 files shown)


packages-next/admin-ui/src/pages/ListPage.tsx
```diff
@@ -14,6 +15,90 @@ type ListPageProps = {
...
 type Filter = { field: string; type: string; value: JSONValue };
 
+const ListPageHeader = ({ listKey }: { listKey: string }) => {
+  const list = useList(listKey);
+  return (
+    <Stack
+      across
+      marginY="large"
+      gap="medium"
+      css={{
```
jossmac, line 25: This makes me think the `Stack` component is incomplete or not the right tool for the job here

PR #3760, opened 2020-09-23
In renovate, skip remark and gatsby repos

Noviny:


timleslie (approved):
👍

.github/renovate.json
```diff
@@ -0,0 +1,8 @@
+{
```
mitchellhamilton, line 1: This should be in the renovate config we already have in the root of the repo

mitchellhamilton (approved):

PR #3768, opened 2020-09-23
Default column name for count differs between Postgres/Mysql-esque

tjbp:
This is a regression introduced by #3616 - the default column name for `count(*)` in MariaDB/MySQL is `count(*)`, not `count` like in Postgres. Thanks!


timleslie:
Hi @tjbp, thanks for the PR. We don't currently support MariaDB or MySQL in the `knex` adapter. Have you been able to get this adapter to work in either of those contexts?

tjbp:
@timleslie yep, appreciate that - I do actually have things up and running in MariaDB with this change, but do understand that it's not supported - I'd hoped this and future PRs might be accepted if they're minor and don't impact support for Postgres.

timleslie:
@tjbp That's great to hear! I know that in the past people have tried to use MySQL and it is a long way away from working. It's good to know that MariaDB works out of the box. I might see if we can add in regression tests to run on MariaDB which would allow us to officially support it.

In the meantime, could I ask you to run `yarn format` on your code (this runs `prettier`) and then add a changeset by running `yarn changeset`. This will make sure that your fix gets included in the next release on Monday morning.

Thanks for the PR, and if you find any other minor issues like this then I'm more than happy to accept PRs 👍

tjbp:
@timleslie awesome, thanks! Haven't tested much beyond the Admin UI so far but will certainly put in future PRs. Changeset added with formatted code.

timleslie (approved):
Thanks for this PR! 🙏

PR #3796, opened 2020-09-26
Quick fix for exposed PassportAuthStrategy SessionList

xakdog:
### Problem

PassportAuthStrategy SessionList is exposed for public. Anyone can read secret key!

### Solution

Quick fix: restrict access to the data, table structure will be still exposed.
Proper solution require read permissions to be 'internal', but it wasn't clear how to implement it fast.


timleslie (approved):
Thank you very much for this PR, @xakdog 🙏

PR #3819, opened 2020-09-28
Update authentication.md

maheraldous:
Error just add this ","


timleslie (approved):
Nice catch, thanks for the PR 👍

PR #3889, opened 2020-10-06
Make unit tests more parallel to increase throughput

timleslie:
A lot of our test time is tied up in a couple of long running tests. This change appears to make things a little bit better. A longer term solution will be to break up/speed up the long running tests (there's definitely a lot of scope for optimisation, but it's also a lot of work). Let's try out this config and see if it a) makes things faster and b) reduces false positives due to memory issues.


JedWatson (approved):
👍 Looks good to me

PR #3924, opened 2020-10-11
Ignore yarn.lock in shouldRunTests

JedWatson:
This ignores changes to yarn.lock in the shouldRunTests check, which should speed up CI for package bumps that only change the new packages

(e.g #3921)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           filters: |
             shouldRunTests:
-              - '!{{website,docs,examples,packages-next,examples-next,design-system}/**,**/*.md}'
+              - '!{{website,docs,examples,packages-next,examples-next,design-system}/**,**/*.md,yarn.lock}'
       - run: echo "::set-env name=SHOULD_RUN_TESTS::$SHOULD_RUN"
         if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


mitchellhamilton:
We can't do this yarn.lock changes could _definitely_ break things.

JedWatson:
@mitchellhamilton 💯 it sure could. But what's the chance that we merge a PR that only changes yarn.lock without also changing any other files in the codebase that would trigger the tests?

I can think of an edge case where we have a PR that _only_ resets the `yarn.lock` and that would skip the tests... which, fair, is something we could do. But to my memory we've never done that, and I didn't change the auto approval paths so it wouldn't get past us anyway.

On a side note I think it probably is worth doing a yarn.lock refresh sometime... but we can roll this back for that since we'd want it to go through CI. 

Given the specific goal of streamlining CI for isolated -next package dep bumps (as per how isolated code changes work), what's the concern?

Is there a loop hope I've missed?

JedWatson:
Hah so just after I wrote that comment I noticed #3805.

So that's a bit of a smack down to my theory 🙂

timleslie:
Yeah, we 💯 can't do this, the tests are the one thing that gives us any feedback at all as to whether or not a yarn.lock change is safe or not.

PR #3929, opened 2020-10-12
Use Emotion 11 in new interfaces

mitchellhamilton:
This is mostly because I want to test Emotion 11(it's already v reliable, don't worry), there's lots of stuff but the big that matters here is improved type defs, it changes the way the css prop is typed from "some prop that you can use on any component" to "you can use it if the component has a className prop AND you've setup the jsx pragma" which is v nice. (the requiring a className prop on the component before allowing the css prop bit found some missing className props on things 🎉 )

### DIFF
--- a/design-system/packages/button/package.json
+++ b/design-system/packages/button/package.json
@@ -10,7 +10,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
-    "@emotion/core": "^10.0.35",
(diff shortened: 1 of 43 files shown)


JedWatson:
Nice, I really like the type improvements. Also very cool to see this, I remember us talking about it _months_ ago `'@emotion/react'` 🎉

JedWatson (approved):

PR #3971, opened 2020-10-17
Update README.md

maheraldous:
I was looking in the docs for a way to upload images via GraphQL and no one really guided me so I tried to do it my way and I wanted to add it to the docs for others to find. Feel free to remove Altair but I really like it's file upload tool.


JedWatson (approved):

JedWatson (approved):
👍 Thanks for this! looks good.

PR #3976, opened 2020-10-18
Deduplicate Babel deps

mitchellhamilton:
I've been doing some perf work on Preconstruct lately and found that sizeable amount of time is spent requiring all of the babel deps because there are lots of duplicates of common dependencies because the lockfile had grown organically so I've refreshed the Babel part of the lockfile here(by removing all the `@babel` deps from the lockfile and running yarn).


timleslie (approved):
👍

PR #4008, opened 2020-10-21
fix(knex-adapter): prepend relation tables with schema name in dropDatabase method

Neonox31:
The `dropDatabase` method doesn't delete relation tables because schema name is not defined.
This P.R simply prepends the schema name to relation tables.

### DIFF
--- a/.changeset/silver-dryers-mate.md
+++ b/.changeset/silver-dryers-mate.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/adapter-knex': patch
+---
+
+prepend relation tables with schema in dropDatabase method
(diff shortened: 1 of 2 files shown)


.changeset/silver-dryers-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/adapter-knex': patch
+---
+
+prepend relation tables with schema in dropDatabase method
```
timleslie, line 5:
```suggestion
Fixed bug in `.dropDatabase()`, relationship join tables are now correctly dropped when using a non-default schema.
```

timleslie (approved):
Great catch, thanks for the PR 🙏

PR #4014, opened 2020-10-22
Rename prisma to prisma_postgresql in multiAdapterRunners

timleslie:
This will change the name of the tests on CI, so this will need to get a ✅ even though some "required" tests appear not to have been run. I'll update the github config once it's into master 👍


molomby (approved):
Makes sense

Issue #4005, opened 2020-10-21
New segmented-control display mode for the Select field breaks with a large number of options


gautamsi:
# New Interface Select Options

I know that this was just first pass in the select options. this is how it looks like currently when you have many options

![image](https://user-images.githubusercontent.com/5769869/96697758-e78b5100-13a9-11eb-98d1-c679aa4e0a65.png)

there are 5-6 more items after scrolling.

@mitchellhamilton

JedWatson:
@gautamsi you need to specifically set the `displayMode` to `segmented-control` to get that to happen, right?

I think we need to find a clear way of conveying that the use case for this new mode is when you only have a small number of options to select from. I'm not sure that a hard technical limit there is the right thing to do (because it's also highly dependent on the length of labels) but that's the intention of how you'd use this option.

With a large number of items, you should be using the default "select" mode.

Having said that, we should totally fix the cropping bug 😅

gautamsi:
I did realise that it was using segmented control in the example. I think I had more gap this time in checking this out. This should flexibly wrap instead of scrolling. I think in prod version we are making a decision to use different select type when there is more than 'x' items. I think we can have that configurable that you can have segmented control but it may revert back to select with large number of items. that large number should be configurable in admin. 

> I use a custom text field which renders select value from different list and I had to use a virtualised list as the number of items were in 100s. Hope i can upgrade that with virtualised option.

There are few other things can be fixed

![image](https://user-images.githubusercontent.com/5769869/96720320-3c3dc480-13c8-11eb-9794-d7cf92a435da.png)

1. select is not properly z-index'ed
2. Relationship is not responsive, it goes off container on shrinking. 

(I know we are still working on it :) )

JedWatson:
Yeah I've noticed the z-index problems, and width of fields / responsiveness is generally a thing we still need to think about holistically.

FWIW because this may not be obvious, we're focusing on **functionality** at the moment with some broad brush strokes when it comes to UX; config, state management, etc are the foundations so we're getting that right first.

Which is also to say we're _deliberately_ not focusing too much on UI design and polish yet, until the foundations are a bit more complete. But also we're going to be onboarding a designer into the project shortly as well which is really exciting 🎉 and I expect we'll circle back to UI polish once they've started establishing a clearer vision for how the new Admin UI should look / navigate / behave etc.

[...]

PR #4027, opened 2020-10-23
Ui tweaks new interface - segmented control wrap and width

gautamsi:
ref #4005 

was able to fix the width as well as wrap on this but not able to fix the relationship field. 

image

tried using token system. (still TODO for theme based width - not my expertise yet)


JedWatson (approved):
Nice work @gautamsi - how the theme handles width does need a bit more review, but you've applied the current pattern perfectly here so whatever changes land in the future should be easy to apply.

Thanks! 🙏🏻

PR #4028, opened 2020-10-23
New Interface Select control zIndex issue fix

gautamsi:
ref #4005 

fixed zIndex issue, there was unnecessary zIndex style.

image


JedWatson (approved):
Thanks @gautamsi

PR #4090, opened 2020-10-30
Move createKeystone to lib

JedWatson:
The `createKeystone` function was left in `classes` when we refactored it from a class to a function. This just moves it into the `lib` folder with the rest of the main keystone functions, which I think makes more sense for people new to the codebase.

(also added a couple of TODOs, not worth an additional PR)

### DIFF
--- a/packages-next/TODO.md
+++ b/packages-next/TODO.md
@@ -24,4 +24,6 @@
 
 ## Next steps after alpha release
 
+- [ ] Review how we're returning errors from the GraphQL API
+- [ ] Review how the search field (and query input) works
 - [ ] Revisit design system and theming
--- a/packages-next/keystone/src/index.ts
+++ b/packages-next/keystone/src/index.ts
@@ -1,1 +1,1 @@
-export { createKeystone } from './classes/Keystone';
+export { createKeystone } from './lib/createKeystone';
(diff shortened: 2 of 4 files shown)


gautamsi:
Hope you are planning to move all new keystone customization will merge with original `Keystone` class after preview.

JedWatson:
@gautamsi I'm going to write another blog post about this soon, but short version of the plan is:

- Publish the new packages under the `@keystone-next` scope early next week and keep iterating on them quickly. It should be considered more or less alpha at the moment, and not feature complete with the main keystone packages
- As people try them out & give us feedback, we'll learn a lot and use that for planning
- At some point they'll become feature complete, stable, and better in all ways than the main keystone packages; at that point we'll start recommending people use them for new projects but they'll represent an optional breaking change upgrade for people with existing keystone projects
- Then the main work left will be to ensure test coverage / comprehensive documentation / migration guides and other things we haven't  really thought about much yet
- Eventually (at the moment I expect early next year, but will depend on a lot of factors) we will move the new packages over the current ones and publish them as a major breaking change
- Finally we'll be able to do a lot of cleanup in the old keystone codebase that will otherwise stay around while we're supporting both versions

How feedback goes will determine how much work we put into easing the upgrade, adding things to create better backwards compatibility, etc. Also how many new features get worked on before we transition back to just one set of Keystone packages.

Just to make sure I'm clear about what you're saying, that means the new `@keystone-next/keystone` **will** replace the current `@keystonejs/keystone` as a major version after preview (along with `-fields`, etc)

However the new customisation and features we've been working on **will not** be back-ported into the current keystone packages, because they're all deeply coupled to the new work we've done and we're focusing all our effort on the transition.

Since the current `@keystonejs/keystone` package is still the core of `@keystone-next/keystone` (which you can think of as a new wrapper, with a new build process / cli and server implementation) it means we haven't forked any of the implementation we don't need to, and any new features that can be (e.g our new Prisma support) are being developed for **both** sets of packages.

Hope that clears it up?

PR #4139, opened 2020-11-05
Stop using deprecated Actions feature

mitchellhamilton:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


timleslie (approved):
👍

PR #4164, opened 2020-11-08
[POC] Cleanup of the CloudinaryImage field type

timleslie:
A quick sketch of some changes we discussed the other day. tl;dr is that this decouples the `CloudinaryImage` field type from the `File` field type. May or may not be worth doing? 🤷


gautamsi:
what is the purpose of keeping adapter in here? So that we can pass more options in adapter constructor??

can they not be merged with field config? this way there could be s3 field without the adapter. 

I was also wondering what is the difference in file field with cloudinary adapter and separate cloudinary field? (ignore the fact that there is no file field in next)

PR #4179, opened 2020-11-10
Fix defaultValue types for autoIncrement and integer fields

JedWatson:
These were string, should be number.

### DIFF
--- a/.changeset/poor-pants-tan.md
+++ b/.changeset/poor-pants-tan.md
@@ -2,4 +2,4 @@
 "@keystone-next/fields": patch
 ---
 
-Fix defaultValue types for autoIncrement and integer fields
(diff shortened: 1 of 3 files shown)


.changeset/poor-pants-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-next/fields": patch
+---
+
+Fix defaultValue types for autoIncrement and integer fields
```
timleslie, line 5:
```suggestion
Fixed defaultValue types for `AutoIncrement` and `Integer` fields.
```

Past tense, full sentence. Leads to better changelogs :-)

JedWatson: Thanks

timleslie (approved):
👍

PR #4185, opened 2020-11-10
Fixed field defaultValue types for new fields package

JedWatson:
These types were wrong, noticed when I was working on the new tracking fields example.

@mitchellhamilton really want you to take a look at what I've done here (after merge, none of it should be problematic 🤞)

Specifically, the big thing I'm fixing here is that the `defaultValue` config for fields can be a plain value (of a type specific to the field type) or a function that returns that type.

### DIFF
--- a/packages-next/fields/src/types/timestamp/index.ts
+++ b/packages-next/fields/src/types/timestamp/index.ts
@@ -2,13 +2,13 @@ import { DateTimeUtc } from '@keystonejs/fields';
 
 import type { FieldConfig } from '../../interfaces';
 import type { FieldType } from '@keystone-next/types';
-import type { BaseGeneratedListTypes } from '@keystone-next/types';
+import type { BaseGeneratedListTypes, FieldDefaultValue } from '@keystone-next/types';
 import { resolveView } from '../../resolve-view';
 
 export type TimestampFieldConfig<TGeneratedListTypes extends BaseGeneratedListTypes> = FieldConfig<
   TGeneratedListTypes
 > & {
-  defaultValue?: string;
+  defaultValue?: FieldDefaultValue<string>;
   isRequired?: boolean;
   isIndexed?: boolean;
   isUnique?: boolean;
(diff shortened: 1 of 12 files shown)


nathsimpson (approved):
👍

packages-next/types/src/schema/hooks.ts
```diff
@@ -77,7 +77,14 @@ type ResolveInputHook<TGeneratedListTypes extends BaseGeneratedListTypes> = (
 ) =>
   | Promise<TGeneratedListTypes['inputs']['create'] | TGeneratedListTypes['inputs']['update']>
   | TGeneratedListTypes['inputs']['create']
-  | TGeneratedListTypes['inputs']['update'];
+  | TGeneratedListTypes['inputs']['update']
+  // TODO: I'm pretty sure this is wrong, but without these additional types you can't define a
+  // resolveInput hook for a field that returns a simple value (e.g timestamp)
```
mitchellhamilton, line 82: yeah, this is def wrong. silly me for trusting the docs instead of looking at the code when writing this originally

packages-next/fields/src/types/timestamp/index.ts
```diff
@@ -2,13 +2,13 @@ import { DateTimeUtc } from '@keystonejs/fields';
...
 import type { FieldType } from '@keystone-next/types';
-import type { BaseGeneratedListTypes } from '@keystone-next/types';
+import type { BaseGeneratedListTypes, FieldDefaultValue } from '@keystone-next/types';
 import { resolveView } from '../../resolve-view';
 
 export type TimestampFieldConfig<TGeneratedListTypes extends BaseGeneratedListTypes> = FieldConfig<
   TGeneratedListTypes
 > & {
-  defaultValue?: string;
+  defaultValue?: FieldDefaultValue<string>;
```
mitchellhamilton, line 11: Not a comment about this but defaultValue should probably live in the FieldConfig since it works for every field.

PR #4186, opened 2020-11-10
Improving the appearance of readonly fields in the Admin UI

JedWatson:
This tweaks the styling in the theme for disabled fields, and fixes the display of simple fields when they're in readonly mode

### DIFF
--- a/packages-next/fields/src/types/integer/views/index.tsx
+++ b/packages-next/fields/src/types/integer/views/index.tsx
@@ -15,15 +15,18 @@ import {
 export const Field = ({ field, value, onChange, autoFocus }: FieldProps<typeof controller>) => (
   <FieldContainer>
     <FieldLabel>{field.label}</FieldLabel>
-    <TextInput
-      autoFocus={autoFocus}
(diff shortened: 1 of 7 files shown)


packages-next/fields/src/types/integer/views/index.tsx
```diff
@@ -15,15 +15,18 @@ import {
...
+      <TextInput
+        autoFocus={autoFocus}
+        type="number"
+        onChange={event => {
+          onChange(event.target.value.replace(/\D/g, ''));
+        }}
+        value={value}
+      />
+    ) : (
+      value
```
nathsimpson, line 28: Does this need to be in a div?

JedWatson: Not at the moment, it's consistent with other fields for now.

Ultimately I do think we should add a wrapper of some sort to readonly output (including the Virtual field) but am going to leave that for a separate design session to work out.

nathsimpson (approved):
👍

PR #4187, opened 2020-11-10
Improve readonly relationship views

JedWatson:
This is a fairly big update to the Relationship field, to render readonly views as buttons instead of in selects.

Still needs some more work for large datasets, but in the meantime it's a big improvement.

image


nathsimpson (approved):
👍

PR #4189, opened 2020-11-10
Add new todo example (WIP)

JedWatson:
This adds a new example project, with a todo list that can be assigned to users.

It's slightly more complex than most todo examples, but I think there's a bunch of useful stuff we can do here including different relationship views, tacking fields, and hooks.

I'd like to add more features to it and use it as a separate use case for testing, and the tracking fields need to be pulled out into their own package, but it's a good start.


nathsimpson (approved):
👍

PR #4206, opened 2020-11-11
Fixing menu cropping in Drawers for multi select

JedWatson:
Same fix as #4204, I missed the `MultiSelect` variant


JedWatson:
Note: doesn't need a changeset because it'll be released with the patch in #4204

PR #4216, opened 2020-11-11
Admin UI Tinkering

jossmac:
Miscellaneous tweaks and tidying

image

### DIFF
--- a/packages-next/fields/src/types/relationship/views/index.tsx
+++ b/packages-next/fields/src/types/relationship/views/index.tsx
@@ -14,13 +14,9 @@ import {
   ListMeta,
 } from '@keystone-next/types';
 import { Button } from '@keystone-ui/button';
-<<<<<<< HEAD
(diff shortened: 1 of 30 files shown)


packages-next/fields/src/types/relationship/views/index.tsx
```diff
@@ -43,18 +49,15 @@ function LinkToRelatedItems({
           .map(({ id }) => id)
           .join(',')}"`}
       >
-        View List of Related Items
+        View related items
       </Button>
     );
   }
+
+  // QUESTION: why is the target `_blank` here?
```
mitchellhamilton, line 57: I believe it's because that's how it is in the current Admin UI - I don't have opinions on whether it should be target _blank or not

The reason it's not using the router Link is because it's target _blank so it wouldn't do anything

mitchellhamilton (approved):
😍

PR #4224, opened 2020-11-12
Fix lazy admin meta being in loading state in Safari after restarting…

mitchellhamilton:
… the dev server but not reloading the page

### DIFF
--- a/packages-next/admin-ui/src/pages/ItemPage/index.tsx
+++ b/packages-next/admin-ui/src/pages/ItemPage/index.tsx
@@ -311,7 +311,9 @@ export const ItemPage = ({ listKey }: ItemPageProps) => {
   let { data, error, loading } = useQuery(query, {
     variables: { id, listKey },
     errorPolicy: 'all',
+    skip: id === undefined,
   });
+  loading ||= id === undefined;
 
   const dataGetter = makeDataGetter<
     DeepNullable<{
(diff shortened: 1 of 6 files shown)


packages-next/admin-ui/src/pages/ItemPage/index.tsx
```diff
@@ -311,7 +311,9 @@ export const ItemPage = ({ listKey }: ItemPageProps) => {
   let { data, error, loading } = useQuery(query, {
     variables: { id, listKey },
     errorPolicy: 'all',
+    skip: id === undefined,
```
mitchellhamilton, line 314: This is unrelated but I found it when debugging this. For some reason, Next renders the page without the query object filled so there is an unnecessary request that fails because the id isn't there

packages-next/admin-ui/src/utils/useLazyMetadata.tsx
```diff
@@ -27,7 +27,7 @@ export function useLazyMetadata(
   visibleLists: VisibleLists;
   createViewFieldModes: CreateViewFieldModes;
 } {
-  let result = useQuery(query, { errorPolicy: 'all' });
+  let result = useQuery(query, { errorPolicy: 'all', fetchPolicy: 'network-only' });
```
mitchellhamilton, line 30: I don't really understand why caching this breaks things but not caching this is basically totally fine because this will never be unmounted so the cache isn't doing anything

JedWatson (approved):
Looks great 👍 thanks @mitchellhamilton, this is going to make a big difference

PR #4225, opened 2020-11-12
Fix dropdown in nav in Firefox

mitchellhamilton:
Not sure why the width and height were there, it looks like Safari and Chrome ignore the height so it looks normal there but Firefox respects it so it squashes all the content


jossmac (approved):
Copy/paste, sorry about that

PR #4229, opened 2020-11-12
Fix margin between toolbar and form in Create modal

JedWatson:
The toolbar was hard up against the form components, so I gave it some margin. cc/ @jossmac 

image


jossmac:
I don't think this is the right solution.  The "body" needs padding, so scrollable views appear flush with the footer
image

PR #4230, opened 2020-11-12
Add Cross-Origin Resource Support to the dev server

JedWatson:
This adds `core` middleware to the new dev server, with the same defalults keystone had previously.

We should make this configurable, but for now it will unblock development against the API.

cc/ @wesbos


wesbos (commented):
looks goods - agreed we need to be able to configure cors

wesbos (approved):

PR #4236, opened 2020-11-12
Remove unnecessary functions calls

timleslie:


JedWatson:
@timleslie not obvious from the implementation yet I guess but there are a few reasons this exists

* we'll have other scripts (e.g `build`, `start`) alongside `dev` that will want to share the "init keystone from config" logic
* the CLI should support flags for options like where to find the config file, e.g `keystone-next --config mykeystoneconfig.ts` and that's an example of something that would be implemented in the `initKeystoneFromConfig` method

I'd prefer to get around to implementing at least that flag, and/or today at least add a comment to that file, rather than removing it altogether

JedWatson (approved):

PR #4239, opened 2020-11-13
Add support for onConnect in new interfaces

timleslie:


packages-next/types/src/index.ts
```diff
@@ -54,6 +54,7 @@ export type KeystoneConfig = {
   db: {
     adapter: 'mongoose' | 'knex';
     url: string;
+    onConnect?: (keystone: Keystone) => any;
```
mitchellhamilton, line 57: `Keystone` refers to a different thing here which is not what the argument to onConnect is

nathsimpson (approved):

packages-next/types/src/index.ts
timleslie: Oh, right... 🤦

timleslie: Can we maybe pick a different name for that thing then? Not for this PR obviously, just flagging it as a thing for the future.

mitchellhamilton (approved):

gautamsi:
Can we make onConnect an array so that plugins in future can also be able to push to onConnect for their own keystone related code. Like extra seeding for the aux list or regular lists.

PR #4241, opened 2020-11-13
Expanded the defaults for labelField to include "label" and "title"

JedWatson:
At the moment, the `labelField` option on lists defaults to the `name` field if there's one defined, and failing that will fall back to `id`.

This is mainly because that's how it's always worked in Keystone and we've never changed it.

Increasingly, I find myself specifying `ui: { labelField: 'label }` on lists which ... just makes no sense, since it's called the *label* field. Why default to "name" but not "label"?

"title" is also (in the context of a CMS, where _posts_ have _titles_) a really common choice.

This PR adds `label` and `title` as default fields to use before we fall back to `id`, and I think with fresh eyes that makes a lot more sense than just `name`. I've left the order as prioritising name first, so that anyone upgrading a list with both fields present doesn't get surprised (unlikely but possible)


packages-next/keystone/src/lib/createKeystone.ts
```diff
@@ -70,8 +70,16 @@ export function createKeystone(config: KeystoneConfig): Keystone {
...
     const labelField =
-      (listConfig.ui?.labelField as string | undefined) ?? (listConfig.fields.name ? 'name' : 'id');
+      (listConfig.ui?.labelField as string | undefined) ??
+      (listConfig.fields.name
+        ? 'name'
+        : listConfig.fields.label
+        ? 'label'
+        : listConfig.fields.title
+        ? 'title'
+        : 'id');
```
timleslie, line 82: Can we hoist this logic up into core Keystone instead of having it here?

jesstelford:
What happens if someone is using `label` as a synonym for `tag` which is a `Relationship`? That could potentially be a breaking change as the query will not work (it doesn't select any sub fields of the `label` field).

packages-next/keystone/src/lib/createKeystone.ts
JedWatson: Nope, because core Keystone doesn't have a concept of a `labelField` option. It has a label resolver, which we could update to match this logic as well, but that's a breaking change for the current system (and potentially a big one, to change the default behaviour of the `_label_` resolver) so imo that's out of scope for this PR.

JedWatson:
@jesstelford we need to add validation to this based on the type of field specified for the label field generally (i.e check it returns a string) but don't actually have a way of doing that readily at the moment. It's on the backlog to work that out for this and other reasons.

But the same logic is true if your `name` field isn't actually a string, and my change here doesn't impact that at all, it just broadens the set of defaults keystone will recognise. So it's a good call out but not in scope for this PR.

jesstelford:
> But the same logic is true if your name field isn't actually a string, and my change here doesn't impact that at all

Yes, and `title` has been the default since Keystone v5's inception, so if anyone setup a list with `title` as a non-`Text`, they'd immediately get the error and fix it.

However, people may have been setting up lists since then without a `title`, but _with_ a `label` that is not a `Text`. As soon as they upgrade, they'd be getting errors and would have to resolve that. Another hurdle for them to jump over in the upgrade path.

> we need to add validation to this based on the type of field specified for the label field generally

Agreed. But either way, the end result is the same: An error occurs that the use has to action.

packages-next/keystone/src/lib/createKeystone.ts
```diff
@@ -70,8 +70,16 @@ export function createKeystone(config: KeystoneConfig): Keystone {
...
       hooks: listConfig.hooks,
     } as any) as any;
+    // Default the labelField to `name`, `label`, or `title` if they exist; otherwise fall back to `id`
     const labelField =
-      (listConfig.ui?.labelField as string | undefined) ?? (listConfig.fields.name ? 'name' : 'id');
+      (listConfig.ui?.labelField as string | undefined) ??
+      (listConfig.fields.name
+        ? 'name'
+        : listConfig.fields.label
+        ? 'label'
```
mitchellhamilton, line 79: I kinda feel like label should be preferred over name

JedWatson: Fair enough

timleslie (approved):

JedWatson:
After a bit more discussion offline yesterday, some notes to wrap up the feedback:

* I think given @mitchellhamilton's point about `label` should come first (which makes sense to me, given the fact it's called "labelField") this would be a hard breaking change to consume in isolation in current keystone. That's different to just adding more defaults to choose from when "name" doesn't exist, before we show the ID.
* In the future context, I'm bullish with keystone-next that we prioritise APIs that make sense for new users. Everything that doesn't make sense intuitively is a bad experience waiting to happen to someone.
* There's a strong argument that where possible, we should back-port changes (including breaking changes) from `@keystone-next/*` to `@keystonejs/*` packages, including because the more we minimise the "Big Bang" breaking changes, the easier it will be for people to upgrade down the track
* However to make this safe in either set of packages, we need to add more work to ensure the label works properly if you set it to a field that's not a basic "string" type under the hood. While the same label _feature_ is implemented in an equivalent way (more or less) between `@keystone-next/*` to `@keystonejs/*` the implementation _details_ and therefore the behaviour for edge cases are quite different in important but subtle ways. For that reason I'm inclined to make the effort once, because this will always represent "something to be checked" in the new Admin UI.

In summary:

* the `labelField` already falls into the category of "same feature, implemented differently" in the new Admin UI, including breaking changes (i.e syntax for configuration has changed, if you want a custom resolver you now have to use a virtual field, and we've removed the default `_label_` field from the GraphQL schema)
* I think the cost of checking the behaviour change in this PR is minor (and the same work as other checks that are unavoidable) in the context of upgrading to keystone-next
* I don't think the effort required to back-port the behaviour change safely to `@keystonejs/*` is worth the minor difference
* Without this change, we have config that behaves in a confusing way, and fixing all those paper cuts is one of the major goals of keystone-next

With all that in mind, I'm going to land this.

PR #4246, opened 2020-11-13
Add addToCart and checkout mutations to ecommerce example

timleslie:
Adds `addToCart` and `checkout` mutations to the ecommerce example.

add_to_cart_and_checkout

### DIFF
--- a/examples-next/ecommerce/package.json
+++ b/examples-next/ecommerce/package.json
@@ -15,7 +15,7 @@
     "@keystone-next/fields": "^3.0.0",
     "@keystone-next/keystone": "^3.0.0",
     "@keystone-next/types": "^3.0.0",
-    "@keystonejs/server-side-graphql-client":"^1.1.2",
+    "@keystonejs/server-side-graphql-client": "^1.1.2",
     "dotenv": "^8.2.0",
     "next": "^9.5.5",
     "nodemailer": "^6.4.16",
--- a/examples-next/ecommerce/package.json
+++ b/examples-next/ecommerce/package.json
@@ -15,6 +15,7 @@
     "@keystone-next/fields": "^3.0.0",
     "@keystone-next/keystone": "^3.0.0",
     "@keystone-next/types": "^3.0.0",
+    "@keystonejs/server-side-graphql-client":"^1.1.2",
     "dotenv": "^8.2.0",
     "next": "^9.5.5",
     "nodemailer": "^6.4.16",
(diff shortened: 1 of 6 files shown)


examples-next/ecommerce/package.json
```diff
@@ -15,6 +15,7 @@
     "@keystone-next/fields": "^3.0.0",
     "@keystone-next/keystone": "^3.0.0",
     "@keystone-next/types": "^3.0.0",
+    "@keystonejs/server-side-graphql-client":"^1.1.2",
```
timleslie, line 18:
```suggestion
    "@keystonejs/server-side-graphql-client": "^1.1.2",
```

examples-next/ecommerce/schema.ts
```diff
@@ -7,8 +9,8 @@ import type { ListsAPI, AccessControl } from './types';
   TODO
     - [ ] Access Control (new Roles system)
     - [ ] Tracking (createdAt, updatedAt, updatedAt, updatedBy)
-    - [ ] Port the addToCard mutation
-    - [ ] Port the checkout mutation
+    - [.] Port the addToCard mutation
+    - [.] Port the checkout mutation
```
timleslie, line 13:
```suggestion
    - [x] Port the addToCard mutation
    - [x] Port the checkout mutation
```

packages-next/keystone/src/lib/createKeystone.ts
```diff
@@ -257,6 +257,10 @@ export function createKeystone(config: KeystoneConfig): KeystoneSystem {
         },
         schema: graphQLSchema,
       } as KeystoneGraphQLAPI<any>,
+      // Note: These two fields let us use the server-side-graphql-client library.
+      // We may want to remove them once the updated graphQL API is available.
```
JedWatson, line 261: For upgraders, I'm seriously considering a config flag like `includeLegacyAPIs` that includes stuff like this, but is off by default. Sort of like a reverse "experimental" flag (which tailwind does really well)

Thinking:
* Want to hide this when there are better ways, so new users only see _the best way_ we have for them to do a thing
* Don't want to force people into upgrading to new changed APIs when all we need to do is continue passing internals through context, if what they've got will continue working

Just a not since you mention it here, not a thing for this PR

JedWatson (approved):
Great 👍

packages-next/keystone/src/lib/createKeystone.ts
timleslie: Yep, that's a plausible pattern, let's keep that on the table as an option and try to call out where we'd want to see it 👍

PR #4247, opened 2020-11-13
Add product status field to the ecommerce example

JedWatson:
@wesbos this is for you to make a call on, I think it's worth adding a `status` field to Products in the eCommerce example.

Means that the front-end would query for "Available" products, and you could draft a new product without sending it live, as well as remove products for sale without deleting them.


wesbos:
I like it!

PR #4253, opened 2020-11-14
Admin UI Tinkering

jossmac:
**Changes**
- tidier segmented control
- darker tooltips
- headings support heading font from theme
- resolve some alignment and spacing issues; mainly relationship field
- more "punchy" toast messages (now they're bottom right)
- attempt to find a layout solution for list + item view
- full-bleed fields
- field legends for "card" relationship fields

**Known issues**
- responsiveness: only really viable for screens >= 1280px

PR #4254, opened 2020-11-14
fix label in item page

gautamsi:
fixed label which was recently updated and defaulted to button instead of label

before
image

after
image


JedWatson (approved):

JedWatson:
Oh heck, thanks for catching that @gautamsi!

I was fixing the types in #4253 and that skipped right past me.

PR #4255, opened 2020-11-14
fixed Stack key prop

gautamsi:
I saw bunch of key prop errors in console log due to .map inside the Stack Component

### DIFF
--- a/design-system/packages/core/src/components/Stack.tsx
+++ b/design-system/packages/core/src/components/Stack.tsx
@@ -67,7 +67,7 @@ export const Stack = forwardRefWithAs<'div', StackProps>(
           .filter(child => isValidElement(child))
           .map((child, index) => {
             return (
-              <Fragment>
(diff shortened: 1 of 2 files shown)


design-system/packages/core/src/components/Stack.tsx
```diff
@@ -67,7 +67,7 @@ export const Stack = forwardRefWithAs<'div', StackProps>(
           .filter(child => isValidElement(child))
           .map((child, index) => {
             return (
-              <Fragment>
+              <Fragment key={index}>
```
gautamsi, line 70: not sure if `index` is right thing here but it does remove the error from console for now.

JedWatson: should be fine here, thanks 👍

JedWatson (approved):
Thanks @gautamsi

PR #4260, opened 2020-11-15
Updating Create Item placement and style

JedWatson:
I've been playing with the placement and style of the "Create Item" button, and I _think_ this is better than what we currently have. On the list screen, it looks like this:

image

It's clearer that there's now **one primary action** on both the list screen (create) and the item screen (save). Also putting it near the list title / filters / pagination makes it easier to see and use (this is based purely on my subjective testing, but what I'm saying is "I think it feels nice")

Separately, I've been playing with putting the "create a new {item}" and "delete" actions in a dropdown menu on the item screen, since these those are secondary actions. **And** I've been thinking about putting the save button up the top right where the create button currently is, as a primary action, instead of a floating toolbar; the dropdown would go next to it.

That needs some more thought and work so it'll land in another PR, but for now I've removed the create button from the item screen because (a) swapping one main button for another will be disorienting so I'd rather remove it until that decision lands, and (b) it's otherwise imo too far away from everything else - as well as too far away from where the button is on the list screen - that it's out of place and hard to use.

While navigating back to the list screen to create new items is more clicks, it feels solid and it's only temporary.


gautamsi:
Some thought on this based on my past experience with UI Hooks

* Placement like this will affect number of primary actions which can be placed on Item List page, I can visualize that it will be ugly when we want to add more item there. 
* if you place this filter and primary action together, Where would you add search. 
    * one option would be to add a global search on header section which can search item based on context (when search is fixed) then it can have GitHub like search context like this
image

* I do have a customer where I have added 4-5 primary action and some secondary actions (when you select items) to be applied on selected items

* I like the idea to add "Save" button on top of Create Item modal, I noticed that we have some view hints in ui config for items, one of the way we can also minimize the number of fields needed for initial create item view, the sequence is also important when rendered.

* I also have request from customer to have full page editing like item details page rather than slide out modal. Basically I have request for both the form being either full page or both like create item modal.

it would be helpful if we can preserve the extensibility to customize the form rather than compose ourself. I know the focus is on composability rather than the ui hooks in new interface.

jossmac (approved):

JedWatson:
@gautamsi that's really great feedback, I've been looking for more use-case information like this 👍 

I'll open a new issue with that so we can collect it somewhere outside this PR.

I'm not sure it'll stay where it is, so merged this PR as a transition, but it may change again in the next week or so as we add more actions.

PR #4263, opened 2020-11-15
Rename type Keystone to KeystoneSystem

timleslie:
Having a type called `Keystone` which is different to the core class `Keystone` is confusing, particularly when a field of the type is an instance of the `Keystone` class. This renames the type `Keystone` to `KeystoneSystem` to avoid this confusion.


JedWatson (changes requested):
👍 LGTM, but only the `types` package should be major, right? `keystone` and `admin-ui` should be patch because they have no API change

timleslie:
> only the `types` package should be major, right? `keystone` and `admin-ui` should be patch because they have no API change

Yeah, I was in two minds about this. In the past we've had some weird issues where we `major` bumped one thing and `patch` bumped a package that used it. Going with a `major` bump is safer, but churnier. Probably not too much of an issue in these packages right now. Happy to change packages to `patch` and see what happens.

.changeset/cyan-terms-unite.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystone-next/admin-ui': major
+'@keystone-next/keystone': major
```
timleslie, line 3:
```suggestion
'@keystone-next/admin-ui': patch
'@keystone-next/keystone': patch
```

JedWatson (approved):
👍

PR #4269, opened 2020-11-16
Rename KeystoneSystem objects to system

timleslie:
Now we know that a variable called `keystone` is indeed a core `Keystone` object.

### DIFF
--- a/.changeset/soft-adults-call.md
+++ b/.changeset/soft-adults-call.md
@@ -1,7 +1,7 @@
 ---
 '@keystone-next/types': major
-'@keystone-next/auth': patch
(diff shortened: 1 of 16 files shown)


.changeset/soft-adults-call.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystone-next/types': major
+'@keystone-next/auth': patch
+'@keystone-next/keystone': patch
```
mitchellhamilton, line 4:
```suggestion
'@keystone-next/auth': major
'@keystone-next/keystone': major
```

mitchellhamilton (approved):

PR #4270, opened 2020-11-16
Simplify auth graphQL generation

timleslie:
This PR changes things up so that the helper functions just have to worry about returning `{ typeDefs, resolvers }`, and all the schema merging is handled in a single step.

### DIFF
--- a/.changeset/happy-pigs-suffer.md
+++ b/.changeset/happy-pigs-suffer.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Simplified the code used to generate graphQL extensions.
--- a/packages-next/auth/package.json
+++ b/packages-next/auth/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.12.5",
+    "@graphql-tools/merge": "^6.2.5",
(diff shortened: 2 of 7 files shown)


JedWatson:
I'm 👎 on this as it currently is

Part of the point here is for _the system_ (at its lowest level) to put _doing anything with the GraphQL schema_ in the hands of solution developers. Expecting `{ typeDefs, resolvers }` at this level limits flexibility.

For example, you could use Prisma + Nexus to build part of your GraphQL schema and merge it with Keystone's using the current API, but with this change you'd be blocked from doing that. 

I think the more friendly we can make Keystone with other tooling in the GraphQL ecosystem the better, which means overlaying less of our own expectations on the lower-level APIs.

The point of publishing `graphQLSchemaExtension` was to make the composition that we think is _most likely_ as easy as possible without building that opinion into the base layer.

I'm down with whatever we can do to make it simpler -- maybe `graphQLSchemaExtension` isn't a great name, or maybe we could even offer two APIs on keystone config (an extension like this, and a lower level `rawExtendGraphQLSchema` or something) but also two ways of doing the same thing is 🤦 and I'm going round in circles.

In summary maybe there's something to do here, but I don't think this is it.

JedWatson (approved):

PR #4275, opened 2020-11-16
New roles based access example

JedWatson:
This adds a new example demonstrating how to implement sophisticated roles-based access control in a Keystone project.

This first PR sets up the example, schema, and behaviour spec with `/* TODO */` markers for behaviour to implement.

I'm going to split the entire example across multiple PRs so the development process is clearer after it's complete.

**Note to reviewers:**

Please check the logic of the permissions I've set up and the documented behaviour, to make sure it's a good spec before we go ahead and implement the actual access control and hooks


wesbos:
Thank Jed, a few questions to clarify:

1. We expose the permissions via the session, right? 
2. I don't see where we are allowing/stopping people from creating items. Is that yet to be coded?

JedWatson:
@wesbos 

(1) Yep - see these lines in the `keystone.js` config file:

```js
     session: withItemData(statelessSessions(sessionConfig), {
       /* This loads the related role for the current user, including all permissions */
       Person: `name role {
         id
         name
         canCreateTodos
         canManageAllTodos
         canSeeOtherPeople
         canEditOtherPeople
         canManagePeople
         canManageRoles
       }`,
     }),
```

To unpack what's happening here: 

* `statelessSessions(sessionConfig)` stores an `itemId` and `listKey` in a cookie, encrypted and signed, and decrypts the cookie data at the start of each request (name is not great tbh but "stateless" refers to "there is no remote state associated with the cookie, it's all in the cookie itself"
* `withItemData` then (wrapping that) takes the `listKey` and `itemId` and queries the GraphQL schema with the specified field selection set to load the item stored in the cookie.
  * If the query returns data, the result is put on `session.data` which is how access control functions can then do checks like `create: ({ session }) => session?.data.role?.canCreateTodos` without doing another query
  * If no item is found, we treat the session cookie as invalid and set the `session` object to `null`.

This is somewhat complex to express in the config, but the upside is that it's composable -- you could, for example, swap in a redis session store that just stores a sessionId in the cookie and loads the `itemId` and `listKey` from redis. Or replace `withItemData` with something else, if what you're storing in session isn't an item, for example if you're integrating with an SSO provider.

(2) That's correct, I wanted to get a review on the spec before I write the access control functions, because it's pretty complex and I wanted to check the logic before writing the code. I've got @timleslie's 👍 now so will merge this and ship a separate PR with the access control implementation in it (theory goes it's easier to follow and review the diff in multiple PRs, rather than just "all this code is new")

wesbos:
Awesome thanks a lot, looks great and all made sense for me!

PR #4296, opened 2020-11-17
Add Roles to ecommerce example

wesbos:
still a WIP.

One thing I find annoying is having to write out files and then also the permissions for each one. I'm experimenting with a `can()` syntax. Not sure if I like it..


wesbos:
Okay ready for review. I scrapped the `can()` idea.

Ended up programmatically creating the permissions functions, TS types, and Session GraphQL Query based on the checkbox fields. I feel this is the least amount duplication possible.

JedWatson (approved):
And done ✅

PR #4300, opened 2020-11-17
Remove connect and disconnect from implementSession

timleslie:
These functions are not being used.


molomby (approved):
If you say so 🤷🏻‍♂️

PR #4302, opened 2020-11-18
Document field

mitchellhamilton:


packages-next/fields-document/src/component-blocks.tsx
```diff
@@ -0,0 +1,206 @@
...
+      [Key in keyof PropsOption]: ExtractPropFromComponentPropField<PropsOption[Key]>;
+    }
+  ) => ReactElement | null;
+  props: PropsOption;
+}): ComponentBlock {
+  return options;
+}
+
+export const NotEditable = ({ children, ...props }: { children: ReactNode }) => (
+  <span contentEditable={false} {...props}>
```
JedWatson, line 203: Why doesn't this have `css={{ userSelect: 'none' }}` on it? I thought that was also important for non-editable things in Slate

mitchellhamilton: Added it back. I was testing things to see when it fails without user-select: none (if you know how to get it to fail, I would love to know)

packages-next/fields-document/src/DocumentEditor/Toolbar.tsx
```diff
@@ -0,0 +1,237 @@
...
+      <BlockComponentsButtons shouldInsertBlock={shouldInsertBlock} />
+      <Button
+        onMouseDown={event => {
+          event.preventDefault();
+          insertColumns(editor);
+        }}
+      >
+        + Columns
+      </Button>
+      {/*
```
JedWatson, line 52: AccessBoundary has been deleted for now, right? Shouldn't we delete this code as well (not just comment it out)?

timleslie (changes requested):
We need to add a `test-fixtures.js` to the field package to make sure we have test coverage of the new field type.

JedWatson (approved):
OK!

That was a monster to review, but I'm stoked to get this in 🎉

PR #4303, opened 2020-11-18
Fix sessionToken return value of auth mutation

timleslie:
No `sessionToken` value was being returned, which also had the side effect of returning the data as the wrong graphQL type.


molomby (approved):
Yeah cool, I think we renamed it at some point to avoid confusion with the other auth tokens (forgotten pass, etc.)

PR #4304, opened 2020-11-18
Use protectIdentities = true by default

timleslie:
This closes a security hole in the new auth implementation.

Update: I now see that `protectIdentities` is a thing. I wonder why this is `false` by default? I think perhaps we should change that. I'm curious as to when we would ever want to set this to `false` 🤔 

Update 2: I have made the change to set `protectIdentities: true` in both the new and current APIs.


packages-next/auth/src/gql/getBaseAuthSchema.ts
```diff
@@ -64,7 +56,11 @@ export function getBaseAuthSchema({
               itemPlural: list.adminUILabels.plural,
               code: result.code,
             });
-            return { code: result.code, message };
+            // We log the failure mode to the console, but do not
+            // expose this info via the graphQL interface.
+            // FIXME: Need to consider how to do server logging in general.
+            console.log({ code: result.code, message, identity: args[identityField] });
```
JedWatson, line 62: We arguably shouldn't do this either; leaking data into production logs is mostly not safe

JedWatson:
I'd rather a PR that sets the default on that option to `true` rather than removing the functionality. This can be genuinely useful for dev, but shouldn't really ever be on in production.

Or, since we don't like booleans that default to true, change it to something like `dangerouslyLeakIdentitiesThroughAPI` 😅 

Renaming it would be a breaking change with current keystone but this area is full of that anyway and imo the clarity is worth it (but I'm not seriously suggesting that name for the record)

timleslie:
> I'd rather a PR that sets the default on that option to `true` rather than removing the functionality. This can be genuinely useful for dev, but shouldn't really ever be on in production.

Yeah, I think that probably makes more sense. 
 
> Renaming it would be a breaking change with current keystone but this area is full of that anyway and imo the clarity is worth it (but I'm not seriously suggesting that name for the record)

Changing the default value in core keystone would also be a breaking change (it's defaulted to `false` in both the current and new systems). I think that's a breaking change that we should make, as the current system is insecure-by-default, which is not a great way to be.

molomby (approved):

packages-next/auth/src/gql/getBaseAuthSchema.ts
molomby: Specifically, logging PII (which the identity field will often contain) is going to be a problem for anyone needing to comply with the GDPR. Although I suppose you could argue that, since auth failed, the value often won't be correct...? Regardless, might be better to exclude it.

molomby (changes requested):

molomby (approved):
Aces 👌🏼

JedWatson (approved):

PR #4308, opened 2020-11-18
Use itemAPI in createInitialItem

timleslie:
Also fixes a bug in `itemAPI.count`, a bug in the `auth` example configuration, and uses `gqlNames` as the resolver name.


molomby (approved):
Good catches

JedWatson (approved):
✅ nice

PR #4316, opened 2020-11-19
Improved cors support

timleslie:
This PR adds support for configuring the `cors` middleware in line with what's currently available in core Keystone. The developer can now specify `server.cors` in their config object. Also, the apollo server no longer hijacks the cors config.

### DIFF
--- a/packages-next/keystone/src/lib/createAdminUIServer.ts
+++ b/packages-next/keystone/src/lib/createAdminUIServer.ts
@@ -45,7 +45,7 @@ export const createAdminUIServer = async (config: KeystoneConfig, system: Keysto
   if (config.server?.cors) {
     // Setting config.server.cors = true will provide backwards compatible defaults
     // Otherwise, the user can provide their own config object to use
-    let corsConfig =
+    const corsConfig =
       typeof config.server.cors === 'boolean'
         ? { origin: true, credentials: true }
         : config.server.cors;
--- a/packages-next/keystone/src/lib/createAdminUIServer.ts
+++ b/packages-next/keystone/src/lib/createAdminUIServer.ts
@@ -6,7 +6,7 @@ import express from 'express';
 import { ApolloServer } from 'apollo-server-express';
 // @ts-ignore
 import { formatError } from '@keystonejs/keystone/lib/Keystone/format-error';
-import type { KeystoneSystem } from '@keystone-next/types';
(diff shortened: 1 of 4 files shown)


packages-next/keystone/src/lib/createAdminUIServer.ts
```diff
@@ -36,14 +36,14 @@ const addApolloServer = ({ server, system }: { server: any; system: KeystoneSyst
...
+  apolloServer.applyMiddleware({ app: server, path: '/api/graphql', cors: false });
 };
 
-export const createAdminUIServer = async (system: KeystoneSystem) => {
+export const createAdminUIServer = async (config: KeystoneConfig, system: KeystoneSystem) => {
   const server = express();
 
   // TODO: allow cors to be configured
-  server.use(cors({ origin: true, credentials: true }));
+  server.use(cors({ origin: true, credentials: true, ...(config.server?.cors || {}) }));
```
JedWatson, line 46: I would not include cors by default like this, aside from it being an opinionated thing to do, you're also not able to turn it off.

Including defaults is (probably) fine, if they make sense, I'm not sure these are the right ones.

I'd basically not include this line at all if `config.server.cors` isn't defined, and _maybe_ (probably _later_) allow it to be `true` for the defaults and an object if you want to specify your own config (specifying would **set** not **spread**)

JedWatson (changes requested):

JedWatson (approved):
Looks good 👍

packages-next/keystone/src/lib/createAdminUIServer.ts
```diff
@@ -36,14 +36,21 @@ const addApolloServer = ({ server, system }: { server: any; system: KeystoneSyst
...
-export const createAdminUIServer = async (system: KeystoneSystem) => {
+export const createAdminUIServer = async (config: KeystoneConfig, system: KeystoneSystem) => {
   const server = express();
 
-  // TODO: allow cors to be configured
-  server.use(cors({ origin: true, credentials: true }));
+  if (config.server?.cors) {
+    // Setting config.server.cors = true will provide backwards compatible defaults
+    // Otherwise, the user can provide their own config object to use
+    let corsConfig =
```
timleslie, line 48:
```suggestion
    const corsConfig =
```

PR #4335, opened 2020-11-20
Lock yarn version to 1.22.10

jesstelford:
The documentation on yarn's homepage defaults to v2 (aka "Berry"), which may lead to contributors accidentally installing a that version of yarn before starting work.

Since we don't yet support v2, this change leverages a feature of both yarn v1 & v2 which allows us to lock the yarn version to 1.22.10 no matter the contributor's installed version.

Note the files in `.yarn` were generated by running `yarn policies set-version 1.22.10`.

Read more on `yarn policies` here: https://classic.yarnpkg.com/en/docs/cli/policies/


timleslie (approved):
This makes sense to me, can always revert if there are unexpected/unintended impacts 👍

PR #4341, opened 2020-11-22
Use explicitly silenced errors rather than warnings for disabled tests

timleslie:


JedWatson:
@timleslie no description... what does this change? (in effect, I mean, I can see what the code does but uncertain what the effect is)

timleslie:
> @timleslie no description... what does this change? (in effect, I mean, I can see what the code does but uncertain what the effect is)

Silences the warnings we were seeing both on the CLI and in CI, which was generating a lot of noise.

JedWatson (approved):

PR #4359, opened 2020-11-23
Minor tweaks to renovate config

timleslie:
I've adjusted the schedule to allow these new configs to run this morning. Once I'm happy with the config I'll put it back to when it was 👍


nathsimpson (approved):
🚢

PR #4372, opened 2020-11-24
Fix return value of sendItemPasswordResetLink

timleslie:


JedWatson (approved):
Man these things are getting clean. Love it.

PR #4374, opened 2020-11-24
Pass through itemAPI rather than full context in auth lib

timleslie:
Greatly reduces the API surface area available to our auth lib functions, making them easier to reason about.


JedWatson (approved):
👍

molomby:
Also 👍

PR #4379, opened 2020-11-24
Refactor auth token handling

timleslie:
This PR refactors the auth code to surface all the system-state-changing code into the mutation. This makes it easier to see directly what is changing in any mutation, which will hopefully make the code easier to understand. It also has the side-effect of causing a couple of modules to completely evaporate, which is nice :-)


packages-next/auth/src/gql/getMagicAuthLinkSchema.ts
```diff
@@ -78,16 +74,30 @@ export function getMagicAuthLinkSchema({
...
+            await itemAPI.updateOne({
+              id: itemId,
+              data: {
+                [`${tokenType}Token`]: token,
+                [`${tokenType}IssuedAt`]: new Date().toISOString(),
+                [`${tokenType}RedeemedAt`]: null,
+              },
+            });
+
+            await magicAuthLink.sendToken({ itemId, identity, token });
```
JedWatson, line 91: Not a nit for this PR, and I know this is typed as required, but should we be guarding this? or validating it when you create the auth? I just worry it'll blow up in unhelpful ways if it's not defined, esp when the consumer isn't using typescript and doesn't see the type error.

JedWatson (approved):

PR #4384, opened 2020-11-25
Add stricter types in auth package

timleslie:
A fair bit of messing about here as I try to tighten things up. I'll probably want to pair with someone at some point.

### DIFF
--- a/.changeset/old-stingrays-divide.md
+++ b/.changeset/old-stingrays-divide.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Added improved types to resolver functions.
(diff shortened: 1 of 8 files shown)


packages-next/auth/src/types.ts
```diff
@@ -92,3 +93,12 @@ export type AuthTokenRedemptionErrorCode =
   | 'TOKEN_MISMATCH'
   | 'TOKEN_EXPIRED'
   | 'TOKEN_REDEEMED';
+
+type T = KeystoneListsAPI<Record<string, BaseGeneratedListTypes>>[string];
+type KeystoneList = { adminUILabels: { singular: string; plural: string } };
+export type Context = {
```
JedWatson, line 99: You'll be able to use the type I've created in #4414 soon, has been bugging me that we re-type this everywhere

PR #4391, opened 2020-11-25
Fix addFieldValidationError lack of validation 

pahaz:
I have code like so:
```
...
const Ticket = keystone.createList({
...
    fields: { ... },
    hooks: {
        validateInput: ({ resolvedData, addFieldValidationError }) => {
                addFieldValidationError(`Some error`)  // DOESN'T REALLY ADD ERROR!
        },
    },
...
```

As you can see I call `addFieldValidationError` instead of `addValidationError`. Which leads to a lack of validation.
In some cases, this may lead to SQL injection. 

The function `addFieldValidationError` is included in arguments, but its calls **do not participate in the validation process**.

This fix lead to `addFieldValidationError is not a function` errors which is better than lack of validation.


packages/keystone/lib/ListTypes/hooks.js
```diff
@@ -119,6 +119,8 @@ class HookManager {
     if (fieldValidationErrors.length) {
       this._throwValidationFailure({ errors: fieldValidationErrors, operation, originalInput });
     }
+    // NOTE: addFieldValidationError should be awailable only for field level hooks
```
one-thunder, line 122: awailable

pahaz: ?

thekevinbrown: He's trying to tell you to change `awailable` to `available` in the comment.

pahaz, line 122:
```suggestion
    // NOTE: addFieldValidationError should be available only for field-level hooks
```

timleslie:
Fixed in #4652

PR #4405, opened 2020-11-26
Enable sendForgottenPassword feature and fix some security things

JedWatson:
@wesbos a couple of things for you!

* the `protectIdentities` config is pretty important to set in production, it means that any unsuccessful call to the auth mutations return with a generic `FAILURE` code rather than what really went wrong; this closes attack vectors, people can't brute force emails that *do* exist in the system by seeing messages like "invalid password for user" or "user not found" etc.
* I made it so users can only access the Admin UI if they have a role assigned, which anyone signed up for orders in the front-end wouldn't but admin users would. before, any user could get to the Admin UI (even though they couldn't actually see or do much, it's better to turn this off)
* I turned on the `sendForgottenPassword` feature in the auth mutation, currently it just logs the data out to the console but next step we can make this actually send the email with the password reset link in it


JedWatson:
Steps for how to use the `sendForgottenPassword` feature (needs to be documented properly elsewhere but this will get things going)

## 1. Turn on the feature as per this PR

That function that logs to the console is where you'd actually send an email, but for dev for now it's actually pretty helpful to log to the console.

## 2. Hit the `sendUserPasswordResetLink` mutation

Looks like this in the GraphQL Playground:

image

That will log the reset info that would be used to send an email to the console, in my case:

```js
Password reset info: {
  itemId: '5fb6956bf38339f25baa07a6',
  identity: 'admin@keystonejs.com',
  token: 'hSOp3bNrZr8RNg2Emj9P'
}
```

## 3. Hit the `redeemUserPasswordResetToken` mutation

image

✅ and you're done!

## A note on what the mutations return

These mutations return `null` when they are successful, mainly because there's nothing we actually need to (or even should) return. e.g we can't return the item (even its ID) because the person using the API may not actually be the owner of the email address.

If you get back `{ code message }` then the code is an error enum, and something went wrong. If `protectIdentities` isn't turned on, the code and message will tell you what _actually_ went wrong, otherwise this is an area the API should basically say as little as possible.

PR #4406, opened 2020-11-26
Factor out createContext from createSystem

JedWatson:
First in a series of refactoring PRs I've got planned to factor out a lot of the code currently in `createSystem` into separate files, to make the whole thing clearer to follow.

This is the important one, as it paves the way for some changes to createContext that will let us generate the new itemAPI including bringing in the functionality from the server-side-graphql package.

@timleslie probably best if we do this together, so I'll leave this PR for your review before opening the next one(s)

### DIFF
--- a/packages-next/keystone/src/lib/createContext.ts
+++ b/packages-next/keystone/src/lib/createContext.ts
@@ -59,8 +59,8 @@ export function makeCreateContext({
         createContext,
         raw: rawGraphQL,
         run: runGraphQL,
+        schema: graphQLSchema,
       } as KeystoneGraphQLAPI<any>,
-      schema: graphQLSchema,
       // Note: These two fields let us use the server-side-graphql-client library.
       // We may want to remove them once the updated itemAPI w/ resolveFields is available.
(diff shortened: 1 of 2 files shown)


packages-next/keystone/src/lib/createContext.ts
```diff
@@ -0,0 +1,76 @@
...
+      ...(skipAccessControl ? skipAccessControlContext : accessControlContext),
+      lists: itemAPI,
+      totalResults: 0,
+      keystone,
+      graphql: {
+        createContext,
+        raw: rawGraphQL,
+        run: runGraphQL,
+      } as KeystoneGraphQLAPI<any>,
+      schema: graphQLSchema,
```
timleslie, line 63:
```suggestion
        schema: graphQLSchema,
      } as KeystoneGraphQLAPI<any>,
```

timleslie (changes requested):
This is consistent with pending changes I have in my yard sale branch 👍

timleslie (approved):
👍

PR #4414, opened 2020-11-27
Type keystone context

JedWatson:
It's been really hard to deal with context because we never actually typed it properly, so it's specified mostly inline as `any`. Now that I'm trying to bring order to chaos with the context object, I decided to sit down and try to type it properly.

Putting this PR up early for feedback from @timleslie and @mitchellhamilton because I want to check the direction before I play it through.

Specifically, I haven't gone and actually updated most of the `context: any` types yet, and there are some TODO types in there that I will get to, but lmk what you think


packages-next/types/src/index.ts
```diff
@@ -142,12 +136,41 @@ export type KeystoneSystem = {
...
+  // Note: session is typed like this to acknowledge the default session shape
+  // if you're using keystone's built-in session implementation, but we don't
+  // actually know what it will look like.
+  session?:
+    | {
+        itemId: string;
+        listKey: string;
+        data?: Record<string, any>;
+      }
+    | any;
```
mitchellhamilton, line 154: This `| any` means the the type is just `any`. If you want to leave the previous thing there for documentation, that's fine but just be aware that it's doing nothing type-wise

JedWatson: Yeah, I know, and I don't know what to do about that 😞 

Would prefer to leave it there for now so if someone is peeking the types, they find it, but we'll probably remove it later when we put another round of focus into generating this stuff.

mitchellhamilton (commented):
✅ to the general thing you're doing here.

JedWatson:
OK, I think this has gone far enough for now, still have some types in the context marked with `TODO` but those can come in a separate PR. This has cleared things up considerably.

mitchellhamilton (approved):

packages-next/auth/src/gql/getPasswordResetSchema.ts
```diff
@@ -59,7 +61,7 @@ export function getPasswordResetSchema({
     `,
     resolvers: {
       Mutation: {
-        async [gqlNames.sendItemPasswordResetLink](root: any, args: any, context: any) {
+        async [gqlNames.sendItemPasswordResetLink](root, args, context) {
```
timleslie, line 64: Do we not want to add a `KeystoneContext` on here (and other places in the auth package)?

PR #4421, opened 2020-11-27
Handling of File state after saving was improved

miguelvvs:
**Handling of File state after saving was improved**

In Admin UI when a File field is updated and saved there are at least 2 scenarios where the user interface is not accurate enough.:

1) User changes the file and the label "save to upload" is shown even after the changes were applied.

2) User deletes a file, save the changes, and the label "save to remove" is still shown.

This Change considers these scenarios and updates the component status accordingly.


JedWatson (approved):
Thanks @miguelvvs! Much appreciated.

PR #4424, opened 2020-11-29
Types for listQuery and similar

timleslie:


packages-next/types/src/index.ts
```diff
@@ -161,7 +161,7 @@ export type SessionContext = {
 
 export type KeystoneContext = {
   schemaName: 'public';
-  lists: any; // TODO: type this (itemAPI),
+  lists: Record<string, KeystoneListsAPI<Record<string, BaseGeneratedListTypes>>[string]>;
```
JedWatson, line 164: This is a bit weird, looks like you're unwrapping and then rewrapping a record here?

examples-next/basic/schema.ts
```diff
@@ -229,7 +227,7 @@ export const extendGraphqlSchema = graphQLSchemaExtension({
       createRandomPosts(root, args, context) {
         // TODO: add a way to verify access control here, e.g
         // await context.verifyAccessControl(userIsAdmin);
-        const lists: KeystoneListsAPI<KeystoneListsTypeInfo> = context.lists;
+        const { lists } = context;
```
JedWatson, line 230: Why have we dropped the usage of the generated types here?

.changeset/soft-impalas-remember.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystone-next/app-basic': patch
+'@keystone-next/auth': patch
+'@keystone-next/types': major
```
JedWatson, line 4: Save argument as in #4427, correct usage of APIs isn't going to cause a problem with the updated types, imo this should be patch.

Otherwise almost every change to the types package is going to result in a major.

packages-next/types/src/index.ts
timleslie: Oh, quite possibly, this was a straight cut and paste from somewhere I don't remember, TBH I didn't actually engage my brain on what the actual type is 😅 What would be a cleaner type here?

packages-next/auth/src/gql/getMagicAuthLinkSchema.ts
```diff
@@ -82,7 +82,7 @@ export function getMagicAuthLinkSchema({
             // Save the token and related info back to the item
             const { token, itemId } = result;
             await itemAPI.updateOne({
-              id: itemId,
+              id: itemId.toString(),
```
JedWatson, line 85: If this is to satisfy the types, I'm not sure how I feel about it, but stylistically I'd prefer

```
id: `${itemId}`,
```

Which functionally has the same effect (including ensuring the type will be `string`)

examples-next/basic/schema.ts
timleslie: Well `lists` is already typed as it comes of the `context` object. The re-typing serves to restrict it (there are a couple of fields that are not included in the generated types, in particular `search` and `orderBy`). This gives it the more correct typing. If we want to restrict usage of `search` and `orderBy` completely then we should do it at the `context` type layer.

I've seen a few examples like there where we have slightly inconsistent types, I'm trying to grind through the backlog and get everything talking the same language.

.changeset/soft-impalas-remember.md
timleslie, line 4:
```suggestion
'@keystone-next/types': patch
```

packages-next/auth/src/gql/getMagicAuthLinkSchema.ts
timleslie, line 85:
```suggestion
              id: `${itemId}`
```

packages-next/auth/src/gql/getPasswordResetSchema.ts
```diff
@@ -83,7 +83,7 @@ export function getPasswordResetSchema({
             // Save the token and related info back to the item
             const { token, itemId } = result;
             await itemAPI.updateOne({
-              id: itemId,
+              id: itemId.toString(),
```
timleslie, line 86:
```suggestion
              id: `${itemId}`,
```

packages-next/types/src/index.ts
JedWatson: Without reading the surrounding code and without being 100% sure, I _think_ that this is equivalent:

```
lists: KeystoneListsAPI<Record<string, BaseGeneratedListTypes>>;
```

You're basically accessing a property referenced by `string` in a Record, then wrapping that in a Record indexed by `string`, the line above removes both.

But also I don't think this is happening the right way, because the argument to `KeystoneListsAPI` should be defaulted as you have it, and is only needed if you're using the more specific generated types. Which is where this gets tricky.

examples-next/basic/schema.ts
JedWatson: Just FYI, I don't believe `search` and `orderBy` were deliberately restricted, they were probably just missed in the first pass of types.

examples-next/basic/schema.ts
```diff
@@ -229,7 +229,7 @@ export const extendGraphqlSchema = graphQLSchemaExtension({
       createRandomPosts(root, args, context) {
         // TODO: add a way to verify access control here, e.g
         // await context.verifyAccessControl(userIsAdmin);
-        const lists: KeystoneListsAPI<KeystoneListsTypeInfo> = context.lists;
+        const lists = context.lists as KeystoneListsAPI<KeystoneListsTypeInfo>;
```
JedWatson, line 232: I'd like to get @mitchellhamilton to check out what's going on here, and why this is necessary

packages-next/types/src/index.ts
timleslie: After much mucking around I'm 99% sure that the type as given is correct. It mirrors what we already have in `itemAPI.js`, and while I'm still getting the hang of these advanced types I don't think there's any alternative that makes everything work. The suggested equivalent did not work.

examples-next/basic/schema.ts
timleslie: In the `ecommerce example` we have:

```
type ListsAPI = KeystoneListsAPI<any /* KeystoneListsTypeInfo */>;
...
const lists: ListsAPI = context.lists;
```

which we would also probably want to get to the bottom of.

examples-next/basic/schema.ts
mitchellhamilton: > Just FYI, I don't believe search and orderBy were deliberately restricted, they were probably just missed in the first pass of types.

orderBy was deliberately excluded since it should really be used anymore though if people want to add it, that's fine(though probs add an `@deprecated`)

packages-next/types/src/index.ts
mitchellhamilton:
```suggestion
  lists: KeystoneListsAPI<any>;
```

examples-next/basic/schema.ts
JedWatson: That one in the `ecommerce` example is because I didn't want to commit the types, which we can fix by having a type generation process that gets run by CI before linting. Mitchell has thoughts on this, for now we should probably just commit the types so this isn't commented out.

mitchellhamilton: > In the ecommerce example we have:

That's probably because the generated types file for the ecommerce example is gitignored, it shouldn't be gitignored

> needing to use `as`

https://github.com/keystonejs/keystone/pull/4424/files#r532976085 will make it unnecessary to use `as`(which is a good thing because `as` is nuanced and you should avoid it where possible(except for `as const` which is totally different and v good))

packages-next/types/src/utils.ts
```diff
@@ -2,13 +2,16 @@ export type BaseGeneratedListTypes = {
...
-    create: GraphQLInput;
-    update: GraphQLInput;
-    where: GraphQLInput;
-  };
+  inputs: { create: GraphQLInput; update: GraphQLInput; where: GraphQLInput };
   args: {
-    listQuery: GraphQLInput;
+    listQuery: {
+      readonly where?: GraphQLInput | null;
+      readonly search?: string;
```
mitchellhamilton, line 9:
```suggestion
      readonly search?: string | null;
```

mitchellhamilton (approved):

PR #4425, opened 2020-11-29
[WIP] Updated Items API / GraphQL Context Implementation

JedWatson:
As @timleslie and I have been discussing, this includes:

* addition of `cloneContext` to `context`
* you can no longer pass a different context in to `graphql.raw` and `graphql.run` or the Items API, the new idea is to _create a new context first_ and use the APIs on that new context
* adding new `resolveFields` argument to the Items API


timleslie:
I think everything we wanted from this has now been integrated 👍

PR #4426, opened 2020-11-29
Use KeystoneContext type where appropriate

timleslie:


examples-next/ecommerce/mutations.ts
```diff
@@ -11,7 +12,7 @@ export const extendGraphqlSchema = graphQLSchemaExtension({
   `,
   resolvers: {
     Mutation: {
-      checkout: async (root, { token }: { token: string }, context) => {
+      checkout: async (root, { token }: { token: string }, context: KeystoneContext) => {
```
JedWatson, line 15: this should already be typed as KeystoneContext by `graphQLSchemaExtension`

timleslie: Yep, there's actually a few of these now I notice, going through and cleaning them up now.

packages-next/admin-ui/src/templates/adminMetaSchemaExtension.ts
```diff
@@ -141,7 +141,12 @@ export function adminMetaSchemaExtension({
...
       },
       KeystoneMeta: {
-        adminMeta(rootVal: any, args: any, context: any) {
+        adminMeta(
+          rootVal: any,
+          args: any,
+          context: any
+          // context: { isAdminUIBuildProcess: true } | KeystoneContext FIXME: Why doesn't this work?
+        ) {
           if (context.isAdminUIBuildProcess || isAccessAllowed === undefined) {
```
JedWatson, line 150: Answer to comment above 

> Why doesn't this work?

Is this line, `context.isAdminUIBuildProcess` is an internal extension just for the build phase, I decided not to include that in the types on the first pass so we could discuss it.

Possibly worth just dropping that in the main context type, but it's a weird one and I'm not sure we want it to be public API.

JedWatson: OH right I see what's happening here, sorry -- I _think_ the answer to the comment above is that you need an `&` not a `|`

timleslie: Oh!!! I misread `context.isAdminUIBuildProcess || isAccessAllowed === undefined` as `context.isAdminUIBuildProcess && isAccessAllowed === undefined`, ok, that explains everything, incoming fix 👍

JedWatson (approved):

PR #4427, opened 2020-11-30
Core keystone type

timleslie:
This is a something-is-better-than-nothing PR. Definitely needs more comprehensive thinking around the best ways to provide typing for the pre-existing parts of the Keystone system.

### DIFF
--- a/.changeset/nasty-dingos-punch.md
+++ b/.changeset/nasty-dingos-punch.md
@@ -1,0 +1,7 @@
+---
+'@keystone-next/auth': major
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+Added a `BaseKeystone` type to replace usage of `any` in all instances.
--- a/.changeset/nasty-dingos-punch.md
+++ b/.changeset/nasty-dingos-punch.md
@@ -1,6 +1,6 @@
 ---
-'@keystone-next/auth': major
-'@keystone-next/keystone': major
+'@keystone-next/auth': patch
+'@keystone-next/keystone': patch
 '@keystone-next/types': major
 ---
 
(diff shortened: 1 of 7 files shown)


.changeset/nasty-dingos-punch.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystone-next/auth': major
```
JedWatson, line 2: `patch` for `/auth` and `/keystone`, surely, and `minor` for `/types`?

timleslie: Yeah, probably right for `auth` and `keystone`, but `types` has changes which restrict existing APIs, so technically they're `major`, right? (I'm not super fussed at this early alpha stage, but probably good to get the hang of versioning our types correctly).

.changeset/nasty-dingos-punch.md
```diff
@@ -0,0 +1,7 @@
...
+'@keystone-next/keystone': major
```
timleslie, line 3:
```suggestion
'@keystone-next/auth': patch
'@keystone-next/keystone': patch
```

.changeset/nasty-dingos-punch.md
JedWatson: Fair, if we're making types more restrictive that's theoretically a breaking change for the types... except I'd say in this instance we're adding types for APIs that haven't actually gone through major revisions so theoretically it's not a breaking change unless someone's passing something invalid which would now result in a type error.

My rule of thumb has been that, for example, changing the _name of a type_ would result in a braking change to the types package because a consumer may have imported it and would now need to import something different.

But typing things we didn't previously type (like here) is _adding something_ (minor) and adding properties to types that already exist is _fixing them_ (patch)

Of course, if an API we're typing goes through a breaking change, that's a major on both the package with the functionality and also the types package because you don't want someone getting the types for the new API with a `^` range.

JedWatson: (also, I'd argue that typing something that wasn't previously typed, even if it results in errors for consumers, is actually _fixing a bug_ with the types, because before they were incorrectly passing -- again, the point is that valid API usage wouldn't see a "breaking change")

If you're happy with this definition of how we version types we should probably add it to the Readme in the types package as documentation...

JedWatson (approved):

PR #4430, opened 2020-11-30
Convert utils package to typescript

timleslie:
This is a POC to help envisage what might be required to convert any or all of our existing packages to typescript.

There's no rush to push this through, I'd like to use this POC to help us capture the general pattern of things which need to be done in order to convert a package.

I'm open to feedback on what's missing, what else we might want to do, and what else might need to be considered here.

I've picked this package because it's got a fairly simple API with very few dependencies, so we can focus on process rather than difficult typing challenges.

Update: I think we should now push this through 👍

### DIFF
--- a/packages/utils/src/index.ts
+++ b/packages/utils/src/index.ts
@@ -3,28 +3,34 @@ import pReflect from 'p-reflect';
 import isPromise from 'p-is-promise';
 import semver from 'semver';
 
-export const noop = x => x;
+export const noop = <T>(x: T): T => x;
 export const identity = noop;
-export const getType = thing =>
+export const getType = (thing: any) =>
   Object.prototype.toString.call(thing).replace(/\[object (.*)\]/, '$1');
 
-export const escapeRegExp = str =>
+export const escapeRegExp = (str: string) =>
   (str || '').replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
 
 // { key: value, ... } => { key: mapFn(value, key), ... }
-export const mapKeys = (obj, func) =>
+export const mapKeys = <T, U>(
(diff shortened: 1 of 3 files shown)


packages/utils/src/index.ts
```diff
@@ -3,28 +3,34 @@ import pReflect from 'p-reflect';
...
 
-export const escapeRegExp = str =>
+export const escapeRegExp = (str: string) =>
   (str || '').replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
 
 // { key: value, ... } => { key: mapFn(value, key), ... }
-export const mapKeys = (obj, func) =>
+export const mapKeys = <T, U>(
+  obj: Record<string, T>,
+  func: (value?: T, key?: string, obj?: Record<string, T>) => U
```
mitchellhamilton, line 17:
```suggestion
  func: (value: T, key: string, obj: Record<string, T>) => U
```

packages/utils/src/index.ts
```diff
@@ -3,28 +3,34 @@ import pReflect from 'p-reflect';
...
+): Record<string, U> =>
   Object.entries(obj).reduce((acc, [key, value]) => ({ ...acc, [key]: func(value, key, obj) }), {});
 
 // { key: value, ... } => { mapFn(key, value): value, ... }
-export const mapKeyNames = (obj, func) =>
+export const mapKeyNames = <T>(
+  obj: Record<string, T>,
+  func: (key?: string, value?: T, obj?: Record<string, T>) => string
```
mitchellhamilton, line 24:
```suggestion
  func: (key: string, value: T, obj: Record<string, T>) => string
```

mitchellhamilton (approved):
Some more of these types could be better but meh, it's fine

PR #4431, opened 2020-11-30
Reorder lint run order for faster failure

timleslie:
It's nice to know ASAP if linting is going to fail, and in my experience this the order in which it's most likely to go wrong.


JedWatson (approved):
👍

PR #4448, opened 2020-11-30
Pass a context through to onConnect

JedWatson:
Opening a PR for easier review, ref #4440


JedWatson:
👍 I'm happy with these changes, does simplify things and is more consistent

packages-next/keystone/src/lib/createContext.ts
```diff
@@ -54,6 +53,11 @@ export function makeCreateContext({
       lists: itemAPI,
       totalResults: 0,
       keystone,
+      // Only one of these will be available on any given context
+      // TODO: Capture that in the type
+      knex: keystone.adapters.KnexAdapter?.knex,
```
gautamsi, line 58: What if this was named as `adapter` instead of individual adapter itself?

There can be prop on adapter whether they are knex, mongoose or prisma. 

Now that there are efforts to get rid of multiple adapter system we could just have keystone.adapter rather than .knex, .prisma etc.

PR #4450, opened 2020-12-01
Skip access control in Order create

timleslie:
This is for forward compatibility with upcoming changes to the items API, which will no longer skip access control by default.

### DIFF
--- a/.changeset/chilled-scissors-worry.md
+++ b/.changeset/chilled-scissors-worry.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/example-ecommerce': patch
+---
+
+Updated `checkout` mutation to use a context which skips access control to create the `Order` object.
--- a/examples-next/ecommerce/mutations.ts
+++ b/examples-next/ecommerce/mutations.ts
@@ -66,14 +66,16 @@ export const extendGraphqlSchema = graphQLSchemaExtension({
 
         // 5. create the Order
         console.log('Creating the order');
-        const order = await context.lists.Order.createOne({
-          data: {
-            total: charge.amount,
-            charge: `${charge.id}`,
-            items: { create: orderItems },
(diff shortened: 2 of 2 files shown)


examples-next/ecommerce/mutations.ts
```diff
@@ -66,14 +66,16 @@ export const extendGraphqlSchema = graphQLSchemaExtension({
...
         console.log('Creating the order');
-        const order = await context.lists.Order.createOne({
-          data: {
-            total: charge.amount,
-            charge: `${charge.id}`,
-            items: { create: orderItems },
-            user: { connect: { id: userId } },
-          },
-        });
+        const order = await context
```
JedWatson, line 69: For the example, I'd prefer to see the context created separately from the order. I can imagine people copying+pasting this pattern and creating a new context for each items API operation, so let's clarify that into two steps.

JedWatson (changes requested):

.changeset/chilled-scissors-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/example-ecommerce': patch
```
JedWatson, line 2: Thinking about #4462, I don't think this sort of refactor is worth a changeset. It's basically duplicating the commit history for the example into the changelog. But maybe that's more accessible for people who want to follow along? I could probably be convinced if you have strong views, but I haven't been adding changesets for similar things and we should be consistent.

JedWatson: Experimented with this in #4463

examples-next/ecommerce/mutations.ts
JedWatson: Experimented with this in #4463

timleslie:
Superseded by #4463.

PR #4456, opened 2020-12-01
Use original config object, not system.config

timleslie:
Moving towards not carrying the `config` property on the `system`. We should be explicitly consuming the original `config` object where required.

### DIFF
--- a/.changeset/chilled-onions-type.md
+++ b/.changeset/chilled-onions-type.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/admin-ui': patch
+'@keystone-next/keystone': patch
+---
+
+Refactored code to use the original `config` object, rather than `system.config`.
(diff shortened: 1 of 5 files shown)


JedWatson (approved):

JedWatson:
Leaving my approval in place for now, but after reviewing #4459 I'm not so sure about this.

It's pretty critical* that the Admin UI is built based on the same config that was used to initialise the system, and with this change that's no longer guaranteed or clear from the API.

Not sure putting config on system is the best way to achieve that, but this feels like a step backwards in terms of clarity.

\* when I say critical, I can't actually think of what would break, but it's absolutely possible it could happen down the track because we haven't put a lot of thought into the config yet, it's bare minimum to run keystone in dev so I'd really like this to be a thing we guarantee from the outset

PR #4459, opened 2020-12-01
Apply applyIdFieldDefaults before createSystem

timleslie:
Having `createSystem` modify the `config` object is somewhat unexpected. This makes it more clear that we're modifying the config object immediately after loading it.

I suspect as we continue factoring the code for testability these couple of lines in `dev.ts` will evolve further, but this is a useful first step.

### DIFF
--- a/.changeset/plenty-donkeys-live.md
+++ b/.changeset/plenty-donkeys-live.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Refactored dev script to apply `applyIdFieldDefaults` before sending `config` object to `createSystem`.
--- a/packages-next/keystone/src/lib/applyIdFieldDefaults.ts
+++ b/packages-next/keystone/src/lib/applyIdFieldDefaults.ts
@@ -2,7 +2,7 @@ import type { KeystoneConfig } from '@keystone-next/types';
 import { autoIncrement, mongoId } from '@keystone-next/fields';
 
 /* Validate lists config and default the id field */
-export function applyIdFieldDefaults(config: KeystoneConfig): KeystoneConfig {
+export function applyIdFieldDefaults(config: KeystoneConfig): KeystoneConfig['lists'] {
   const lists: KeystoneConfig['lists'] = {};
   Object.keys(config.lists).forEach(key => {
     const listConfig = config.lists[key];
@@ -31,6 +31,5 @@ export function applyIdFieldDefaults(config: KeystoneConfig): KeystoneConfig {
     const fields = { id: idField, ...listConfig.fields };
     lists[key] = { ...listConfig, fields };
   });
-
-  return { ...config, lists };
+  return lists;
 }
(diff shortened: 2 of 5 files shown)


JedWatson (changes requested):
I'm not a fan of this one.

1. Over time, I expect we'll add more of these "validate and mutate the config" functions; I originally had this as more of a generic "sortOutConfig" function but landed on it being clearer if each mutation is a separate functional step.

2. As we add more scripts than just "dev" (e.g to build the Admin UI, start the production server, start just the GraphQL API, build the types, etc) one of the first thing they'll do is create the system.

Applying the ID field defaults (and any other config validation and mutation steps we add) is a critical step in starting the system correctly. It belongs with creating the system, not starting the dev server - no alternate mode you can start Keystone in should ever do anything different.

Also, I expect it's possible that people may want to implement their own scripts (this is a fairly opinionated part of keystone) which is why the steps in `dev.ts` are organised the way they are:

* Load the config
* Create a System from the config
* Generate the GraphQL Schema
* Generate the Admin UI
* Start the Express Server

Each of these should be like a lego block, and (aside from passing the config and/or system into each) they're fairly independent of each other by design.

---

Thinking about it more, and considering how #4456 relates to this, I'd be on board with making "sort out the config" one of those first class steps (separate from loading the config)

As in, we could take it further and make _initialised config_ a first class thing alongside _the system_, or pull it back to _the config is part of the system_.

It's the implication of not enforcing that you're creating the system from an initialised config that's bothering me.

timleslie:
Perhaps the best way of thinking about this PR and the #4458 is as a defactoring. We're pulling things apart to make it clear what pieces we have before we put them back together again in a more organised fashion.

For example, in this instance, this now isolates the "loading-and-preparing-config" from the "config => system" step. This means that the second step is now something that we can potentially test in isolation. It also helps us think about where config comes from in the systems and we we do to it before actually using it. Lastly, it also makes it slightly more clear that there are other stages in the dev script which depend on prepared config.

As mentioned in the PR comment, this isn't intended as a final step, but as a stepping stone as we move towards a more complete solution.

JedWatson:
That makes sense and I think I'm just sticking on a minor part of it.

> We're pulling things apart to make it clear what pieces we have before we put them back together again in a more organised fashion.

I'm uneasy about having `applyIdFieldDefaults` hanging about in `dev`, I just don't feel like this makes it clear what the pieces are. `dev` should be doing high level things like "get the config", "create a system", and "generate an Admin UI", which will be shared building blocks with other scripts. `applyIdFieldDefaults` is an implementation detail of "get the config" and the dev script shouldn't know about it. TBH if the function had been named `initConfig` in the first place, I wouldn't have any issues with what's going on in this PR.

> this now isolates the "loading-and-preparing-config" from the "config => system" step. This means that the second step is now something that we can potentially test in isolation. It also helps us think about where config comes from in the systems and we we do to it before actually using it. Lastly, it also makes it slightly more clear that there are other stages in the dev script which depend on prepared config.

Check out #4479 and see if you can get on board with that, hopefully it strikes a balance between achieving your goal and the bit I'm concerned about.

(also think of that as a transitional step, we can organise it better when we work out what else we have, but it puts a line in the sand and says "there is a consolidated initConfig process")

JedWatson (approved):
👍

PR #4463, opened 2020-12-01
Skip access control in Order create (alt)

JedWatson:
Alternative take on  #4450

tbh not really sure what to call the second context, I went with `superContext` but it's not my finest work.

I do think it solves for my comment there, but I'm not sure what a _great_ pattern looks like here yet.


timleslie:
TBH part of the reason I did it inline rather than naming it is because naming is hard :-) Here some are alts that come to mind. I'm not sold on them, but I also don't think `superContext` is quite right either 🤷‍♂️ 

* `allAccess`
* `allAccessContext`
* `skipAccess`
* `skipAccessContext`
* `adminContext`

JedWatson:
So.. I can't tell if this is a really good idea or a really bad idea.

What do you think of the name `sudoContext`?

I struggle a bit with `adminContext` because of the Admin UI connection, even though "admin" is usually associated with "all access" in _our_ Admin UI's case it's in front of access control and I don't want to do anything that would confuse that association.

`superContext` was meant to draw on the `superuser` association, but I actually kind of prefer "sudo" because with reference to the Wikipedia description:

> sudo (/suːduː/[4] or /ˈsuːdoʊ/[4][5]) is a program for Unix-like computer operating systems that allows users to run programs with the security privileges of another user, by default the superuser.

that's... kind of actually pretty close to what we're doing here. And people using node will be fairly familiar with "if you can't execute something in the current user context, use `sudo` to break outside those restrictions, but be careful with it"

timleslie:
I can live with `sudoContext`. It's not a slam-dunk, but I think it conveys the right info to enough people, while being readable, searchable and unambiguous.

timleslie (approved):
👍

molomby:
Not at all a fan of `sudoContext` – It doesn't actually describe what the variable _is_ it just kind of _eludes_ to it by reference. There are/will be Keystone/nodejs devs on Windows that have literally never used or heard of the `sudo` command. Why use metaphor when you can use plain english instead? `skipAccessContext` or maybe `elevatedContext` would be much clearer/better.

PR #4465, opened 2020-12-01
[RFC] Make SessionStrategy.start and SessionStrategy.end required

timleslie:
Having these methods being optional leads to some slightly awkward arrangements in our code. For example the `enableSignout` property of `adminMeta` is conditional on the `.end` property existing. This has implications all the way through how we setup out `system` object.

We could greatly simplify our internal code by making these methods required on the type. If a strategy doesn't have anything to do then they can always provide a no-op function.

If there something that I'm missing here? For example is there a world in which we want to have a session implemented but explicitly don't want the user to ever be able to signout?

If we decide to go ahead with this change I'll follow up with PRs to simplify the code that follows from this.

### DIFF
--- a/.changeset/eight-taxis-lick.md
+++ b/.changeset/eight-taxis-lick.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/types': major
+---
+
+Updated `SessionStrategy.start` and `SessionStrategy.end` to be required attributes.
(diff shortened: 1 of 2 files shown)


JedWatson:
> For example is there a world in which we want to have a session implemented but explicitly don't want the user to ever be able to signout?

In theory, yes, we may not be able to (e.g session implementation is 3rd-party and we don't actually have an API to end it)

The point is not so much that it could be a no-op, fine, but how do we know not to generate a mutation for a no-op? that's the reasoning behind them being optional.

Having said that if this simplifies things a lot I'm happy to drop that for now, it's a theoretical feature. We can always solve for it later if it comes up in the real world.

JedWatson (approved):

PR #4467, opened 2020-12-01
Add type for BaseKeystone.createApolloServer

timleslie:


packages-next/types/src/index.ts
```diff
@@ -215,6 +215,7 @@ export type BaseKeystone = {
   ) => BaseKeystoneList;
   connect: () => Promise<void>;
   lists: Record<string, BaseKeystoneList>;
+  createApolloServer: (a: { schemaName: string; dev: boolean }) => any;
```
JedWatson, line 218:
```suggestion
  createApolloServer: (args: { schemaName: string; dev: boolean }) => any;
```

afaik we've been using `args` everywhere we do this, lets be consistent

JedWatson (changes requested):

JedWatson (approved):

PR #4468, opened 2020-12-01
Precompute arg-parsing functions

timleslie:


.changeset/silent-donkeys-work.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Optimised `createContext` by precomputing arg-parsing functions at system initialisation.
```
JedWatson, line 5: 👍 excellent use of a patch changeset here imo

packages-next/keystone/src/lib/createContext.ts
```diff
@@ -75,7 +82,7 @@ export function makeCreateContext({
       gqlNames: (listKey: string) => keystone.lists[listKey].gqlNames,
     };
     for (const [listKey, list] of Object.entries(keystone.lists)) {
-      itemAPI[listKey] = itemAPIForList(list, contextToReturn, graphQLSchema);
+      itemAPI[listKey] = itemAPIForList(list, contextToReturn, getArgsByList[listKey]);
```
JedWatson, line 85: I generally prefer the constructor pattern in #4425, this feels to me like it reduces code clarity. We've got two tightly coupled functions that are only separate for performance reasons; but with the precompute/pass pattern it feels like they are more independent than they are.

Not a blocking comment, since there's no functional difference, and this probably simplifies testing.

JedWatson (approved):

PR #4471, opened 2020-12-02
Simplify code with SessionStrategy.end and .start being required

timleslie:


JedWatson:
Aah. Now I see the implication, I realise what was niggling my brain with #4465

Maybe we should roll that back instead of moving forward with this.

Basically, the use case is that your session could be provided by another system that the keystone instance doesn't have the privilege to log you out of. That might be something like IP-based auth on an internal network, or a session service that's implemented in a reverse-proxy layer in front of keystone where the process of login and logout is handled entirely outside of the Admin UI.

(It clicked when I reviewed this code, and realised the logic for "do we show a signout button" is linked to "can we log someone out", and this replaces that with "is there a session". The two answers are not necessarily the same)

With our previous session implementation all this was impossible because you couldn't externalise session management and use it in any meaningful way inside of Keystone (e.g inside access control / hooks / etc).

The new one is a lot more powerful and I think we're still getting our heads around how it might be used.

I'm still open to moving forward if you feel strongly enough that the simplification is worth cutting off those use cases, because tbh I don't know how much this will come up in the real world, and there are probably other ways we could solve for it, but I'm less inclined than I was yesterday.

timleslie:
> the logic for "do we show a signout button" is linked to "can we log someone out", and this replaces that with "is there a session". The two answers are not necessarily the same

I think there are actually three things that might not be the same:
 * a) Can we log someone out (e.g. does `.end()` exist and work)
 * b) Is there a keystone-generated signout mutation
 * c) Is there a logout button on the admin UI

I can imagine a world in which any of `[a]`, `[a,b]` or `[a,b,c]` are true. I think if we want to provide the ability to control and manipulate these things we'd be better off doing it with explicit configuration, rather than hingeing on the `.end` method. I think we should go ahead with this PR, which cleans everything up and gets us to a nice baseline from which we can go in and add configuration.

I think the world in which anyone genuinely needs these config options is pretty small, so I would be in favour of deferring those design decisions until well into the future.

JedWatson (approved):
👍 agreed, let's do this

PR #4473, opened 2020-12-02
Add a resolveFields argument to the item API methods

timleslie:


.changeset/brown-melons-jump.md
```diff
@@ -0,0 +1,47 @@
...
+```js
+export const extendGraphqlSchema = graphQLSchemaExtension({
+  typeDefs: `
+    type Mutation {
+      topPost(userId: ID): Post
+    }
+  `,
+  resolvers: {
+    Mutation: {
+      topPost: async (root, { userId } : { userId: string }, context) => {
```
JedWatson, line 36: Nit: this doesn't need to be async, it'll just return the promise.

That's not going to block approval, because I want to get this unblocked but (unless I missed something) it'd be good to fix that before merge.

JedWatson (approved):
Approved with two comments:

1. Check my note on the changeset
2. For completeness, we should also update the rest of the GraphQL queries in the ecommerce example to use the new API (and remove the server-side-graphql-client dependency) but I'm OK if that happens in a separate PR

.changeset/brown-melons-jump.md
timleslie: Good catch 👍 
```suggestion
      topPost: (root, { userId } : { userId: string }, context) => {
```

Issue #4261, opened 2020-11-15
Input validation is not working properly


osterreich:
# Bug report

## Describe the bug

Strange behavior of `addFieldValidationError()` function in `validateInput` hook.

## To Reproduce

1. Add the following field to Keystone List:

```javascript
    test: {
        type: Text,
        hooks: {
            validateInput: async ({ addFieldValidationError }) => {
                addFieldValidationError("This error message expected to be displayed...");
            }
        },
    },
```

2. In Admin UI edit any field of the document and click Save.

3. In the console logged the following output: 
```[GraphQL error]: Message: You attempted to perform an invalid mutation, Location: undefined, Path: updateBill```

4. Document resets to initial state **BUT** in the top right corner appears the green success toast.

## Expected behaviour

After saving document either error toast should be displayed, or the error message appears under the "Test" field. 
Green success toast is confusing in this case.

Issue #4173, opened 2020-11-09
Admin UI wrong success message in case of error


pahaz:
# Bug report

## Describe the bug

In admin UI I don't have access to perform `Save Changes` action. (`access: {update: false}`)
![image](https://user-images.githubusercontent.com/1394025/98604034-171be200-22f4-11eb-8156-1d27e15a2331.png)

After clicking the button I've received the success message.

I also have a console error:
```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
    in input (created by Context.Consumer)
    in EmotionCssPropInternal (created by ForwardRef)
    in ForwardRef (created by Context.Consumer)
    in EmotionCssPropInternal (created by CalendarDayField)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by FieldInput)
    in FieldInput (created by CalendarDayField)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by FieldContainer)
    in FieldContainer (created by CalendarDayField)
    in CalendarDayField (at Item/index.js:278)
    in ErrorBoundary (at Item/index.js:277)
    in Render (at Item/index.js:248)
    in form (created by Context.Consumer)
    in EmotionCssPropInternal (at Item/index.js:43)
    in Form (at Item/index.js:245)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by Card)
    in Card (created by Context.Consumer)
    in EmotionCssPropInternal (at Item/index.js:244)
    in ItemDetails (at Item/index.js:426)
    in div (created by Context.Consumer)
    in Styled(div) (at Item/index.js:425)
    in main (at Item/index.js:423)
    in ItemProvider (at Item/index.js:422)
    in ItemPage (at Item/index.js:444)
    in Suspense (at Item/index.js:443)
    in Unknown (at client/index.js:55)
    in Route (at client/index.js:55)
    in Switch (at client/index.js:53)
    in ListProvider (at client/index.js:52)
    in ListPageWrapper (at client/index.js:85)
    in Route (at client/index.js:85)
    in Switch (at client/index.js:80)
    in Suspense (at client/index.js:79)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (at Nav.js:55)
    in Page (at Nav.js:560)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (at Nav.js:58)
    in PageWrapper (at Nav.js:505)
[...]

PR #4481, opened 2020-12-02
Implemented a workaround to show mutation errors

miguelvvs:
Looks like after the update of apollo client, apollographql/apollo-client#5708 is preventing to show error messages in mutation as reported #4173 #3476 and #4261, this change is a workaround since the issue in the apollo client has been open for almost a year.


miguelvvs:
@mitchellhamilton sorry to bother, but I wonder if you had a chance to take a look at this, is there something I can improve or do differently? we are very interested in getting this fixed in a future version release since this particular error is preventing us to upgrade our keystone implementation. Any help on this would be really appreciated thanks.

mitchellhamilton (approved):

PR #4483, opened 2020-12-03
Disabled the debug output by default in the Document field

JedWatson:
This snuck in with a recent commit, shouldn't be turned on by default.

Might be nice to add an option for it that's accessible without changing code, but that can happen later, just turn it off for now.


timleslie (approved):
👍

PR #4484, opened 2020-12-03
Move admin UI code into admin UI package

timleslie:


packages-next/keystone/src/lib/createExpressServer.ts
```diff
@@ -1,15 +1,11 @@
...
 import cors from 'cors';
-import next from 'next';
 import express from 'express';
 import { ApolloServer } from 'apollo-server-express';
 // @ts-ignore
 import { formatError } from '@keystonejs/keystone/lib/Keystone/format-error';
 import type { KeystoneSystem, KeystoneConfig } from '@keystone-next/types';
-
-const dev = process.env.NODE_ENV !== 'production';
+import { createAdminUIServer } from '@keystone-next/admin-ui/system';
```
gautamsi, line 8: should we also decouple the admin-ui part from the keystone package?

I suggest we provide the AdminUi information in config for dev/start script to work on. similar to existing keystone way. 

this will help in using a different admin-ui (in case)

packages-next/keystone/src/scripts/dev.ts
```diff
@@ -2,10 +2,11 @@ import path from 'path';
 import express from 'express';
 import { printSchema } from 'graphql';
 import * as fs from 'fs-extra';
+import prettier from 'prettier';
+import { generateAdminUI } from '@keystone-next/admin-ui/system';
```
gautamsi, line 6: this should come from `config` exported by keystone.ts (see above comment)

packages-next/keystone/src/lib/createExpressServer.ts
timleslie: More fine grained control over this kind of thing will come in future PRs 👍 . The goal of this PR is to have the right code sitting in the right place so we can make better decisions around exactly how/where the developer manages their admin UI.

This probably won't be the last PR which shuffles the existing code around, now that we've got a better feel for how this new code fits together.

JedWatson (approved):

PR #4485, opened 2020-12-03
Updating auth extensions to skip access control after the context API changes

JedWatson:
Having changed the itemAPI to use the current context by default, we need to update our internal use of it to create a new context that skips access control.

### DIFF
--- a/.changeset/few-pillows-lie.md
+++ b/.changeset/few-pillows-lie.md
@@ -1,0 +1,5 @@
+---
+"@keystone-next/auth": patch
+---
+
+Updating auth extensions to skip access control after the context API changes
--- a/packages-next/auth/src/gql/getBaseAuthSchema.ts
+++ b/packages-next/auth/src/gql/getBaseAuthSchema.ts
@@ -54,7 +54,8 @@ export function getBaseAuthSchema({
           }
 
           const list = context.keystone.lists[listKey];
-          const itemAPI = context.lists[listKey];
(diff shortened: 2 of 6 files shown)


timleslie (approved):
👍

timleslie (changes requested):
Actually, we should also make sure we correctly specify `resolveFields` in all of these use cases, but I don't think they're correct that the moment.

timleslie:
I think we only need:
```
          // Update system state
          const item = await itemAPI.createOne({
            data: { ...data, ...itemData },
            resolveFields: false,
          });
```

in `getInitFirstItemSchema.ts`.

JedWatson:
> Actually, we should also make sure we correctly specify resolveFields in all of these use cases, but I don't think they're correct that the moment.

I was going to do that in a separate PR, this gets master working again and I want to make sure none of the usages get missed

JedWatson:
I did take a look at the usage and it's fine as is, because all that's being used is the ID which is the default value of `returnFields`

```
const sessionToken = await context.startSession({ listKey, itemId: item.id });
```

timleslie:
The `item` is also in the return value of the mutation, which definitely needs to have `resolveFields: false` for it to work properly.

JedWatson:
@timleslie I've added `resolveFields: false` to all the itemsAPI usage, which means it's gone back to the original behaviour (I ended up finding another place where the item was returned expecting unresolved fields)

timleslie (approved):
Looks good 👍

PR #4492, opened 2020-12-03
Pass createContext rather than system to session functions

timleslie:
Narrows down the API surface area that we're exposing in this corner of the system to just what we need.

### DIFF
--- a/packages-next/keystone/src/session/index.ts
+++ b/packages-next/keystone/src/session/index.ts
@@ -1,13 +1,12 @@
 import { IncomingMessage, ServerResponse } from 'http';
 import * as cookie from 'cookie';
 import Iron from '@hapi/iron';
-import { execute, parse } from 'graphql';
 import {
   SessionStrategy,
   JSONValue,
   SessionStoreFunction,
   SessionContext,
-  KeystoneSystem,
+  CreateContext,
 } from '@keystone-next/types';
 
 // uid-safe is what express-session uses so let's just use it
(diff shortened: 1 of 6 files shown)


packages-next/keystone/src/session/index.ts
```diff
@@ -78,38 +77,28 @@ export function withItemData(createSession: CreateSession, fieldSelections: Fiel
...
-        const session = await get({ req, system });
+      get: async ({ req, createContext }) => {
+        const session = await get({ req, createContext });
+        const sudoContext = createContext({ skipAccessControl: true });
         if (
           !session ||
           !session.listKey ||
           !session.itemId ||
-          !system.adminMeta.lists[session.listKey]
+          !sudoContext.lists[session.listKey]
```
JedWatson, line 87: Yeah, this is much clearer, that should absolutely not have been depending on `adminMeta`

packages-next/keystone/src/session/index.ts
```diff
@@ -78,38 +77,28 @@ export function withItemData(createSession: CreateSession, fieldSelections: Fiel
...
-        const fields = fieldSelections[session.listKey] || 'id';
-        const query = parse(`query($id: ID!) {
-          item: ${gqlNames.itemQueryName}(where: { id: $id }) {
-            ${fields}
-          }
-        }`);
-        const args = { id: session.itemId };
-        const result = await execute(system.graphQLSchema, query, null, context, args);
-        // TODO: This causes "not found" errors to throw, which is not what we want
-        // TODO: Also, why is this coming back as an access denied error instead of "not found" with an invalid session?
```
JedWatson, line 104: Separate topic: I think we bungled this erroring stuff.

Right now we can't return errors because they're misleading. A "not found" comes back as "you don't have access" which leads to questions "what have I done wrong that I don't have access?".

When I say we bungled it, I mean "you don't have access" should come back as "this does not exist" and not the other way around. In my experience that's how every other system handles it when the difference is not technically discernible. And I say this because I've spent time recently debugging what I'm doing wrong that the code I wrote doesn't have the right access control on it, until I learned the quirk.

(didn't really mean to start this conversation here, just this comment was my marker for it, not the right place for that, but we should backlog it elsewhere)

cc/ @molomby to put this on your radar too

JedWatson (approved):

PR #4502, opened 2020-12-04
Fixed key missed in #4493, rename views to viewsIndex

JedWatson:
This key was missed in #4493 when `views` was renamed to `viewsIndex`, fixes it so the Admin UI loads again 😅


dominikwilkowski (approved):
LGTM

PR #4503, opened 2020-12-04
Added resolveFields: false to the authenticatedItem query

JedWatson:
Add a missing `resolveFields` argument in the `authenticatedItem` query so that the fields resolve properly again.

Fixes this bug in the Admin UI:

image


dominikwilkowski (approved):
LGTM

PR #4507, opened 2020-12-07
Changed resolveFields default to false to remove a footgun

JedWatson:
After sitting with our decision around `resolveFields`, and working through updating our usage of the items API to use the new option, I now think we ended up with a footgun and behaviour that's harder to document and predict.

Defaulting the argument to false removes an entire "mode" the API operates in, which simplifies the explanation, and also removes the footgun.

### The additional mode

Before this PR, the items API has three modes, and using it correctly requires understanding all of them and how to choose between them:

* Default mode (you just get the ID) - probably what you want when you're not doing anything with the data that is returned, e.g from an update or create where the result won't be returned from the query/mutation
* Resolve fields mode - what you want when you're doing something with the returned data, and have a need for data that the field resolvers provide like virtual field values, nested related data, etc.
* Unresolved fields mode - what you want when you're going to return the data from the query/mutation

As I've started to try and document / explain this, it's become clear that we could implement a much simpler model instead:

* Default mode, fields aren't resolved - what you want when you're going to return the data from the query/mutation, or if you don't need data from the fields after performing the operation
* Opt-in resolve fields mode - if you need this you'll know it, because you'll be trying to query a set of fields from the graph after performing an operation. Should not be returned from a query/mutation unless you have a custom type matching the fields selected.

### The foot gun

I've found (and fixed) a bunch of bugs where we were returning items queried through the items API that didn't have the `resolveFields: false` argument passed correctly.

This is subtle and important: there's absolute no indicator to the developer they're doing something wrong, by using the default implementation of the API and returning it. The result even looks like it could be a legitimate data condition (why doesn't my User have a name? did I set one on this data I'm looking at? oh, yeah I did. Why isn't that coming through? am I not passing it correctly? ... and so forth until the real answer presents itself: you didn't pass this argument to this function)

It's a bad experience to do _probably not what you meant by default_, and requires the developer to understand the intricacies of the `resolveFields` argument in _all three modes_ to use it correctly; and there's no error we can catch, or hint we can provide to help them.

Compared to after this change: naive usage of the lists API to load data and return it from a custom query/mutation will just work. Then, when you want to load more data from the graph (e.g in the `ecommerce` example: https://github.com/keystonejs/keystone/blob/master/examples-next/ecommerce/mutations.ts#L23-L34)

* It'll be immediately apparent in the function you're writing that this data doesn't exist on the returned object;
* The developer says "oh hey how do I get the additional data?"
* There's an obvious answer: specify the fields you want to resolve and we'll resolve them for you

### Summary

I know we had our reasons for designing this API the way we did, but I believe "how easy is this to explain vs. use incorrectly" should override them, and to solve the problems I've outlined above, we should change the default.


.changeset/five-students-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-next/keystone": major
+---
+
+Changed resolveFields default to false
```
timleslie, line 5:
```suggestion
Changed `resolveFields` default from `"id"` to `false` in the items API.
```

jesstelford:
I agree strongly with this change 👍

> - The developer says "oh hey how do I get the additional data?"
> - There's an obvious answer: specify the fields you want to resolve and we'll resolve them for you

Could we make `There's an obvious answer` even _more_ obvious? I'm worried someone might miss the meaning of the parameter `resolveFields` when scanning through the docs.

_Aside: I am totally open to this just being a thing in my head though... For some reason I think of `resolve` as different and distinct to `run the resolver`. ie; in executing `.findOne(...)`, my mental model is "I want to fetch the data for this item from the database", and the word "fetch" here is a synonym for "resolve"._

Suggestions:

1. Rename `resolveFields` to `withFieldResolvers` or `asGraphQL` or something else which gives a stronger indication that the parameter will answer the question you posed _"oh hey how do I get the additional data?"_
2. Rename the existing functions to suffix `Raw`; `.findOne()` becomes `.findOneRaw()`. Then, add back in the original functions, but with a slightly altered implementation:

```javascript
  const findOneRaw = ({ resolveFields = false, ...rawArgs }) => {
    const args = getArgs.findOne(rawArgs) as { where: { id: string } };
    if (resolveFields) {
      return getItem({ listKey, context, returnFields: resolveFields, itemId: args.where.id });
    } else {
      return list.itemQuery(args, context);
    }
  },
  return {
    findOneRaw,
    findOne({ resolveFields, ...rawArgs }) {
      if (!resolveFields) { throw new Error('Must specify fields to resolve') }
      return findOneRaw(rawArgs);
    }
  }
```

gautamsi:
I am in favour of `.findOneRaw()` so that it is obvious. I do like the feature to get raw fields so that I don't have to update returned fields every time I add logic with more fields in list.

also in favour of renaming `resolveFields` to `returnedFields` it may be obvious that if this is not provided then raw fields are returned.

timleslie:
This PR has been superseded by the decision to split the API into `lists` and `db.lists`. See #5456

PR #4510, opened 2020-12-07
Add support for new interfaces to @keystonejs/test-utils

timleslie:
The aim of this PR is to allow us to write tests targeting the GraphQL layer of projects written with the new interfaces.


JedWatson:
👍 Looks good

One question: right now we have a really clear fork between the new and current packages, this is the first thing (afaik) that breaks that. i.e the only way _into_ the new packages is _through_ the new packages, if that makes sense.

Not saying it's a problem, but it's a line we haven't crossed yet. Are we comfortable blurring this now? or would it be better to create an alternate entry into `tests-next` and expose this new API for the test-utils package from `@keystone-next/test-utils`?

timleslie:
I don't think there's any need to make our packages any more complicated, this should be fine as is.

JedWatson (approved):

PR #4525, opened 2020-12-08
Use adminMeta query in init item page

timleslie:
The aim of this page is to have the init item page in the auth package use a dynamic adminMeta query, so that we don't need to pass the static adminMeta object around at build time.

~This cracks open a bit of a can of worms, so there will need to be a few other changes happen before this can land, including:~
 * ~Adding `req` to the context object~
 * ~Passing `context` to `isAccessAllowed`~
 * ~Expanding the `FieldMeta` type.~


packages-next/auth/src/index.ts
```diff
@@ -339,6 +334,17 @@ export function createAuth<GeneratedListTypes extends BaseGeneratedListTypes>({
           return (await adminPageMiddleware(args)) ?? keystoneConfig?.ui?.pageMiddleware?.(args);
         },
         enableSessionItem: true,
+        isAccessAllowed: async (context: KeystoneContext) => {
+          // Allow access to the adminMeta data from the /init path to correctly render that page
+          // even if the user isn't logged in (which should always be the case if they're seeing /init)
+          const headers = context.req?.headers;
+          const url = headers?.referer ? new URL(headers.referer) : undefined;
+          return (
+            (url?.pathname === '/init' && url?.host === headers?.host) ||
```
mitchellhamilton, line 343: This should check that there are no items in the list otherwise people can just fake this and get the admin meta.

packages-next/auth/src/index.ts
```diff
@@ -339,6 +334,17 @@ export function createAuth<GeneratedListTypes extends BaseGeneratedListTypes>({
...
+            keystoneConfig.ui?.isAccessAllowed?.(context) ||
+            false
```
mitchellhamilton, line 345: This should be like this:

https://github.com/keystonejs/keystone/blob/481e456ac4158207436ddd9be18fdca0f27b6409/packages-next/admin-ui/src/system/createAdminUIServer.ts#L29-L31

packages-next/auth/src/pages/InitPage.tsx
```diff
@@ -20,20 +20,21 @@ import {
...
+  const { adminMeta } = useKeystone();
   const fields = useMemo(() => {
     const fields: Record<string, FieldMeta> = {};
-    Object.keys(serializedFields).forEach(fieldPath => {
+    fieldPaths.forEach(fieldPath => {
       // note that we're skipping the validation since we don't know the list key and
       // the validation will happen after the user has the created the initial item anyway
-      const field = serializedFields[fieldPath];
+      const field = adminMeta.lists[listKey].fields[fieldPath];
       const views = fieldViews[field.viewsIndex];
```
mitchellhamilton, line 38: Everything inside this useMemo should be replaced with
```ts
const fields: Record<string, FieldMeta> = {};
fieldPaths.forEach(fieldPath => {
  fields[fieldPath] = adminMeta.lists[listKey].fields[fieldPath];
})
return fields
```

.changeset/modern-taxis-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Updated `/item` page to use the `adminMeta` query.
```
mitchellhamilton, line 5:
```suggestion
Updated `/init` page to use the `adminMeta` query.
```

mitchellhamilton (approved):

PR #4542, opened 2020-12-09
Update keystone class API README.md

davideberlein:
Added missing `await` to the `executeGraphQL` examples so they are valid


timleslie (approved):
Nice catch, thanks for the PR!

PR #4544, opened 2020-12-10
Replace useCompiledBcrypt option with bcrypt option

mitchellhamilton:
This is in preparation for production builds of the new interfaces which will likely involve bundling the server-side code and bundlers are generally unhappy with requires for packages that aren't installed so this changes the way people specify that they want to use native bcrypt by providing the native bcrypt module rather than a boolean option.


.changeset/dirty-spies-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-next/fields': major
+'@keystonejs/fields': major
+---
+
+Replaced `useCompiledBcrypt` option with `bcrypt` option which accepts an alternative implementation of bcrypt(such as the native `bcrypt` npm package) in the password field type
```
timleslie, line 6: Could you add if an example here of how a user should change their code. e.g

If you had the following field definition:

```js
password: { type: Password, useCompiledBcrypt: true }
```

you will need to change it to:

```js
password: { type: Password, bcrypt: require('bcrypt') }
```

timleslie (approved):
👍

PR #4547, opened 2020-12-10
Remove allViews from KeystoneSystem

timleslie:
This PR avoids having to pass around the `allViews` array as part of the `system` object. Views can now be looked up by a precomputed hash value, which is a function of config object.


packages-next/admin-ui/src/admin-meta-graphql.ts
```diff
@@ -152,8 +152,8 @@ type KeystoneAdminUIFieldMeta = {
   label: Scalars['String'];
   isOrderable: Scalars['Boolean'];
   fieldMeta: Maybe<Scalars['JSON']>;
-  viewsIndex: Scalars['Int'];
-  customViews: Maybe<Scalars['Int']>;
+  viewsIndex: Scalars['String'];
```
mitchellhamilton, line 155: This shouldn't be called viewsIndex if it's not an index and whatever the naming is, it should be consistent with customViews

timleslie: Yeah, I agree. I didn't want to include a change to this in this PR a) because there are already a lot of moving parts and b) I don't know what a good name would be 😅 If you have good ideas for consistent names I'm happy to ship a follow-up PR with the new and improved names 👍

mitchellhamilton: > I didn't want to include a change to this in this

I'd much prefer it all in one PR tbh.

`viewsHash`

timleslie: Cool, I can get on board with `viewsHash`, will make the change 👍 . Are we happy with `customViews`? Should it be `customViewsHash`?

mitchellhamilton: `customViewsHash`

packages-next/types/src/admin-meta.ts
```diff
@@ -5,7 +5,7 @@ export type FieldControllerConfig<FieldMeta extends JSONValue | undefined = unde
   listKey: string;
   path: string;
   label: string;
-  customViews: Record<string, any>;
+  customViewsHash: Record<string, any>;
```
mitchellhamilton, line 8: This should just be customViews, it's not the hash

timleslie: Nice catch, overcooked the search and replace :-)

mitchellhamilton (approved):

PR #4548, opened 2020-12-10
build and start commands for new interfaces

mitchellhamilton:


timleslie (approved):
This all looks good to me 👍 

Could you follow up with a PR to add `dev`/`build`/`start` support to the various `examples-next` packages 🙏

PR #4550, opened 2020-12-11
Force the Admin UI to use consistent versions of react, react-dom and @keystone-next/admin-ui

mitchellhamilton:
This is basically the same as we do in the current Admin UI


timleslie (approved):
LGTM 👍

PR #4553, opened 2020-12-12
Fixed typo in initial-data.md

jackchengck:
In the many to one relationship part, there is a typo missing an s that will cause the code error.


JedWatson (approved):

JedWatson:
Good catch @jackchengck, thanks!

PR #4556, opened 2020-12-14
Document field normalization, fixes and tests

mitchellhamilton:


packages-next/fields-document/src/DocumentEditor/component-blocks/form.tsx
```diff
@@ -0,0 +1,223 @@
+import { useKeystone } from '@keystone-next/admin-ui/src';
```
JedWatson, line 1: Shouldn't have a `src`

```suggestion
import { useKeystone } from '@keystone-next/admin-ui';
```

JedWatson (approved):

PR #4570, opened 2020-12-14
Show error on mutation errors 2

miguelvvs:
Exactly the same solution proposed on https://github.com/keystonejs/keystone/pull/4481 for https://github.com/keystonejs/keystone/issues/4173 and https://github.com/keystonejs/keystone/issues/4261  but now implemented in CreateItemModal.js  and UpdateManyItemsModal.js. In case we don't feel so comfortable duplicating code I also created this branch https://github.com/miguelvvs/keystone/commit/cb1d9ecc530bdf0ffecd6b438a78208a54993e1c, however, I consider this is a cleaner solution. Please let me know.


timleslie:
Fixed in #4651

PR #4573, opened 2020-12-14
Copy rather than mutate in initConfig

timleslie:
This is needed if we want to use the same config object multiple times, which is the case when running tests.


gwyneplaine (approved):
nice, lgtm

PR #4577, opened 2020-12-16
Document field renderer

mitchellhamilton:
I'd quite like this to be a separate package so that people who have their frontend and backend separate don't have to pull in all the deps on fields-document but curious to hear your thoughts @JedWatson


gautamsi:
I would want this in the new package rather than core packages

JedWatson:
>  curious to hear your thoughts @JedWatson

Yep, let's make this a `@keystone-next/document-renderer` package

JedWatson (approved):

PR #4584, opened 2020-12-17
Remove script element from debugging

mitchellhamilton:
I was using this a while ago for some debugging and probably forgot to remove it.


timleslie (approved):
👍

PR #4605, opened 2020-12-22
Convert session package to TypeScript

timleslie:


packages/session/lib/session.ts
```diff
@@ -1,9 +1,45 @@
...
+
+// FIXME: In the future this types will need to come from Keystone itself.
+type _Item = { id: string };
+type _List = {
+  key: string;
+  adapter: { itemsQuery: (args: { where: { id: string } }) => Promise<_Item[]> };
+};
+type _Keystone = { lists: Record<string, _List> };
+
+declare module 'express' {
```
mitchellhamilton, line 16: This module augmentation(and the one below it) will probably cause issues. This means that anyone who imports @keystonejs/session(including transitively) will see these definitions and it could be wrong/conflict with the users types, you should probably remove it.

packages/session/lib/session.ts
```diff
@@ -1,9 +1,32 @@
-const cookieSignature = require('cookie-signature');
```
mitchellhamilton, line 1: Sorry, missed this last time, could you move this file to src?

mitchellhamilton (approved):

PR #4608, opened 2020-12-23
Fix a test and always run package unit tests

mitchellhamilton:
~~Not sure why this only started failing after #4606 was merged and not in the PR but ¯\\\_(ツ)_/¯~~

It was because the package unit tests were skipped for packages-next. I've changed it so they always run now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -112,22 +112,18 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -112,22 +112,18 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
-        if: needs.should_run_tests.outputs.shouldRunTests == 'true'
```
timleslie, line 115: 🚀

timleslie (approved):
👍

PR #4611, opened 2020-12-25
[WIP] List Plugins for new-interface

gautamsi:
trying to port list plugins for new interface

- [x] atTracking - in progress - initial try in this wip so far
- [x] byTracking
- [ ] logging
- [ ] singleton

```ts
list(withAtTracking({...listConfig}, {...atTrackingOptions}));
```

caveat is that we would have to wrap it multiple time and also makes it difficult for adding config.

@timleslie @JedWatson  can you provide feedback on this approach?


gautamsi:
@timleslie can I get some feedback to complete this? if i am going in right direction or not

gautamsi:
@timleslie is this in roadmap, should I continue work on this?

timleslie:
Hi @gautamsi, at the moment we're not looking to support these features until we've finalised the new API implementation.

gautamsi:
Okay

gautamsi:
it is live here - https://www.npmjs.com/package/@keystonejs-contrib-next/list-plugins

PR #4623, opened 2020-12-29
pattern matching search that works with sqlite

7opf:
Admin UI for TODO demo app fails to display Todos due to unsupported operator `~*` when using a SQLite backend. [Using `like`](http://knexjs.org/#Builder-where) instead fixes this.


timleslie:
This change does not result in equivalent code. The `~*` operator performs insensitive regex searches, while the `LIKE` operator performs a case sensitive pattern match.

The knex adapter does not support sqlite, so this is not an issue we plan to address.

PR #4657, opened 2021-01-13
Replace system argument to createAdminUIServer with createContext

timleslie:


gwyneplaine (approved):
@timleslie lgtm, so we're no longer giving createAdminUIServer access to the entire system object, just the createContext fn?

timleslie:
> @timleslie lgtm, so we're no longer giving createAdminUIServer access to the entire system object, just the createContext fn?

@gwyneplaine Yep, that's the plan. Passing in the minimal thing required, rather than the kitchen sink 👍

PR #4660, opened 2021-01-14
Convert test-utils package to TypeScript

timleslie:
This is a first pass of adding types to this package. We'll need to revisit this once more of the core of keystone is correctly typed.

### DIFF
--- a/packages-next/types/src/base.ts
+++ b/packages-next/types/src/base.ts
@@ -18,6 +18,7 @@ export type BaseKeystone = {
     }
   ) => BaseKeystoneList;
   connect: (args?: any) => Promise<void>;
+  disconnect: () => {};
   lists: Record<string, BaseKeystoneList>;
   createApolloServer: (args: { schemaName: string; dev: boolean }) => any;
   getTypeDefs: (args: { schemaName: string }) => any;
--- a/packages-next/types/src/base.ts
+++ b/packages-next/types/src/base.ts
@@ -18,7 +18,7 @@ export type BaseKeystone = {
     }
   ) => BaseKeystoneList;
   connect: (args?: any) => Promise<void>;
-  disconnect: () => {};
+  disconnect: () => Promise<void>;
   lists: Record<string, BaseKeystoneList>;
   createApolloServer: (args: { schemaName: string; dev: boolean }) => any;
   getTypeDefs: (args: { schemaName: string }) => any;
--- a/packages/test-utils/package.json
+++ b/packages/test-utils/package.json
@@ -20,10 +20,5 @@
     "mongodb-memory-server-core": "^6.9.2",
     "supertest-light": "^1.0.3"
   },
-  "preconstruct": {
-    "entrypoints": [
-      "index.ts"
-    ]
-  },
   "repository": "https://github.com/keystonejs/keystone/tree/master/packages/test-utils"
 }
--- a/packages/test-utils/package.json
+++ b/packages/test-utils/package.json
@@ -4,6 +4,8 @@
   "version": "10.1.3",
   "author": "The KeystoneJS Development Team",
   "license": "MIT",
+  "main": "dist/test-utils.cjs.js",
+  "module": "dist/test-utils.esm.js",
   "engines": {
     "node": ">=10.0.0"
   },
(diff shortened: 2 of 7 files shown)


packages/test-utils/package.json
```diff
@@ -18,5 +20,10 @@
     "mongodb-memory-server-core": "^6.9.2",
     "supertest-light": "^1.0.3"
   },
+  "preconstruct": {
+    "entrypoints": [
+      "index.ts"
+    ]
+  },
```
mitchellhamilton, line 27: This is the default so it's not necessary

```suggestion
```

packages/test-utils/src/index.ts
```diff
@@ -160,8 +187,13 @@ function teardownMongoMemoryServer() {
   return stopping;
 }
 
-function _keystoneRunner(adapterName, tearDownFunction) {
-  return function (setupKeystoneFn, testFn) {
+type Setup = { keystone: Keystone<string> | BaseKeystone; context: KeystoneContext };
```
mitchellhamilton, line 190: `KeystoneContext` is the wrong type, no? It's for the new interfaces context which is quite different to current keystone context?

packages/test-utils/src/index.ts
```diff
@@ -189,16 +221,18 @@ function _keystoneRunner(adapterName, tearDownFunction) {
...
+  ) {
     const { keystone, app } = await setupKeystone(adapterName);
     await keystone.connect();
     return { keystone, app };
   };
 }
 
-function _after(tearDownFunction) {
-  return async function (keystone) {
+function _after(tearDownFunction: () => Promise<any> | void) {
```
mitchellhamilton, line 234:
```suggestion
function _after(tearDownFunction: () => Promise<void> | void) {
```

packages/test-utils/src/index.ts
```diff
@@ -160,8 +187,13 @@ function teardownMongoMemoryServer() {
   return stopping;
 }
 
-function _keystoneRunner(adapterName, tearDownFunction) {
-  return function (setupKeystoneFn, testFn) {
+type Setup = { keystone: Keystone<string> | BaseKeystone; context: KeystoneContext };
+
+function _keystoneRunner(adapterName: AdapterName, tearDownFunction: () => Promise<any> | void) {
```
mitchellhamilton, line 192:
```suggestion
function _keystoneRunner(adapterName: AdapterName, tearDownFunction: () => Promise<void> | void) {
```

packages-next/types/src/base.ts
```diff
@@ -18,6 +18,7 @@ export type BaseKeystone = {
     }
   ) => BaseKeystoneList;
   connect: (args?: any) => Promise<void>;
+  disconnect: () => {};
```
mitchellhamilton, line 21:
```suggestion
  disconnect: () => Promise<void>;
```

packages/test-utils/package.json
timleslie: Ah cool, that's what I thought. I was having trouble getting VS code to play nice and this was part of the experiment to make it fly. I shall remove 👍

packages/test-utils/src/index.ts
timleslie: No, this is for the new interfaces. The test-utils package support both old and new packages.

mitchellhamilton: So you're only planning to make tests for the new interfaces use TS?

packages/test-utils/src/index.ts
timleslie: I had this initially, but some of our tear down functions don't necessarily return void, so changed to any out of laziness. I think void is the right type though, I'm going to go in and fix those tear down functions to return void rather than arbitrary promises 👍

packages/test-utils/src/index.ts
timleslie: As above 👍

packages/test-utils/src/index.ts
timleslie: For now, yes.

packages/test-utils/src/index.ts
mitchellhamilton: void is a bit of a weird type because essentially anything is assignable to it in some cases like below(but it's still useful over any because it conveys that the return value should be ignored)

```ts
function something(cb: () => void) {}

something(() => '')
```

mitchellhamilton (approved):

PR #4662, opened 2021-01-14
Remove KeystoneSystem type

timleslie:
This is no longer a useful grouping of objects in the system.

### DIFF
--- a/packages-next/keystone/src/lib/createSystem.ts
+++ b/packages-next/keystone/src/lib/createSystem.ts
@@ -3,7 +3,7 @@ import { MongooseAdapter } from '@keystonejs/adapter-mongoose';
 import { KnexAdapter } from '@keystonejs/adapter-knex';
 // @ts-ignore
 import { PrismaAdapter } from '@keystonejs/adapter-prisma';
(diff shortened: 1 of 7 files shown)


packages-next/keystone/src/lib/createSystem.ts
```diff
@@ -73,7 +73,7 @@ export function createKeystone(config: KeystoneConfig) {
   return keystone;
 }
 
-export function createSystem(config: KeystoneConfig): KeystoneSystem {
+export function createSystem(config: KeystoneConfig) {
```
gwyneplaine, line 76: Should this be replaced with a substitute return type, instead of having the return type removed entirely?

gwyneplaine (changes requested):

packages-next/keystone/src/lib/createSystem.ts
timleslie: The return type is correctly inferred from the rest of the type system, so there's no need for an explicit return type.

gwyneplaine (approved):

PR #4675, opened 2021-01-17
Define static (non-configurabale) paths

timleslie:
These particular paths can't ever be part of the config, because we need them in order to find the config itself!


gwyneplaine (approved):
lgtmtm

molomby (approved):
Sure

PR #4691, opened 2021-01-19
Update PrismaAdapter with new migration support

timleslie:
Prisma migrate has been upgraded from "experimental" (alpha) to "preview" (beta), and so our prisma adapter needs some updates to go along for the ride.


gwyneplaine (approved):
lgtmtm

PR #4707, opened 2021-01-25
Add support Decimal field type in Prisma adapter

timleslie:


gwyneplaine (approved):
lgtm, but do we need tests for scale and precision work as expected  as per your slack comment @timleslie ?

timleslie:
> lgtm, but do we need tests for scale and precision work as expected as per your slack comment @timleslie ?

So it turns out we do have some very rudimentary coverage of these properties, although they could definitely be improved. I think they're sufficient to unblock this for now, but I would definitely like to add them in the future.

PR #4709, opened 2021-01-25
Restrict the ways in which a listAdapterClass can be specified

timleslie:
Our APIs don't need to support a use case of a user setting a custom list adapter class.

This PR doesn't prevent them from doing it themselves, they just need to set `.listAdapterClass` after constructing the base adapter object.


molomby (approved):
Sure

Issue #3164, opened 2020-06-16
Relationships throwing error


junaid33:
# Bug report

## Describe the bug

I'm trying to create relationships between multiple lists but the Admin UI is throwing an error.

## To Reproduce

Steps to reproduce the behaviour. Please provide code snippets or a repository:

1. Clone this repo: https://github.com/junaid33/ks-error-repro
2. Add local postgres connection URL to .env
3. npm install, npm run create-tables, npm run dev
4. Go to localhost:3000/admin/graphql
5. Enter this to create a user: 

mutation {
  createUser(
    data: {
      name: "Keystone"
      email: "admin@keystonejs.com"
      password: "password"
    }
  ) {
    id
    name
    email
  }
}

6. Log into localhost:3000/admin using created log-in.
7. Click on OrderItem, ChannelItem, and Match and you will see errors

## Expected behaviour

OrderItem, ChannelIem, and Match should not have errors and should be able to be created.

## Screenshots

These are screenshots on the error 

<img width="1030" alt="Screen Shot 2020-06-16 at 5 36 25 PM" src="https://user-images.githubusercontent.com/34615258/84837703-5de40500-affe-11ea-8f36-6695dc8705f2.png">
<img width="986" alt="Screen Shot 2020-06-16 at 5 33 06 PM" src="https://user-images.githubusercontent.com/34615258/84837705-5e7c9b80-affe-11ea-8603-f1914dc7478d.png">

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]

junaid33:
May be related to: https://github.com/keystonejs/keystone/issues/2828

And it seems to work fine with MongoDB, just not Knex/Postgres

junaid33:
I was able to fix the "Cannot destructure propety 'rel' of 'fieldAdapter'" error. I didn't have the User relationship added to the ChannelItem and ShopItem. Once added, the error went away.

The 

`GraphQL error: select "t0".* from "public"."Match" as "t0" left outer join "public"."User" as "t0__user" on "t0__user"."id" = "t0"."user" where true and ("t0__user"."id" = $1) order by "input" ASC limit $2 - column "input" does not exist` '

error is still there for the Match list. It works on Mongoose but throws error on Knex/Postgres.

josemf:
@junaid33 in case you are still wondering,

[...]

PR #4719, opened 2021-01-26
Update adapter-knex.js

josemf:
Fixes an issue where List that have a relationship field as its first field will have this field used for orderBy clause in query. This will fail the query for those relationships where fields don't exist in the List.

Issue: https://github.com/keystonejs/keystone/issues/3164


timleslie (changes requested):
Hi @josemf , thanks for tracking down this bug. I think you're changes will help solve the issue.

In the future we will need to address the issue of sorting by relationship fields which do not have a `realKey`, but that can be solved once this fix has gone through.

To get through CI could I ask you to run `yarn format` so that `yarn lint` passes, and also to add a `changeset` to capture the changes for the changelog (`yarn changeset`). Thanks!

josemf:
@timleslie thank you for following up

I followed your instructions on a new PR for this issue.
https://github.com/keystonejs/keystone/pull/4791

You can close this one,

PR #4726, opened 2021-01-27
Simplify code in parseAccess

timleslie:
The inlining of code from `validateGranularConfigTypes` and `parseAccessCore` intentionally violates DRY for now in order to a) make adding types easier and b) to allow for a future refactoring with hopefully simpler code.

### DIFF
--- a/.changeset/two-pots-vanish.md
+++ b/.changeset/two-pots-vanish.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/access-control': patch
+---
+
+Refactored `parseAccess` and inlined the code from `validateGranularConfigTypes` and `parseAccessCore`.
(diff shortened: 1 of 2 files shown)


gwyneplaine (approved):
lgtm
> The inlining of code from `validateGranularConfigTypes` and `parseAccessCore` intentionally violates DRY for now in order to a) make adding types easier and b) to allow for a future refactoring with hopefully simpler code.

is the fact that we should / need to do b) after the typescript conversion tracked in our backlog?

timleslie:
> lgtm
> 
> > The inlining of code from `validateGranularConfigTypes` and `parseAccessCore` intentionally violates DRY for now in order to a) make adding types easier and b) to allow for a future refactoring with hopefully simpler code.
> 
> is the fact that we should / need to do b) after the typescript conversion tracked in our backlog?

Not really, I anticipate it will be done as part of the conversion process, e.g. this series of PRs. If we close out the conversion and it's still not completed then we can log is as a future task.

PR #4729, opened 2021-01-28
Update types in utils package

timleslie:


packages/utils/src/index.ts
```diff
@@ -12,27 +12,32 @@ export const escapeRegExp = (str: string) =>
...
   );
 
-export const resolveAllKeys = async <T>(obj: Record<string, Promise<T>>) => {
-  const returnValue: Record<string, T> = {};
-  const errors: Record<string, unknown> = {};
+export const resolveAllKeys = async <V, T extends Record<string, Promise<V>>, K extends keyof T>(
+  obj: T
+) => {
+  const returnValue = {} as Record<K, V>;
+  const errors = {} as Record<K, unknown>;
```
mitchellhamilton, line 38: This will make it correctly type inputs like `{thing: Promise.resolve(1),another: Promise.resolve("something")}` as `Promise<{thing:number,another: string}>` rather than `Record<"thing" | "another", number | string>`

```suggestion
type Await<TPromise extends Promise<any>> = TPromise extends Promise<infer Value> ? Value : never;

export const resolveAllKeys = async <T extends Record<string, Promise<any>>>(obj: T) => {
  const returnValue = {} as { [Key in keyof T]: Await<T[Key]> };
  const errors = {} as Record<keyof T, unknown>;
```

packages/utils/src/index.ts
```diff
@@ -12,27 +12,32 @@ export const escapeRegExp = (str: string) =>
   (str || '').replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
 
 // { key: value, ... } => { key: mapFn(value, key), ... }
-export const mapKeys = <T, U>(
-  obj: Record<string, T>,
-  func: (value: T, key: string, obj: Record<string, T>) => U
-): Record<string, U> =>
-  Object.entries(obj).reduce((acc, [key, value]) => ({ ...acc, [key]: func(value, key, obj) }), {});
+export const mapKeys = <T, K extends keyof T, R>(
```
mitchellhamilton, line 15: This `K` type parameter isn't really useful here and the ones serving a similar since it'll also be keyof T, never more specific. It'd be better to do `T extends Record<string, any>` and when you need the key, do `keyof T`

packages/utils/src/index.ts
```diff
@@ -69,34 +71,33 @@ export const unique = <T>(arr: T[]): T[] => [...new Set(arr)];
...
-export const omit = <T>(obj: Record<string, T>, keys: string[]) =>
-  omitBy(obj, value => keys.includes(value));
+export const omit = <T, K extends keyof T>(obj: T, keys: K[]) =>
+  omitBy(obj, value => keys.includes(value as K)) as Omit<T, K>;
 
 // [{ k1: v1, k2: v2, ...}, { k3: v3, k4: v4, ...}, ...] => { k1: v1, k2: v2, k3: v3, k4, v4, ... }
 // Gives priority to the objects which appear later in the list
-export const objMerge = <T>(objs: Record<string, T>[]) =>
-  objs.reduce((acc, obj) => ({ ...acc, ...obj }), {});
+export const objMerge = <T>(objs: T[]) => objs.reduce((acc, obj) => ({ ...acc, ...obj }), {} as T);
```
mitchellhamilton, line 91: This won't give you the type you want. tbh, I think this should just be removed and the usages replaced with `Object.assign({}, ...objs)` but if you want to keep it:
```suggestion
type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (
  x: infer R
) => any
  ? R
  : never;

export const objMerge = <T extends any[]>(
  objs: [...T]
): UnionToIntersection<T[number]> => Object.assign({}, ...objs);
```

mitchellhamilton (approved):

PR #4738, opened 2021-02-01
Added APP_URL for ability to rewrite app url

MurzNN:
At now KeystoneJS application url is hardcoded as `http://localhost:${port}`. But very often KeystoneJS instances works not directly at port, but behind reverse proxies, so they have working URLs like `https://example.com/api` or `https://api.example.com`.
And when launching Keystone on remote server, the provided urls with "localhost" isn't working.

This PR adds ability to set right external URL to access Keystone interface, via `APP_URL` environment variable or `--app_url` cli argument.

As result, users will see right URL in output after launching Keystone, here is example:
```
$ yarn dev --app_url https://staging.mycoolproject.com/v2 --port=12345
...
✔ Keystone server listening on port 12345
...
✔ Keystone instance is ready at http://localhost:12345 🚀
🔗 Keystone Admin UI:   https://staging.mycoolproject.com/v2/admin
🔗 GraphQL Playground:  https://staging.mycoolproject.com/v2/admin/graphiql
🔗 GraphQL API:         https://staging.mycoolproject.com/v2/admin/api
```

Please review my proposal, and comment if something is implemented via not so good way. Thanks!


timleslie:
Hi @MurzNN, this looks good to me. Could you please add a changeset with a `minor` bump to the `keystone` package (see changeset bot comment above for instructions). Also, there's a file `test.patch` which I don't think was meant to be included in this PR. Thanks!

gautamsi:
Cli option should be `--app-url`

MurzNN:
@timleslie Thanks for reviewing, fixed! Also I renamed cli option to `--app-url`.

.changeset/happy-peas-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/keystone': minor
+---
+
+Added APP_URL option for set custom URL of app web interface
```
timleslie, line 5:
```suggestion
Added an `APP_URL` environment variable and `--app-url` command line option to a set custom URL to be reported when starting the server.
```

timleslie (approved):
👍

PR #4754, opened 2021-02-02
Init welcome to keystone

JedWatson:
This screen will welcome new users to Keystone, linking to the docs / twitter / GitHub project and offering to sign them up for our newsletter.

Still to do:

- [x] Mailing list integration
- [x] Design and copy
- [x] Boyscout comment code
- [x] Create next app for signup API and deploy to signup.keystonejs.com via vercel


packages-next/auth/src/pages/InitPage.tsx
```diff
@@ -12,20 +14,104 @@ import { SigninContainer } from '../components/SigninContainer';
...
+              // any status that isn't 200 we assume is a failure
+              // which we want to surface to the user
+              res.json().then(({ error }) => setError(error));
+            } else {
+              router.push((router.query.from as string | undefined) || '/');
+            }
+          })
+          .catch(err => {
+            // network errors or failed parse
+            setError(err);
```
mitchellhamilton, line 67:
```suggestion
            setError(err.toString());
```

packages-next/auth/src/pages/SigninPage.tsx
```diff
@@ -58,6 +58,7 @@ export const SigninPage = ({
       router.push((router.query.from as string | undefined) || '/');
     }
   }, [rawKeystone.authenticatedItem, router.query.from]);
+
   return (
```
timleslie, line 62:
```suggestion
  return (
```

packages-next/auth/src/pages/InitPage.tsx
```diff
@@ -65,39 +193,61 @@ export const InitPage = ({
...
+          // // Create the first item in the database.
           createFirstItem({
             variables: {
               data,
             },
           })
             .then(() => {
+              // refetch admin meta
               reinitContext();
-              router.push('/');
```
gwyneplaine, line 100: No longer needed, this gets handled by the effect further up

packages-next/auth/src/components/SigninContainer.tsx
```diff
@@ -22,7 +22,7 @@ export const SigninContainer = ({ children }: SigninContainerProps) => {
       <Box
         css={{
           background: colors.background,
-          minWidth: 600,
+          width: 600,
```
gwyneplaine, line 25: changed this to have uniform width between screens.
this component is only used by 3 components, Welcome, InitPage and Signin 

Before the change InitPage and Signin were the same length coincidentally due to having similar contents. However Welcome was far too large due to much copy. 

This change resolves this issue.

packages-next/auth/src/components/Icons.tsx
```diff
@@ -0,0 +1,81 @@
+/* @jsx jsx */
+
+import { forwardRefWithAs, jsx } from '@keystone-ui/core';
+
+const Icon: React.ComponentType = ({ children, ...props }) => {
```
mitchellhamilton, line 5: pls don't use `ComponentType`, type the props instead since `ComponentType` includes children as a prop always(which sure, is correct in this case but it's wrong generally and someone reading it probably won't know that `ComponentType` includes `children`). I'll make a PR enforcing that it's not used with ESLint.

gwyneplaine: sure thing, no idea that was a rule for us 👍

.changeset/slimy-dogs-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/auth': minor
+---
+
+Add Welcome page with mailing list signup flow to InitPage
```
timleslie, line 5:
```suggestion
Added Welcome page with mailing list signup flow to InitPage.
```

.changeset/eighty-geckos-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+minor refactors to router events on InitPage, minor changes to SignupContainer styles
```
timleslie, line 5: Because changeset text ends up in changelogs we write it in past tense and full sentences (capital letters, full stops).

```suggestion
Refactored router events on InitPage, and made minor changes to SignupContainer styles.
```

packages-next/auth/src/types.ts
```diff
@@ -52,7 +52,7 @@ export type AuthConfig<GeneratedListTypes extends BaseGeneratedListTypes> = {
     /** Array of fields to collect, e.g ['name', 'email', 'password'] */
     fields: GeneratedListTypes['fields'][];
     /** Suppresses the second screen where we ask people to subscribe and follow Keystone */
-    skipKeystoneSignup?: boolean;
+    skipKeystoneWelcome?: boolean;
```
timleslie, line 55: I think this makes this into a `major` version bump, since it's potentially a breaking change for users.

yarn.lock
```diff
@@ -5979,7 +5979,7 @@ babel-plugin-macros@2.8.0, babel-plugin-macros@^2.0.0, babel-plugin-macros@^2.8.
     cosmiconfig "^6.0.0"
     resolve "^1.12.0"
 
-babel-plugin-remove-graphql-queries@2.7.2, babel-plugin-remove-graphql-queries@^2.7.2:
+babel-plugin-remove-graphql-queries@^2.7.2:
```
timleslie, line 5982: These `yarn.lock` changes look a bit suss to me. Can you revert `yarn.lock` and then re-run `yarn` to see if these changes are actually needed?

timleslie (changes requested):

.changeset/eighty-geckos-rush.md
gwyneplaine: argh i keep forgetting this, thanks Tim

packages-next/auth/src/types.ts
gwyneplaine: yep, good point 
Added a major changeset to account for this 👍

timleslie (approved):
Looking good 👍

PR #4761, opened 2021-02-04
Hide config.graphql.path until implemented

timleslie:
Exposing this on the type implies functionality that isn't currently implemented.


gwyneplaine (approved):
lgtm, if applicable and not there already, can we track this in JiRA?

timleslie:
> lgtm, if applicable and not there already, can we track this in JiRA?

Done 👍

PR #4763, opened 2021-02-04
Remove SchemaConfig type

timleslie:
The `SchemaConfig` type wasn't really helping, it's much more clear to have each of its components directly on the `KeystoneConfig` object.


gwyneplaine (approved):
lgtm

PR #4765, opened 2021-02-04
Reorganise type definitions to easier documenting

timleslie:
This PR groups all of the types required for `KeystoneConfig` to work into the `config` directory, which will make documenting this API much easier.

The location of types, and the order of fields within them have been changed, but there are no functional changes.


gwyneplaine (approved):
looks like a standard lift and shift to me, lgtm

PR #4768, opened 2021-02-04
Create indexes for relation adjacency table fields

kschingiz:
Seems when keystone creates a schema using mongoose adapter for N:N relations like `user_projects_project_users` we are missing indexes for `left_id` and `right_id` fields.
This fix will add indexes for left_id and right_id fields.


kschingiz:
I am not sure why Cloudinary tests were failed, I did not change anything there.

timleslie:
This looks like a useful addition, it looks like there's something not working with the specific implementation, but if you can get this to work then I'd definitely like to include it.

> I am not sure why Cloudinary tests were failed, I did not change anything there.

These tests require access to certain credentials which we don't expose for community PRs for security reasons. If those are the only failing tests then we can go ✅ the PR.

kschingiz:
@timleslie don't merge, that's simple fix broke my local project

kschingiz:
@timleslie I improved my patch:
1. Create index for any Relationship fields by default, currently it works only if you provide `isIndexed = true`, but now you can omit that field, indexes will be created by default
2. Create 2 way compound indexes for N:N relationship, indexes are:
```
    schema.index({ [leftFkPath]: 1, [rightFkPath]: 1 })
    schema.index({ [rightFkPath]: 1, [leftFkPath]: 1 })
```

packages/adapter-mongoose/lib/adapter-mongoose.js
```diff
@@ -88,6 +88,11 @@ class MongooseAdapter extends BaseKeystoneAdapter {
 
     schema.add({ [leftFkPath]: {} });
     schema.add({ [rightFkPath]: {} });
+
+    // create 2 way indexes for reference fields
+    schema.index({ [leftFkPath]: 1, [rightFkPath]: 1 });
```
gautamsi, line 93: is this creating combined index? 

why not create index for each field (left and right) separately without combining them, they are non unique so it would not have issues.

kschingiz: @gautamsi yep, I was also thinking if we really need to make it a compound index. But I found it useful, see this case:
Users and Projects have N:N relation.
```
query() {
    allProjects() {
       users() {
       }
    }
}
```
Currently keystone will request by project_left_id only, but if we add additional filter to users:
```
query() {
    allProjects(where: { id: 1 }) {
       users(where: { id: 2 }) {
       }
    }
}
```
mongodb will use IXSCAN only by one field Project_left_id, but to retrieve nested users with `users(id = 1)` it has to apply additional in memory filter because the second field is not indexed. So making compound index would resolve that. Anyway the second field is not going to take a lot of space.
Let me know if I am at the wrong direction.

gautamsi: I meant there could be two individual indexes. one for each field. So they both be indexed.

kschingiz: @gautamsi I am creating for both of them:
```
schema.index({ [leftFkPath]: 1, [rightFkPath]: 1 });
schema.index({ [rightFkPath]: 1, [leftFkPath]: 1 });
```

gautamsi: why not do this

```js
schema.index({ [leftFkPath]: 1 });
schema.index({ [rightFkPath]: 1 });
```

kschingiz: @gautamsi fixed

gautamsi: thanks, hopefully @timleslie likes the split index

kschingiz:
@timleslie @gautamsi any news?

.changeset/friendly-countries-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs/adapter-mongoose': patch
+---
+
+Create indexes for relation adjacency collection fields
```
timleslie, line 5:
```suggestion
Created indexes for relation adjacency collection fields
```

timleslie (approved):
Looks good, thanks for the PR 🙏

PR #4769, opened 2021-02-04
docs-next design update

JedWatson:
A few things going on here -

* I switched over to tailwind for the new website
* As part of that, I had to update to Next v10
* I cleaned up the site design so it's nicer, more to come

The next upgrade seems to work just fine across the website *and* the Admin UI so I'm pretty comfortable with it, @mitchellhamilton would be good to get you to check it as well


JedWatson:
Also I am not sure what to do about the type failure, looks like something has gone wrong -- it's failing on the webpack dependency of next 😕

timleslie:
Just leaving a note here that we had previously tried upgrading to Next 10 and had to revert it because $something went wrong. I'll see if I can track down the history of that.

timleslie:
Related to the types error: https://github.com/vercel/next.js/issues/21679

.changeset/happy-garlics-admire.md
```diff
@@ -0,0 +1,7 @@
+---
```
mitchellhamilton, line 1: This changeset should be removed since the update happened in another PR and that PR had a changeset

JedWatson: ✅ done, thanks

timleslie (approved):
Looks good 👍

PR #4123, opened 2020-11-02
Revert "fix(deps): update dependency next to v10 (#4055)"

mitchellhamilton:
This reverts commit 557639a0bc1d636ff723aa5b2a1d960a362978a3.

PR #4770, opened 2021-02-04
Upgrade Next.js to 10.0.5

timleslie:
We'll pin to `10.0.5` for now, because of https://github.com/vercel/next.js/issues/21679.

Also addresses some warnings in the example projects that arise from this upgrade.

Previous attempt at this upgrade was reverted in #4123.

Previous reversion was due to an issue in the `meetup` example project: https://github.com/keystonejs/keystone/pull/4055#issuecomment-720265756. I've manually tested this example with this update and everything seems to be working well now 👍


gwyneplaine (approved):
✅

JedWatson:
Thanks @timleslie 🙏

PR #4775, opened 2021-02-05
Update schema API docs

timleslie:


docs-next/pages/apis/schema.mdx
```diff
@@ -2,10 +2,219 @@ import { Markdown } from '../../components/Page';
...
+});
 ```
-import type { ListSchemaConfig } from '@keystone-next/types';
+
+This document will explain the configuration options which can be used with the `list()` function.
+
+## `fields`
+
+The `fields` option defines the names, types, and configuration of the fields in the list.
+The configuration option takes an object with field names as keys, and configured field types as values.
```
gwyneplaine, line 34: which configuration option are you referring to here? are you just referring to the `fields` option here? 

if yes 
```suggestion
This configuration option takes an object with field names as keys, and configured field types as values.
```

gwyneplaine (approved):

PR #4784, opened 2021-02-08
Reorder type definitions

timleslie:
Also remove unused `ListAdminUIConfig.path`.

### DIFF
--- a/packages-next/types/src/config/lists.ts
+++ b/packages-next/types/src/config/lists.ts
@@ -77,9 +77,9 @@ export type ListAdminUIConfig<
 > = {
   /**
    * The field to use as a label in the Admin UI. If you want to base the label off more than a single field, use a virtual field and reference that field here.
-   * @default 'name' if it exists, otherwise 'id'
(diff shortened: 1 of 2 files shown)


packages-next/types/src/config/lists.ts
```diff
@@ -79,20 +79,22 @@ export type ListAdminUIConfig<
    * The field to use as a label in the Admin UI. If you want to base the label off more than a single field, use a virtual field and reference that field here.
    * @default 'name' if it exists, otherwise 'id'
    */
-  labelField?: keyof Fields; // path of the field to use as the label for items in the list, defaults to 'name' or 'id'
+  labelField?: keyof Fields; // path of the field to use as the label for items in the list, defaults to 'label', 'name', 'title' or 'id'
```
mitchellhamilton, line 82: Can you update the @default bit above(it'll be shown to people in their editor) and probs just remove this line comment?

timleslie: Good catch, done.

mitchellhamilton (approved):

PR #4805, opened 2021-02-09
[POC] Allow fields to be grouped in admin app

kschingiz:
Hi, I have some django framework background, it has very good feature to group item fields into separated blocks. I was playing with keystone and decided to try to implement it.
This PR adds new `fieldGroups` field to schema config:
```
keystone.createList('Event', {
  fields: {
    name: { type: Text },
    status: { type: Select, options: 'draft, active', defaultValue: 'draft' },
    themeColor: { type: Text },
    startTime: { type: DateTime },
    durationMins: { type: Integer },
    description: { type: Text },
    locationAddress: { type: Text },
    locationDescription: { type: Text },
    maxRsvps: { type: Integer, defaultValue: 120 },
    isRsvpAvailable: { type: Checkbox, defaultValue: true },
  },
  fieldGroups: [
    {
      label: 'Base',
      fields: ['name', 'status', 'description', 'startTime']
    },
    {
      label: 'Location',
      fields: ['locationAddress', 'locationDescription']
    },
    {
      label: 'Misc',
      fields: ['maxRsvps', 'isRsvpAvailable', 'themeColor']
    },
  ]
});
```
Admin app will separate fields into 3 blocks: Base, Location, Misc and render their appropriate fields.
This feature will give us more control over:
1. Which fields should be hidden, in this case you just remove it from the fields array
2. In which order fields should be displayed
3. Logically separate fields into blocks

This is a very first commit, code still needs to be cleaned up.
The result of the schema above:
Screenshot from 2021-02-10 00-16-53

### DIFF
--- a/packages/app-admin-ui/client/pages/Item/index.js
+++ b/packages/app-admin-ui/client/pages/Item/index.js
@@ -55,6 +55,30 @@ const getCurrentValues = memoizeOne(getValues);
 
 const deserializeItem = memoizeOne((list, data) => list.deserializeItemData(data));
 
+const getFieldGroups = memoizeOne(list => {
+  const fieldGroups = list.fieldGroups || [];
+  if(fieldGroups.length) {
+    return fieldGroups.map(({ label, fields }) => {
+      const fieldsToRender = fields.map((fieldName) => {
+        const field = list.fields.find((field) => field.path === fieldName);
+        if(!field) {
+          throw new Error(`Field ${fieldName} not found`);
+        }
+        return field;
+      });
+      return {
+        label,
+        fields: fieldsToRender
+      };
+    });
+  }
+  return [
+    {
+      fields: getRenderableFields(list)
+    }
+  ];
+});
+
 const getRenderableFields = memoizeOne(list =>
   list.fields.filter(({ isPrimaryKey }) => !isPrimaryKey)
 );
@@ -76,6 +100,8 @@ const ItemDetails = ({ list, item: initialData, itemErrors, onUpdate }) => {
     onError: error => handleCreateUpdateMutationError({ error, addToast }),
   });
 
+  const fieldGroups = getFieldGroups(list);
+  console.log(fieldGroups);
   const getFieldsObject = memoizeOne(() =>
     arrayToObject(
       // NOTE: We _exclude_ read only fields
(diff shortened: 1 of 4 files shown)


gautamsi:
I like this.

alternate config (do not make changes before hearing from the core team, I am just giving my opinion)

```js
keystone.createList('Event', {
  fields: {
    name: { type: Text, group:  'Base' },
    status: { type: Select, options: 'draft, active', defaultValue: 'draft', group:  'Base'  },
    themeColor: { type: Text, group:  'Misc'  },
    startTime: { type: DateTime, group:  'Base'  },
    durationMins: { type: Integer  },
    description: { type: Text, group:  'Base'  },
    locationAddress: { type: Text, group:  'Location'  },
    locationDescription: { type: Text, group:  'Location'  },
    maxRsvps: { type: Integer, defaultValue: 120, group:  'Misc'  },
    isRsvpAvailable: { type: Checkbox, defaultValue: true, group:  'Misc'  },
  },
});
```
Only thing is devs would have to sort fields in way they want in display.

It would also be nice feature to have these groups collapsible and default collapse option as in original configs.

I believe this also qualifies for the UI customization rather than built in field feature. UI customization would be handled in separate file `./admin-ui/index.js` and would not clutter schema definition.

packages/app-admin-ui/client/pages/Item/index.js
```diff
@@ -55,6 +55,30 @@ const getCurrentValues = memoizeOne(getValues);
 
 const deserializeItem = memoizeOne((list, data) => list.deserializeItemData(data));
 
+const getFieldGroups = memoizeOne(list => {
+  const fieldGroups = list.fieldGroups || [];
+  if(fieldGroups.length) {
```
gautamsi, line 60: logic for `getRenderableFields()` is missing in this condition. fields should be filtered in the group based on output of `getRenderableFields`, also the empty groups should be filtered out.

kschingiz: I think my conf format is more comfortable for developers :) Because you quickly see how many groups are there + we can add new fields to the config like:
```
    {
      label: () => i18.t('Misc') // label resolver function
      fields: ['maxRsvps', 'isRsvpAvailable', 'themeColor'],
      isCollapsible: true, // can be collapsed or not
      collapsed: true // default collapsed state
    },
```

VinayaSathyanarayana: Thanks @kschingiz . I like this.

Please consider the following suggestions
``` javascript
fieldGroups: [
  // fieldGroups to be displayed in the same sequence as they are mentioned in the array
  {
    label: 'Base',
    fieldsInSequence: ['name', 'status', 'description', 'startTime']
  },
  {
    label: 'Location',
    fieldsInSequence: ['locationAddress', 'locationDescription']
  },
  {
    label: () => i18.t('Misc') // label resolver function
    fieldsInSequence: ['maxRsvps', 'isRsvpAvailable', 'themeColor'], 
    // Changed to indicate that the fields will be shown in the UI in the sequence specified
    isCollapsible: true, // can be collapsed or not
    collapsed: () => { return true } , // function that returns true or false
    displayFieldGroup: () => {  
      //  function that returns true or false
      //  This logic can be based on fields filled in some other group
      //  e.g. If a person is single, do not ask details of the spouse, 
      //  the spouseFieldGroup is not displayed
      if ( /*  logic to check if field group has to be displayed */ ) {
        return true;
      } else {
	return false
      }, 
    },
]
```

VinayaSathyanarayana: Another thought: Should the fieldGroups be part of the UI block as these are related to the UI / display?

kschingiz:
```
UI customization would be handled in separate file ./admin-ui/index.js
```
@gautamsi  I am new to keystone, do we have any UI customization already, if yes I can refactor it so we don't clutter schema. How to use this admin-ui/index.js file? I cannot find any references to it.

dcousens:
This pull request wasn't relevant now anyway, but this feature has been added in https://github.com/keystonejs/keystone/pull/8088

PR #4806, opened 2021-02-09
Add scalar field API docs

timleslie:
This is the start of documenting the fields API, beginning with the scalar fields. There are still a bunch of placeholders which need to be fleshed out, but the core stuff is now there 👍

### DIFF
--- a/docs-next/pages/apis/fields.mdx
+++ b/docs-next/pages/apis/fields.mdx
@@ -2,6 +2,564 @@ import { Markdown } from '../../components/Page';
 
 # Fields API
 
+The `fields` option of a [list configuration](./schema) defines the names, types, and configuration of the fields in the list.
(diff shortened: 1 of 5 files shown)


docs-next/pages/apis/fields.mdx
```diff
@@ -2,6 +2,564 @@ import { Markdown } from '../../components/Page';
...
+    ...
+  }),
+  ...
+});
+```
+
+## Common configuration
+
+All field types accept a common set of configuration options.
+All options are optional.
```
gwyneplaine, line 54: no nit, just found this phrase funny 😂

gwyneplaine (approved):
lgtm, common thread for all these PRs is that we can probably make the definitions clearer with a type signature later down the road. 

But everything is great so far for least effort / most bang for buck

PR #4807, opened 2021-02-10
Update prisma.md

yantakus:
Fix typo.


timleslie (approved):
Good catch, thanks for the PR!

PR #4812, opened 2021-02-11
First pass of architecture doc for document field

mitchellhamilton:


packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
+# Document editor architecture
+
+The document editor is using [Slate](https://github.com/ianstormtaylor/slate). To learn about Slate generally, go through https://docs.slatejs.org/walkthroughs and https://docs.slatejs.org/concepts. The sections on [Nodes](https://docs.slatejs.org/concepts/02-nodes) and [Locations](https://docs.slatejs.org/concepts/03-locations) contain especially important terminology.
+
+## Plugins
+
+For normalization and handling user input and some other things(you can see what they are by looking at the type definition for the `Editor` in Slate), you write [Slate "plugins"](https://docs.slatejs.org/concepts/07-plugins), plugins are not some special API, it means accepting an `Editor` object and returning it but modifying some of the properties.
```
timleslie, line 7:
```suggestion
For normalization and handling user input and some other things(you can see what they are by looking at the type definition for the `Editor` in Slate), you write [Slate "plugins"](https://docs.slatejs.org/concepts/07-plugins).
plugins are not some special API, it means accepting an `Editor` object and returning it but modifying some of the properties.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Rendering
+
+The rendering of the editor is mostly like any other bit of React code. See https://docs.slatejs.org/concepts/08-rendering for how rendering the document works in Slate.
```
timleslie, line 11: > mostly like any other bit of React code

Are there specific differences we should know about?

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+The editor is very performance-sensitive because it will re-render on every keystroke so keep these things in mind while writing code for rendering the editor:
+
+- Renderers for elements/leaves should never use `useSlate` or `useToolbarState` because they will change on every editor change, if they need to get the editor to do something in response to a user action, they should use `useStaticEditor` which will not cause re-renders
```
timleslie, line 15:
```suggestion
- Renderers for elements/leaves should never use `useSlate` or `useToolbarState` because they will change on every editor change. If they need to get the editor to do something in response to a user action, they should use `useStaticEditor` which will not cause re-renders.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+- All state derived from the document/selection that is shown in the toolbar should come from `useToolbarState` so that it is only computed once
+- Components that use `useToolbarState` and are shown when the user is typing(i.e. things that aren't in dialogs) should memoize the React elements they return with `useMemo` to avoid re-rendering when the state they consume from `useToolbarState` hasn't changed
```
timleslie, line 17:
```suggestion
- Components that use `useToolbarState` and are shown when the user is typing (i.e. things that aren't in dialogs) should memoize the React elements they return with `useMemo` to avoid re-rendering when the state they consume from `useToolbarState` hasn't changed
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+- Calls to `useToolbarState` should be as deep as possible to re-render the smallest number of components change when the relevant part of the toolbar state changes(e.g. a tooltip should be above the component that renders the tooltip so that it doesn't need to re-render even if the button inside it needs to re-render)
```
timleslie, line 18:
```suggestion
- Calls to `useToolbarState` should be as deep as possible to re-render the smallest number of components change when the relevant part of the toolbar state changes (e.g. a tooltip should be above the component that renders the tooltip so that it doesn't need to re-render even if the button inside it needs to re-render)
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Normalization
+
+Normalization is used in the editor to enforce a specific structure. Slate runs normalization by going through all the changed or "dirty" nodes and keeps calling `editor.normalizeNode` from the deepest node that changed up to the `Editor` until the normalization doesn't make any more changes.
```
timleslie, line 22:
```suggestion
Normalization is used in the editor to enforce a specific structure. Slate runs normalization by going through all the changed or "dirty" nodes and calling `editor.normalizeNode` from the deepest node that changed up to the `Editor` until the normalization doesn't make any more changes.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+For example, if you have two lists next to each other, you want to merge them so that they're one list. You don't want to do this in response to a specific user action because any number of things could result in two lists being next to each other so this is where you would use normalization.
+
+Besides bespoke normalization for the different types of nodes, there is a set of generic normalization which enforces which block types can have which children, search for `editorSchema` in the to see how this works.
```
timleslie, line 26:
```suggestion
Besides bespoke normalization for the different types of nodes, there is a set of generic normalization which enforces which block types can have which children. Search for `editorSchema` in the to see how this works.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+If you want to ensure that the parent of a node is always of a particular type, you must do this by checking the children of a node, not the parent of a node because normalization only runs on nodes that have changed and when a node is moved, normalization will not happen on that node because it did not actually change, normalization will happen on the parent it moved from and to though because both of those node's children array changed.
```
timleslie, line 28: This sentence is rather long and confusing. I'm not sure sure which node you're talking about at each point. Could you try to rephrase it?

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## User Input
+
+Some behaviour should happen in response to user input, for example markdown shortcuts. These shouldn't happen in normalization because you still want people to be able to write `> ` for example without it becoming a blockquote by undoing it so for these kind of things, they're handled by overriding `editor.insertText`, `editor.insertBreak`, `editor.insertData` or `onKeyDown`(which is a prop to the `Editable` component rather than on the editor instance)
```
timleslie, line 32:
```suggestion
Some behaviour should happen in response to user input, for example markdown shortcuts. These shouldn't happen in normalization because you still want people to be able to write `> ` for example without it becoming a blockquote by undoing it so for these kind of things, they're handled by overriding `editor.insertText`, `editor.insertBreak`, `editor.insertData` or `onKeyDown` (which is a prop to the `Editable` component rather than on the editor instance).
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## DocumentFeatures
+
+- `DocumentFeatures` is the configuration of what is enabled in the editor
+- Disabling a feature in `DocumentFeatures` disables it globally and means that it should not be shown in the toolbar and all instances of it should be normalized away
+- Features can also be disabled locally for the current selection(e.g. most things are disabled in code blocks, features can be disabled inside of component child props) which will be represented in the `ToolbarState`
```
timleslie, line 38:
```suggestion
- Features can also be disabled locally for the current selection (e.g. most things are disabled in code blocks, features can be disabled inside of component child props) which will be represented in the `ToolbarState`
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Component Blocks
+
+Component blocks are the main way that solution developers customise the editor. See the docs for the document field for how they're used. The way they're represented in the structure is that nodes with the type `component-block` store the form data and they have children of type `component-block-prop` and `component-inline-prop` with a `propPath` which points to where in the props structure they are. (note that they are both blocks, inline and block in the name refers to what kind of children they contain)
+
+## Tests
+
+The tests in the document editor are mainly about testing normalization and user input because most of the complexity is there though the UI is also rendered by helpers as a smoke test to ensure nothing immediately blows up. In tests, the document structure is written in jsx, use your editor's autocomplete to see what you can use in the jsx. When making assertions about the document structure of the editor, you should generally use inline snapshots. If you're abstractly testing some behaviour, for example testing markdown shortcuts for each of the marks, use `toEqualEditor`.
```
timleslie, line 47:
```suggestion
The tests in the document editor are mainly about testing normalization and user input because most of the complexity is there. The UI is also rendered by helpers as a smoke test to ensure nothing immediately blows up. In tests, the document structure is written in jsx, use your editor's autocomplete to see what you can use in the jsx. When making assertions about the document structure of the editor, you should generally use inline snapshots. If you're abstractly testing some behaviour, for example testing markdown shortcuts for each of the marks, use `toEqualEditor`.
```

timleslie (changes requested):
A few minor grammar nits, on the whole looks pretty good 👍 

Also, as per our STYLE_GUIDE.md we should be putting each sentence on a new line.

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
+# Document editor architecture
+
+The document editor is using [Slate](https://github.com/ianstormtaylor/slate). To learn about Slate generally, go through https://docs.slatejs.org/walkthroughs and https://docs.slatejs.org/concepts. The sections on [Nodes](https://docs.slatejs.org/concepts/02-nodes) and [Locations](https://docs.slatejs.org/concepts/03-locations) contain especially important terminology.
```
timleslie, line 3:
```suggestion
The document editor uses [Slate](https://github.com/ianstormtaylor/slate). To learn about Slate generally, go through https://docs.slatejs.org/walkthroughs and https://docs.slatejs.org/concepts. The sections on [Nodes](https://docs.slatejs.org/concepts/02-nodes) and [Locations](https://docs.slatejs.org/concepts/03-locations) contain especially important terminology.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,53 @@
...
+
```
timleslie, line 4: Could you add some brief notes on the data format? I think having a simple example of what a basic document looks like would be really helpful here. Something like...

```suggestion
## Data storage

The document field type stores its data as a JSON blob, representing the Slate formatted data.
A trivial document with a single heading and paragraph would like like

... insert JSON here
```

packages-next/fields-document/ARCHITECTURE.md
mitchellhamilton: Besides the performance stuff and the stuff explained in the Slate docs, no

timleslie: In that case can we remove the word `mostly`? It makes me wonder if there are specific things I need to be looking for.

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,54 @@
...
+The editor is very performance-sensitive because it will re-render on every keystroke so keep these things in mind while writing code for rendering the editor:
+
+- Renderers for elements/leaves should never use `useSlate` or `useToolbarState` because they will change on every editor change. If they need to get the editor to do something in response to a user action, they should use `useStaticEditor` which will not cause re-renders.
+- All state derived from the document/selection that is shown in the toolbar should come from `useToolbarState` so that it is only computed once
+- Components that use `useToolbarState` and are shown when the user is typing (i.e. things that aren't in dialogs) should memoize the React elements they return with `useMemo` to avoid re-rendering when the state they consume from `useToolbarState` hasn't changed
+- Calls to `useToolbarState` should be as deep as possible to re-render the smallest number of components change when the relevant part of the toolbar state changes (e.g. a tooltip should be above the component that renders the tooltip so that it doesn't need to re-render even if the button inside it needs to re-render)
+
+## Normalization
```
timleslie, line 21: Reading the code it looks like normalization/validation is applied in the `resolveInput` hooks on the server. Can you call out here when/where normalization is applied. e.g. is it happening on the both the client and the server?

Also, might be worth mentioning what the `hydrateRelationships` does and how the related data is injected into the document on the server before being returned to the client. Maybe in a different section of the document?

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,107 @@
...
+Plugins are not some special API, it means accepting an `Editor` object and returning it but modifying some of the properties.
+
+## Rendering
+
+The rendering of the editor is like any other bit of React code.
+See https://docs.slatejs.org/concepts/08-rendering for how rendering the document works in Slate.
+
+The editor is very performance-sensitive because it will re-render on every keystroke so keep these things in mind while writing code for rendering the editor:
+
+- Renderers for elements/leaves should never use `useSlate` or `useToolbarState` because they will change on every editor change. If they need to get the editor to do something in response to a user action, they should use `useStaticEditor` which will not cause re-renders.
```
JedWatson, line 44:
```suggestion
- Renderers for elements/leaves should never use `useSlate` or `useToolbarState` because those hooks will cause a re-render on every editor change. If they need to get the editor to do something in response to a user action, they should use `useStaticEditor` which will not cause re-renders.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,107 @@
...
+The entire document is also normalized in `resolveInput`.
+
+For example, if you have two lists next to each other, you want to merge them so that they're one list.
+You don't want to do this in response to a specific user action because any number of things could result in two lists being next to each other so this is where you would use normalization.
+
+Besides bespoke normalization for the different types of nodes, there is a set of generic normalization which enforces which block types can have which children. Search for `editorSchema` in the to see how this works.
+
+When a node is moved, normalization will not happen on that node because it did not actually change.
+Normalization will happen on the old and the new parent nodes of the node that moved because both of the parent nodes children arrays changed.
+If you want to enforce something about the relationship between a parent and child node, you must do this by doing normalization when normalizeNode is passed the parent node since the child node won't be passed to normalizeNode when it is moved.
```
timleslie, line 62:
```suggestion
If you want to enforce something about the relationship between a parent and child node, you must do this by doing normalization when `normalizeNode` is passed the parent node since the child node won't be passed to `normalizeNode` when it is moved.
```

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,107 @@
...
+
+## User Input
+
+Some behaviour should happen in response to user input, for example markdown shortcuts.
+These shouldn't happen in normalization because you still want people to be able to write `> ` for example without it becoming a blockquote by undoing it so for these kind of things, they're handled by overriding `editor.insertText`, `editor.insertBreak`, `editor.insertData` or `onKeyDown` (which is a prop to the `Editable` component rather than on the editor instance).
```
timleslie, line 67: This sentence is bit long and confusing, I'm not quite sure what it's trying to say. Could you rephrase it?

packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -0,0 +1,107 @@
...
+- Disabling a feature in `DocumentFeatures` disables it globally and means that it should not be shown in the toolbar and all instances of it should be normalized away
+- Features can also be disabled locally for the current selection (e.g. most things are disabled in code blocks, features can be disabled inside of component child props) which will be represented in the `ToolbarState`
+- The disabled status in the `ToolbarState` refers to the local disabled status only, it doesn't care about the global disabled status so a feature could have `isDisabled: false` locally and be disabled globally
+
+## Component Blocks
+
+Component blocks are the main way that solution developers customise the editor.
+See the docs for the document field for how they're used.
+The way they're represented in the structure is that nodes with the type `component-block` store the form data and they have children of type `component-block-prop` and `component-inline-prop` with a `propPath` which points to where in the props structure they are. (note that they are both blocks, inline and block in the name refers to what kind of children they contain)
```
timleslie, line 80:
```suggestion
The way they're represented in the structure is that nodes with the type `component-block` store the form data and they have children of type `component-block-prop` and `component-inline-prop` with a `propPath` which points to where in the props structure they are. (note that they are both blocks, "inline" and "block" in the name refers to what kind of children they contain)
```

timleslie (changes requested):
Looking good, just a few nits, and one confusing sentence that needs attention 👍

timleslie (approved):
Looks good 🚀

PR #4815, opened 2021-02-11
Add context.sudo() method

timleslie:
Doing `context.createContext({ skipAccessControl: true })` in order to get a context with access control completely disable is a) very common and b) very verbose 😅 This helper method makes for much more concise and easy to read/write code. It saves the method `context.createContext({ ... })` for more advanced situations where the developer needs to manipulate the request or session objects. It will also make it easier to document these methods, as the very common `.sudo()` method can be easily explained in isolation from the extremely advanced `.createContext()` method.


JedWatson:
👍 👍 I am two thumbs up for this, great API update.

One comment on the PR though, are you actually intending to remove the `skipAccessControl` option entirely? Because while a lot of the code in this refactor is more elegant with `.sudo()` I'd argue other parts would be simpler just using the old option. 

Wanted to be clear on the thinking behind why you're eradicating the option usage entirely, because that seems different to the description above:

> It saves the method context.createContext({ ... }) for more advanced situations where the developer needs to manipulate the request or session objects

Implication here is that when you're doing the more advanced thing, it's still cool to use `skipAccessControl` as one of the options, right?

i.e replacing [this usage](https://github.com/keystonejs/keystone/pull/4815/files#diff-1d8405eecac9f983c368d21d11e229f93e0a1446b971c7a71a2c84083e906edeR245-R247) w/ the `.sudo()` chaining doesn't seem simpler to me, implies that chaining is _always_ the way to get a context w/o access control:

```
          const context = keystone.createContext({
             schemaName: 'public'
             authentication: {},
             skipAccessControl: true,		
           });
```

timleslie:
> One comment on the PR though, are you actually intending to remove the `skipAccessControl` option entirely?

Maybe... but not just yet? The other main use case for calling `createContext` is in the context of either unit test or custom mutations where you might want to run a particular mutation *as* a particular user. In this case the general strategy is to manipulate the `context.session` property. Right now this is a bit fiddly, and I have a follow up PR to this one which explores this issue. I'm not sure if we'll completely get rid of `skipAccessControl`, it will depend on how the rest of the chips fall.

> > It saves the method context.createContext({ ... }) for more advanced situations where the developer needs to manipulate the request or session objects
> 
> Implication here is that when you're doing the more advanced thing, it's still cool to use `skipAccessControl` as one of the options, right?

Yep, that's definitely the case. We have an example of this in `mutations.test.js`, which exists in a unit testing context, so the `context` object is already a sudo context, and we need to set `{ skipAccessControl: false }` for when we're testing access control behaviour.

> 
> i.e replacing [this usage](https://github.com/keystonejs/keystone/pull/4815/files#diff-1d8405eecac9f983c368d21d11e229f93e0a1446b971c7a71a2c84083e906edeR245-R247) w/ the `.sudo()` chaining doesn't seem simpler to me, implies that chaining is _always_ the way to get a context w/o access control:
> 
> ```
>           const context = keystone.createContext({
>              schemaName: 'public'
>              authentication: {},
>              skipAccessControl: true,		
>            });
> ```

Yeah, chaining here is probably not optimal (although the marginal difference is pretty negligible). The motivation for replacing all instances in *our* repo is it let's me see what's left over in terms of `createContext()` usage, which is helping with my thinking about the follow up changes.

The higher level aim here is to have a small handful of obvious interfaces to use for common tasks, so that we can point users at them, and save the "low level" `.createContext()` as an escape hatch of last resort, since it's a very subtle beast and documenting it is very hard 😅 . Perhaps it can even go away entirely at some stage 🤷

JedWatson:
Oh to clarify

> Maybe... but not just yet?

Didn't actually think you were going for this, not really of the mind that we should, more saying the diff makes me think you are going for this, but also I realised that by looking at the diff I'm not seeing the cases where you _didn't_ refactor `skipAccessControl: true` to `.sudo()`

JedWatson (approved):

mitchellhamilton (commented):
All the packages with usages of `.sudo()` need major bumps

.changeset/weak-parrots-remember.md
```diff
@@ -0,0 +1,17 @@
...
+'@keystone-next/admin-ui': patch
+'@keystone-next/auth': patch
+'@keystone-next/types': patch
+'@keystonejs/auth-passport': patch
+'@keystonejs/example-projects-starter': patch
+'@keystonejs/fields': patch
+'@keystonejs/list-plugins': patch
+'@keystonejs/server-side-graphql-client': patch
+'@keystonejs/test-utils': patch
+'@keystonejs/api-tests': patch
```
timleslie, line 14:
```suggestion
'@keystonejs/demo-project-meetup': major
'@keystone-next/admin-ui': major
'@keystone-next/auth': major
'@keystone-next/types': major
'@keystonejs/auth-passport': major
'@keystonejs/example-projects-starter': major
'@keystonejs/fields': major
'@keystonejs/list-plugins': major
'@keystonejs/server-side-graphql-client': major
'@keystonejs/test-utils': major
'@keystonejs/api-tests': major
```

PR #4826, opened 2021-02-13
Allow custom idField types and pass id field value on GraphQL create mutations

### LINKED ISSUE #4824: Can't set a custom ID value when inserting new item in List via GraphQL mutation query
# Bug report

## Describe the bug

Can't set a custom ID value when inserting new item in List via GraphQL mutation query:
```graphql
mutation {
  createCountry(data: {
    id: "RU"
    name: "Russia"
  }) {
    id, name
  }
}
```
 - got the error as result:
```
GraphQLError: Field \"id\" is not defined by type \"CountryCreateInput\".
    at ObjectField (keystone/node_modules/graphql/validation/rules/ValuesOfCorrectTypeRule.js:73:29)
    at Object.enter (keystone/node_modules/graphql/language/visitor.js:323:29)
    at Object.enter (keystone/node_modules/graphql/utilities/TypeInfo.js:370:25)
    at visit (keystone/node_modules/graphql/language/visitor.js:243:26)
    at Object.validate (keystone/node_modules/graphql/validation/validate.js:69:24)
    at validate (keystone/node_modules/apollo-server-core/dist/requestPipeline.js:221:34)
    at Object.<anonymous> (keystone/node_modules/apollo-server-core/dist/requestPipeline.js:118:42)
    at Generator.next (<anonymous>)
    at fulfilled (keystone/node_modules/apollo-server-core/dist/requestPipeline.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
I use the MongoDB database with MongooseAdapter, which have no problems with custom id values, here is raw query that do this thing:
```
> db.countries.insert({_id:"RU", "Name" : "Russia"})
WriteResult({ "nInserted" : 1 })
```

## To Reproduce

1. Create a simple List, for example "Country":
```js
keystone.createList('Country', {
  fields: {
    id: { type: Text },
    name: { type: Text },
  },
});
```
2. Try to insert new city via GraphQL mutation query:
```graphql
mutation {
  createCountry(data: {
    id: "RU"
    name: "Russia"
  }) {
    id, name
  }
}
```
3. See the error

## Expected behaviour

This mutation query must create the new item with sent id value.

MurzNN:
Very often, when we syncing data from some external storage, we need to set exact id of list item on `create` mutation, instead of using the generated `autoIncrement` or `uuid` value.

But at now isn't possible to set desired `id` field value on Keystone when creating List items via GraphQL `create` mutation query, because `id` field is excluded from schema, even if we manually implement the custom `id` field in custom List schema and giving access to edit it on create operations.

This PR fix this problem via deleting forced removal of the `id` field in all query types, staying this check to access function, that is already successfully remove the `id` field by default.

How to reproduce the problem:
1. Get the `examples/blog` project, and give the `edit` permission to `id` field via lines:
```js
...
  Post: list({
    fields: {
...
    },
    idField: integer({
      isUnique: true,
      isRequired: true,
      isIndexed: false,
      access: {
        create: true,
      },
      ui: {
        createView: { fieldMode: 'edit' },
        itemView: { fieldMode: 'hidden' },
      },
    }),
```

2. Try to create the `Post` item with desired id:
```graphql
mutation {
  createPost(data: {
    id: "777",
    title: "My prosperous post!",
  }) {
    id, title
  }
}
```
You will receive the error:
```
TypeError: Cannot assign to read only property 'path' of object '#<Object>'",
```

3. Apply this PR and repeat the query, it will execute well and successfully create the list item with id = 777.
Tested successfully also with "text" field as idField.

Resolves #4824.

### DIFF
--- a/packages-next/fields/src/types/integer/index.ts
+++ b/packages-next/fields/src/types/integer/index.ts
@@ -30,7 +30,7 @@ export const integer =
   meta =>
     fieldType({
       kind: 'scalar',
-      mode: 'optional',
+      mode: isRequired ? 'required' : 'optional',
       scalar: 'Int',
       index: getIndexType({ isIndexed, isUnique }),
     })({
--- a/packages-next/keystone/src/lib/core/field-assertions.ts
+++ b/packages-next/keystone/src/lib/core/field-assertions.ts
@@ -69,30 +69,12 @@ function assertIdFieldGraphQLTypesCorrect(list: ListForValidation) {
       `The idField on a list must define a uniqueWhere GraphQL input with the ID GraphQL scalar type but the idField for ${list.listKey} does not define one`
     );
   }
-  if (idField.input.uniqueWhere.arg.type !== schema.ID) {
-    throw new Error(
-      `The idField on a list must define a uniqueWhere GraphQL input with the ID GraphQL scalar type but the idField for ${
-        list.listKey
-      } defines the type ${idField.input.uniqueWhere.arg.type.graphQLType.toString()}`
-    );
-  }
-  // we may want to loosen these constraints in the future
-  if (idField.input.create !== undefined) {
-    throw new Error(
-      `The idField on a list must not define a create GraphQL input but the idField for ${list.listKey} does define one`
-    );
-  }
-  if (idField.input.update !== undefined) {
-    throw new Error(
-      `The idField on a list must not define an update GraphQL input but the idField for ${list.listKey} does define one`
-    );
-  }
   if (idField.access.read === false) {
     throw new Error(
       `The idField on a list must not have access.read be set to false but ${list.listKey} does`
     );
   }
-  if (idField.output.type.kind !== 'non-null' || idField.output.type.of !== schema.ID) {
+  if (idField.output.type.kind !== 'non-null' && idField.dbField.mode !== 'required') {
     throw new Error(
       `The idField on a list must define a GraphQL output field with a non-nullable ID GraphQL scalar type but the idField for ${
         list.listKey
--- a/packages-next/keystone/src/lib/core/prisma-schema.ts
+++ b/packages-next/keystone/src/lib/core/prisma-schema.ts
@@ -21,6 +21,10 @@ const modifiers = {
 };
 
 function printIndex(fieldPath: string, index: undefined | 'index' | 'unique') {
+  // For id fields we already have the primary index
+  if (fieldPath === 'id') {
+    return '';
+  }
   return {
     none: '',
     unique: '@unique',
@@ -177,13 +181,13 @@ function assertDbFieldIsValidForIdField(
       } field`
     );
   }
-  if (field.index !== undefined) {
+  if (field.index !== undefined && field.index !== 'unique') {
     throw new Error(
       `id fields must not specify indexes themselves but the id field for the ${listKey} list specifies an index`
     );
   }
   // this will likely be loosened in the future
-  if (field.default === undefined) {
+  if (field.mode !== 'required' && field.default === undefined) {
     throw new Error(
       `id fields must specify a Prisma/database level default value but the id field for the ${listKey} list does not`
     );
--- a/.changeset/dry-ears-tell.md
+++ b/.changeset/dry-ears-tell.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/keystone': patch
+---
+
+Allow set id value on GraphQL create mutations input
--- a/packages-next/fields/src/types/text/index.ts
+++ b/packages-next/fields/src/types/text/index.ts
@@ -33,7 +33,7 @@ export const text =
   meta =>
     fieldType({
       kind: 'scalar',
-      mode: 'optional',
+      mode: isRequired ? 'required' : 'optional',
       scalar: 'String',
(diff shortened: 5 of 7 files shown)


MurzNN:
And here https://github.com/keystonejs/keystone/compare/master...MurzNN:todo-custom-ids?expand=1 is example of todo app, that works with custom ids of Todo List items.

MurzNN:
After several tests I reveal, that the access control function already filtering-out right the `id` fields via line:
```js
      .filter(field => field.access[schemaName][access]); // If it's globally set to false, makes sense to never let it be updated
```
So previous line:
```
      .filter(({ path }) => path !== 'id') // Exclude the id fields update types
```
is not even needed! 

After removing it in all examples nothing is changes, but API becomes allowing pass the `id` value from GraphQL if they are manually enabled, here is example how to check this on `examples/basic/schema.ts` - add to the "Posts" list this block:
```js
    idField: autoIncrement({
      isUnique: true,
      access: {
        create: true,
        // update: true,
      },
      ui: {
        createView: { fieldMode: 'edit' },
        // itemView: { fieldMode: 'edit' },
      },
    }),
```
Via same manner we can also allow changing the `id` value on `update` queries via uncommenting lines.

MurzNN:
I updated the PR to match new core, there is much more changes needed, I'll comment them in the diff.

packages-next/fields/src/types/integer/index.ts
```diff
@@ -30,7 +30,7 @@ export const integer =
   meta =>
     fieldType({
       kind: 'scalar',
-      mode: 'optional',
+      mode: isRequired ? 'required' : 'optional',
```
MurzNN, line 33: We must match the `isRequired` value to set correct mode, this is needed to check that idField is required. Maybe other fields need the same fix, but this is outside of current PR.

packages-next/keystone/src/lib/core/field-assertions.ts
```diff
@@ -69,30 +69,12 @@ function assertIdFieldGraphQLTypesCorrect(list: ListForValidation) {
       `The idField on a list must define a uniqueWhere GraphQL input with the ID GraphQL scalar type but the idField for ${list.listKey} does not define one`
     );
   }
-  if (idField.input.uniqueWhere.arg.type !== schema.ID) {
```
MurzNN, line 72: This check disallowing changing the type of ID field, so we can't change, for example, `autoIncrement` to `integer`. I see no reasons to keep it, so remove is right decision?

packages-next/keystone/src/lib/core/field-assertions.ts
```diff
@@ -69,30 +69,12 @@ function assertIdFieldGraphQLTypesCorrect(list: ListForValidation) {
...
-  if (idField.input.create !== undefined) {
-    throw new Error(
-      `The idField on a list must not define a create GraphQL input but the idField for ${list.listKey} does define one`
-    );
-  }
-  if (idField.input.update !== undefined) {
-    throw new Error(
-      `The idField on a list must not define an update GraphQL input but the idField for ${list.listKey} does define one`
-    );
-  }
```
MurzNN, line 89: Those already can be controlled by `access` property of field:
```js
      access: {
        create: true,
        update: true,
      },
```
so we must not disable this.

packages-next/keystone/src/lib/core/prisma-schema.ts
```diff
@@ -177,13 +181,13 @@ function assertDbFieldIsValidForIdField(
       } field`
     );
   }
-  if (field.index !== undefined) {
+  if (field.index !== undefined && field.index !== 'unique') {
```
MurzNN, line 184: We should not throw the error if field is market as unique.

packages-next/keystone/src/lib/core/prisma-schema.ts
```diff
@@ -177,13 +181,13 @@ function assertDbFieldIsValidForIdField(
...
     throw new Error(
       `id fields must not specify indexes themselves but the id field for the ${listKey} list specifies an index`
     );
   }
   // this will likely be loosened in the future
-  if (field.default === undefined) {
+  if (field.mode !== 'required' && field.default === undefined) {
```
MurzNN, line 190: For required fields we don't need the default value.

MurzNN:
At the backend side with those changes all works well. But the problem is in frontend side, because in some places the id field type is hard-coded as `ID!`, I commented some of those places in last commit.
So I see two solutions for this problem:
1. Keep all `id` fields in GraphQL schema as `ID` type, because GraphQL schema allows any singular type in it, regarding to https://graphql.org/learn/schema/
> ID: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an ID signifies that it is not intended to be human‐readable.

2. Pass the real type of id field (Int, String, maybe other?) to GraphQL schema. This already works for id fields in list, but there are many places in frontend, where the `ID!` field type is hard-coded, see last commit.

So the question to developers: which way is better to solve this task?

MurzNN:
After recent changes in Kestone core, related to idFields, I decided to rework this feature from scratch, here is new PR with implementation: https://github.com/keystonejs/keystone/pull/6048

PR #4838, opened 2021-02-15
Add heading anchors to @keystone-next/website

gwyneplaine:
## Description
This PR contains the following:

- Direct port of the `Heading` component from @keystone/website with a few tweaks. 
- Remove back ticks from headings in mdx

## Heading component port
Besides a few changes this is an almost identical port of the Heading component in the @keystone/website package. 

### Changes
First, the id isn't passed in as a prop, as this isn't information immediately available to the MDX parser. Instead we deduce the id from the children of the heading component. We make two assumptions here:
a) There is only one child
b) The child is a string value. This means that heading values in mdx should only be formatted as such. Any additional tags like bold, code or pre will result in no id being appended to that heading. 

Second, previously the heading LinkIcon would replace the url directly on click. I've modified this behaviour slightly to be inline with the common user experience. Namely, onClick of a Link, the anchor path is copied to the users clipboard. We use the toast component here to give users proper feedback. 

### Copy to clipboard functionality:
As of early last year, document.execCommand has been deprecated for all (most) major browsers in favor of the asynchronous navigator.clipboard API. 

Our implementation accounts for this by doing the following: 
* leverage the navigator api if it exists 
* If it doesn't (in the case of older versions of IE) we fallback to the copy-to-clipboard library, which operates on top of the old document.execCommand API.

## Remove back ticks from headings in mdx
As per above, we do this because back ticks add additional mark up to children, which makes it difficult to resolve to the text node without a noticeable time differential. 

As of writing this, @timleslie is ok with this change. 

Important to note that this implementation similar to all other implementations of anchor links in our repo are subject to the same bug. As the id is at the moment extrapolated from the id, headings *must* be unique. 

Non-unique headings occurring in the same page will not produce the desired anchoring result. 

## Demo:

https://user-images.githubusercontent.com/12119389/107909951-836e4780-6fad-11eb-9320-74da77b4f377.mov


docs-next/pages/apis/config.mdx
```diff
@@ -7,20 +7,36 @@ The `keystone-next` CLI expects to find a module named `keystone.ts` with a defa
 ```typescript filename=keystone.ts
 import { config } from '@keystone-next/keystone/schema';
 
-export default config({ /* ... */ });
+export default config({
```
timleslie, line 10: I know I'm going to lose this battle at some point, but could you review this PR for all the places our dots get spread over three lines and put them back on one? 🙏

gwyneplaine: will diddly do

docs-next/pages/apis/config.mdx
```diff
@@ -7,20 +7,36 @@ The `keystone-next` CLI expects to find a module named `keystone.ts` with a defa
...
+  /* ... */
+});
```
gwyneplaine, line 12:
```suggestion
export default config({  /* ... */ });
```

docs-next/pages/apis/config.mdx
```diff
@@ -7,20 +7,36 @@ The `keystone-next` CLI expects to find a module named `keystone.ts` with a defa
...
+  },
+  session: () => {
+    /* ... */
+  },
+  graphql: {
+    /* ... */
+  },
+  extendGraphqlSchema: {
+    /* ... */
+  },
```
gwyneplaine, line 39:
```suggestion
  lists: { /* ... */ },
  db: { /* ... */ },
  ui: { /* ... */ },
  server: { /* ... */ },
  session: () => { /* ... */ },
  graphql: { /* ... */ },
  extendGraphqlSchema: { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -81,7 +97,9 @@ export default config({
   db: {
     adapter: 'prisma_postgresql',
     url: 'postgres://dbuser:dbpass@localhost:5432/keystone',
-    onConnect: async context => { /* ... */ },
+    onConnect: async context => {
+      /* ... */
+    },
```
gwyneplaine, line 102:
```suggestion
    onConnect: async context => { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -103,7 +121,9 @@ export default config({
   db: {
     adapter: 'knex',
     url: 'postgres://dbuser:dbpass@localhost:5432/keystone',
-    onConnect: async context => { /* ... */ },
+    onConnect: async context => {
+      /* ... */
+    },
```
gwyneplaine, line 126:
```suggestion
    onConnect: async context => { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -205,14 +225,21 @@ In general you will use functions from the `@keystone-next/keystone/session` pac
...
 export default config({
-  session: withItemData(statelessSessions({ /* ... */ }), { /* ... */ }),
+  session: withItemData(
+    statelessSessions({
+      /* ... */
+    }),
+    {
+      /* ... */
+    }
+  ),
```
gwyneplaine, line 235:
```suggestion
  session: withItemData(
    statelessSessions({ /* ... */ }),
    { /* ... */ }
  ),
```

docs-next/pages/apis/schema.mdx
```diff
@@ -12,12 +12,24 @@ import { config, createSchema, list } from '@keystone-next/keystone/schema';
...
+      },
+      ui: {
+        /* ... */
+      },
+      hooks: {
+        /* ... */
+      },
+      graphql: {
+        /* ... */
+      },
```
gwyneplaine, line 32:
```suggestion
      fields: { /* ... */ },
      idField: { /* ... */ },
      access: { /* ... */ },
      ui: { /* ... */ },
      hooks: { /* ... */ },
      graphql: { /* ... */ },
```

docs-next/pages/apis/schema.mdx
```diff
@@ -66,7 +80,9 @@ import { autoincrement, mongoId } from '@keystone-next/fields';
 export default config({
   lists: createSchema({
     ListName: list({
-      fields: { /* ... */ },
+      fields: {
+        /* ... */
+      },
```
gwyneplaine, line 85:
```suggestion
      fields: { /* ... */ },
```

docs-next/components/Heading.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return '';
+};
+
+interface StringOnlyChildren {
+  children: string;
+}
+
+interface HeadingProps extends StringOnlyChildren {
+  as: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  css: Record<string, any>;
```
gwyneplaine, line 29: @mitchellhamilton  not really sure what to do here besides import the CSSObject. 
Let me know if there's a better way to infer this type in this particular use case

mitchellhamilton: It shouldn't be there, className should be there(and it should be optional). the jsx pragma automatically adds the css prop if a component accepts className.

gwyneplaine: Perfect, ty!

JedWatson (approved):

JedWatson:
Just leaving this here for later - @gwyneplaine the `remark-heading-id` [plugin](https://github.com/imcuttle/remark-heading-id) looks handy if we ever have clashes in the heading ids

PR #4840, opened 2021-02-15
[WIP] Refactor createContext for self-referential clarity

timleslie:


mitchellhamilton (commented):
This just seems more complicated and easier to accidentally break because of the type casting?

timleslie:
> This just seems more complicated and easier to accidentally break because of the type casting?

Yeah, I'm actually kind 50/50 on this change... it unblocks a future change I have planned, but also makes me wonder if that future chance is actually worth it. I'm going to put this back into [WIP] and reconsider the direction.

PR #4845, opened 2021-02-15
Updating docs styling

JedWatson:
A few tweaks:

* Fixed up the prism theme to match the site more naturally
* Fixes a few layout issues w/ code snippets
* Adds special syntax for info / warning / error callout boxes (we had a hack for this in the previous site, but it required specific language, I think this is neater)
* Fixes styles when you hover the Keystone Next link in the header


timleslie (approved):
LGTM 👍

PR #4853, opened 2021-02-15
Paramterise access control types

timleslie:
This is working towards getting correct access control types across keystone-5 and keystone-next packages, which use different arguments to imperative access control.


gwyneplaine (approved):
lgtm

PR #4854, opened 2021-02-15
Improve access-control types

timleslie:
This isn't the final form of these types, but it gets things a bit more organised before we shuffle things across files.


gwyneplaine (approved):
lgtm

PR #4856, opened 2021-02-16
Stop enabling lignatures with the `optimizeLegibility` prop in `Core`

mitchellhamilton:
Mainly doing this because of the docs website, have a look at `config` inside a code block on the currently deployed version vs this PR.

Before:
Before

After:
After


timleslie (approved):
Much nicer!

PR #4857, opened 2021-02-16
Remove all non ES5/ES Modules syntax from generated Admin UI files

mitchellhamilton:
This is so that Keystone doesn't have an implicit dependency on the user's Babel config.


timleslie (approved):
LGTM 👍

PR #4858, opened 2021-02-16
Dim out Placeholder pages in the new docs nav

JedWatson:
This helps the new docs site a lot, because you can see which pages actually have content. Makes it easier for us to see which bits still need to be written, and should frustrate our audience a lot less because they're not clicking around pages that have no content.

### DIFF
--- a/docs-next/components/Navigation.tsx
+++ b/docs-next/components/Navigation.tsx
@@ -13,15 +13,16 @@ const Section = ({ label, children }: SectionProps) => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


timleslie (approved):
Very nice 👍 Just needs a quick `yarn format`

JedWatson:
oof, forgot I turned "format on save" off while writing docs in markdown.

pushed the fix 👍

timleslie (approved):
LGTM 👍

PR #4860, opened 2021-02-16
Automatically include Next's Babel config in the dev require hook if the user has no Babel config

mitchellhamilton:
the main result of this: people won't have to have a Babel config to use TypeScript in their config.


timleslie (approved):
👍

PR #4861, opened 2021-02-16
Remove default values from validation functions

timleslie:
Part of the effort to consolidate access control across keystone-5 and keystone-next.


.changeset/large-dots-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/access-control': major
+'@keystonejs/keystone': major
```
mitchellhamilton, line 3: How is this a major for `@keystonejs/keystone`?

timleslie: Good point, will change 👍

timleslie, line 3:
```suggestion
'@keystonejs/keystone': patch
```

mitchellhamilton (approved):

PR #4864, opened 2021-02-16
Port more API tests to new interfaces

timleslie:


packages-next/fields-document/ARCHITECTURE.md
```diff
@@ -63,7 +63,7 @@ If you want to enforce something about the relationship between a parent and chi
 
 ## Changing the default handling of user input
 
-For some behaviour like markdown shortcuts, you need to change the default handling of user input like converting a paragraph to a blockquote when a user types `> `.
+For some behaviour like markdown shortcuts, you need to change the default handling of user input like converting a paragraph to a blockquote when a user types `>`.
```
mitchellhamilton, line 66: The space here is intentional, in the example, the blockquote is created when pressing space, not `>`

timleslie: Hmm, this is the result of a `yarn format`. Looks like we'll need to work out a way to run `yarn format` and not have it break this bit of the docs. For now I'll revert it on this branch.

timleslie, line 66:
```suggestion
For some behaviour like markdown shortcuts, you need to change the default handling of user input like converting a paragraph to a blockquote when a user types `> `.
```

mitchellhamilton (approved):

PR #4865, opened 2021-02-17
Add operation argument to validateAuthAccessControl

timleslie:
This change will allow for more consistent type definitions across the package.


gwyneplaine (approved):
lgtm

Issue #461, opened 2018-10-19
Discussion: Boolean variable names --`blah` vs `isBlah`


molomby:
In dynamically typed languages, I've always preferred to prefix my Boolean variable names with something that makes their type explicit. Eg: `isEnabled`, `hasChildren`, etc. Just a few additional characters carry so much useful meaning.

A variable name like, for example, `required` could mean the thing _is_ or it could be the mask, regex or value the thing is required to match or be. Likewise, Imagine you're reading unfamiliar code and see a variable `foo.table`. Is it a Boolean, indicating the `foo` _is_ a table? Is it the _name_ of the table `foo` represents? Is `.table` an object with attributes of it's own? How long do you now waste digging back to find the origin of the value, or imply it's type from other (hopefully correct, unabigious!) usage? An explicit variable name like `foo.isTable` makes the intention blindingly clear.

The most obvious places this would affect are the list/field config structures:

* `require` > `isRequired`
* `indexed` > `isIndexed`
* `unique` > `isUnique`

Benefits:

* More explicit and specific; clearer, easier to search for

Downsides: 

* Diverges from KS4; slightly complicates upgrade path (or we support both versions? 😕)
* As pointed out in the `unique` context, [diverges from the equivilant Mongoose config](https://github.com/keystonejs/keystone-5/pull/412#pullrequestreview-162666323)

This is part of a broader issue discussion about explicitness -vs- verbosity in variable naming and I'll admit there are trade offs. 

I'm strongly for, as you might have guessed, but acknowledge that consistency should trump correctness in this case. Let's pick a convention and stick to it.

What do you people (@JedWatson, @timleslie, @jesstelford) think?

PR #4866, opened 2021-02-17
Add config.ui.isDisabled option

timleslie:
In particular, this is needed for our testing framework, where we want to spin up an express server for just the graphQL API.

### DIFF
--- a/packages-next/keystone/src/scripts/build/build.ts
+++ b/packages-next/keystone/src/scripts/build/build.ts
@@ -2,12 +2,31 @@ import Path from 'path';
 import prettier from 'prettier';
 import fs from 'fs-extra';
 import { buildAdminUI, generateAdminUI } from '@keystone-next/admin-ui/system';
+import { AdminFileToWrite } from '@keystone-next/types';
 import { createSystem } from '../../lib/createSystem';
 import { initConfig } from '../../lib/initConfig';
 import { requireSource } from '../../lib/requireSource';
 import { saveSchemaAndTypes } from '../../lib/saveSchemaAndTypes';
 import { CONFIG_PATH } from '../utils';
 import type { StaticPaths } from '..';
+
+// FIXME: Duplicated from admin-ui package. Need to decide on a common home.
+async function writeAdminFile(file: AdminFileToWrite, projectAdminPath: string) {
+  const outputFilename = Path.join(projectAdminPath, file.outputPath);
+  if (file.mode === 'copy') {
+    if (!Path.isAbsolute(file.inputPath)) {
+      throw new Error(
+        `An inputPath of "${file.inputPath}" was provided to copy but inputPaths must be absolute`
+      );
+    }
+    await fs.ensureDir(Path.dirname(outputFilename));
+    // TODO: should we use copyFile or copy?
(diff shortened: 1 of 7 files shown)


mitchellhamilton (commented):
It should be `isDisabled` as per #461, no?

timleslie:
> It should be `isDisabled` as per #461, no?

Good point, it 💯 should. Will change 👍

packages-next/keystone/src/scripts/build/build.ts
```diff
@@ -16,11 +16,15 @@ export async function build({ dotKeystonePath, projectAdminPath }: StaticPaths)
   console.log('✨ Generating graphQL schema');
   await saveSchemaAndTypes(graphQLSchema, keystone, dotKeystonePath);
 
-  console.log('✨ Generating Admin UI code');
-  await generateAdminUI(config, graphQLSchema, keystone, projectAdminPath);
+  if (config.ui?.isDisabled) {
+    console.log('✨ Skipping Admin UI build');
```
mitchellhamilton, line 20: This is problematic now that I think about it. The build also builds the API so we can't just skip it like this.

timleslie:
Currently blocked on #4867

packages-next/keystone/src/scripts/build/build.ts
```diff
@@ -23,21 +42,52 @@ export function serializePathForImport(path: string) {
...
+        {
+          mode: 'copy' as const,
+          inputPath: Path.join(pkgDir, 'static', 'tsconfig.json'),
+          outputPath: 'tsconfig.json',
+        },
+        {
+          mode: 'write' as const,
+          src: `import React from 'react'; export default () => <></>`,
+          outputPath: Path.join('pages', 'index.js'),
+        },
```
mitchellhamilton, line 86:
```suggestion
```

I just tested this locally without the generation of this file and it works so I don't think it's necessary.

Also a nit: `files.push(...[stuff,stuff])` -> `files.push(stuff,stuff)`

packages-next/keystone/src/scripts/run/start.ts
```diff
@@ -28,7 +28,11 @@ export const start = async ({ dotKeystonePath, projectAdminPath }: StaticPaths)
     false,
     projectAdminPath
   );
-  console.log(`👋 Admin UI and graphQL API ready`);
+  if (config.ui?.isDisabled) {
+    console.log(`👋 GraphQL API ready`);
+  } else {
+    console.log(`👋 Admin UI and graphQL API ready`);
```
mitchellhamilton, line 34:
```suggestion
    console.log(`👋 Admin UI and GraphQL API ready`);
```

packages-next/keystone/src/scripts/build/build.ts
timleslie: Hmm, I had issues without this file. Let me double check 🤔 

nit: 👍

timleslie: OK, have confirmed that `index.js` isn't needed 👍

mitchellhamilton (approved):

PR #4868, opened 2021-02-17
Docsnext/local nav

gwyneplaine:
## UPDATE:
A less hacky way to iterate through and collate headings was found via the following tweet (thanks @jaredpalmer)
 https://twitter.com/jaredpalmer/status/1362037404199243781?s=21
 
We've now moved the iteration logic to React's runtime , which means it'll run at least once per request/render, but there are ways for us to optimise this should it be a performance bottleneck later. For now it means we can remove exorbitant complexity and brittleness created by the remark plugin. 

---------------------------------------------------------------

The aim of this PR is to port over the Table of Contents from the old keystone website and have it work in @keystone-next/website. 

Unfortunately as per the jira issue, the Table of Contents component relies on having a pre-collated list of headings from the mdx to be rendered. This is automagically provided by gatsby via its gql layer, however that's not the case with Next (afaik). 

Thus this PR includes two bits of work: 
1. A remark plugin for parsing mdx files and passing the list of headings down to the TableOfContents component. 
2. A port of the TableOfContents component into `@keystone-next/website`

## Remark plugin
At the moment our remark plugin does the following: 
* Traverse each mdx file and collate headings into a list in the order that they appear. 
* Add a named export `meta` object that includes the array of headings we've collated 
* Append that `meta` to the `meta` prop of the <Markdown> component in the markup of the default export in the file. 

The unfortunate assumption the plugin making here is that there is indeed a <Markdown> component in our markup and that it accepts a meta prop of the same type as the meta object we've created. This means that:
* any changes to the Markdown components API will result in no ToC being rendered
* Not having a Markdown component in your export default will result in no ToC being rendered

As of writing this, I can't find any alternative way to properly pass this information through to the TableOfContents app in a similar fashion to Gatsby, any thoughts here are welcome. 

### Other considered approaches:
We could store the collated lists in a map indexed to the path from which the mdx would be served from. i.e.
```javascript
{ 'whats-new' : [{ type: 'heading'  ... }] }
```
and then writing this to disk for look up later by the next app. 
The issue with this is that it's heavily non-performant for the dev environment, as each mdx request triggers the remark plugin (dev env transforms files request by request). In other words we'd be writing to disk once per request. (this is not a problem in prod) 

Alternatively we can abandon the withMDX plugin entirely, and sacrifice statically served mdx for more control over the parsed output of our mdx. 

```jsx
export. const getStaticProps = (contents) => { 
   const headings = collateHeadings(contents);
  /** .... transform MDX */
 return {
   props: {
       mdx, headings
   }
 }
}

export default Guides = ({ props, headings }) => (
  <Page>
      <TableOfContents headings={headings}> 
       <MDXRenderer>
          {mdx}
       </MDXRenderer>
 </Page>
) 

```
This would mean our docs app would be no longer static, and we'd also have to write all the boilerplate associated with parsing and rendering the MDX. Which imo is a lot of work to get a TableOfContents working 🤷 

**Leveraging front matter is also no bueno unless we do the above, since the out of the box mdx plugin for next doesn't have frontmatter support. (frontmatter will just be rendered as content on the page)**


docs-next/remark-plugins/headings.js
```diff
@@ -0,0 +1,60 @@
...
+const slugify = require('@sindresorhus/slugify');
+const fs = require('fs');
+
+function insertAtEnd(node, { children }) {
+  // find index of last import
+  return children.push(node);
+}
+
+module.exports = () => {
+  const fileMap = {
```
mitchellhamilton, line 11: Is this being used?

docs-next/remark-plugins/headings.js
```diff
@@ -0,0 +1,60 @@
...
+    visit(tree, 'export', node => {
+      if (node.default) {
+        const key = 'Markdown';
+        const index = node.value.indexOf(key);
+        if (index > -1) {
+          const offset = index + key.length;
+          // console.log(node.value, typeof node.value);
+          const beginning = node.value.slice(0, offset);
+          const end = node.value.slice(offset);
+          node.value = `${beginning} meta={meta} ${end}`;
```
mitchellhamilton, line 44: lmao

---

Could you add `meta={meta}` to the Page component in the document field doc? (it doesn't render Markdown for Reasons)

gwyneplaine: yeah I'll add it, but dear god this feels gross 😂

docs-next/pages/apis/schema.mdx
```diff
@@ -69,7 +69,9 @@ import { autoincrement, mongoId } from '@keystone-next/fields';
 export default config({
   lists: createSchema({
     ListName: list({
-      fields: { /* ... */ },
+      fields: {
+        /* ... */
+      },
```
gwyneplaine, line 74:
```suggestion
      fields: { /* ... */ },
```

docs-next/pages/apis/session.mdx
```diff
@@ -14,7 +14,9 @@ export default config({
   session: withItemData(
     statelessSessions({
       secret: 'ABCDEFGH1234567887654321HGFEDCBA',
-      ironOptions: { /* ... */ },
+      ironOptions: {
+        /* ... */
+      },
```
gwyneplaine, line 19:
```suggestion
      ironOptions: { /* ... */ },
```

docs-next/pages/apis/session.mdx
```diff
@@ -146,7 +148,9 @@ import { statelessSessions, withItemData } from '@keystone-next/keystone/session
 
 export default config({
   session: withItemData(
-    statelessSessions({ /* ... */ }),
+    statelessSessions({
+      /* ... */
+    }),
```
gwyneplaine, line 153:
```suggestion
    statelessSessions({ /* ... */ }),
```

docs-next/pages/apis/config.mdx
```diff
@@ -208,8 +228,12 @@ import { statelessSessions, withItemData } from '@keystone-next/keystone/session
...
 export default config({
   session: withItemData(
-    statelessSessions({ /* ... */ }),
-    { /* ... */ }
+    statelessSessions({
+      /* ... */
+    }),
+    {
+      /* ... */
+    }
```
gwyneplaine, line 236:
```suggestion
    statelessSessions({ /* ... */ }),
    { /* ... */ }
```

docs-next/pages/apis/config.mdx
```diff
@@ -103,7 +121,9 @@ export default config({
   db: {
     adapter: 'knex',
     url: 'postgres://dbuser:dbpass@localhost:5432/keystone',
-    onConnect: async context => { /* ... */ },
+    onConnect: async context => {
+      /* ... */
+    },
```
gwyneplaine, line 126:
```suggestion
    onConnect: async context => { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -81,7 +97,9 @@ export default config({
   db: {
     adapter: 'prisma_postgresql',
     url: 'postgres://dbuser:dbpass@localhost:5432/keystone',
-    onConnect: async context => { /* ... */ },
+    onConnect: async context => {
+      /* ... */
+    },
```
gwyneplaine, line 102:
```suggestion
    onConnect: async context => { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -7,20 +7,36 @@ The `keystone-next` CLI expects to find a module named `keystone.ts` with a defa
...
+  },
+  session: () => {
+    /* ... */
+  },
+  graphql: {
+    /* ... */
+  },
+  extendGraphqlSchema: {
+    /* ... */
+  },
```
gwyneplaine, line 39:
```suggestion
  lists: { /* ... */ },
  db: { /* ... */ },
  ui: { /* ... */ },
  server: { /* ... */ },
  session: () => { /* ... */ },
  graphql: { /* ... */ },
  extendGraphqlSchema: { /* ... */ },
```

docs-next/pages/apis/config.mdx
```diff
@@ -7,20 +7,36 @@ The `keystone-next` CLI expects to find a module named `keystone.ts` with a defa
 ```typescript filename=keystone.ts
 import { config } from '@keystone-next/keystone/schema';
 
-export default config({ /* ... */ });
+export default config({
+  /* ... */
+});
```
gwyneplaine, line 12:
```suggestion
export default config({ /* ... */ });
```

mitchellhamilton (approved):

JedWatson:
Two thoughts on reviewing the preview deployment:

### We should probably include H1

Sometimes there's quite a long intro, before the first H2, so the first item in the nav (which starts off the screen) is already highlighted when you land on the page, and it's not obvious how to get back to the top. (not that you can't just scroll, but it feels weird that the top section isn't acknowledged in the nav)

e.g https://keystone-next-docs-mvzuhhnm2.vercel.app/apis/config

### H4 (?) links are a bit wacky

Seems like after a certain H level (I'm guessing 4 but haven't checked) the nav items stop being indented, and don't pick up that you've scrolled to them. So on the page below, the smaller headings look like main headings and the main headings look like small headings, and only the main headings that look like small heading have the selected state applied.

Sound confusing? it is 🙂 

https://keystone-next-docs-mvzuhhnm2.vercel.app/guides/cli

(alternatively we could exclude H4 or lower headings from the nav)

---

Hopefully fixing both of these is just a matter of tweaking how different heading levels are handled, since on pages that coincidentally have the right heading structure it seems to work well 👍

JedWatson (changes requested):
See my comment about buggy behaviour with certain combinations of heading levels

mitchellhamilton:
If you're gonna include h1s in it, the h1 on the document field page isn't actually in the mdx so make sure to handle that.

gwyneplaine:
> Two thoughts on reviewing the preview deployment:
> 
> ### We should probably include H1
> Sometimes there's quite a long intro, before the first H2, so the first item in the nav (which starts off the screen) is already highlighted when you land on the page, and it's not obvious how to get back to the top. (not that you can't just scroll, but it feels weird that the top section isn't acknowledged in the nav)
> 
> e.g https://keystone-next-docs-mvzuhhnm2.vercel.app/apis/config
> 
> ### H4 (?) links are a bit wacky
> Seems like after a certain H level (I'm guessing 4 but haven't checked) the nav items stop being indented, and don't pick up that you've scrolled to them. So on the page below, the smaller headings look like main headings and the main headings look like small headings, and only the main headings that look like small heading have the selected state applied.
> 
> Sound confusing? it is 🙂
> 
> https://keystone-next-docs-mvzuhhnm2.vercel.app/guides/cli
> 
> (alternatively we could exclude H4 or lower headings from the nav)
> 
> Hopefully fixing both of these is just a matter of tweaking how different heading levels are handled, since on pages that coincidentally have the right heading structure it seems to work well 👍

H1 inclusion +++ 

The second issue seems to be conflating the styling of the H4 on the page with how h4s are operating in local nav, could you clarify which it is you want to resolve? I'm all for removing h4s from the local-nav

> Two thoughts on reviewing the preview deployment:
> 
> ### We should probably include H1
> Sometimes there's quite a long intro, before the first H2, so the first item in the nav (which starts off the screen) is already highlighted when you land on the page, and it's not obvious how to get back to the top. (not that you can't just scroll, but it feels weird that the top section isn't acknowledged in the nav)
> 
> e.g https://keystone-next-docs-mvzuhhnm2.vercel.app/apis/config
> 
> ### H4 (?) links are a bit wacky
> Seems like after a certain H level (I'm guessing 4 but haven't checked) the nav items stop being indented, and don't pick up that you've scrolled to them. So on the page below, the smaller headings look like main headings and the main headings look like small headings, and only the main headings that look like small heading have the selected state applied.
> 
> Sound confusing? it is 🙂
> 
> https://keystone-next-docs-mvzuhhnm2.vercel.app/guides/cli
> 
> (alternatively we could exclude H4 or lower headings from the nav)
> 
> Hopefully fixing both of these is just a matter of tweaking how different heading levels are handled, since on pages that coincidentally have the right heading structure it seems to work well 👍

@JedWatson That vercel link is out of date, the h4 problem isn't an issue in the latest 
https://keystone-next-docs-git-docsnext-local-nav-keystonejs.vercel.app/guides/cli
Including h1 in the local-nav ++++ though

gwyneplaine:
> If you're gonna include h1s in it, the h1 on the document field page isn't actually in the mdx so make sure to handle that.

oh dear, lemme see what i can do, thnx for the heads up Mitchell 🎉

JedWatson (approved):

PR #4874, opened 2021-02-18
Update test-utils to support network request based tests

timleslie:


tests/api-tests/auth-header.test.js
```diff
@@ -102,51 +107,25 @@ multiAdapterRunners().map(({ runner, adapterName }) =>
       );
 
       describe('logged in', () => {
-        test(
```
timleslie, line 105: We no longer support `Bearer` tokens, so this test can go away 👍

mitchellhamilton (approved):

PR #4875, opened 2021-02-18
Use more generic types in access control

timleslie:
These changes will make this code agnostic to the arguments being passed to imperative access control functions.

### DIFF
--- a/.changeset/plenty-goats-end.md
+++ b/.changeset/plenty-goats-end.md
@@ -1,6 +1,6 @@
 ---
 '@keystonejs/access-control': major
-'@keystonejs/keystone': major
+'@keystonejs/keystone': patch
 ---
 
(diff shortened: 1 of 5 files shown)


mitchellhamilton (commented):
This passes in `access` to the access control functions where it previously wasn't. I think it'd be better to do `{ access: ..., args: Args }` instead of `{ access: ... } & Args`

.changeset/plenty-goats-end.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystonejs/access-control': major
+'@keystonejs/keystone': major
```
mitchellhamilton, line 3: How is this a major for `@keystonejs/keystone`?
```suggestion
'@keystonejs/keystone': patch
```

mitchellhamilton (approved):

PR #4880, opened 2021-02-18
Set port and db.url from env in examples-next

MurzNN:
For easier testing add expanding examples I added ability to declare port number, database adapter and url in `.env` file.


timleslie:
Thanks for the PR @MurzNN. We're planning to spend a bit of time and attention on our example projects next week. I'm going to leave this PR here for now and we'll either merge it in next week when we're looking at these projects, or else incorporate the ideas into a more general PR across all the examples.

MurzNN:
Thanks, will wait for updates of examples-next, they are really helpful not only for building new app, but also for testing core features! I use them actively for testing things before implement in my custom app.

Also on each PR I see the "CloudinaryAdapter requires cloudName, apiKey, and apiSecret" error, seems this is because of missing actual "CLOUDINARY_SECRET" in examples-next/ecommerce/.env file. But adding real credentials into git is bad idea, maybe for solve this problem we can add it to gitignore and fill via real data on some test stage?

MurzNN:
And will be good to have in examples-next some example with LocalFileAdapter.

timleslie:
> Also on each PR I see the "CloudinaryAdapter requires cloudName, apiKey, and apiSecret" error, seems this is because of missing actual "CLOUDINARY_SECRET" in examples-next/ecommerce/.env file. But adding real credentials into git is bad idea, maybe for solve this problem we can add it to gitignore and fill via real data on some test stage?

Yeah, this is slightly annoying. We do actually have the cloudinary keys in the secure storage area of github, it's just that github doesn't let non-core contributors access these keys when running actions on their behalf. As such, you'll see that PRs from core contributors will pass this test, while it will fail for community contributors.

The way we currently handle this is to double check the failure when reviewing community PRs, and if it's this known failure mode then we will simply override the failing PR. It's not ideal, but it's better than nothing. If we becomes a serious impediment to getting PRs shipped we will look for a more streamlined solution.

timleslie:
> And will be good to have in examples-next some example with LocalFileAdapter.

At the moment we're looking at rethinking the way we handle file uploads, so for the time being there won't be a `File` type available until we complete that piece of the system design.

gautamsi:
@timleslie  can you convert those static cloudinary configs and use github env for them too. I think if Iadd those env in my fork, it may help run those tests fine

timleslie:
> @timleslie can you convert those static cloudinary configs and use github env for them too. I think if Iadd those env in my fork, it may help run those tests fine

Making our cloudinary access key/secret available outside of the core team would be a security risk which we're not willing to accept.

gautamsi:
I am asking something else:
currently you have key and cloudname in the tests.yaml workflow action, if you use secrets.CLOUDINARY_KEY and secrets.CLOUDINARY_CLOUD_NAME then these two config can be taken from my fork secrets rather than hard coded. you will not have to expose any keys

MurzNN:
@timleslie Maybe simply add checker to the test code for CLOUDINARY_KEY and skip test (report success), if it is absent (or have default example value)?

dcousens:
I think everything in this pull request has been merged in other ways

PR #4883, opened 2021-02-18
Make the document field toolbar wrap

JedWatson:
This change make the toolbar wrap when the viewport is small enough.

@mitchellhamilton I'm not sure why it was a grid in the first place, this doesn't seem to have changed anything else... I also removed the spacing, we could add it back but it looks fine and is slightly less likely to wrap this way


gautamsi:
I also noticed that the document field on maximizing does not cover the full height of page. should that be fixed too?

mitchellhamilton (approved):

PR #4892, opened 2021-02-19
Add configuration for custom search fields

timleslie:


mitchellhamilton (approved):

gautamsi:
I like this, there is one improvement I can think of is `searchTemplate` which can take a form of function or string literal which returns the query fragment based on search input. This way the search is very useful.

My customer wanted to search from make and model field from list, also the title of ecommerce items.

PR #4893, opened 2021-02-19
Convert test to new interfaces

timleslie:


tests/api-tests/extend-graphql-schema/extend-graphql-schema.test.js
```diff
@@ -1,59 +1,62 @@
...
+        },
+      }),
+    }),
   });
 }
+
 multiAdapterRunners().map(({ runner, adapterName }) =>
   describe(`Adapter: ${adapterName}`, () => {
     describe('keystone.extendGraphQLSchema()', () => {
-      it(
```
timleslie, line 40: Removed because this only tests implementation details.

tests/api-tests/extend-graphql-schema/extend-graphql-schema.test.js
```diff
@@ -1,59 +1,62 @@
-const { Text } = require('@keystonejs/fields');
-const { multiAdapterRunners, setupServer } = require('@keystonejs/test-utils');
+const { multiAdapterRunners, setupFromConfig } = require('@keystonejs/test-utils');
+import { createSchema, list, graphQLSchemaExtension, gql } from '@keystone-next/keystone/schema';
+import { text } from '@keystone-next/fields';
 
 const falseFn = () => false;
 
+const withAccessCheck = (access, resolver) => {
```
timleslie, line 7: The new interfaces don't specifically support access control. This gives one quick pattern for how a dev might wrap their resolvers in a general access control handling mechanism.

Of course the simplest solution would be to include the access control logic inline within the resolver itself.

This could can be revisited if/when we decide on a best practice for this type of thing.

mitchellhamilton (approved):

PR #4896, opened 2021-02-19
Fix toolbar layout issue for dropdowns, reverts #4883

JedWatson:
That wrapping fix introduced a problem for dropdowns, I suspect this component is a bit overloaded so it'll take a more comprehensive update to let the main toolbar wrap without any other side-effects.

In the meantime, this reverts the change


packages-next/fields-document/src/DocumentEditor/primitives/toolbar.tsx
```diff
@@ -40,15 +40,24 @@ type DirectionType = keyof typeof autoFlowDirection;
 const ToolbarGroupContext = createContext<{ direction: DirectionType }>({ direction: 'row' });
 const useToolbarGroupContext = () => useContext(ToolbarGroupContext);
 
-type ToolbarGroupProps = { direction?: DirectionType } & MarginProps &
+type ToolbarGroupProps = { direction?: DirectionType; wrap?: boolean } & MarginProps &
```
mitchellhamilton, line 43: Is wrap consumed by Box?

gautamsi:
I can fix this without reverting

change `Fragment` to `Box` in  https://github.com/keystonejs/keystone/blob/master/packages-next/fields-document/src/DocumentEditor/component-blocks/index.tsx#L76-L96
```tsx
export const BlockComponentsButtons = ({ onClose }: { onClose: () => void }) => {
  const editor = useStaticEditor();
  const blockComponents = useContext(ComponentBlockContext)!;
  const relationships = useDocumentFieldRelationships();
  return (
    <Box>
      {Object.keys(blockComponents).map(key => (
        <ToolbarButton
          key={key}
          onMouseDown={event => {
            event.preventDefault();
            insertComponentBlock(editor, blockComponents, key, relationships);
            onClose();
          }}
        >
          {blockComponents[key].label}
        </ToolbarButton>
      ))}
    </Box>
  );
};
```

timleslie (approved):
👍

PR #4905, opened 2021-02-21
fix types for hooks

gautamsi:
I found soem inconsistencies in the afterDelete and beforeDelete hook types.

found missing `updatedItem` in the hook for afterChange and beforeChange hook

### DIFF
--- a/packages-next/types/src/config/hooks.ts
+++ b/packages-next/types/src/config/hooks.ts
@@ -92,7 +92,9 @@ type ValidateInputHook<TGeneratedListTypes extends BaseGeneratedListTypes> = (
 ) => Promise<void> | void;
 
 type BeforeOrAfterChangeHook<TGeneratedListTypes extends BaseGeneratedListTypes> = (
-  args: ArgsForCreateOrUpdateOperation<TGeneratedListTypes> & ({ updatedItem: TGeneratedListTypes['inputs']['update']})
+  args: ArgsForCreateOrUpdateOperation<TGeneratedListTypes> & {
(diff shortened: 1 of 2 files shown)


packages-next/types/src/config/hooks.ts
```diff
@@ -92,7 +92,7 @@ type ValidateInputHook<TGeneratedListTypes extends BaseGeneratedListTypes> = (
 ) => Promise<void> | void;
 
 type BeforeOrAfterChangeHook<TGeneratedListTypes extends BaseGeneratedListTypes> = (
-  args: ArgsForCreateOrUpdateOperation<TGeneratedListTypes>
+  args: ArgsForCreateOrUpdateOperation<TGeneratedListTypes> & ({ updatedItem: TGeneratedListTypes['inputs']['update']})
```
gautamsi, line 95: should this be properly added with `TGeneratedListTypes['inputs']['create']` as well?

gautamsi:
@timleslie also found that there is no typings for authentication hooks

packages-next/types/src/config/hooks.ts
timleslie: Yeah, I think this should be:

```
updatedItem: TGeneratedListTypes['inputs']['create'] | TGeneratedListTypes['inputs']['update'];
```

timleslie (changes requested):
Good catches, just one change as per comments below 👍

packages-next/types/src/config/hooks.ts
gautamsi: done

timleslie (approved):
👍

PR #4907, opened 2021-02-21
Add float field type for keystone-next

MurzNN:
I created a `float` field for Keystone-next, based on code from integer field type with some modifications. Please review it and comment if something done wrong.
If all is done right, I will try to implement `double` field type via same way.

How to test:
1. In `examples-next/basic/schema.ts` add the float field type, for example, to "Posts" list:
```patch
   Post: list({
     fields: {
       ...
+      float: float(),
        ...
     }
   }),
```

2. Try to create, update, remove items in "Posts" list - all should work well.

Tested with mongoose and knex adapters.


timleslie:
This all looks good to me, thanks for the PR @MurzNN 🙏 

@mitchellhamilton Could you please review the Admin UI component and if you're happy with it, give this a ✅ Thanks!

packages-next/fields/src/types/float/views/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export const controller = (config: FieldControllerConfig): FieldController<string, string> => {
+  return {
+    path: config.path,
+    label: config.label,
+    graphqlSelection: config.path,
+    defaultValue: '',
+    deserialize: data => (parseFloat(data[config.path]) || '') + '',
+    serialize: value => ({ [config.path]: parseFloat(value.replace(',', '.')) || null }),
```
mitchellhamilton, line 52: This logic should be identical to the logic in integer except it should call parseFloat

packages-next/fields/src/types/float/views/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+  );
+};
+
+export const controller = (config: FieldControllerConfig): FieldController<string, string> => {
+  return {
+    path: config.path,
+    label: config.label,
+    graphqlSelection: config.path,
+    defaultValue: '',
+    deserialize: data => (parseFloat(data[config.path]) || '') + '',
```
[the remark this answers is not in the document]
MurzNN: Thanks for figuring out this, I've synced the logic with integer.

packages-next/fields/src/types/float/views/index.tsx
MurzNN: In this logic I wanted to add support for `,` as decimal separator together with `.`, because it is popular in many countries
https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma
So pasting something like `3,14159` should be converted to `3.14159` instead of `314159` or `3`.
But this is debatable change, so now let's keep the previous logic to merge this, I've reverted this line to integer logic.

MurzNN:
I created the PR https://github.com/keystonejs/keystone/pull/4928 that fix this problem.

mitchellhamilton (approved):

PR #4911, opened 2021-02-21
[TEST] Try using my own cloudinary secrets for CI

gautamsi:


gautamsi:
I am not sure why this still fails, I already have secrets for cloudinary env configured from my account.

PR #4913, opened 2021-02-21
Add fake values for cloudinary keys in ecommerce example project so that tests don't fail on PRs from forks

mitchellhamilton:
Note this PR is from a fork to test that this does indeed fix it


timleslie (approved):
👍

PR #4922, opened 2021-02-22
Access control API docs

timleslie:
Preview deployment here: https://keystone-next-docs-git-access-control-api-keystonejs.vercel.app/apis/access-control

PR #4935, opened 2021-02-24
Remove itemId from FieldAccessArgs

timleslie:
Found this gap while documenting the access-control API in #4922

### DIFF
--- a/.changeset/mean-lions-suffer.md
+++ b/.changeset/mean-lions-suffer.md
@@ -2,4 +2,4 @@
 '@keystonejs-next/keystone': major
 ---
 
-Removed `itemIds` from `FieldAccessArgs` and no longer pass this value into field level imperative access control rules.
+Removed `itemId` from `FieldAccessArgs` and no longer pass this value into field level imperative access control rules.
--- a/.changeset/mean-lions-suffer.md
+++ b/.changeset/mean-lions-suffer.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs-next/keystone': major
+---
+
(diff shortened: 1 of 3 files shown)


emmatown (commented):
What's the rationale behind this if they already receive the item?

timleslie:
> What's the rationale behind this if they already receive the item?

Very good question 🤔 it ostensibly makes the arguments for `FieldAccessArgs` match those for `ListAccessArgs`.

On reflection, there are slight differences in the semantics of `itemId` in each of these cases, which might lead to confusion. I think you're right, we'd be better off completely removing the `itemId` from `FieldAccessArgs`.

Once the API docs PR has landed I'll change this PR to remove `itemId` and make sure the docs come along for the ride 👍

.changeset/mean-lions-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystonejs-next/keystone': major
+---
+
+Removed `itemIds` from `FieldAccessArgs` and no longer pass this value into field level imperative access control rules.
```
mitchellhamilton, line 5:
```suggestion
Removed `itemId` from `FieldAccessArgs` and no longer pass this value into field level imperative access control rules.
```

emmatown (commented):
Do you know if there is a FieldAccessArgs or something in @keystone-next/types that would also have itemId on it?

timleslie:
> Do you know if there is a FieldAccessArgs or something in @keystone-next/types that would also have itemId on it?

So... there are, but they're already a little bit broken and somewhat tricky to work with. I have another branch in the works which brings all this stuff up to date with reality, so I'll make sure that this change is reflected in there when it lands.

emmatown (approved):

PR #4936, opened 2021-02-24
Remove itemIds from FieldAccessArgs

timleslie:
Field level imperative access control is applied to each item individually, even in multi-valued mutations. `itemIds` don't make sense to pass into these functions. Developers should be using `itemId` for any checks.

Found this issue as part of documenting the access-control API in #4922


emmatown (commented):
test

emmatown (approved):

PR #4938, opened 2021-02-24
Added decimal field to keystone-next with precision, scale

MurzNN:
Added decimal field type like in #4907, but also extended base class (`packages/fields/src/types/Decimal/Implementation.js` for passing precision, scale values to database column in knex.

In GraphQL query field is represented as string, because of missing "Decimal" type in default GraphQL schema.

There are exists extension https://www.npmjs.com/package/graphql-type-decimal to solve this, maybe add it to keystone-next? Or keep decimal as string, like now?


packages-next/fields/src/types/decimal/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const decimal = <TGeneratedListTypes extends BaseGeneratedListTypes>(
+  config: DecimalFieldConfig<TGeneratedListTypes> = {}
+): FieldType<TGeneratedListTypes> => ({
+  type: Decimal,
+  config,
+  views: resolveView('decimal/views'),
+  getAdminMeta: () => ({
+    precision: config.precision || null,
+    scale: config.precision || null,
```
timleslie, line 27:
```suggestion
    scale: config.scale || null,
```

packages-next/fields/src/types/decimal/index.ts
```diff
@@ -0,0 +1,29 @@
+import { Decimal } from '@keystonejs/fields';
+import type { FieldType, BaseGeneratedListTypes, FieldDefaultValue } from '@keystone-next/types';
+import { resolveView } from '../../resolve-view';
+import type { FieldConfig } from '../../interfaces';
+
+export type Int = number & { __int__: void };
+
```
timleslie, line 7:
```suggestion
```

I'd prefer not to use branded types here. While they send a soft signal to the reader, they make for more verbose code for the user, who has to cast their argument to `Int` wherever they use this. They also don't provide any type safety, as the user can case any value to `Int`, whether it's actually an integer value or not.

.changeset/good-melons-whisper.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-next/fields': minor
+'@keystonejs/fields': minor
+---
+
+Added decimal field type for keystone-next with precision, scale
```
timleslie, line 6:
```suggestion
Added a `decimal` field type to keystone-next.
```

timleslie (changes requested):
Hi @MurzNN, this is all looking pretty good. A couple of minor changes in comments below.

Could I also ask you to update the docs at `docs-next/pages/apis/fields.mdx` to include the `decimal` type.

Now I look I see we're also missing docs for the `float` type. Do you think you could add some docs for that field type as well as a separate PR?

Thanks! 🙏

MurzNN:
I added the documentation for both (`decimal` and `float`) types here, hope my English is not so bad ;)

docs-next/pages/apis/fields.mdx
```diff
@@ -178,6 +178,84 @@ export default config({
...
+});
+```
+
+### decimal
+
+A `decimal` field represents a decimal value.
+
+Options:
+
+- `defaultValue` (default: `undefined`): Can be either a decimal value or an async function which takes an argument `({ context, originalItem })` and returns an decimal value.
```
timleslie, line 224:
```suggestion
- `defaultValue` (default: `undefined`): Can be either a decimal value or an async function which takes an argument `({ context, originalItem })` and returns a decimal value.
```

docs-next/pages/apis/fields.mdx
```diff
@@ -178,6 +178,84 @@ export default config({
...
+  This value will be used for the field when creating items if no explicit value is set.
+  `context` is a [`KeystoneContext`](./context) object.
+  `originalItem` is an object containing the data passed in to the `create` mutation.
+- `precision` (default: `null` treated as `18`): Maximum number of digits that are present in the number.
+- `scale` (default: `null` treated as `4`): Maximum number of decimal places.
```
timleslie, line 229:
```suggestion
- `precision` (default: `18`): Maximum number of digits that are present in the number.
- `scale` (default: `4`): Maximum number of decimal places.
```

packages/fields/src/types/Text/Implementation.js
```diff
@@ -4,10 +4,11 @@ import { KnexFieldAdapter } from '@keystonejs/adapter-knex';
 import { PrismaFieldAdapter } from '@keystonejs/adapter-prisma';
 
 export class Text extends Implementation {
-  constructor(path, { isMultiline }) {
+  constructor(path, { isMultiline, maxLength }) {
```
timleslie, line 7: I think these `Text` changes slipped in from your other PR.

timleslie (changes requested):

docs-next/pages/apis/fields.mdx
```diff
@@ -178,6 +178,84 @@ export default config({
 });
 ```
 
+### float
```
timleslie, line 181: I've pulled these changes into #4961 👍

MurzNN: Thanks!

packages/fields/src/types/Text/Implementation.js
MurzNN: Sorry for this, I remove those code.

timleslie:
This is looking good to me now, thanks @MurzNN.

@mitchellhamilton Could you please take a look at the frontend code? Thanks!

mitchellhamilton (changes requested):
Can you remove all the parseFloat? parseFloat will lose precision

MurzNN:
> Can you remove all the parseFloat? parseFloat will lose precision

@mitchellhamilton, but `.toFixed()` works only with numbers, and we got the string from input widget, not?
```js
'3.14159'.toFixed(2)
// TypeError: "3.14159".toFixed is not a function
```
I can't find the better way for convert string to fractional number, than `parseFloat`.

mitchellhamilton:
You don't need to convert it to anything, the GraphQL API accepts a string

MurzNN:
> You don't need to convert it to anything, the GraphQL API accepts a string

But we need to round the fractional digits to selected scale, for pass the `3.14159` as `3.142` with scale = 3 in field settings, that done via `toFixed(config.fieldMeta.scale)` function.

Or you talk only about `graphql` function, where we can keep the string untouched?

mitchellhamilton:
> But we need to round the fractional digits to selected scale, for pass the 3.14159 as 3.142 with scale = 3 in field settings, that done via toFixed(config.fieldMeta.scale) function.

That should be done by the API/database, not the frontend.

MurzNN:
Ok, I remove parseFloat.

packages-next/fields/src/types/decimal/index.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
gautamsi, line 1: why add it at the top, you can add it where it is needed.

mitchellhamilton: It's not to the whole file(`@ts-ignore` doesn't work like that), it's to the import on the next line.

MurzNN: I simply copy-pasted this trick from `packages-next/fields/src/types/integer/index.ts` file :)

mitchellhamilton (approved):

PR #4939, opened 2021-02-24
Ability to set maxLength and fixedLength for text fields

MurzNN:
Sometimes we need to limit the length of text fields on database storage side, especially when using idField as text, or build index by text field.
This PR is implementing new config options `maxLength` and `fixedLength` - if they exists, it set native type (for PostgreSQL and MySQL) to `Varchar` and `Char` respectively.


timleslie:
Hi @MurzNN, thanks for the PR.

This is a really interesting PR, and it opens up a few questions. There are a few things we need to account for before shipping this as a feature:
 * Do we want to enforce this at the database level?
 * Do we want to validate this at the keystone validation hook layer?
 * What behaviour do we want to enforce if someone attempts to submit a value beyond the max length? Trim and no error? Trim + error? Just an error?
 * How are we going to test this behaviour.

I'm going to add the `needs-end-design` tag to this PR and raise an internal issue to resolve these questions. Once we've got answers to these questions we can move forward with this PR 👍

MurzNN:
First task for me is enforcing this at the database level, to restrict saving too long values on fields, that used as primary key or indexes, and this is easily solved via minimum of code. So, I think, this is a good first step. It not break any previous behavior, because without describing the limit, all works like before.

And the database level do the back-report of the error to upper functions, so users can understand why this item can't be saved, and for GraphQL level this is enough, as I think.

For Admin UI now I implement the simple trim of string, as quick solution. 

Via next steps we can improve this via adding hints that this value have maxLength limit, and showing the error on save (no automatic trim) - this will be definitely better that silent trim like now :)

For testing - we can implement the unit test, that try to define some field with maxLength=3 and try to save the string with length=4 to it, and catch the error.

MurzNN:
And in future extensions also will be good to have `minLength` and `fixedLength` options too, where `fixedLength` stores data as `char(N)` instead of `varchar(N)`. I can add them in next PR, after this.

MurzNN:
I rebase my PR to new core, please review it.
Also I removed the validation from frontend, because it is done well at backend side via displaying error like:
```
 Invalid `prisma.post.create()` invocation: The provided value for the column is too long for the column's type. Column: (not available)
```
I think this will be enough.

timleslie:
After reviewing the open questions on this I've decided to close this PR for now. There are a diverse range of desired behaviours which a user might have when it comes to restricting the length of a text field. Possible solutions include custom UI views, validation hooks, input resolver hooks, and custom fields types. Each of these solutions provides a different set of trade offs which the user can opt into. We may review this design in the future if there is sufficient demand, or if a single compelling solution presents itself, but for now we are better served leaving the decision up to the developer to make the decision about the exact behaviour they require in their system.

MurzNN:
The profit of current improvement is to store strings in database with more optimal way, because, for example, storing country id like `EN`, `AU`, `RU` in `Char(2)` field is much more effective (for storage size, index and query performance), than use unlimited `Text` field type.

PR #4943, opened 2021-02-25
Make DatabaseAPI a union type

timleslie:
This makes explicit what was previously implicit.

### DIFF
--- a/.changeset/hip-kangaroos-float.md
+++ b/.changeset/hip-kangaroos-float.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/types': patch
+---
+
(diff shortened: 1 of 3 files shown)


emmatown (changes requested):
This would make the API basically unusable
https://www.typescriptlang.org/play?ssl=7&ssc=2&pln=1&pc=1#code/CYUwxgNghgTiAEEQBd5gPYDtkgB7IC54BveAa0zyKkwE94BfeAHxPgFssBzddAZxDU6jFmwAOMAJZ92UIfQYBueACgVGbHmQA6CdNmK1kgGbwAFAHI9MqBfiTMaLDnwBKEivhPN+XVJsqDEA

timleslie:
As per the comment above, this actually make things worse, so I'm going to leave it alone for now.

PR #4945, opened 2021-02-25
Remove context arg from raw() and run()

timleslie:
This argument was mostly for backwards compatibility which is no longer required. None of our remaining use cases need this argument, and it makes the documentation much more confusing to explain.

### DIFF
--- a/.changeset/modern-squids-move.md
+++ b/.changeset/modern-squids-move.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+Removed the `context` argument from `KeystoneContext.graphql.raw` and `KeystoneContext.graphql.run`.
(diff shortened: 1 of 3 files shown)


gautamsi:
What happens if the sub query needs to make another query? Like in access control or in any hooks. Like I want to get some data in afterChange hook and put it in queue

timleslie:
> What happens if the sub query needs to make another query? Like in access control or in any hooks. Like I want to get some data in afterChange hook and put it in queue

If you call `context.graphql.raw()` then that `context` object will be the one used as the context for the query. You don't need to pass it in explicitly.

gautamsi:
Make sense, thought it may be removing the context itself. I guess the context will inherit the sudo or exitSudo state

timleslie:
> Make sense, thought it may be removing the context itself. I guess the context will inherit the sudo or exitSudo state

That's correct. This will be made explicit in the documentation.

JedWatson (approved):

PR #4950, opened 2021-02-25
initial cypress setup for @keystone-next/website

gwyneplaine:
As specified on the tin. 
Basic cypress setup for the @keystone-next/website package. 

I notice we're using 'execa' for managing cypress and server processes in other CI tests. 
Couldn't get this working properly with `next dev` so i've opted with the recommendation in the docs instead 

`start-server-and-test` found [here](https://github.com/bahmutov/start-server-and-test)

There's some boilerplate in here from running `cypress open` initially, haven't deleted them yet because it may be handy as a reference later when creating actual tests / adding necessary config.


JedWatson:
The empty file for `docs-next/cypress/videos/examples/home.spec.js.mp4` is a bit weird, videos should be ignored right?

JedWatson:
Also there's a failing cypress test:

```
docs-next/cypress/integration/examples/home.spec.js
```

I'm assuming that's not expected @gwyneplaine?

timleslie (changes requested):
In the past we've found that the cypress tests are sufficiently different from our non-cypress unit tests that it's helpful to have them run as a separate group of things in CI.

Can you tweak the CI config so that these tests get run in the Cypress block, rather than under the generic `package unit tests`?

gwyneplaine:
> In the past we've found that the cypress tests are sufficiently different from our non-cypress unit tests that it's helpful to have them run as a separate group of things in CI.
> 
> Can you tweak the CI config so that these tests get run in the Cypress block, rather than under the generic `package unit tests`?

Yeah this is what i've been working on this morning, realised this is why that one cypress test was failing

gwyneplaine:
> Also there's a failing cypress test:
> 
> ```
> docs-next/cypress/integration/examples/home.spec.js
> ```
> 
> I'm assuming that's not expected @gwyneplaine?

Looks like this is from the bump from 6.x down from 5.6,

timleslie (approved):
This is great!

Issue #4674, opened 2021-01-17
Cannot set `domain` for cookie


wires:
# Bug report

The documentation https://www.keystonejs.com/keystonejs/keystone/#cookie suggests that the `domain` property of cookie can be set (as it refers to https://github.com/expressjs/session#cookie), but this is not the case, only three properties are defined:

```typescript
interface KeystoneOptions {
  cookie?: {
      secure?: boolean;
      maxAge?: number;
      sameSite?: boolean;
  }
}
```

## Describe the bug

I should be able to set the domain property of the authentication cookie, but I can not.

## To Reproduce

Steps to reproduce the behaviour:

1. Set the cookie properties as documented here https://github.com/expressjs/session#cookie

```js
const keystone = new Keystone({
  cookieSecret: 'hello',
  cookie: {
    domain: 'example.com',
    sameSite: "Strict",
    secure: true,
    httpOnly: true,
    maxAge: 1000 * 60 * 60 * 24 * 30, // 30 days
    path: '/'
  },
```

2. Login to Admin UI and check headers

```
set-cookie: keystone.sid=s%3AeIM2QFVjwON-K_5OE-iNu31sZjlMVlSi.3BXYlzylQ6hm90qeOTG7jRHYqmAhRcFZGoIk2kVbD%2B0;
   Path=/; Expires=Tue, 16 Feb 2021 22:19:17 GMT; HttpOnly
```

## Expected behaviour

I was hoping to have `Domain=...` in the header.

If you could point me in the direction where a fix could be applied, I can try to write a PR. Thank you :)

PR #4951, opened 2021-02-25
Added domain option to session

yannick1691:
This adds the option for users to specify a domain for the session that is then used in the cookie.

## Use case of this feature
We ran into an issue where we had keystone running at api.domain.com, this caused the cookie domain to be set to api.domain.com.
We then had a use case for the cookie at example.domain.com, however this domain doesn't have access to the cookie.

Allowing the user to set the domain allows them to set it to: `domain.com`

This would also solve #4674 for keystone-next


timleslie (changes requested):
Hi @yannick1691, thanks for the PR. I think this is a helpful change. To get it through could I get you to address the linting issues. I suspect it just needs trailing commas after `domain`, or you can run `yarn format`.

Also, could you please update the docs in `docs-next/pages/api/session.mdx` to include this option.

Thanks!

yannick1691:
> 
> 
> Hi @yannick1691, thanks for the PR. I think this is a helpful change. To get it through could I get you to address the linting issues. I suspect it just needs trailing commas after `domain`, or you can run `yarn format`.
> 
> Also, could you please update the docs in `docs-next/pages/api/session.mdx` to include this option.
> 
> Thanks!

Hi @timleslie , thank you for the fast response. I have implemented the suggested changes and have updated the documentation. For the documentation I looked at the MDN specs of the domain property. I also link to the MDN page, but still mentioned some of the key parts of the domain property so that users don't miss those. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes

docs-next/pages/apis/session.mdx
```diff
@@ -62,6 +63,8 @@ Options
   For Firefox, the `https:` requirements are ignored when the `secure` attribute is set by localhost (since Firefox 75).
 - `path` (default: `'/'`): A path that must exist in the requested URL, or the browser won't send the cookie header.
   The forward slash (`/`) character is interpreted as a directory separator, and subdirectories will be matched as well: for `path: '/docs'`, `/docs`, `/docs/Web/`, and `/docs/Web/HTTP` will all match.
+- `domain` (default: current document URL): Host to which the cookie will be send [Domain](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes).
+  **Note** Only one domain is allowed, if a domain is specified then subdomains are always included.
```
timleslie, line 67:
```suggestion
- `domain` (default: current document URL): Host to which the cookie will be sent [Domain](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes).
  **Note** Only one domain is allowed. If a domain is specified then subdomains are always included.
```

docs-next/pages/apis/session.mdx
```diff
@@ -62,6 +63,8 @@ Options
   For Firefox, the `https:` requirements are ignored when the `secure` attribute is set by localhost (since Firefox 75).
 - `path` (default: `'/'`): A path that must exist in the requested URL, or the browser won't send the cookie header.
   The forward slash (`/`) character is interpreted as a directory separator, and subdirectories will be matched as well: for `path: '/docs'`, `/docs`, `/docs/Web/`, and `/docs/Web/HTTP` will all match.
+- `domain` (default: current document URL): Host to which the cookie will be sent [Domain](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes).
+  **Note** Only one domain is allowed. If a domain is specified then subdomains are always included.
```
timleslie, line 67:
```suggestion
- `domain` (default: current document URL): Host to which the cookie will be sent. See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes) for more details on the `domain` cookie attribute..
  **Note** Only one domain is allowed. If a domain is specified then subdomains are always included.
```

docs-next/pages/apis/session.mdx
```diff
@@ -62,6 +63,8 @@ Options
   For Firefox, the `https:` requirements are ignored when the `secure` attribute is set by localhost (since Firefox 75).
 - `path` (default: `'/'`): A path that must exist in the requested URL, or the browser won't send the cookie header.
   The forward slash (`/`) character is interpreted as a directory separator, and subdirectories will be matched as well: for `path: '/docs'`, `/docs`, `/docs/Web/`, and `/docs/Web/HTTP` will all match.
+- `domain` (default: current document URL): Host to which the cookie will be sent. See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes) for more details on the `domain` cookie attribute..
+  **Note** Only one domain is allowed. If a domain is specified then subdomains are always included.
```
timleslie, line 67:
```suggestion
  **Note**: Only one domain is allowed. If a domain is specified then subdomains are always included.
```

timleslie (approved):
This looks great, thanks @yannick1691 🙏

PR #4965, opened 2021-03-01
Fix bug in passwordResetLink

timleslie:
These field names need to correspond to `const tokenType = 'passwordReset';` in `getPasswordSchemaReset.ts`. They aren't meant to be dynamic on the field name.

### DIFF
--- a/.changeset/soft-stingrays-happen.md
+++ b/.changeset/soft-stingrays-happen.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Fixed bug when using `passwordResetLink` with a secretField other than 'password'.
--- a/.changeset/soft-stingrays-happen.md
+++ b/.changeset/soft-stingrays-happen.md
@@ -1,5 +1,5 @@
 ---
-'@keystone-next/auth': patch
+'@keystone-next/auth': major
 ---
 
-Fixed bug when using `passwordResetLink` with a secretField other than 'password'.
+Fixed a bug when using `passwordResetLink` with a `secretField` other than `'password'`. If you used a `secretField` other than `'password'` then the generated fields in your schema will change.
--- a/packages-next/auth/src/index.ts
+++ b/packages-next/auth/src/index.ts
@@ -62,9 +62,9 @@ export function createAuth<GeneratedListTypes extends BaseGeneratedListTypes>({
   } as const;
 
   const additionalListFields = {
-    [`${secretField}ResetToken`]: password({ ...fieldConfig }),
-    [`${secretField}ResetIssuedAt`]: timestamp({ ...fieldConfig }),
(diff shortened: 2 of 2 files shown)


JedWatson:
You mean `getPasswordSchemaReset.ts`

JedWatson:
So overall I'm in favour of this, but I'd like to go more comprehensively in one direction or the other.

These fields were _supposed_ to be variable, because it's a "secret" and may not be called "password".

But can we be consistent? it's configured as a `secret` field, which is almost always called `password`, and we're hard-coding `password` in some places while it's still `secret` in others. Confusing.

I don't think I'm unhappy with this PR, since among other things it fixes a bug, but I do think I'm unhappy with the overall balance of apparently-configurable vs not-actually-configurable combined with (I think) _we use two words for the same thing_.

@timleslie if you agree, we should probably backlog another pass to clean this up properly even further, align on one consistent term, and be more opinionated overall in the name of simplicity (tbh at this point I'd go for "password" over "secret", and I think "secret" is a hangover from Keystone 5 when it genuinely was less opinionated)

(to be clear, I'm not suggesting we hard code the two primary fields - currently, `identity` and `secret` - they are defined by the list schema and only referred to here, that's fine. but you configure the "secret" field, and then we call it "password" in the context of resetting it, and that's confusing... except that devs using keystone will, if they follow the defaults / examples, see "password" and "passwordReset" columns in their database and that actually does make sense 🤦‍♂️)

.changeset/soft-stingrays-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/auth': patch
```
JedWatson, line 2: This isn't a patch; it changes database structure for anyone using `auth` with a field path other than `password` for the secret field.

We can decide to ship this as a breaking change, or we can fix the bug in the other direction (by making `getPasswordResetSchema.ts` generate the field path dynamically based on the `secretField` config when the `tokenType` is `passwordReset`)

JedWatson (changes requested):

.changeset/soft-stingrays-happen.md
timleslie, line 2:
```suggestion
'@keystone-next/auth': major
```

.changeset/soft-stingrays-happen.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fixed bug when using `passwordResetLink` with a secretField other than 'password'.
```
timleslie, line 5:
```suggestion
Fixed bug when using `passwordResetLink` with a `secretField` other than 'password'. If you used a `secretField` other than `password` this change the name of the automatically generated fields, which will result in a change to your GraphQL and database schema.
```

JedWatson (approved):

PR #4969, opened 2021-03-01
Add universal cypress tests for @keystone-next/website

gwyneplaine:
This PR consists of a few different things: 
1. Markup adjustments, there's a lot of `divs` in our markup, which makes asserting nav and content block existence hard. This has been amended for better accessibility and more reliable tests. 
2. We test that each page has a nav, a header and a main content block to ensure that users are wrapping their mdx content in the Markdown / Page component. 
3. We test that markdown and/or react page content doesn't include any empty anchors i.e. `[Hello](#)` or `<a href="#"/>` 

These test are REALLY SLOW. Because we're navigating to ALL pages to make these assertins. 
Screen Shot 2021-03-01 at 4 08 54 pm

This might be a Jurassic park scientists moment if i'm honest, but I'm hoping its because I'm inefficient with cypress and one of this PR's reviewers knows of a simple test optimisation.


docs-next/components/Page.tsx
```diff
@@ -23,8 +23,8 @@ export const Page = ({
   const [mobileNavCollapsed, setMobileNavCollapsed] = useState(true);
   const contentRef = useRef<HTMLDivElement | null>(null);
   return (
-    <div className="antialiased pb-24">
-      <div className="py-4 border-b border-gray-200">
+    <div className="pb-24">
```
gwyneplaine, line 26: does this wrapping div exist solely for this pb-24 class? 
Is it safe to assume that we can add that class to the body of our __document.tsx file @JedWatson ? 

I didn't move it assuming it needs to be here, but would be nice for us to be able to reduce the amount of wrapping divs in our markup.

docs-next/pages/_document.tsx
```diff
@@ -27,7 +27,7 @@ class MyDocument extends Document {
             }}
           />
         </Head>
-        <body>
+        <body className="antialiased">
```
gwyneplaine, line 30: moved over from the above file, we probably want this class applied to the body (@JedWatson  to confirm)

docs-next/cypress/integration/general_spec.js
```diff
@@ -0,0 +1,27 @@
...
+    return cy.get('nav a').each(navItem => {
+      cy.visit(navItem[0].href).get('nav');
+      cy.get('header');
+      cy.get('main');
+    });
+  });
+  it('shouldn not have any empty links', () => {
+    return cy.get('nav a').each(navItem => {
+      cy.visit(navItem[0].href);
+      cy.document().then(document => {
```
gwyneplaine, line 15: cy.get() will throw if it times out on a query. 
In this case we want to run an assertion if anchors DO exist, and not fail if they don't. 

I hate this because there's a LOT of ways this can go wrong, but according to cypress docs this is probably the most reliable way to make this kind of conditional assertion.

docs-next/components/Navigation.tsx
```diff
@@ -52,6 +52,7 @@ export const Navigation = () => {
         <NavItem href="/guides/hooks" isPlaceholder>
           Hooks
         </NavItem>
+
```
timleslie, line 55:
```suggestion
```

mitchellhamilton:
> These test are REALLY SLOW.

If all we're doing is checking that some things exist, we could just do some HTTP requests in parallel and use an html parser and check that the things we want to exist do in fact exist?

gwyneplaine:
> > These test are REALLY SLOW.
> 
> If all we're doing is checking that some things exist, we could just do some HTTP requests in parallel and use an html parser and check that the things we want to exist do in fact exist? (not saying that we necessarily should just that it might be a good option if the test time causes problems)

yeah this is a good alternative, cypress includes a cy.request, since this is a next app we can probably just test markup.

docs-next/pages/_document.tsx
JedWatson: I'm fine with this 👍 

(I'm also not 100% clear on whether we need this at all, because it may have come over from a previous example, but that's a different issue for a different PR)

timleslie (approved):
Happy to ship this for now and loop back on performance issues in the (possibly near) future 👍

PR #4982, opened 2021-03-01
remove getting started and fake link in getting started

gwyneplaine:
fake link for validating test failure made it into master, 
removed from master along with getting-started in nav


timleslie (approved):
👍

PR #4984, opened 2021-03-02
Fix crash if the prisma client directory exists but the prisma schema doesn't

mitchellhamilton:
Found this when writing tests for the examples. Also made it so the checking logic does fewer file system operations.

I would write a test for this but I couldn't find a nice place for it where there are tests for generating migrations or something?

### DIFF
--- a/.changeset/wet-nails-move.md
+++ b/.changeset/wet-nails-move.md
@@ -1,0 +1,5 @@
+---
+'@keystonejs/adapter-prisma': patch
+---
+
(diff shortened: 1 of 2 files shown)


timleslie:
> I would write a test for this but I couldn't find a nice place for it where there are tests for generating migrations or something?

At the moment we don't have tests of the migration paths. We exercise the adapter through the API tests, but these take advantage of the `prototype` mode, not any of the actual migration code.

Writing tests for actual migrations is something we definitely need, but can be considered out of scope for this PR 👍

PR #4992, opened 2021-03-02
Table of Contents tests

gwyneplaine:
* Add table of contents tests
* Remove typescript <reference/> tags for js files (oopsies)


timleslie (approved):
👍

PR #4993, opened 2021-03-02
Smoke tests for examples

emmatown:
These are intended to be quite minimal and just ensure that the projects can be started in dev and prod and you can do a few basic things.

The ecommerce example tests are disabled in this PR because they're currently failing in production because it seems as though the wrong field view is used for a field, we can look into that later, IMO it shouldn't block this PR.


examples-next/todo/schema.ts
```diff
@@ -1,6 +1,7 @@
 import { createSchema, list } from '@keystone-next/keystone/schema';
 import { checkbox, password, relationship, text, timestamp } from '@keystone-next/fields';
 
+// this implementation for createdBy and updatedBy is currently wrong so they're disabled for now
```
timleslie, line 4: @mitchellhamilton Could you please create a ticket to capture this 👍

tests/examples-smoke-tests/ecommerce.test.disabled.ts
```diff
@@ -0,0 +1,18 @@
+import { Browser, Page } from 'playwright';
+import { exampleProjectTests, initUserTest } from './utils';
+
+// this is disabled currently because it's currently failing and we want to get the tests in without being blocked on this
```
timleslie, line 4: @mitchellhamilton And a ticket for this 👍

mitchellhamilton: There already is.

tests/examples-smoke-tests/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@keystonejs/examples-smoke-tests",
+  "description": "A set of tests for smoke testing our examples",
+  "private": true,
+  "version": "9.0.0",
```
timleslie, line 5:
```suggestion
  "version": "0.0.0",
```
Should we start at `0.0.0` and let the changeset bot do its thing? We probably want a changeset on this PR as well in that case.

tests/examples-smoke-tests/utils.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const promiseSignal = (): Promise<void> & { resolve: () => void } => {
+  let resolve;
+  let promise = new Promise<void>(_resolve => {
+    resolve = _resolve;
+  });
+  return Object.assign(promise, { resolve: resolve as any });
+};
+
+export const initUserTest = (getPage: () => playwright.Page) => {
```
timleslie, line 30: Help make it clear that this is testing the `initFirstItem` part of the API.

```suggestion
export const initFirstItemTest = (getPage: () => playwright.Page) => {
```

timleslie (changes requested):
A couple of small changes, other than that looks good 👍

tests/examples-smoke-tests/package.json
mitchellhamilton: Updated it to 0.0.0 but I'm not going to add a changeset because this package isn't published so it does not matter.

timleslie (approved):
👍

PR #5006, opened 2021-03-03
form submit tests

gwyneplaine:
This PR includes:
* enabling the experimental fetch polyfill feature flag, so that cypress can stub fetch requests. **
* adding basic tests for the submission form on the welcome page. 

** This flag is marked for deprecation, and as such we should upgrade cypress from 5.6 --> 6 ASAP to make use of the new (less hacky) intercept() api.

### DIFF
--- a/docs-next/cypress.json
+++ b/docs-next/cypress.json
@@ -1,1 +1,3 @@
-{}
+{
+  "experimentalFetchPolyfill": true
+}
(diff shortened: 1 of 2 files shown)


timleslie:
> ** This flag is marked for deprecation, and as such we should upgrade cypress from 5.6 --> 6 ASAP to make use of the new (less hacky) intercept() api.

Yep, we should be able to pull down a lot of the existing cypress tests during the controlled demolition, which will make this transition much easier in the very near future 👍

timleslie (approved):
Looks good 👍

PR #5007, opened 2021-03-03
Add custom idField types support to Prisma adapter

MurzNN:
Now Keystone's Prisma adapter force converting all values for `id` field to number via `Number(value)`.

Here is my attempt to solve this via implementing function `_formatValue(value, path)` that try to detect field type and do the right conversion. I replace the `Number()` calls to it hope in all right places. 

The function `_getFieldStorageFormat(path)` now trying to extract the field type from Prisma's schema string via regex:
```
     let prismaSchemaFormat = prismaSchema[0].replace(/^(\w+)\s+(\w+).+$/, '$2');
```
This is dubious way that might be better, if you have idea how to implement this better - please describe, I'll try to rewrite this.


packages/adapter-prisma/lib/adapter-prisma.js
```diff
@@ -312,14 +312,39 @@ class PrismaListAdapter extends BaseListAdapter {
...
   }
 
+  _formatValue(value, path) {
+    let format = this._getFieldStorageFormat(path);
+    switch (format) {
+      case 'Boolean':
+        return Boolean(value);
+      case 'DateTime':
+      case 'String':
+        return value + '' || '';
```
gautamsi, line 322: should this be
```suggestion
        return (value || '') + '';
```

packages/adapter-prisma/lib/adapter-prisma.js
```diff
@@ -312,14 +312,39 @@ class PrismaListAdapter extends BaseListAdapter {
...
+      case 'Int':
+      case 'Float':
+      case 'Decimal':
+        return Number(value) || null;
```
gautamsi, line 326: should this change from Number to parseInt, parseFloat etc. also does this decimal represented as string in database or number?

MurzNN: Fixed, thanks!

MurzNN:
I also added fix from https://github.com/keystonejs/keystone/issues/5014 - with it all tests seems passed! :rocket:

timleslie:
Hi @MurzNN, thanks for the PR. This is another one which we're going to need to sit on for a while. In particular, we need to properly define which fields types are and are not supported as potential ID fields, and make sure that we have proper testing in place. I'm going to update the docs now to point out that we currently only support the `autoincrement` field so that users can fully understand the current limitations of this feature.

MurzNN:
With #5665 (New core) there are a lot of things changed, so I implemented this fix into #4826, so let's close this.

PR #5009, opened 2021-03-03
Rename updateAuthToken to createAuthToken

timleslie:
This better captures what this function actually does.


gwyneplaine (approved):
lgtm

PR #5011, opened 2021-03-03
Update auth package cosmetics

timleslie:
This PR contains a bit of a grab-bag of cosmetic changes which came from the process of writing docs for the auth package. There aren't any functional changes in any of this.

### DIFF
--- a/packages-next/auth/src/index.ts
+++ b/packages-next/auth/src/index.ts
@@ -32,18 +32,22 @@ export function createAuth<GeneratedListTypes extends BaseGeneratedListTypes>({
   // to match this. -TL
   const protectIdentities = true;
   const gqlNames: AuthGqlNames = {
-    CreateInitialInput: `CreateInitial${listKey}Input`,
-    createInitialItem: `createInitial${listKey}`,
+    // Core
     authenticateItemWithPassword: `authenticate${listKey}WithPassword`,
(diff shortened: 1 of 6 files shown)


packages-next/auth/src/index.ts
```diff
@@ -185,74 +180,61 @@ export function createAuth<GeneratedListTypes extends BaseGeneratedListTypes>({
...
+      const msg = `A createAuth() invocation for the "${listKey}" list specifies ${s} as its secretField but no field with that key exists on the list.`;
+      throw new Error(msg);
     }
-    const secretPrototype =
-      secretFieldConfig.type &&
-      secretFieldConfig.type.implementation &&
-      secretFieldConfig.type.implementation.prototype;
-    const secretTypename = secretFieldConfig.type && secretFieldConfig.type.type;
+
+    // FIXME: Until we explicity support user defined field types, should we just check against .type.type === 'Password'?
```
gwyneplaine, line 207: are FIXME's and TODO's something that we reliably action in our workflow?
If yes, then lgtm, if no then we should either pair this with a matching JiRA issue, or ask the question directly in slack / confluence

gwyneplaine (changes requested):

gwyneplaine (approved):
lgtm

PR #5012, opened 2021-03-04
improve SEO 

gwyneplaine:
Minor upkeep things:
* Add og info
* Add twitter card info 
* Add title per page**
* Port over rest of the relevant meta and head config from Keystone docs 

** Front-matter support sucks for withMDX plugin, and I'm not really willing to trust a third party extension atm, so right now we're pulling the first title off each document in React and making that the title. 

We're already iterating over children to find headings anyways, so this is computationally fine despite being a little icky feeling.


docs-next/components/Page.tsx
```diff
@@ -13,110 +14,119 @@ import { TableOfContents } from './TableOfContents';
...
   isProse,
 }: {
   headings?: Heading[];
   children: ReactNode;
+  title?: string;
   isProse?: boolean;
 }) => {
   const [mobileNavCollapsed, setMobileNavCollapsed] = useState(true);
   const contentRef = useRef<HTMLDivElement | null>(null);
+  const metaTitle = title ? `Keystone Next Documentation - ${title}` : `Keystone Next`;
```
jesstelford, line 28: nit: Can this be swapped around? Ie; `${title} - Keystone Next Documentation`. Reason: The browser tab shows the title, but if I have multiple docs pages open, it's impossible to know which is which because the phrase` Keystone Next Documentation` fills the entire tab width.

docs-next/components/Page.tsx
```diff
@@ -13,110 +14,119 @@ import { TableOfContents } from './TableOfContents';
...
-              className="w-8 h-8 mr-2"
-            />
-            <Link href="/" passHref>
-              <a className="bg-clip-text text-transparent hover:text-transparent bg-gradient-to-r from-lightblue-500 to-indigo-600 hover:from-lightblue-400 hover:to-indigo-700 text-xl font-semibold">
-                Keystone Next
+    <Fragment>
+      <Head>
+        <meta property="og:site_name" content={metaTitle} />
```
jesstelford, line 32: Pro tip: If you throw a `key` on this, it'll allow other components further down the tree to override the value without causing a duplicate tag to appear (as long as the `key`s match)

Similarly for the `<title>` below, but Next.js special-cases `<title>` already for you, so no need to manually add a `key` there.

jesstelford (approved):
Can definitely be merged as-is. I left a couple of comments that may interest you also 👍

docs-next/components/Page.tsx
gwyneplaine: good call, thnx Jess 🎉

PR #5021, opened 2021-03-04
Remove empty line at end of code examples

emmatown:
It looks like there was a fix for this previously but it didn't work for things that didn't really have syntax highlighting. Look at the command line page for an example once the preview deployment is up.


timleslie (approved):
Excellent!

PR #5033, opened 2021-03-08
Add ability to enable a graphql API endpoint in the Admin UI. Add example for this feature in examples-next

rohan-deshpande:
### Added 

* Config option for toggling the graphql api endpoint in the admin ui
* Template for graphql api endpoint to be generated for the admin ui

### Changed 

* The ApolloServer is no longer added to express if the new config option is enabled as the API is made available through a next-js API route

### DIFF
--- a/.changeset/slimy-elephants-hear.md
+++ b/.changeset/slimy-elephants-hear.md
@@ -1,0 +1,8 @@
+---
+'keystone-next-app': patch
+'@keystone-next/admin-ui': patch
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
+---
+
+Added graphql api route example for admin UI
--- a/.changeset/slimy-elephants-hear.md
+++ b/.changeset/slimy-elephants-hear.md
@@ -1,8 +1,7 @@
 ---
-'keystone-next-app': patch
 '@keystone-next/admin-ui': patch
 '@keystone-next/keystone': patch
 '@keystone-next/types': patch
 ---
 
-Added graphql api route example for admin UI
+Added `experimental` config namespace and `enableNextJsGraphqlApiEndpoint` property to support the GraphQL API being served from a Next.js API route rather than Express
--- a/docs-next/pages/apis/access-control.mdx
+++ b/docs-next/pages/apis/access-control.mdx
@@ -12,15 +12,9 @@ export default config({
   lists: createSchema({
     ListKey: list({
       fields: {
-        fieldName: text({
-          access: {
-            /* ... */
-          },
-        }),
+        fieldName: text({ access: { /* ... */ }, }),
       },
-      access: {
-        /* ... */
-      },
+      access: { /* ... */ },
     }),
   }),
 });
--- a/docs-next/pages/apis/access-control.mdx
+++ b/docs-next/pages/apis/access-control.mdx
@@ -12,9 +12,15 @@ export default config({
   lists: createSchema({
     ListKey: list({
       fields: {
-        fieldName: text({ access: { /* ... */ }, }),
+        fieldName: text({
+          access: {
+            /* ... */
+          },
+        }),
       },
-      access: { /* ... */ },
+      access: {
+        /* ... */
+      },
     }),
   }),
 });
--- a/packages-next/admin-ui/src/system/buildAdminUI.ts
+++ b/packages-next/admin-ui/src/system/buildAdminUI.ts
@@ -1,5 +1,6 @@
 export async function buildAdminUI(projectAdminPath: string) {
   // importing next/dist/build is quite expensive so we're requiring it lazily
-  const build = (require('next/dist/build') as typeof import('next/dist/build')).default;
+  const next = 'next/dist/build';
+  const build = require(next).default;
   await build(projectAdminPath);
 }
--- a/packages-next/admin-ui/src/system/buildAdminUI.ts
+++ b/packages-next/admin-ui/src/system/buildAdminUI.ts
@@ -1,5 +1,6 @@
 export async function buildAdminUI(projectAdminPath: string) {
   // importing next/dist/build is quite expensive so we're requiring it lazily
+  /** We do this to stop webpack from bundling next inside of next */
   const next = 'next/dist/build';
   const build = require(next).default;
   await build(projectAdminPath);
--- a/packages-next/admin-ui/src/system/createAdminUIServer.ts
+++ b/packages-next/admin-ui/src/system/createAdminUIServer.ts
@@ -1,5 +1,4 @@
 import url from 'url';
-import next from 'next';
 import express from 'express';
 import type { KeystoneConfig, SessionStrategy, CreateContext } from '@keystone-next/types';
 import { createSessionContext } from '@keystone-next/keystone/session';
@@ -11,6 +10,8 @@ export const createAdminUIServer = async (
   projectAdminPath: string,
   sessionStrategy?: SessionStrategy<any>
 ) => {
+  const thing = 'n' + 'ext';
+  const next = require(thing);
   const app = next({ dev, dir: projectAdminPath });
   const handle = app.getRequestHandler();
   await app.prepare();
--- a/packages-next/admin-ui/src/system/createAdminUIServer.ts
+++ b/packages-next/admin-ui/src/system/createAdminUIServer.ts
@@ -19,7 +19,7 @@ export const createAdminUIServer = async (
   const publicPages = ui?.publicPages ?? [];
   return async (req: express.Request, res: express.Response) => {
     const { pathname } = url.parse(req.url);
-    if (pathname?.startsWith('/_next')) {
+    if (pathname?.startsWith('/_next') || pathname === '/api/graphql') {
       handle(req, res);
       return;
     }
--- a/packages-next/admin-ui/src/system/createAdminUIServer.ts
+++ b/packages-next/admin-ui/src/system/createAdminUIServer.ts
@@ -10,7 +10,8 @@ export const createAdminUIServer = async (
   projectAdminPath: string,
   sessionStrategy?: SessionStrategy<any>
 ) => {
-  const thing = 'n' + 'ext';
+  /** We do this to stop webpack from bundling next inside of next */
+  const thing = 'next';
   const next = require(thing);
   const app = next({ dev, dir: projectAdminPath });
   const handle = app.getRequestHandler();
--- a/packages-next/admin-ui/src/templates/api.ts
+++ b/packages-next/admin-ui/src/templates/api.ts
@@ -1,0 +1,1 @@
+export const apiTemplate = '';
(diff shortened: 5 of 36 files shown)


packages-next/admin-ui/src/templates/index.ts
```diff
@@ -37,4 +38,5 @@ export const writeAdminFiles = (
     ({ adminUILabels: { path }, key }) =>
       ({ mode: 'write', src: itemTemplate(key), outputPath: `pages/${path}/[id].js` } as const)
   ),
+  { mode: 'write', src: apiTemplate, outputPath: 'pages/api/graphql.js' },
```
mitchellhamilton, line 41: Should be conditional based on the option.

rohan-deshpande: Yah I tried this but typescript was very angry at me. Will try again.

packages-next/admin-ui/src/system/createAdminUIServer.ts
```diff
@@ -11,14 +10,16 @@ export const createAdminUIServer = async (
   projectAdminPath: string,
   sessionStrategy?: SessionStrategy<any>
 ) => {
+  const thing = 'n' + 'ext';
```
mitchellhamilton, line 13: Probs worth adding a comment above this and the other one to say this is to stop webpack from bundling next inside neext.

rohan-deshpande: Is the string concat actually needed here? Is that thing that will _trick_ webpack?

packages-next/admin-ui/src/templates/index.ts
mitchellhamilton: The answer to TypeScript complaining will probably be to write `as const` after `mode: 'write'`

packages-next/admin-ui/src/system/createAdminUIServer.ts
mitchellhamilton: Don't think so since it works on the other one without the string concat. Having the string in a variable and then calling require with it is def needed though.

mitchellhamilton:
You're gonna have to run `yarn manypkg fix` to fix CI

packages-next/admin-ui/src/system/createAdminUIServer.ts
rohan-deshpande: Yah cool, will just put it in a var and dump the concat then, cheers!

mitchellhamilton (commented):
Looks fine to me(other than it needs changesets and CI needs to pass) now but I feel like @timleslie will have opinions on this

packages-next/keystone/src/lib/createKeystone.ts
```diff
@@ -25,7 +25,7 @@ export function createKeystone(config: KeystoneConfig, dotKeystonePath: string,
     adapter = new MongooseAdapter({ mongoUri: db.url, ...db.mongooseOptions });
   } else if (db.adapter === 'prisma_postgresql') {
     adapter = new PrismaAdapter({
-      getPrismaPath: () => path.join(dotKeystonePath, 'prisma'),
+      getPrismaPath: () => '.keystone/prisma',
```
timleslie, line 28:
```suggestion
      getPrismaPath: () => path.join(dotKeystonePath, 'prisma'),

```

packages-next/types/src/config/index.ts
```diff
@@ -85,6 +85,8 @@ export type AdminUIConfig = {
     isValidSession: boolean;
     createContext: CreateContext;
   }) => MaybePromise<{ kind: 'redirect'; to: string } | void>;
+  /** Enables nextjs graphql api route mode */
+  enableGraphQlApiEndpoint?: boolean;
```
timleslie, line 89: This will need to be documented in the  API docs in the "Advanced configuration:" section of `docs-next/pages/apis/config.mdx`.

packages-next/admin-ui/src/templates/api.ts
```diff
@@ -0,0 +1,36 @@
+export const apiTemplate = `
```
timleslie, line 1: This code is a long way away from the very similar code in `createExpressServer`.

I wonder if there's a way we can bring those two things closer together? 🤔 

Also, since this isn't paramterised, should we put in in `static` and use `copy` operation rather than a `write`?

timleslie (changes requested):
A few minor nits and thoughts to follow up on.

packages-next/types/src/config/index.ts
JedWatson: oof, is that how we capitalise `enableGraphQlApiEndpoint`?

Took a look at the types and we're not super consistent here, but an equivalent config option is `extendGraphqlSchema` so I'm pretty sure we decided that "Graphql" is just a word in terms of camelcasing

Also I get why this is nested under `ui` but it's ... weird. I misunderstood what this option meant on the first read. We should come back to that later, in the meantime my preferred name for this is

```
  /** Enables nextjs graphql api route mode */
  addNextjsGraphqlEndpoint?: boolean;
```

Basically, tells the ui to "add a Next.js GraphQL Endpoint", and then when we're creating an express server, if that's the case we don't also add an express handler for that.

JedWatson:
Actually, carrying forward with my config nits, it's slightly outside the scope of this PR but maybe we can make a quick call.

This config option really _should_ be added now, but also, we need to loop back and think more about how the whole API works in terms of where we're generating the graphql endpoint (and how that might change between dev, prod, and different deployment targets)

So I propose we add a top-level `experimental` grab bag of config that lets us say "this is a thing we're trying out"

The change to this PR would be moving the config key `enableGraphQlApiEndpoint` here:

```
type KeystoneConfig = {
  /* ... */
  experimental?: {
    /** Adds an API route to the Admin UI to serve the graphql endpoint, and skips adding express middleware for it */
    useNextjsRouteForGraphqlEndpoint?: boolean;
  }
}
```

I think that resolves all of my concerns about getting this shipped; it's an experimental feature, and also I'm not sure where the config should live yet.

packages-next/keystone/src/lib/createKeystone.ts
mitchellhamilton: not resolving this path is important so that adapter-prisma will do it because `@keystone-next/keystone` will be bundled in the next app(but `@keystonejs` packages won't be, re: why just `@keystonejs`? because it's an easy line to draw because it's all backend code) and webpack will essentially mock `process.cwd()` to return `"/"` so without this change, the prisma adapter would try to write to `/.keystone/prisma`(as in at the root of the file system).

packages-next/admin-ui/src/templates/api.ts
rohan-deshpande: > This code is a long way away from the very similar code in createExpressServer.

Hmm but there does seem to be a fairly heavy codependency between `admin-ui` and `system` it's already there as `createAdminUIServer` is imported into `createExpressServer` 

> Also, since this isn't paramterised, should we put in in static and use copy operation rather than a write?

Easily done!

rohan-deshpande:
- [ ] Add experimental config and put the new config option in there
- [ ] Put the graphql api route page template inside of `static` and use `copy` instead of `write`

packages-next/types/src/config/index.ts
rohan-deshpande: Just saw this comment, okay can easily change this

rohan-deshpande: I think enable is better for boolean values though as a key

rohan-deshpande: also do we need to document stuff under `experimental` _right now_?

packages-next/admin-ui/src/templates/api.ts
rohan-deshpande: On second thought I think this should live in templates... I can't see a `static` dir in there but it's definitely a template even if it's not dynamic currently. I think moving it away from there might be a bit weird. 

Lemme know if I've misunderstood something.

packages-next/keystone/src/lib/createKeystone.ts
JedWatson: @mitchellhamilton do you mean we _should_ merge @timleslie's suggestion? or not?

mitchellhamilton: I mean that we very much _shouldn't_

docs-next/pages/apis/access-control.mdx
```diff
@@ -12,9 +12,15 @@ export default config({
   lists: createSchema({
     ListKey: list({
       fields: {
-        fieldName: text({ access: { /* ... */ }, }),
```
JedWatson, line 15: I'm not sure why these got changed, looks like our prettier ignore might not be configured correctly? or this is maybe something that happened locally

Because this shows up w/ formatting in the docs though, we'll want to roll all the mdx changes back in this PR

@gwyneplaine do you know if we've made any changes to the repo config that would have caused this to happen?

gwyneplaine: @JedWatson not aware of any prettier config changes, however these single-line formatted comments were manual commits via suggestion iirc.

I honestly i think we should stop fighting prettier at this point cc @timleslie

JedWatson: @gwyneplaine agree on not fighting prettier generally

In this case I'd rather we tell it to ignore `mdx` files entirely if we can't just tell it to ignore `tsx` in `mdx`

This formatting ends up in the docs, and all the extra line breaks hurt readability much more than the benefit having prettier runs on our examples adds, at least at this point imo

rohan-deshpande: @JedWatson how do you want me to resolve? Just checkout the MDX files from origin/master and push back up?

packages-next/admin-ui/src/system/buildAdminUI.ts
```diff
@@ -1,5 +1,6 @@
 export async function buildAdminUI(projectAdminPath: string) {
   // importing next/dist/build is quite expensive so we're requiring it lazily
-  const build = (require('next/dist/build') as typeof import('next/dist/build')).default;
+  const next = 'next/dist/build';
```
mitchellhamilton, line 3:
```suggestion
  /** We do this to stop webpack from bundling next inside of next */
  const next = 'next/dist/build';
```

.changeset/slimy-elephants-hear.md
```diff
@@ -0,0 +1,8 @@
+---
+'keystone-next-app': patch
+'@keystone-next/admin-ui': patch
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
+---
+
+Added graphql api route example for admin UI
```
mitchellhamilton, line 8: You don't need a changeset for `keystone-next-app` since it's not published but for the other packages, you should provide a more descriptive summary since seeing `Added graphql api route example for admin UI` in the changelog for any of the other packages wouldn't really make sense

docs-next/pages/apis/access-control.mdx
rohan-deshpande: Reverted all to `origin/master`

.changeset/slimy-elephants-hear.md
rohan-deshpande: Done!

timleslie (approved):
This all looks good to me 👍

Issue #5025, opened 2021-03-07
Remove unused keystone-5 packages from this repo


JedWatson:
As outlined in #4962 (what's next for KeystoneJS)

Now that we have the [keystone-5 repo](https://github.com/keystonejs/keystone-5) ready to go, we can start cleaning up this codebase to prepare for our next round of changes in keystone-next.

The first step is to remove packages and code paths that aren't being used by the `@keystone-next` packages, for example:

* The Keystone 5 docs and website
* The create-keystone-app package
* Apps, deprecated field types, etc

If you're watching this repo, it'll mean that you'll see us deleting a lot of the code that has been moved over to the new repo for maintenance and is now unreferenced by the new Keystone Next packages and website.

In this first pass we'll leave in place the packages and code paths that are currently referenced in the new packages.

Once that's complete, we'll start refactoring the remaining Keystone 5 code into the new packages, so we have a clean codebase to build on again.

JedWatson:
With all the PRs above merged, I think we're good to close this issue.

The remaining `keystone-5` packages are all being used in `keystone-next`, and we'll collapse them as we continue to refactor the codebase during feature work.

PR #5035, opened 2021-03-08
Remove arch-ui packages and website

JedWatson:
See #5025

Note these packages now live over in [keystonejs/keystone-5](https://github.com/keystonejs/keystone-5)

We've created a new design system for keystone-next, while there are still a lot of legacy uses of these components in the codebase they'll get installed from npm now rather than being linked in the monorepo and that's fine 👍


package.json
```diff
@@ -200,6 +195,7 @@
       "packages/**/*.js",
       "!packages/**/dist/**",
       "!packages/app-admin-ui/client/**",
+      "!packages/create-keystone-app/templates/**",
       "!packages/arch/**",
```
timleslie, line 199:
```suggestion
```

timleslie (changes requested):

timleslie (approved):
👍

PR #5037, opened 2021-03-08
Remove Keystone 5 website

JedWatson:
See #5025

The source for this website has been moved to [keystonejs/keystone-5](https://github.com/keystonejs/keystone-5)

I've added a new top-level `assets` folder to home misc. assets like the logo at the top of the readme, that was previously in the website static folder (which I think is clearer)


timleslie (approved):
👍

PR #5056, opened 2021-03-08
Reorder tests to kick off critical path early

timleslie:
I'm hoping this will get our longest running tests kicked off first to reduce the length of the critical path in CI.

Update: Moving `docs-next` to the top of the matrix worked. Reordering blocks in the file, and changing the `name` field of the jobs made no difference. Change the actual `job` name (e.g. `api-tests` -> `graphql-api-tests`) made a difference. It would seem that jobs run in alphabetical order based on their `job` name, so we now kick off `docs-next` as the first test in the first job that gets run, which is what we want 👍


gwyneplaine (approved):
lgtm

PR #5060, opened 2021-03-09
Tell prettier to leave our docs code alone

timleslie:
The way we write code in docs does not conform to our regular prettier config, as it needs to present code in a more concise way than we normally would in fully fledged code.

This PR tells prettier to ignore the inline code snippets, while still checking the docs themselves.

### DIFF
--- a/docs-next/pages/apis/graphql.mdx
+++ b/docs-next/pages/apis/graphql.mdx
@@ -196,7 +196,7 @@ type User {
 }
 ```
 
-### _allUsersMeta
+### \_allUsersMeta
 
 ```graphql
 type Query {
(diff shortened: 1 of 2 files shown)


gwyneplaine:
@timleslie, should we just get prettier to ignore mdx again?

timleslie:
> @timleslie, should we just get prettier to ignore mdx again?

Nah, it picks up on some genuinely useful things, like list formatting, etc, that I wouldn't want to lose. I just want it to stay out of the code snippets.

docs-next/pages/apis/graphql.mdx
```diff
@@ -196,7 +196,7 @@ type User {
 }
 ```
 
-### _allUsersMeta
+### \_allUsersMeta
```
gwyneplaine, line 199: I'm assuming this is just to prove that it works, and can be removed pre merge

gwyneplaine (approved):
lgtm

PR #5059, opened 2021-03-09
Run db push against programmatic Prisma API and remove faulty optimisation

mitchellhamilton:

PR #5061, opened 2021-03-09
Convert Prisma adapter to TypeScript

mitchellhamilton:
This is very basic and very not good typing at all but it was frustrating me not having TS in #5059.


mitchellhamilton:
I'm not gonna do this actually because I don't want to spend time fixing these problems, will do what I want in another way

PR #5068, opened 2021-03-09
Stop prisma migrate dev running when building

rohan-deshpande:
### Changed 

* `createKeystone` will no longer run `prisma migrate dev` when `yarn build` is run. This is another thing we need to deploy KS apps via Vercel


timleslie (approved):
Ah, good catch 👍

PR #5069, opened 2021-03-09
Remove unused heroku config files

timleslie:


emmatown (commented):
Can you remove the `heroku-postbuild` package.json script as well?

emmatown (approved):

PR #5072, opened 2021-03-09
Prebuild docs site before tests

timleslie:
Experiment to see if building the docs site up front will speed up the cypress tests.

Update: Success! This change brings the total time for the test down from ~9 minutes to under 5 minutes, which means this test is no longer on the critical path 👍


gwyneplaine (approved):
ah nice!

PR #5073, opened 2021-03-09
Fix viewport meta tag

gwyneplaine:
Viewport meta tags shouldn't be in `_document.tsx` for the following [reasons](https://github.com/vercel/next.js/blob/master/errors/no-document-viewport-meta.md)

Viewport meta tag now moved to the `_app.tsx` file instead


timleslie (approved):
Nice 👍

PR #5087, opened 2021-03-10
Change createKeystone and createSystem to accept a migration mode rather than script

emmatown:
This is extracted out of #5059. There's no behavior change here, I just feel like accepting a migration mode is more clear than the current script arg.


timleslie (approved):
👍

PR #5095, opened 2021-03-11
Add simple JSON field type

MurzNN:
I've added the simple JSON field type, that use jsonb/object database storage format, and represented in Admin UI as plaintext textarea / input (for short single-line jsons).

In future it can be extended via adding codemirror-like editor, or some more complex widget, but at first start - this simple realization is better than nothing.

Filtering queries and GraphQL are copied from text field, and can be extended in future via filtering using json matches.

Please review the code and merge, if all is done right.


timleslie:
Thanks for the PR @MurzNN, we're not quite ready to add support for this field type just yet. We are currently in the process of removing the legacy field types which we don't plan to immediately support in `keystone-next`. This in turn will allow us to simplify the field type implementation which will then make it easier to add new field types like `Json` including robust testing, documentation, and front end support. Unfortunately we're not quite there yet, so I'm going to put this PR on hold for the time being 👍

MurzNN:
Thanks, will glad to see the simplified field type implementation! In current PR I already try to not create the `packages/fields/src/types/Json` folder, but still have dependency on `@keystone-next/fields-legacy`. 

Is there any PR or issue to track the progress of simplified field type implementation?

MurzNN:
I updated PR to last version of Keystone and convert all changes to typescript. @timleslie can you take a look, is all done in right way?

timleslie:
We won't be merging any new fields until the work in #5665 is complete. Once we got that work complete we'll be able to follow up on PRs like this one 👍

PR #5096, opened 2021-03-11
Support access: false in items API and schema type printer

timleslie:
At the moment, using [static `false` access control](https://next.keystonejs.com/apis/access-control#static-list) breaks the items API and the schema type printer, because we completely parts of the graphQL schema that these systems expect to exist.

This PR is a first pass to unblock the porting of the existing access control tests #4931 and #5097 to keystone-next. These changes are not intended to reflect a final locked in design, but will unblock us from moving forward so that we can follow up with design and testing of this use case in the items API and schema type printer.

### DIFF
--- a/.changeset/friendly-timers-bathe.md
+++ b/.changeset/friendly-timers-bathe.md
@@ -1,6 +1,5 @@
 ---
 '@keystone-next/keystone': patch
-'@keystone-next/types': patch
 ---
 
 Updated items API to handle static `false` access control.
--- a/.changeset/friendly-timers-bathe.md
+++ b/.changeset/friendly-timers-bathe.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
+---
+
+Updated items API to handle static `false` access control.
--- a/packages-next/keystone/src/lib/schema-type-printer.tsx
+++ b/packages-next/keystone/src/lib/schema-type-printer.tsx
@@ -147,7 +147,11 @@ export type ${listTypeInfoName} = {
     update: ${gqlNames.updateInputName};
   };
   args: {
-    listQuery: ${listQuery ? printArgs(listQuery.arguments!) : undefined}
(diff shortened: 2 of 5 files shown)


mitchellhamilton (dismissed):

packages-next/types/src/utils.ts
```diff
@@ -4,7 +4,7 @@ export type BaseGeneratedListTypes = {
   backing: BackingTypeForItem;
   inputs: { create: GraphQLInput; update: GraphQLInput; where: GraphQLInput };
   args: {
-    listQuery: {
+    listQuery?: {
```
mitchellhamilton, line 7: Could you make this non-optional and for the printing case where we don't have args, instead of printing undefined, print `import("@keystone-next-types").BaseGeneratedListTypes["args"]["listQuery"]`?

packages-next/keystone/src/lib/schema-type-printer.tsx
```diff
@@ -146,7 +147,11 @@ export type ${listTypeInfoName} = {
     update: ${gqlNames.updateInputName};
   };
   args: {
-    listQuery: ${printArgs(queryNodeFieldsByName[gqlNames.listQueryName].arguments!)}
+    listQuery: ${
+      listQuery
+        ? printArgs(listQuery.arguments!)
+        : 'import("@keystone-next-types").BaseGeneratedListTypes["args"]["listQuery"]'
```
mitchellhamilton, line 153:
```suggestion
        : 'import("@keystone-next/types").BaseGeneratedListTypes["args"]["listQuery"]'
```

.changeset/friendly-timers-bathe.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
```
mitchellhamilton, line 3:
```suggestion
```

mitchellhamilton (approved):

PR #5102, opened 2021-03-12
Skip generating prisma client in keystone-next start

mitchellhamilton:
(note this is tested in the smoke tests for the examples and should marginally decrease the test time there)


timleslie (approved):
Yep, makes sense 👍

PR #5107, opened 2021-03-12
[POC] for theming

gautamsi:
ref #5106 

this only does a poc of how to enable theming with admin ui.

I could make changes to font and colors. This font changes needed custom logo component but we should be able to add a Global style without the need to change the components.
image

not sure if this is in immediate roadmap or not but could be useful and does not change anything if devs do not change the code.


JedWatson:
Hey @gautamsi I just wanted to get back to you with some feedback on this.

First up, you're spot on, this is really close to a POC I did a couple of months back when thinking through "how would we want theming to work in the new Admin UI?"

Second, probably the only reason something very similar to this hasn't been shipped yet is that we have _very little_ confidence in the current theme structure or design system. What's there is very much "just enough to build an Admin UI" and we're conscious that opening it up with a first-class API for theming could cause a really difficult upgrade jump down the track.

So I guess that while I'd say "this is probably how we'd do it" I also mean "unless we meaningfully change how theming works, which we may want to when we actually come to focus on it"

If there's enough demand for theming before we get to that part of our roadmap with v6 we could potentially open up an experimental flag to enable it though, now that we have that pattern, so lets see how it goes.

zuhair-naqvi:
@JedWatson Though this is a POC, I feel enabling it as experimental makes a tonne of sense as theming support (however basic it may be) is one of the most common expectations from a framework. So if there's no major issues with this approach, please merge!

PR #5110, opened 2021-03-14
Convert the relationship CRUD tests to typescript

timleslie:


gwyneplaine (approved):

emmatown (changes requested):
Where you're annotating the result of executeGraphQL with a type that has `errors: undefined` in it, could you replace the  `executeGraphQL` call with one to `graphql.run`? Saying that something is `undefined` isn't right if you then need to assert that it's undefined. (If you don't want to do that right now in this PR, replacing `errors: undefined` with `errors: unknown` would be fine as well)

PR #5115, opened 2021-03-15
Removed skin color for emojis and fixed typos

dominikwilkowski:
Removed some additional skin color information from some emojis and fixed some typos.

image

If this is controversial I'm happy to pull the typos into another PR.


JedWatson (approved):
Thanks @dominikwilkowski, good catch! I think these slipped through because of my recently-used emojis.

PR #5116, opened 2021-03-15
Removed some unused types from roles example

dominikwilkowski:
Looks like neither `ListsAPI`, `GraphqlAPI` nor `FieldAccessArgs` was being used in the role example so I removed `FieldAccessArgs` and left the other two in the comment with them pointing to `KeystoneGraphQLAPI` and `KeystoneListsAPI` respectively.


timleslie (approved):
LGTM 👍

emmatown (changes requested):
Can you delete the stuff rather than commenting it out?

dominikwilkowski:
@mitchellhamilton I'm happy to but there was a comment in there I assumed has value. All I did was make it work and removed the not-used things. Are those types now all generated?

mitchellhamilton:
We need to look at the type generation stuff again in detail sometime but right now, having those bits in comments isn't really adding value.

dominikwilkowski:
Removed @mitchellhamilton

emmatown (approved):

PR #5120, opened 2021-03-15
Convert relationship tests to typescript

timleslie:


tests/api-tests/relationships/shared-names.test.ts
```diff
@@ -20,29 +29,29 @@ const createInitialData = async context => {
...
+  const employees: { id: IdType; name: string }[] = await createItems({
     context,
     listKey: 'Employee',
     items: [
       {
         data: {
           name: 'EmployeeA',
-          company: { connect: { id: companies.find(({ name }) => name === 'CompanyA').id } },
-          role: { connect: { id: roles.find(({ name }) => name === 'RoleA').id } },
+          company: { connect: { id: companies.find(({ name }) => name === 'CompanyA')?.id } },
```
mitchellhamilton, line 39: Could you replace all the optional chaining with `!.`? `!` is TypeScript syntax for "hey type system, pretend that this type isn't null or undefined." It's pretty much a horrible thing to use in regular code but for tests it's quite good https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-

emmatown (changes requested):

emmatown (approved):

PR #5123, opened 2021-03-15
Add CodeSandbox CI

emmatown:
Now that we support SQLite I'm seeing if we can get CodeSandbox CI working so we can have deploy previews of an actual Keystone instance on every PR

---

So the actual sandboxes seem to be having some problems though having CodeSandbox CI also means that on every PR we'll have published versions of the packages available at `https://pkg.csb.dev/keystonejs/keystone/commit/commit-here/package-name-here`(in a package.json it looks like `"@keystone-next/fields": "https://pkg.csb.dev/keystonejs/keystone/commit/ff2cc995/@keystone-next/fields",`) so IMO we should merge this so we can test things in other repos without having to do a release and I've created a ticket for making the actual sandboxes work.


timleslie (approved):
Looks very useful!

PR #5135, opened 2021-03-16
Updated keystone-next dev so that it interactively prompts for creating and applying a migration

mitchellhamilton:
This is mostly based on [the implementation of the actual `prisma migrate dev` command](https://github.com/prisma/prisma/blob/63840b22acb006d3fb8ed2b0e07caa3e3548e223/src/packages/migrate/src/commands/MigrateDev.ts) but not doing things that prisma does(e.g. reading the schema from the file system)

https://user-images.githubusercontent.com/11481355/111262733-727a2a00-8670-11eb-8d24-874b71b280a9.mp4


package.json
```diff
@@ -77,12 +77,14 @@
     "trailingComma": "es5",
     "printWidth": 100,
     "arrowParens": "avoid",
-    "overrides": [{
-      "files": "docs-next/**",
-      "options": {
-        "embeddedLanguageFormatting": "off"
+    "overrides": [
```
mitchellhamilton, line 80: (this is because of `manypkg fix` which does a JSON.stringify)

.changeset/tame-ducks-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/adapter-prisma-legacy': major
+---
+
+Changed default migrationMode from `dev` to `prototype`
```
mitchellhamilton, line 5: This is because of cypress tests (we could technically change the migrationMode in the test projects because but meh, no one should be using the PrismaAdapter directly anyway so the default doesn't really matter)

packages/adapter-prisma/src/migrations.ts
```diff
@@ -41,6 +47,137 @@ export async function runPrototypeMigrations(dbUrl: string, schema: string, sche
...
+        console.log('✅ The migration has been applied');
+      } else {
+        console.log(
+          'Please edit the migration and run keystone-next dev again to apply the migration'
+        );
+        process.exit(1);
+      }
+    } else if (!appliedMigrationNames.length) {
+      console.log('✨ Your database is already up to date, no migrations need to be applied');
+    }
```
timleslie, line 154:
```suggestion
    } else {
      console.log('✨ Your database is now up to date, no further migrations need to be applied');
    }
```

mitchellhamilton: The if is intentional, why do you want to remove it? This message shouldn't be shown when there were migrations applied.

timleslie: I want to give the user a consistent experience. If we're not creating any additional migrations then they should be informed either with 
```
('✨ Your database is already up to date, no migrations need to be applied')
``` 
or
```
('✨ Your database is now up to date, no further migrations need to be applied')
```
depending on whether or not any outstanding migrations were applied during this run.

mitchellhamilton: wdyt of this?
https://github.com/keystonejs/keystone/blob/43b2f19908d0b6f18c8c07644111d24ce7985238/packages/adapter-prisma/src/migrations.ts#L153-L161

packages/adapter-prisma/src/migrations.ts
```diff
@@ -41,6 +47,145 @@ export async function runPrototypeMigrations(dbUrl: string, schema: string, sche
...
+// TODO: don't have process.exit calls here
+export async function devMigrations(dbUrl: string, prismaSchema: string, schemaPath: string) {
+  await ensureDatabaseExists(dbUrl, path.dirname(schemaPath));
+  let migrate = new Migrate(schemaPath);
+  try {
+    // see if we need to reset the database
+    // note that the other action devDiagnostic can return is createMigration
+    // that doesn't necessarily mean that we need to create a migration
+    // it only means that we don't need to reset the database
+    const devDiagnostic = await migrate.devDiagnostic();
```
timleslie, line 59:
```suggestion
    const devDiagnostic = await runMigrateWithDbUrl(dbUrl, () => migrate.devDiagnostic());
```

packages/adapter-prisma/src/migrations.ts
```diff
@@ -41,6 +47,148 @@ export async function runPrototypeMigrations(dbUrl: string, schema: string, sche
...
+      );
+    }
+    // evaluateDataLoss basically means "try to create a migration but don't write it"
+    // so we can tell the user whether it can be executed and if there will be data loss
+    const evaluateDataLossResult = await runMigrateWithDbUrl(dbUrl, () =>
+      migrate.evaluateDataLoss()
+    );
+    // if there are no steps, there was no change to the prisma schema so we don't need to create a migration
+    if (evaluateDataLossResult.migrationSteps.length) {
+      console.log('✨ There was a change to your Keystone schema that requires a migration');
```
timleslie, line 101:
```suggestion
      console.log('✨ There has been a change to your Keystone schema that requires a migration');
```

packages/adapter-prisma/src/migrations.ts
```diff
@@ -41,6 +47,148 @@ export async function runPrototypeMigrations(dbUrl: string, schema: string, sche
...
+      let shouldApplyMigration =
+        migrationCanBeApplied && (await confirmPrompt('Would you like to apply this migration?'));
+      if (shouldApplyMigration) {
+        await runMigrateWithDbUrl(dbUrl, () => migrate.applyMigrations());
+        console.log('✅ The migration has been applied');
+      } else {
+        console.log(
+          'Please edit the migration and run keystone-next dev again to apply the migration'
+        );
+        process.exit(1);
```
timleslie, line 153:
```suggestion
        process.exit(0);
```

packages/adapter-prisma/src/migrations.ts
```diff
@@ -41,6 +47,148 @@ export async function runPrototypeMigrations(dbUrl: string, schema: string, sche
...
+      }
+    } else {
+      if (appliedMigrationNames.length) {
+        console.log(
+          '✨ Your migrations are already up to date, no new migrations need to be created'
+        );
+      } else {
+        console.log(
+          '✨ Your database is already up to date, no migrations need to be created or applied'
```
timleslie, line 162:
```suggestion
          '✨ Your migrations are up to date, no new migrations need to be created'
        );
      } else {
        console.log(
          '✨ Your database is up to date, no migrations need to be created or applied'
```

timleslie (approved):
A few final nits, but other than that I think we can 🚢 this 👍

PR #5147, opened 2021-03-17
Remove custom access control

timleslie:
See #5025 

Now that #5145 (remove keystone.extendGraphQLSchema) is complete, we no longer need to custom access control parsers and checkers.

### DIFF
--- a/.changeset/rich-llamas-smash.md
+++ b/.changeset/rich-llamas-smash.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone-legacy': major
+---
+
+Removed `context.getCustomAccessControlForUser()`.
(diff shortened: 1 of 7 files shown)


mitchellhamilton (approved):

gautamsi:
Do we not need this with `config.extendGraphQLSchema` ? I am sure with extend we also want to restrict access to those query and mutation.

timleslie:
Since the user is writing the resolver directly in `config.extendGraphQLSchema` they can implement any access control they need inline in their resolver, there's not need for system level support.

gautamsi:
That access was easy and inline with the other access control config.

Anyways, I guess the internal list api (if not using sudo context) would still restrict access using access control, right?

PR #5148, opened 2021-03-17
Use programmatic Prisma API for keystone-next deploy 

mitchellhamilton:
This is basically identical to what `prisma migrate deploy` does except using the programmatic APIs so it's more performant.


timleslie (approved):
Works for me 👍

PR #5149, opened 2021-03-17
remove hash-link offset

gwyneplaine:
Heading components ported from the keystone website had a hash link offset, most likely for adding a bit of padding to scroll anchors to. 

Don't feel like we need this for the new site, moreover the padding is spacing the content out in such a way as to hamper readability for long form content. 

before:
Screen Shot 2021-03-17 at 11 38 31 am

after: 
Screen Shot 2021-03-17 at 11 38 19 am


dominikwilkowski (approved):
LGTM

PR #5150, opened 2021-03-17
Use import/order eslint check

timleslie:
We now apply the `import/order` rule from https://github.com/benmosher/eslint-plugin-import so that we can have consistent ordering of our imports. Order below: 

```javascript
// 1. node "builtin" modules
import fs from 'fs';
import path from 'path';
// 2. "external" modules
import _ from 'lodash';
import chalk from 'chalk';
// 3. "internal" modules
// (if you have configured your path or webpack to handle your internal paths differently)
import foo from 'src/foo';
// 4. modules from a "parent" directory
import foo from '../foo';
import qux from '../../foo/qux';
// 5. "sibling" modules from the same or a sibling's directory
import bar from './bar';
import baz from './bar/baz';
// 6. "index" of the current directory
import main from './';
// 7. "object"-imports (only available in TypeScript)
import log = console.log;
```


packages-next/admin-ui/src/next-config.ts
```diff
@@ -1,6 +1,6 @@
 // @ts-ignore
-import withPreconstruct from '@preconstruct/next';
 import Path from 'path';
+import withPreconstruct from '@preconstruct/next';
```
gwyneplaine, line 3: Can't add a suggestion due to deleted line, @timleslie  the import was moved but not the ts-ignore

gwyneplaine (changes requested):
fix typescript error as noted in comments, otherwise lgtm

gwyneplaine (approved):

PR #5152, opened 2021-03-17
Use Prisma's programmatic API in keystone-next reset and make keystone-next reset actually work

mitchellhamilton:
Same as #5148 but for reset. This also makes the reset command actually runnable, it wasn't in the cli file before.


packages/adapter-prisma/src/adapter-prisma.js
```diff
@@ -269,20 +274,18 @@ class PrismaAdapter extends BaseKeystoneAdapter {
...
-        this._runPrismaCmd(`migrate reset --force --preview-feature`);
-        await runPrototypeMigrations(this._url(), this.prismaSchema, path.resolve(this.schemaPath));
-      }
-    } else if (this.provider === 'sqlite') {
-      const tables = await this.prisma.$queryRaw(
-        "SELECT name FROM sqlite_master WHERE type='table';"
-      );
-      for (const { name } of tables) {
-        await this.prisma.$queryRaw(`DELETE FROM "${name}";`);
+        throw new Error('Only "postgresql" and "sqlite" providers are supported');
```
mitchellhamilton, line 285: Given we need to have a case for all the providers we'll want to support to run the tests fast enough, having a general case that we never actually use or test seems silly so I changed to an error.

packages/adapter-prisma/src/migrations.ts
```diff
@@ -103,6 +103,30 @@ ${chalk.greenBright('All migrations have been successfully applied.')}`);
...
 
+// https://github.com/prisma/prisma/blob/527b6bd35e7fe4dbe854653f872a07b25febeb65/src/packages/migrate/src/commands/MigrateReset.ts
+export async function resetDatabaseWithMigrations(dbUrl: string, schemaPath: string) {
+  await ensureDatabaseExists(dbUrl, path.dirname(schemaPath));
+  const migrate = new Migrate(schemaPath);
+
+  try {
+    await migrate.reset();
+
+    const { appliedMigrationNames: migrationIds } = await migrate.applyMigrations();
```
timleslie, line 114:
```suggestion
    await runMigrateWithDbUrl(dbUrl, () => migrate.reset());

    const { appliedMigrationNames: migrationIds } = await runMigrateWithDbUrl(dbUrl, () =>
      migrate.applyMigrations()
    );
```

timleslie (changes requested):

packages/adapter-prisma/src/adapter-prisma.js
timleslie: Oh, nice catch, this was causing our sqlite tests to be way slower than they needed to be on CI!

timleslie (approved):
Looks good 👍

PR #5153, opened 2021-03-18
Factor out a withMigrate wrapper

timleslie:
This PR adds a wrapper function `withMigration` to capture the common pattern used to setup/teardown prisma `Migrate` objects.

### DIFF
--- a/packages/adapter-prisma/src/migrations.ts
+++ b/packages/adapter-prisma/src/migrations.ts
@@ -22,7 +22,7 @@ function runMigrateWithDbUrl<T>(dbUrl: string, cb: () => T): T {
   }
 }
 
-async function withMigration<T>(
+async function withMigrate<T>(
   dbUrl: string,
   schemaPath: string,
   cb: (migrate: Migrate) => Promise<T>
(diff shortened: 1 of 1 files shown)


packages/adapter-prisma/src/migrations.ts
```diff
@@ -22,21 +22,34 @@ function runMigrateWithDbUrl<T>(dbUrl: string, cb: () => T): T {
...
+async function withMigration<T>(
+  dbUrl: string,
+  schemaPath: string,
+  cb: (migrate: Migrate) => Promise<T>
+) {
+  await ensureDatabaseExists(dbUrl, path.dirname(schemaPath));
+  const migrate = new Migrate(schemaPath);
+
+  try {
+    return cb(migrate);
```
mitchellhamilton, line 34: You need to await the return value here otherwise the stop() is happening at the wrong time

timleslie: Nice catch, thanks!

packages/adapter-prisma/src/migrations.ts
```diff
@@ -22,21 +22,34 @@ function runMigrateWithDbUrl<T>(dbUrl: string, cb: () => T): T {
   }
 }
 
+async function withMigration<T>(
```
mitchellhamilton, line 25: Can you rename this to withMigrate?

timleslie: Sure :-)

mitchellhamilton (approved):

PR #5155, opened 2021-03-18
Use Prisma's programmatic API for keystone-next generate

mitchellhamilton:
Same as the previous PRs about using Prisma's programmatic APIs. See the changeset for the options that it accepts.


yarn.lock
```diff
@@ -12843,13 +12843,6 @@ prism-react-renderer@^1.2.0:
   resolved "https://registry.yarnpkg.com/prism-react-renderer/-/prism-react-renderer-1.2.0.tgz#5ad4f90c3e447069426c8a53a0eafde60909cdf4"
   integrity sha512-GHqzxLYImx1iKN1jJURcuRoA/0ygCcNhfGw1IT8nPIMzarmKQ3Nc+JcG0gi8JXQzuh0C5ShE4npMIoqNin40hg==
 
-prisma@2.18.0:
```
timleslie, line 12846: Is this right?

mitchellhamilton: Yes, we don't need the Prima CLI anymore

timleslie: Oh, of course, this is what they call the old `@prisma/cli` now 👍

timleslie (approved):
Looks good 👍

PR #5162, opened 2021-03-19
Experiment with removing getDbSchemaName

mitchellhamilton:
Note this does not remove the ability to specify a schema name, it just means there's only a single way to configure it, adding  `?schema=` to the database url.


mitchellhamilton:
Superseded by #5276

PR #5168, opened 2021-03-19
Fix withItemData treating sessions that don't match an item as valid

emmatown:
To test this, start one of the examples with a new database, create a user then restart the example with a new database and go the admin ui again and you should be prompted to create a new user rather than be logged in


packages-next/keystone/src/session/index.ts
```diff
@@ -104,15 +104,11 @@ export function withItemData<T extends { listKey: string; itemId: string }>(
             where: { id: session.itemId },
             resolveFields: fieldSelections[session.listKey] || 'id',
           });
-          // If there is no matching item found, return the session without a `data value
-          if (!item) {
-            return session;
-          }
```
mitchellhamilton, line 110: Note this case never happened because findOne throws when it doesn't find an item

timleslie (approved):
👍

PR #5173, opened 2021-03-22
SQLite support for the document field

mitchellhamilton:
This uses the String Prisma type rather than Json since Prisma doesn't support Json on SQLite right now.


timleslie (approved):
👍

PR #5174, opened 2021-03-22
Rework precision-scale options check to allow scale=nought

MurzNN:
I detected that passing `scale = 0` to field options in current logic results to set the default value (4) instead of nought (0).
I reworked the logic to fix this problem, also make it little simpler to parse precision value too.


packages-next/fields/src/types/decimal/index.ts
```diff
@@ -23,6 +23,6 @@ export const decimal = <TGeneratedListTypes extends BaseGeneratedListTypes>(
   views: resolveView('decimal/views'),
   getAdminMeta: () => ({
     precision: config.precision || null,
-    scale: config.scale || null,
+    scale: config.scale == null ? null : config.scale,
```
gautamsi, line 26: do you want to check if data type is integer rather than null?

JedWatson: Yeah I'd flip this around, check that `config.scale` is an integer or otherwise set it to null

JedWatson (changes requested):
Good catch, thanks!

Left one change request, agree with @gautamsi, otherwise keen to get this in 👍

packages-next/fields/src/types/decimal/index.ts
MurzNN: Yeah, I agree

MurzNN: Yes, I fixed this.

MurzNN: Hmm, trying to add `isInteger` check produces the typescript error:
```js
    scale: Number.isInteger(config.scale) ? config.scale : null,
```
```
$ tsc
Error: packages-next/fields/src/types/decimal/index.ts(24,3): error TS2322: Type '() => { precision: number | null; scale: number | null | undefined; }' is not assignable to type '(listKey: string, path: string, adminMeta: AdminMetaRootVal) => JSONValue'.
  Type '{ precision: number | null; scale: number | null | undefined; }' is not assignable to type 'JSONValue'.
    Type '{ precision: number | null; scale: number | null | undefined; }' is not assignable to type '{ [key: string]: JSONValue; }'.
      Property 'scale' is incompatible with index signature.
        Type 'number | null | undefined' is not assignable to type 'JSONValue'.
          Type 'undefined' is not assignable to type 'JSONValue'.

Comment (MurzNN) [reply packages-next/fields/src/types/decimal/index.ts]: Maybe simple checking like this will be enough?
```
    scale: config.scale != null ? config.scale : null,
```
I committed this variant now, but glad to see better fix.

MurzNN:
This issue is already fixed in new core, so closing.

PR #5177, opened 2021-03-22
Allow showing id field in list view UI

MurzNN:
Sometimes we need to show the ID field value in Admin UI - something like this:
```js
  User: list({
    ui: {
      listView: {
        initialColumns: ['id', 'name', 'posts'],
      },
    },
```
This works well, but produces TypeScript error:
```
Type '"id"' is not assignable to type ....
```
I fix this via adding the `id` as hard-coded field name to allowed values, hope this is right way.

### DIFF
--- a/.changeset/long-foxes-suffer.md
+++ b/.changeset/long-foxes-suffer.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/types': minor
+---
+
+Allow showing id field in list view UI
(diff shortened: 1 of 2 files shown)


MurzNN:
I updated my PR to match latest core changes, test it on example projects - all works well. Can anybody review it?

timleslie:
Thanks for the PR! This issue, and some similar ones, have now been fixed in #6249.

PR #5194, opened 2021-03-23
Remove legacy Admin UI for core field types

timleslie:
See #5025.

Third in the series to remove legacy field views.


JedWatson (approved):

design-system/packages/core/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.13.10",
+    "@emotion/cache": "^10.0.27",
```
mitchellhamilton, line 19: Why did you add this?

timleslie: I was getting CI failures in the `document` field and the `docs-next` page. It seems like these was an implicit dependency being provided by one of the `@arch-ui` packages, which I traced back this far.

Not sure if this is the right solution, but so far it seems to at least make our tests pass 🤷‍♂️ .

mitchellhamilton (approved):

PR #5212, opened 2021-03-24
Move test-fixtures into the new field packages

timleslie:
This moves all the test-fixtures for the supported fields into `tests` subdirectories in the new packages. This is a preparation for more of the controlled demolition process in #5025.

### DIFF
--- a/packages-next/fields/src/types/select/tests/test-fixtures.ts
+++ b/packages-next/fields/src/types/select/tests/test-fixtures.ts
@@ -1,14 +1,16 @@
-import Text from '../Text';
-import Select from './';
+// @ts-ignore
+import { Text, Select } from '@keystone-next/fields-legacy';
+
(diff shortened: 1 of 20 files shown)


packages-next/fields/src/types/select/tests/test-fixtures.ts
```diff
@@ -1,14 +1,16 @@
-import Text from '../Text';
-import Select from './';
+// @ts-ignore
+import { Text, Select } from '@keystone-next/fields-legacy';
+
+type MatrixValue = 'enum' | 'string' | 'integer';
```
mitchellhamilton, line 4: (here + in all the other places)
```suggestion
type MatrixValue = typeof testMatrix[number];
```

emmatown (changes requested):

emmatown (approved):

PR #5215, opened 2021-03-24
Fix autoincrement -> autoIncrement in docs

timleslie:
Turns out this field type uses camel case. Who knew! 😅


raveling (approved):
review and approved!

JedWatson (approved):

PR #5217, opened 2021-03-24
Fix select field type

timleslie:
With this bug we would generally expect default usage with the `string` data type to work more or less as expected, most of the time, but `enum` an `interger` would be completely broken.

Fixing this will potentially change peoples graphQL and database schema.

### DIFF
--- a/.changeset/moody-apricots-hear.md
+++ b/.changeset/moody-apricots-hear.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/fields': major
+---
+
+Fixed the `select` field type to always use the correct underlying types.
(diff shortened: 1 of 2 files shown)


emmatown (approved):

JedWatson:
@timleslie does the Select field default to `Text` or `Enum`? because if it's `Enum` then this is a breaking change for current `keystone-next` users in a way that's not being called out in the notes

separately I'd actually be in favour of defaulting to `Text` because it's less opinionated, and less of a foot gun, in terms of values... we don't need to rehash a big conversation here but we do need to check the kind of breaking change this is

PR #5227, opened 2021-03-24
Ensure select uses string as a default datatype

timleslie:
This will minimise the impact of the change made in #5217.

### DIFF
--- a/.changeset/moody-apricots-hear.md
+++ b/.changeset/moody-apricots-hear.md
@@ -1,5 +1,5 @@
 ---
-'@keystone-next/fields': major
+'@keystone-next/fields': minor
 ---
 
-Fixed the `select` field type to always use the correct underlying types.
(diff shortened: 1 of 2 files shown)


JedWatson (changes requested):
Should update the changelog from #5217 to be a minor now, right?

timleslie:
> Should update the changelog from #5217 to be a minor now, right?

Perhaps? Everything is getting majored next release anyway.

JedWatson:
I'd also update the change entry

> Fixed the `select` field type to always use the correct underlying types.

that's really not a good description, it's something more like

**Select now uses the correct underlying type, allowing use of `Enum` and `Integer`  types again**

JedWatson (approved):

JedWatson:
Much better, thanks 👍

PR #5228, opened 2021-03-24
Remove legacy S3Adapter

timleslie:
See #5025


JedWatson (approved):

BHARATHCK:
Hi I just migrated my app from v5 to v6[next] and came across this thread where S3FileAdapter support is removed.
Was unable to to find any S3FileAdapter related docs for v6 [CloudinaryImage is available] , any idea if this would be supported or is there any other way to upload files to S3 bucket and i am missing something.

gautamsi:
https://github.com/keystonejs-contrib/k6-contrib/tree/main/packages

There is one s3 field and bunch of another stuffs

PR #5230, opened 2021-03-25
Update example projects to use the Prisma adapter

timleslie:
We're about to remove `knex` and `mongoose` adapter support, so this PR moves the example projects over to the Prisma adapters.


mitchellhamilton (changes requested):
Can you make all the db urls something generic like app.db rather than a description of the project since they're already within the project's directory?

examples-next/roles/keystone.ts
```diff
@@ -42,10 +42,8 @@ export default withAuth(
   config({
     db: process.env.DATABASE_URL
       ? { adapter: 'prisma_postgresql', url: process.env.DATABASE_URL }
-      : {
-          adapter: 'mongoose',
-          url: 'mongodb://localhost/keystone-examples-roles',
-        },
+      : { adapter: 'prisma_sqlite', url: 'file:./keystone-examples-next-roles.db' },
```
mitchellhamilton, line 45:
```suggestion
      : { adapter: 'prisma_sqlite', url: 'file:./keystone.db' },
```

mitchellhamilton (approved):

PR #5241, opened 2021-03-25
Use new interfaces for document field tests

timleslie:
This field type got missed when porting over the field tests to the new APIs.

### DIFF
--- a/packages-next/fields-document/src/tests/test-fixtures.ts
+++ b/packages-next/fields-document/src/tests/test-fixtures.ts
@@ -1,24 +1,24 @@
-// @ts-ignore
-import { Text } from '@keystone-next/fields-legacy';
-import { DocumentFieldType } from '../base-field-type';
+import { text } from '@keystone-next/fields';
+import { document } from '..';
 
 export const name = 'Document';
-export const type = DocumentFieldType;
+export const typeFunction = document;
 export const exampleValue = () => [{ type: 'paragraph', children: [{ text: '' }] }];
 export const exampleValue2 = () => [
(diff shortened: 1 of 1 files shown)


packages-next/fields-document/src/index.ts
```diff
@@ -168,9 +168,9 @@ export const document = <TGeneratedListTypes extends BaseGeneratedListTypes>(
   return {
     type: DocumentFieldType,
     config: {
-      ...config,
       ___validateAndNormalize: (data: unknown) =>
         validateAndNormalizeDocument(data, documentFeatures, componentBlocks, relationships),
+      ...config,
```
mitchellhamilton, line 173: You should revert this and don't pass `___validateAndNormalize` in the tests.

packages-next/fields-document/src/tests/test-fixtures.ts
```diff
@@ -15,10 +14,8 @@ export const subfieldName = 'document';
 
 export const fieldConfig = () => ({ ___validateAndNormalize: (x: any) => x });
```
mitchellhamilton, line 15: I'm assuming fieldConfig gets passed to `document`?

```suggestion
export const fieldConfig = () => ({});
```

timleslie:
Now that we're applying the built in transforms the expected results needed to be updated. @mitchellhamilton please take a quick look and verify that the thing that we're now testing is indeed the correct behaviour.

mitchellhamilton (approved):
Yeah, this is expected since everything but paragraphs is disabled by default so the headings are unwrapped.

PR #5256, opened 2021-03-29
Remove knex and mongoose from public APIs

timleslie:
As per our [roadmap](https://next.keystonejs.com/roadmap#streamlined-back-end-with-prisma) we're moving towards `Prisma` being the core of our backend database support.

This PR removes `knex` and `mongoose` as supported adapters from the public API. It also removes the `mongoId` field type, which was used exclusively by the `mongoose` adapter.

By focusing on a single backend database adapter we will be able to provide a more consistent experience for all users while still supporting a range of backends, including MongoDB, which is [currently being worked on](https://www.notion.so/Prisma-Roadmap-50766227b779464ab98899accb98295f).

It will also significantly reduce the complexity of our backend code, allowing us to add support for new field types and features much more easily than is currently possible.

Notes for reviewer: This PR is the minimal set of changes required to remove the `knex`, `mongoose`, and `mongoId` values from the public API. Once these are gone there will be a series of follow up PRs to clean up the code which is now inaccessible.

### DIFF
--- a/.changeset/rare-avocados-mate.md
+++ b/.changeset/rare-avocados-mate.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/fields': major
+---
+
+Removed support for the `mongoId` field type.
--- a/.changeset/smart-lizards-mate.md
+++ b/.changeset/smart-lizards-mate.md
@@ -1,0 +1,8 @@
+---
+'@keystone-next/fields': major
+'@keystone-next/keystone': major
+'@keystone-next/types': major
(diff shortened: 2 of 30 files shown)


gautamsi:
Can you release this after #5184

I can use that release with mongo until prisma supports it.

timleslie:
@gautamsi Yep, we're going to ship a release today before this goes through 👍

JedWatson (approved):

PR #5276, opened 2021-03-30
Remove usage of getDbSchemaName, getPrismaPath, migrationMode and dropDatabase from tests

mitchellhamilton:
This means that we'll be able to rip out all that stuff soon.


packages/adapter-prisma/src/migrations.ts
```diff
@@ -37,19 +37,27 @@ async function withMigrate<T>(
   }
 }
 
-export async function runPrototypeMigrations(dbUrl: string, schema: string, schemaPath: string) {
+export async function runPrototypeMigrations(
+  dbUrl: string,
+  schema: string,
+  schemaPath: string,
+  shouldDropDatabase = false
```
mitchellhamilton, line 44: Note that I didn't add a changeset for this because there hasn't yet been a release where `runPrototypeMigrations` was publicly exported.

timleslie (approved):
Yep, this all checks out 👍

PR #5277, opened 2021-03-30
refactor pagination component in admin-ui

gwyneplaine:
## Description
This PR aims to refactor the pagination component to reflect the new designs in [figma](https://www.figma.com/file/piYJ7uG5aLBwP0Lt8xr8iN/Keystone-DS-%5BWIP%5D?node-id=0%3A1) 

In doing so the PR covers the following: 
* Stylistic changes to the pagination component 
* Creating a custom hook for calculating and assembling pagination information 
* Changing the listKey prop to a list prop, so that we don't have to call useList from within the pagination component
* Deletion of old pagination code 
* Moving the pagination components into the `/components` folder
* Adding @testing-library/react and adjacent testing packages to @keystone-next/admin-ui 
* Amending the root jest configs to accomodate the new browser based unit tests 
* Initial rudimentary unit tests for the pagination component. 

What this PR does not include: 
* Tests for the select drop down in pagination. As I'd rather opt out of testing this functionality for now in favour of not hacking our way into a brittle test path. 

To review this component locally: 
* boot up any keystone example 
* Create enough items to allow you to play around with / render the pagination (default pagesize is 25)


tests/jest/setup.js
```diff
@@ -1,24 +0,0 @@
-// `mongodb-memory-server` downloads a binary on first run in CI, which can take
-// a while, so we bump up the timeout here.
-jest.setTimeout(60000); // in milliseconds
-
-// AVA-like test wrapper for known failing tests
-// see: https://github.com/avajs/ava#failing-tests
-test.failing = (title, testFn) => {
```
gwyneplaine, line 7: wasn't used by any thing in our codebase as far as search was concerned, happy to add this back in if anyone else has contrary context.

packages-next/admin-ui/src/pages/ListPage/index.tsx
```diff
@@ -676,7 +676,7 @@ function ListTable({
           })}
         </tbody>
       </TableContainer>
-      <Pagination listKey={listKey} total={count} currentPage={currentPage} pageSize={pageSize} />
+      <Pagination list={list} total={count} currentPage={currentPage} pageSize={pageSize} />
```
gwyneplaine, line 679: Seemed unnecessary for the Pagination component itself to be reaching into context for the list in this scenario. So i've passed the entire list to the component instead. 

Again happy to revert this if there's a good reason for doing this

packages-next/admin-ui/src/components/__tests__/Pagination.test.tsx
```diff
@@ -0,0 +1,117 @@
+/**
+ * @jest-environment jsdom
```
gwyneplaine, line 2: We aren't able to augment the testEnvironment per package yet, so unfortunately this will have to happen per file till we have a solution for this

tests/jest/setup.ts
```diff
@@ -0,0 +1,4 @@
+import '@testing-library/jest-dom';
```
gwyneplaine, line 1: importing jest assertion extensions from @testing-library at the root level: 
a) because we can't call jest-setup from individual packages yet
b) this should probably be a global inclusion, I imagine we'll want to use this for most client side packages.

examples-next/basic/.keystone/schema-types.ts
```diff
@@ -26,10 +26,6 @@ export type UserWhereInput = {
   readonly OR?: ReadonlyArray<UserWhereInput | null> | null;
```
gwyneplaine, line 26: I have no idea why these schema-types.ts changes are here. 
looking at the .gitignore for this example, not ignoring schema-types seems to be intentional. 

IF someone could fill me in about this, that would be great

mitchellhamilton: The short version is "this is fine, this file will be removed soon".

packages-next/admin-ui/src/components/Pagination.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      if (total > 1 && list.plural) {
+        setStats(`${total} ${list.plural}`);
+      } else if (total === 1 && list.singular) {
+        setStats(`${total} ${list.singular}`);
+      }
+    }
+  }, [plural, singular, total, pageSize, currentPage]);
+
+  return { stats };
```
mitchellhamilton, line 32: Are you intentionally using state + an effect here for some reason or could these things be calculated in render and avoid the state?

gwyneplaine: @mitchellhamilton mostly just to avoid calculating this multiple times if other nodes up the tree are liable to update.
Happy for this to occur in render and avoid state if that's preferred

mitchellhamilton: yeah, doing this in render would be preferred. (if this was actually expensive which I'm sure it's not then useMemo would be the right thing over state)

gwyneplaine: done, function renamed as well so we don't confuse this with actual hook functionality

mitchellhamilton (approved):

PR #5283, opened 2021-03-30
Remove experimental.prismaSqlite

timleslie:
As of the upcoming release we're confident that this feature is generally useful, so no longer needs the experimental flag.

We'll ship this PR on the morning of the release to get the timing of the docs update right, as nothing else is blocked by it.


emmatown (changes requested):
Needs a changeset

emmatown (approved):

PR #5284, opened 2021-03-30
Convert prisma adapter to typescript

timleslie:
This is a first pass at converting this module to TypeScript. We will need to loop back and clean up a lot of the types once other areas of the system have been correctly typed.

### DIFF
--- a/packages/adapter-prisma/src/adapter-prisma.ts
+++ b/packages/adapter-prisma/src/adapter-prisma.ts
@@ -1,26 +1,98 @@
 import pWaterfall from 'p-waterfall';
 import { defaultObj, mapKeys, identity, flatten } from '@keystone-next/utils-legacy';
 
+// Note: These type definitions are preliminary while we're working towards
+// a full TypeScript conversion.
+type Rel = {
+  left: Field<any>;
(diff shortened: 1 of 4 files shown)


packages/adapter-prisma/src/adapter-prisma.ts
```diff
@@ -1,26 +1,98 @@
...
 class PrismaAdapter {
+  config: {
+    prismaClient?: any;
+    provider?: 'postgresql' | 'sqlite';
+    enableLogging?: boolean;
+    url?: string;
+  };
+  listAdapters: Record<string, PrismaListAdapter>;
+  listAdapterClass?: any;
+  _prismaClient?: any;
```
mitchellhamilton, line 67: Where is this used?

JedWatson (approved):

packages/adapter-prisma/src/adapter-prisma.ts
timleslie: Good catch, legacy from when branch was first made. Will remove 👍

emmatown (approved):

PR #5286, opened 2021-03-30
Missing changeset for pagination PR 

gwyneplaine:
on the tin


JedWatson (changes requested):
Lets spell out QoL as "quality of life" (or better, explain what they are)

Got to assume not all the people reading our change sets are EFL and that abbreviation isn't so common

JedWatson (approved):

PR #5287, opened 2021-03-30
Remove getDbSchemaName, getPrismaPath and migration things on adapter

mitchellhamilton:
Note this depends on #5276 and the tests will be broken until it is merged.


timleslie (approved):
Fantastic, such a nice clean up!

PR #5290, opened 2021-03-30
[WIP] Fixing DB data type for DateTimeUtc field type

molomby:
Adding native database type attribute for the DateTimeUtc field to produce the `timestamp(3) with time zone` type in pg.

https://thinkmill.atlassian.net/browse/KEY-345


JedWatson:
I'm going to close this until the new Fields work lands and we review Date fields, so we don't churn breaking changes

PR #5291, opened 2021-03-30
Update Slack channel links in roadmap.mdx

borisno2:
update roadmap.mdx so KeystoneJS Slack has the correct links


JedWatson (approved):
I made a minor edit, because I believe (@bladey to confirm) we're trying to point visitors at Slack over GH discussions.

Reason being all the channels are pretty fragmented at the moment so we're tying to consolidate on a few before expanding it again.

PR #5302, opened 2021-03-31
Move usages of @prisma/migrate and @prisma/sdk from adapter to @keystone-next/keystone

mitchellhamilton:
This means that we don't have to load this code in production(technically we still will be with this PR but this PR makes it possible to not do that) and stop exposing some of it publicly so we can change things however we want without having to do majors.


packages/test-utils/src/index.ts
```diff
@@ -8,7 +8,7 @@ import supertest from 'supertest-light';
 // @ts-ignore
 import { Keystone } from '@keystone-next/keystone-legacy';
 import { initConfig, createSystem, createExpressServer } from '@keystone-next/keystone';
-import { runPrototypeMigrations } from '@keystone-next/adapter-prisma-legacy';
+import { pushPrismaSchemaToDatabase } from '@keystone-next/keystone/migrations';
```
mitchellhamilton, line 11: Note no changeset for this because there hasn't yet been a release where the test utils used this util from the adapter package.

timleslie (approved):
👍

PR #5319, opened 2021-03-31
Cleanups in the prisma adapter

timleslie:
This PR cleans up some methods and properties we no longer need, along with some unused code paths. Makes it much easier to follow what happens during the `connect()` operation.

### DIFF
--- a/packages/adapter-prisma/src/adapter-prisma.js
+++ b/packages/adapter-prisma/src/adapter-prisma.js
@@ -36,26 +36,13 @@ class PrismaAdapter {
       log: this.enableLogging && ['query'],
       datasources: { [this.provider]: { url: this.url } },
     });
-    await this.prisma.$connect();
 
-    // Set up all list adapters
-    try {
-      Object.values(this.listAdapters).forEach(listAdapter => {
-        listAdapter._postConnect({ rels, prisma: this.prisma });
-      });
-    } catch (error) {
-      // close the database connection if it was opened
-      try {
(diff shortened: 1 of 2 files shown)


packages/adapter-prisma/src/adapter-prisma.js
```diff
@@ -26,22 +22,27 @@ class PrismaAdapter {
...
-
-      if (errors.length) {
-        if (errors.length === 1) throw errors[0];
-        const error = new Error('Multiple errors in PrismaAdapter.postConnect():');
-        error.errors = errors;
-        throw error;
-      }
+      Object.values(this.listAdapters).forEach(listAdapter => {
+        listAdapter._postConnect({ rels, prisma: this.prisma });
+      });
```
mitchellhamilton, line 45: Could you move this to before the connect(and you might want to rename _postConnect too then)? Since then you can remove the catching and disconnecting stuff below and it's important for the Next-lite stuff that the list APIs can be called synchronously(even though they'll ofc return asynchronously) and moving it to before the connect enables that (in that PR currently, I call postConnect on the adapter to make it work)

mitchellhamilton (changes requested):

packages/adapter-prisma/src/adapter-prisma.js
timleslie: Cool, will do 👍

mitchellhamilton (approved):

PR #5321, opened 2021-03-31
--reset-db flag on keystone-next dev

mitchellhamilton:
This lets people easily start Keystone with a fresh database.


timleslie (approved):
👍

PR #5338, opened 2021-03-31
Allow pass custom config object to dev and start run scripts

MurzNN:
Sometimes we need to initialize Keystone with custom config, especially for debug reasons. At now seems isn't possible to pass custom `KeystoneConfig` to run scripts such as `dev` and `start`, the config is always read from `keystone.ts` file at root.

This PR adds the ability to pass custom KeystoneConfig object to those run scripts, here is example how to use it:
```js
import { default as config } from "keystone";

config.db.onConnect = async function (context) {
  console.log('My custom code is launched!');
  let users = await context.lists.User.findMany({resolveFields: false});
  console.log(users);
}
dev('', false, config);
```

### DIFF
--- a/.changeset/ten-timers-matter.md
+++ b/.changeset/ten-timers-matter.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Allow pass custom config object to dev run
--- a/packages-next/keystone/src/scripts/run/dev.ts
+++ b/packages-next/keystone/src/scripts/run/dev.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 import express from 'express';
 import { generateAdminUI } from '@keystone-next/admin-ui/system';
+import { KeystoneConfig } from '@keystone-next/types';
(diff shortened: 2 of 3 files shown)


timleslie:
Hmm, I'm not sure I see how this is different to simply setting `db.onConnect` in `keystone.ts` itself? The `dev()` functions aren't really intended to be run directly outside of the action `keystone-next` CLI itself.

MurzNN:
I need to have several "entry points" to keystone-core instance:
1. default full featured startup mode
2. "light" mode for local development, with disabled some cron jobs (that are initialized via `db.onConnect` function)
3. cron-only mode (for start heavy cron jobs only as separate instance)
4. debug mode (for example to simply do the `lists.MyList.findMany()` query, dump output and die).

I can put all this logic in `db.onConnect` or generate different "default" object in `keystone.ts`, but more comfortable is to have separate scripts, that launches keystone in different modes.

Also maybe you can provide some better solution for my task?

timleslie:
I've re-reviewed the proposed API changes here and decided that for now this isn't an API we want to expand, as adequate workarounds currently exist to achieve the desired functionality.

PR #5341, opened 2021-03-31
Next-lite example project 

mitchellhamilton:
This is #5187 but opened by me rather than @JedWatson so that he can review it


packages-next/keystone/src/artifacts.ts
```diff
@@ -105,20 +105,93 @@ export async function generateCommittedArtifacts(
...
+
+// note the export default config is just a lazy way of going "this is also any"
+const nextGraphQLAPIDTS = `export const config: any;
+export default config;
+`;
+
 export async function generateNodeModulesArtifacts(
   graphQLSchema: GraphQLSchema,
   keystone: BaseKeystone,
+  config: KeystoneConfig,
```
mitchellhamilton, line 168: Note no changeset for this because there hasn't been a release since this was added.

packages-next/types/src/config/index.ts
```diff
@@ -36,6 +36,10 @@ export type KeystoneConfig = {
     enableNextJsGraphqlApiEndpoint?: boolean;
     /** Enable Prisma+SQLite support */
     prismaSqlite?: boolean;
+    /** Creates a file at `node_modules/.keystone/api` with a `lists` export */
+    generateNodeAPI?: boolean;
```
timleslie, line 40: Could we make sure that the `config.mdx` has at least one sentence about each of these new options 👍

JedWatson (approved):

PR #5343, opened 2021-04-01
Remove legacy stdout logging

timleslie:
The existing logging is only partially implemented and can be a source of confusion to users. We are going to remove what's currently there and re-evaluate the design of this feature.

### DIFF
--- a/packages/keystone/lib/Keystone/format-error.js
+++ b/packages/keystone/lib/Keystone/format-error.js
@@ -63,7 +63,7 @@ const flattenNestedErrors = error =>
     []
   );
 
-export const formatError = error => {
+const formatError = error => {
   const { originalError } = error;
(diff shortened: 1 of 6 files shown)


packages/keystone/lib/Keystone/format-error.js
```diff
@@ -67,64 +63,23 @@ const flattenNestedErrors = error =>
     []
   );
 
-const formatError = error => {
+export const formatError = error => {
```
mitchellhamilton, line 66: You can't use ESM export unless you make the whole package use Preconstruct. (this is probably why the example smoke tests are hanging)

timleslie: Ah cool, just noticed they were being weird. Will investigate 👍

emmatown (approved):

PR #5344, opened 2021-04-01
Prompt before reset without migrations

mitchellhamilton:
A note with the unexecutable steps case: you can't really test that right now because you can't actually make a field non-nullable in Keystone right now 🙃 . (unexecutable steps are basically all making a field with nulls non-nullable) Removing a field that has non-null values(this bit is important) is the easiest way to test the warnings case.


timleslie (approved):
👍

PR #5368, opened 2021-04-06
Add db.provider to replace db.adapter

timleslie:
Now that we only support Prisma the question isn't "which keystone adapter do you want?" as much as "which Prisma provider do you want to use?". This deprecates `db.adapter`, which will be removed in the next release.

### DIFF
--- a/packages-next/types/src/config/index.ts
+++ b/packages-next/types/src/config/index.ts
@@ -63,16 +63,28 @@ export type DatabaseCommon = {
 
 export type DatabaseConfig = DatabaseCommon &
   (
-    | {
-        adapter: 'prisma_postgresql';
+    | ((
+        | {
+            /** @deprecated: The `adapter` option is deprecated. Please use `{ provider: 'postgresql' }` */
+            adapter: 'prisma_postgresql';
+            provider?: undefined;
+          }
+        | { adapter?: undefined; provider: 'postgresql' }
+      ) & {
         useMigrations?: boolean;
         enableLogging?: boolean;
-      }
-    | {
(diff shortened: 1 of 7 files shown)


packages-next/types/src/config/index.ts
```diff
@@ -63,16 +63,28 @@ export type DatabaseCommon = {
...
   (
-    | {
-        adapter: 'prisma_postgresql';
+    | ((
+        | {
+            /** @deprecated: The `adapter` option is deprecated. Please use `{ provider: 'postgresql' }` */
+            adapter: 'prisma_postgresql';
+            provider?: undefined;
+          }
+        | { adapter?: undefined; provider: 'postgresql' }
```
timleslie, line 72: @mitchellhamilton Is there a better way of saying "one or the other, but not both"?

emmatown (changes requested):
Just want to explicitly block this on my review while I check things because I think I will have comments

packages-next/types/src/config/index.ts
mitchellhamilton: Not really

emmatown (changes requested):
Looks fine except for the `@deprecated` bit: can you remove the colon(it doesn't make sense and looks weird when shown in editors) and add the `@deprecated` comment to both the `adapter?: undefined` cases so that the adapter option shows as deprecated in autocomplete (currently in this PR, it doesn't look like this)? 
Screen Shot 2021-04-06 at 11 13 11 am

emmatown (approved):

PR #5372, opened 2021-04-06
Heading tweaks for document-editor

gwyneplaine:
Actually we're making these changes in keystone-ui/core. 
As to my understanding type styles in the keystone document editor field should be a mirror of the styles instantiated in the <Core> component. 

Happy to change this if necessary.


mitchellhamilton (changes requested):
Could you put these on the actual headings so this doesn't interfere with headings that are inside component blocks?

packages-next/fields-document/src/DocumentEditor/heading.tsx
```diff
@@ -5,14 +5,30 @@ import { RenderElementProps } from 'slate-react';
 
 import { Editor, Transforms, Range, Point, Path, Node, Element, Text } from 'slate';
 
+const headingStylesMap: Record<string, any> = {
```
mitchellhamilton, line 8: Is `Record<string, any>` necessary? that just makes the type less specific?

mitchellhamilton (approved):

PR #5373, opened 2021-04-06
Update slate to 0.62

mitchellhamilton:
There's no particularly interesting things here, it's just updating slate and dealing with the mainly TS errors and some other things noted below.

### DIFF
--- a/packages-next/fields-document/src/DocumentEditor/component-blocks/index.tsx
+++ b/packages-next/fields-document/src/DocumentEditor/component-blocks/index.tsx
@@ -2,7 +2,7 @@
 
 import { Fragment, ReactElement, createContext, useContext, useState, useMemo } from 'react';
 import { ReactEditor, RenderElementProps, useFocused, useSelected } from 'slate-react';
-import { Editor, Element, Transforms } from 'slate';
+import { Editor, Element, Transforms, Node } from 'slate';
 
 import { jsx, useTheme } from '@keystone-ui/core';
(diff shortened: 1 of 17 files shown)


packages-next/fields-document/src/DocumentEditor/component-blocks/index.tsx
```diff
@@ -326,21 +327,31 @@ function ComponentBlockRender({
...
-  children: _children,
+  children,
 }: {
   element: Element & { type: 'component-block' };
   onElementChange: (element: Partial<Element>) => void;
   componentBlock: ComponentBlock;
   children: any;
 }) {
   const childrenByPath: Record<string, ReactElement> = {};
-  const children = _children.type(_children.props).props.children;
```
mitchellhamilton, line 337: The children that Slate passes to element renderers is an array now so we don't have to do this hack now.

timleslie:
We disabled renovate for these packages in #5482. We should un-disable renovate as part of this PR.

JedWatson:
We're going to look at this after more of the current fields work lands, because it needs more review and thinking to work out how to best deal with the updated types in Slate 0.62

NoTuxNoBux:
It's worth mentioning that Slate 0.63 updated to immer 8.0.1, which fixes [this security advisory](https://www.npmjs.com/advisories/1603) reported by npm audit and seems to have never been fixed for immer 7.

PR #5393, opened 2021-04-07
Use passed in directory to read config file instead of process.cwd()

emmatown:
This a precursor to tests for CLI things.


packages-next/keystone/package.json
```diff
@@ -4,6 +4,17 @@
...
+    "session",
+    "bin",
+    "artifacts",
+    "___internal-do-not-use-will-break-in-patch",
+    "migrations",
+    "next",
+    "schema",
+    "scripts",
+    "session"
+  ],
```
mitchellhamilton, line 17: This is a thing for future PRs that I forgot I did now and don't really feel like reverting so that we don't publish `src` to npm because it will contain tests soon and they could get large.

timleslie (approved):
👍

PR #5396, opened 2021-04-07
Image field type

rohan-deshpande:
### Added 

* Image field type and API for handling image I/O
* Image config to Keystone config
* Required packages for image I/O processes

### Notes 

* This is completely untested and may or may not work. I want to get into writing tests for it but would like someone's eyes across it first who is maybe a bit more familiar with the `sharp` API. 
* Some of the blurhash code is lifted from https://github.com/woltapp/blurhash/issues/43#issuecomment-597674435 there's maybe some stuff they are doing here that we don't need to do... I'm not too sure
* If the REST endpoint is also going to use `getDataFromStream` then it probably makes sense for the file to get written to the storage layer within this function. If that's the case, then we should change the name of the function to something like `writeStreamToFile`
* I'm really, really not a fan of the `upload` key in the config. While I am also not a fan of `mode` as it is just super generic, I feel it's easier to document than `upload` so I would be in favour of changing this key to `mode`. We at least wouldn't have to do something like this since we are calling it `mode` in every other instance

```js
const { upload: mode } = config;
```


mitchellhamilton:
Given we only need to get the type and size of the iamge, could we replace sharp with https://www.npmjs.com/package/file-type + https://www.npmjs.com/package/image-size? There is one problem with that which is svg support which kinda makes things weird but we might just want to avoid supporting svg for now given it's not a binary format so it's kinda totally different to the others and doesn't(at least in my understanding) have a defined width and height in the way the other formats we have do. I'd like to avoid a dependency on a native module like sharp in keystone if we can avoid it.

packages-next/keystone/src/lib/createImagesContext.ts
```diff
@@ -101,13 +103,13 @@ export function createImagesContext(config: KeystoneConfig['images']): ImagesCon
       if (isLocal(mode)) {
         const buffer = await fs.readFile(path.join(storagePath, `${id}.${extension}`));
         const metadata = await getImageMetadataFromBuffer(buffer);
-        const blurhash = await getBlurhashFromBuffer(buffer, metadata.width, metadata.height);
+        const blurHash = await getBlurhashFromBuffer(buffer, metadata.width, metadata.height);
```
rohan-deshpande, line 105: If we are gonna camel case then the function should be `getBlurHashFromBuffer` I guess

packages-next/keystone/src/lib/createImagesContext.ts
```diff
@@ -5,6 +5,7 @@ import { v4 as uuid } from 'uuid';
 import fs from 'fs-extra';
 import { fromBuffer } from 'file-type';
 import imageSize from 'image-size';
```
rohan-deshpande, line 7: Do we still need this?

packages-next/keystone/src/lib/createImagesContext.ts
```diff
@@ -48,14 +49,17 @@ const getImageMetadataFromBuffer = async (buffer: Buffer) => {
...
 };
 
-const getBlurhashFromBuffer = async (
-  buffer: Buffer,
-  width: number,
-  height: number
-): Promise<string> => {
-  return '';
+const getBlurhashFromBuffer = async (buffer: Buffer): Promise<string> => {
+  const SIZE = 50;
```
rohan-deshpande, line 52: maybe move this out of the function scope?

packages-next/keystone/src/lib/createImagesContext.ts
gwyneplaine: @mitchellhamilton  should we revert things to using just sharp again, given that we still need to leverage sharp to get blurHash working?

packages-next/keystone/src/lib/createImagesContext.ts
gwyneplaine: @rohan-deshpande  ideally we want to not have the size as a magic number at all. We added this because we weren't able to get automatic width and height calculations that were compliant with blurHash at the time. (it was throwing errors). 

Currently the blurHash won't respect the aspect ratio of the original image, is this something you could pick up Rohan? 
Otherwise I'm happy to take a look at it after the remaining UI work is done

packages-next/keystone/src/lib/createImagesContext.ts
rohan-deshpande: I think we need sharp for future ideas as well

packages-next/keystone/src/lib/createImagesContext.ts
rohan-deshpande: It shouldn't be but you may have to end up just picking something. I'll check it out.

packages-next/fields/src/types/image/views/Field.tsx
```diff
@@ -0,0 +1,297 @@
...
+  // Generate a random input key when the value changes, to ensure the file input is unmounted and
+  // remounted (this is the only way to reset its value and ensure onChange will fire again if
+  // the user selects the same file again)
+  const inputKey = useMemo(() => Math.random(), [value]);
+
+  const onSuccess = () => {
+    addToast({ title: 'Copied image ref to clipboard', tone: 'positive' });
+  };
+  const onFailure = () => {
+    addToast({ title: 'Faild to copy image ref to clipboard', tone: 'negative' });
```
mitchellhamilton, line 77:
```suggestion
    addToast({ title: 'Failed to copy image ref to clipboard', tone: 'negative' });
```

packages-next/keystone/package.json
```diff
@@ -35,19 +35,24 @@
     "@types/pluralize": "^0.0.29",
     "@types/prettier": "^2.2.3",
     "@types/prompts": "^2.0.10",
+    "@types/sharp": "^0.28.0",
```
mitchellhamilton, line 38: sharp has been removed, yeah?

packages-next/keystone/package.json
```diff
@@ -35,19 +35,24 @@
...
     "@types/source-map-support": "^0.5.3",
     "@types/uid-safe": "^2.1.2",
+    "@types/uuid": "^8.3.0",
     "apollo-server-express": "^2.22.2",
     "apollo-server-micro": "^2.22.2",
     "apollo-server-types": "^0.7.0",
+    "blurhash": "^1.1.3",
```
mitchellhamilton, line 45: We don't use this anymore, yeah?

emmatown (changes requested):
Can you(@rohan-deshpande and/or @gwyneplaine) resolve the conflicts, add changesets, convert the Implementation to TypeScript like the other fields now are and originally the plan was to store the data in seperate columns rather than JSON, do we want to do that now?

packages-next/fields/src/types/image/Implementation.js
```diff
@@ -0,0 +1,130 @@
...
+        webp
+        gif
+      }
+      type ImageFieldOutput {
+        mode: ImageMode!
+        id: ID!
+        filesize: Int!
+        width: Int!
+        height: Int!
+        blurHash: String!
```
mitchellhamilton, line 36:
```suggestion
```

packages-next/keystone/package.json
rohan-deshpande: thought I removed it from here... hmm

rohan-deshpande: oh types.. damn ok my bad

packages-next/fields/src/types/image/Implementation.js
rohan-deshpande: Done in 1d742ae33

packages-next/keystone/package.json
rohan-deshpande: Fixed in 1d742ae33

packages-next/keystone/package.json
rohan-deshpande: Fixed in 1d742ae33

packages-next/fields/src/types/image/views/Field.tsx
rohan-deshpande: Fixed in 04154283e

packages-next/keystone/src/lib/createImagesContext.ts
rohan-deshpande: No longer relevant as we are descoping blurhash

packages-next/types/src/config/index.ts
```diff
@@ -39,6 +39,17 @@ export type KeystoneConfig = {
...
     generateNextGraphqlAPI?: boolean;
   };
+  images?: ImagesConfig;
+};
+
+export type ImagesConfig = {
+  upload: ImageMode;
+  local?: {
+    /** The path local images are uploaded to */
+    storagePath: string;
```
rohan-deshpande, line 49: make these nullable so that you can have one or the other to leave the default

packages-next/keystone/src/lib/createImagesContext.ts
```diff
@@ -0,0 +1,164 @@
...
+  ImageMode,
+} from '@keystone-next/types';
+import { v4 as uuid } from 'uuid';
+import fs from 'fs-extra';
+import { fromBuffer } from 'file-type';
+import imageSize from 'image-size';
+
+const SUPPORTED_IMAGE_EXTENSIONS = ['jpeg', 'png', 'webp', 'gif'];
+const MODE_LOCAL = 'local';
+const DEFAULT_BASE_URL = 'http://localhost:3000';
```
rohan-deshpande, line 15: change the default here to `/images`

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,130 @@
+import { PrismaFieldAdapter } from '@keystone-next/adapter-prisma-legacy';
+import { Implementation } from '../../Implementation';
+import { ImageData, KeystoneContext } from '@keystone-next/types';
+// eslint-disable-next-line import/no-unresolved
```
mitchellhamilton, line 4:
```suggestion
```

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  getPrismaSchema() {
+    // ${this.config.isUnique ? '@unique' : ''}
+    const { isRequired } = this.config;
+    return [
+      `${this.path}_filesize    Int${isRequired ? '' : '?'} `,
+      `${this.path}_extension   String${isRequired ? '' : '?'}`,
+      `${this.path}_width   Int${isRequired ? '' : '?'}`,
+      `${this.path}_height   Int${isRequired ? '' : '?'}`,
+      `${this.path}_mode   String${isRequired ? '' : '?'}`,
```
mitchellhamilton, line 150: Could you make this an enum when using postgres?

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,247 @@
...
+      return null;
+    }
+    if (data === undefined) {
+      return undefined;
+    }
+    const imageData = await handleImageData(data, context);
+    if (this.adapter.listAdapter.parentAdapter.provider === 'sqlite') {
+      // we store image data as a string on sqlite because Prisma doesn't support Json on sqlite
+      // https://github.com/prisma/prisma/issues/3786
+      return JSON.stringify(imageData);
```
mitchellhamilton, line 106: This should be removed now, yes?

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,247 @@
...
+    };
+  }
+  // Called on `User.avatar` for example
+  gqlOutputFieldResolvers() {
+    return {
+      [this.path]: (item: Record<P, any>) => {
+        return item[this.path];
+      },
+    };
+  }
```
timleslie, line 68:
```suggestion
  gqlOutputFieldResolvers() {
    return { [`${this.path}`]: (item: Record<P, any>) => item[this.path] };
  }
```

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+    if (data === undefined) {
+      return undefined;
+    }
+    const imageData = await handleImageData(data, context);
+    if (this.adapter.listAdapter.parentAdapter.provider === 'sqlite') {
+      // we store image data as a string on sqlite because Prisma doesn't support Json on sqlite
+      // https://github.com/prisma/prisma/issues/3786
+      return JSON.stringify(imageData);
+    }
```
timleslie, line 107: Do we still need this now that we're no longer storing in Json?

timleslie (changes requested):
.

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,247 @@
...
+    resolvedData: Record<P, any>;
+    addFieldValidationError: (msg: string) => void;
+  }) {
+    if (!resolvedData[fieldPath as P]) return;
+    const img = resolvedData[fieldPath as P];
+    Object.keys(img).forEach(key => {
+      if (!img[key]) {
+        addFieldValidationError(`${fieldPath}.${key} is a non-nullable property`);
+      }
+    });
```
mitchellhamilton, line 85: This seems wrong, one of the keys has to be nullable in the input?

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,238 @@
...
+  getPrismaSchema() {
+    // ${this.config.isUnique ? '@unique' : ''}
+    const { isRequired } = this.config;
+    return [
+      `${this.path}_filesize    Int${isRequired ? '' : '?'} `,
+      `${this.path}_extension   String${isRequired ? '' : '?'}`,
+      `${this.path}_width   Int${isRequired ? '' : '?'}`,
+      `${this.path}_height   Int${isRequired ? '' : '?'}`,
+      `${this.path}_mode   String${isRequired ? '' : '?'}`,
+      `${this.path}_id   String${isRequired ? '' : '?'}`,
```
mitchellhamilton, line 142: All of these should always be optional, isRequired does not mean non-nullable in the database

packages-next/fields/src/types/image/Implementation.ts
gwyneplaine: Done 🎉 with a caveat that we fallback to a string type for sqlite

packages-next/fields/src/types/image/Implementation.ts
gwyneplaine: Clarified with @mitchellhamilton, removing this entire check, since we don't want to be running this on the input value

packages-next/fields/src/types/image/Implementation.ts
```diff
@@ -0,0 +1,235 @@
+import { PrismaFieldAdapter, PrismaListAdapter } from '@keystone-next/adapter-prisma-legacy';
+import { ImageData, KeystoneContext, BaseKeystoneList } from '@keystone-next/types';
+import { Implementation } from '../../Implementation';
+import { handleImageData } from './handle-image-input';
+
+const SUPPORTED_IMAGE_EXTENSIONS = ['jpeg', 'png', 'webp', 'gif'];
```
gwyneplaine, line 6: We have this list in here and in the `create-image-context` file, in the future when fields is part of keystone, we should probably have a single reference to this stored in a constants file of some variety.

packages-next/fields/src/Implementation.ts
```diff
@@ -149,8 +149,29 @@ class Field<P extends string> {
...
+  async validateInput(data: {
+    resolvedData: Record<P, any>;
+    existingItem: Record<string, any>;
+    context: KeystoneContext;
+    originalInput: any;
+    listKey: string;
+    fieldPath: string;
+    operation: 'create' | 'update';
+    addFieldValidationError: (msg: string) => void;
+    addValidationError: (msg: string) => void;
```
timleslie, line 173:
```suggestion
```

packages-next/fields/src/Implementation.ts
```diff
@@ -149,8 +149,29 @@ class Field<P extends string> {
...
+   * (no relationships or defaults resolved)
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async validateInput(data: {
+    resolvedData: Record<P, any>;
+    existingItem: Record<string, any>;
+    context: KeystoneContext;
+    originalInput: any;
+    listKey: string;
+    fieldPath: string;
```
timleslie, line 170:
```suggestion
    fieldPath: P;
```

packages-next/fields/src/Implementation.ts
```diff
@@ -149,8 +149,29 @@ class Field<P extends string> {
...
+   * be the existing data in the database
+   * @param {Object} data.context The graphQL context object of the current
+   * request
+   * @param {Object} data.originalInput The raw incoming item from the mutation
+   * (no relationships or defaults resolved)
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async validateInput(data: {
+    resolvedData: Record<P, any>;
+    existingItem: Record<string, any>;
```
timleslie, line 166:
```suggestion
    existingItem?: Record<string, any>;
```

timleslie (dismissed):
.

.changeset/rude-parrots-train.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/app-basic': patch
+---
+
+Added example of image field support to basic keystone-next example.
```
mitchellhamilton, line 5: It doesn't need to be removed but just an fyi, changesets are not necessary for things that aren't published like examples.

.changeset/shaggy-dots-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/fields': patch
```
mitchellhamilton, line 2:
```suggestion
'@keystone-next/fields': minor
```

packages-next/fields/src/types/image/views/Field.tsx
```diff
@@ -0,0 +1,392 @@
...
+        onChange={onUploadChange}
+        type="file"
+        disabled={onChange === undefined}
+      />
+    </FieldContainer>
+  );
+}
+
+export function validateRef({ ref }: { ref: string }) {
+  if (!isValidImageRef(ref)) {
```
gwyneplaine, line 339: @mitchellhamilton  ended up exporting isValidImageRef from utils-legacy because calling parseIamgeRef from here has a couple of issues: 
* `parseImageRef` throws an error both on incorrect regex and on incorrect extensions. I assume we want this behaviour on the backend but we probably don't want this in our Field.
* `parseImageRef` expects to return an object if successful, in this case all we want is a simple boolean check, exposing the check from utils-legacy seems more semantically correct and less error prone. 

Not married to this idea, I think the first point is the only real problem

packages/utils/src/index.ts
```diff
@@ -256,3 +257,43 @@ export const humanize = (str: string) => {
...
+  SUPPORTED_IMAGE_EXTENSIONS.includes(extension);
+
+export const parseImageRef = (
+  ref: string
+): { mode: ImageMode; id: string; extension: ImageExtension } => {
+  const throwInvalidRefError = () => {
+    throw new Error('Invalid image reference');
+  };
+  const [mode, idAndExt] = ref.split(':');
+  const [id, ext] = idAndExt.split('.');
```
mitchellhamilton, line 284: This isn't right, this won't support file names with `.` in it. You should match on the regex here

packages/utils/src/index.ts
```diff
@@ -256,3 +257,43 @@ export const humanize = (str: string) => {
     .map(upcase)
     .join(' ');
 };
+
+export const getImageRef = (mode: ImageMode, id: string, extension: ImageExtension) =>
+  `${mode}:${id}.${extension}`;
+
+export const isValidImageRef = (ref: string): boolean => {
+  if (/^(local):(.+)\.(gif|jpg|png|webp)$/.test(ref)) {
```
mitchellhamilton, line 265: The exact regex should be `/^(local):([^:\n]+)\.(gif|jpg|png|webp)$/`

packages/utils/src/index.ts
```diff
@@ -256,3 +257,24 @@ export const humanize = (str: string) => {
...
+export const getImageRef = (mode: ImageMode, id: string, extension: ImageExtension) =>
+  `${mode}:${id}.${extension}`;
+
+export const SUPPORTED_IMAGE_EXTENSIONS = ['jpg', 'png', 'webp', 'gif'];
+
+export const parseImageRef = (
+  ref: string
+): { mode: ImageMode; id: string; extension: ImageExtension } | undefined => {
+  if (ref.match(REFREGEX)) {
+    const [__, mode, id, ext] = ref.match(REFREGEX) as RegExpMatchArray;
```
mitchellhamilton, line 272:
```suggestion
  const match = ref.match(REFREGEX);
  if (match) {
    const [__, mode, id, ext] = match;
```

emmatown (approved):

PR #5397, opened 2021-04-07
Bumping node versions to 12.x

bladey:
https://thinkmill.atlassian.net/jira/software/projects/KEY/boards/1?selectedIssue=KEY-266

Bumping all package.json node engines to use 12.x instead of 10.x due to 10.x being end of life end of April.

When using ">=12.0.0" and installing "12.0.0" specifically, this error occurs:

`error tailwindcss@2.1.0: The engine "node" is incompatible with this module. Expected version ">=12.13.0". Got "12.0.0"`

Therefore minor has been bumped to "12.13.0", and installing "12.13.0" specifically builds and runs test successfully.

Therefore is it fair to say "12.13.0" is needed given tailwind can't function without it?


timleslie:
> Therefore is it fair to say "12.13.0" is needed given tailwind can't function without it?

Yep, that's fine 👍

timleslie (changes requested):
This will be a breaking change, so we'll want to have a `major` changeset for all of our packages.

.changeset/large-beers-drive.md
```diff
@@ -0,0 +1,44 @@
...
+'@keystone-next/auth': major
+'@keystone-next/cloudinary': major
+'@keystone-next/fields': major
+'@keystone-next/keystone': major
+'@keystone-next/session-store-redis': major
+'@keystone-next/types': major
+'@keystone-next/access-control-legacy': major
+'@keystone-next/adapter-prisma-legacy': major
+'@keystone-next/fields-legacy': major
+'@keystone-next/file-adapters-legacy': major
```
timleslie, line 34:
```suggestion
```

These two packages are being removed in PRs that are about to land, so we don't want them going into any more changesets, since we won't be publishing them.

timleslie (approved):
👍

PR #5402, opened 2021-04-08
Start of testing the CLI

emmatown:
This is mostly testing infrastructure things to make it easy to write tests for the CLI and then one test to make sure it all works.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,15 @@ module.exports = {
     jest: true,
     'cypress/globals': true,
   },
-  plugins: ['react', 'react-hooks', 'jest', 'cypress', 'import', '@typescript-eslint'],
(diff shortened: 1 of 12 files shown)


.eslintrc.js
```diff
@@ -103,6 +111,7 @@ module.exports = {
         },
       },
     ],
+    '@preconstruct/format-js-tag/format': 'error',
```
mitchellhamilton, line 114: This ESLint rule formats things like this with prettier so we can write JS code in JS that we need to write to a file without having to format it manually. (it's not used in the test in this PR but will be used in the future for tests around compilation)

```js
js`
   console.log(1);
`
```

timleslie (approved):
Looks good 👍

PR #5403, opened 2021-04-08
Improve field label from key heuristics

gwyneplaine:
## Description
Field labels were previously using inflection to coerce the field key into a human readable label. 
This caused incongruity with acronyms in particular.  To address this we've done the following: 
* Remove inflection from the fields package 
* Moved the core logic in the `keyToLabel` utility in `@keystone-next/keystone-legacy` to `@keystone-next/utils-legacy`
* Made minor adjustments to the logic. **
* Replace the invocation of `inflection` in the fields implementation with the new `humanize` utility. 

**The previous logic in keyToLabel would  turn the following string `standardHTMLThing`  to `Standard HTMLThing` 
This is not consistent with the asserted behaviour, we've augmented this logic now to ensure the transformation is as follows: 
`standardHTMLThing` --> `Standard HTML Thing` 

Before: 
Screen Shot 2021-04-08 at 10 59 51 am

After: 
Screen Shot 2021-04-08 at 12 24 53 pm

I'm not really sure if this should be considered a breaking change or not, as I'm unclear if the label also affects non ui specific operations. (for example the returned value of keyToLabel is used both in pluralize and singularize in the list implementation) 
perhaps @timleslie  or @mitchellhamilton  may have a semver suggestion here.

----- Edit ------ 
Removing the logic changes to keyToLabel / humanize, as this would be a major breaking change. 
The field label logic should still be a drastic improvement from before.


mitchellhamilton:
Yeah, this is a major because the label is then passed through pluralize and then some more things happen to it and it ends up in the GraphQL API.

gwyneplaine:
> Yeah, this is a major because the label for lists that comes from keyToLabel is then passed through pluralize and then some more things happen to it and it ends up in the GraphQL API.

@mitchellhamilton mm, releasing a major for a slightly more correct label utility seems bad to me. 
I'll revert the logical changes made for this, and it should still satisfy the field bug correction. 

At which point we can *probably* release this as a patch, provided the label in the `packages/fields/src/Implementation.js` is isolated to just UI concerns. 🤞

emmatown (changes requested):
Needs a changeset

packages/utils/src/index.ts
```diff
@@ -241,3 +241,20 @@ export const versionGreaterOrEqualTo = (
...
+export const humanize = (str: string) => {
+  let label = str
+    .replace(/([a-z])([A-Z]+)/g, '$1 $2')
+    .split(/\s|_|\-/)
+    .filter(i => i)
+    .map(upcase)
+    .join(' ');
+
+  return label;
+};
```
timleslie, line 260: Any reason not to do this as a simple return value?
```suggestion
export const humanize = (str: string) => str
  .replace(/([a-z])([A-Z]+)/g, '$1 $2')
  .split(/\s|_|\-/)
  .filter(i => i)
  .map(upcase)
  .join(' ');
```

timleslie (approved):
LGTM 👍

emmatown (approved):

PR #5404, opened 2021-04-08
Postinstall command tests

emmatown:
This is pretty much all tests. The one notable thing besides that is that I made it so we format the GraphQL schema that we write with Prettier and I made that a major because it means `postinstall` and etc. will fail immediately after they update.


timleslie (approved):
Looks good 👍

PR #5405, opened 2021-04-08
Convert Implementation.js to TypeScript

timleslie:
This PR converts the Implementation files for core field types to TypeScript. We don't do the Relationship field type in this PR as it's a bit more complex and will come through in its own PR.

No changeset for this PR as its covered by existing changesets.


emmatown (changes requested):
Can you replace `as string[]` and that sort of thing with typing the return type explicitly? It's generally best to avoid `as` unless you have a specific need for it and a type annotation doesn't solve your problem. When I see `as`, I think "what's the reason why `as` specifically is used here?"

emmatown (approved):

PR #5415, opened 2021-04-11
Typescript document field

timleslie:
Converts the Implementation of the `Document` field to TypeScript.


mitchellhamilton (dismissed):

packages-next/fields-document/src/Implementation.ts
```diff
@@ -1,41 +1,76 @@
-import { PrismaFieldAdapter } from '@keystone-next/adapter-prisma-legacy';
+import { PrismaFieldAdapter, PrismaListAdapter } from '@keystone-next/adapter-prisma-legacy';
 import { Implementation } from '@keystone-next/fields';
+import { FieldConfigArgs, FieldExtraArgs } from '@keystone-next/fields/src/Implementation';
```
mitchellhamilton, line 3: You shouldn't ever import from `src`, can you export these from somewhere that can be imported from an actual entrypoint?

mitchellhamilton (changes requested):

mitchellhamilton (approved):

PR #5416, opened 2021-04-11
Convert the validation module to TypeScript

timleslie:
Typescript conversion of this helper module used in out `limits.test.ts` tests.

### DIFF
--- a/tests/api-tests/queries/validation.ts
+++ b/tests/api-tests/queries/validation.ts
@@ -1,30 +1,50 @@
-const { visit, Kind } = require('graphql');
+import { visit, Kind, ValidationContext, ASTNode } from 'graphql';
 
-const definitionLimit = maxDefinitions => validationContext => {
+type Fragment = { name: string; atDepth: number };
+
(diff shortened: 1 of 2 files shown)


tests/api-tests/queries/validation.ts
```diff
@@ -1,30 +1,50 @@
...
+
+export const definitionLimit = (maxDefinitions: number) => (
+  validationContext: ValidationContext
+) => {
   const doc = validationContext.getDocument();
   if (doc.definitions.length > maxDefinitions) {
     validationContext.reportError(
-      new Error(`Request contains ${doc.definitions.length} definitions (max: ${maxDefinitions})`),
-      doc
+      // @ts-ignore
```
timleslie, line 20: `validationContext.reportError` claims that it wants a `GraphQLError`, but when we give it one we end up with internal server errors. I couldn't work out why without going way deeper than I have time for, so for now we'll stick with what works.

emmatown (approved):

PR #5417, opened 2021-04-11
fixing renderer props being overwritten

alexmgrant:
Looks to me that passing the renderers prop to `<DocumentRenderer />` is overwritten by `defaultRenderers` due to the order of the spread operators. I'm swapping them here so passing in custom `props.renderers` will work.


mitchellhamilton (approved):
Thanks!

PR #5418, opened 2021-04-12
Remove incorrect usage of the CLI in the examples

emmatown:
This doesn't really make the usage of the CLI in the examples correct, it just makes it not totally wrong. Note I've removed some usages of the generated types since they're in node_modules now and we can't use `keystone-next postinstall` properly in the examples until we make various decisions about the examples like if we want them to be able to run on different databases.


timleslie (approved):
LGTM 👍

PR #5420, opened 2021-04-12
Flatten field adapter config

timleslie:
We only support one adapter now, so this no longer needs to be a nested object.

### DIFF
--- a/.changeset/soft-cherries-hide.md
+++ b/.changeset/soft-cherries-hide.md
@@ -1,0 +1,8 @@
+---
+'@keystone-next/cloudinary': major
+'@keystone-next/fields': major
+'@keystone-next/fields-document': major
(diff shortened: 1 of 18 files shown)


.changeset/soft-cherries-hide.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystone-next/cloudinary': major
+'@keystone-next/fields': major
+'@keystone-next/fields-document': major
+'@keystone-next/keystone-legacy': major
```
mitchellhamilton, line 5:
```suggestion
'@keystone-next/keystone-legacy': major
'@keystone-next/keystone': major
```

emmatown (changes requested):

emmatown (approved):

PR #5444, opened 2021-04-13
Prisma command tests

emmatown:
Just found one little problem with how we handled the Prisma CLI exiting with a non-zero exit code.


timleslie (approved):
👍

PR #5447, opened 2021-04-13
Update docs and other assets to refer to new `create-keystone-app`

bladey:
Related PR: https://github.com/keystonejs/create-keystone-next-app/pull/65


examples-next/graphql-api-endpoint/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# keystone-next-app
+# keystone-app
```
timleslie, line 1: I don't think we want to change this package.

examples-next/graphql-api-endpoint/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "keystone-next-app",
+  "name": "keystone-app",
```
timleslie, line 2: I don't think we want to change this package.

timleslie (changes requested):
Just the two reversions and then I think we're good 👍 

Once the CKA churn has died down a bit I'd like us to fully regenerate the cover SVG, but no need to worry about that for now.

examples-next/graphql-api-endpoint/CHANGELOG.md
bladey: Reverted.

examples-next/graphql-api-endpoint/package.json
bladey: Reverted.

timleslie (approved):
Looks good 👍

PR #5450, opened 2021-04-13
Tests for build and start command

emmatown:
Nothing particularly interesting here, the only observable change here is users won't see a stack trace when start fails because they need to run build.


timleslie (approved):
👍

PR #5451, opened 2021-04-13
Rename resolveFields to query in the itemAPI

JedWatson:
After feedback and an internal discussion, we're going to rename the `resolveFields` argument in the Items API to `query`.

Hopefully this helps make the Items API more intuitive and self-documenting.

For background, the Items API makes Keystone's GraphQL API available to Node.js directly, without the overhead of making a full GraphQL query over http. Because of the more direct syntax, you don't need to write out a full GraphQL query, but you do still need to nominate the fields you want resolved, in GraphQL syntax (the "field selection"). Unlike with most ORMs (which this provides a similar interface to) you get the full functionality of nested queries through relationships, and can use virtual fields.

It also offers a more direct option of returning the raw data from the database (effectively equivalent to `select *` in SQL) which we refer to as the "unresolved list item". This is required for certain back-end functionality to to work, e.g comparing a password hash (which is never available through the GraphQL API).

The challenge is given an API that can return two significantly different types, how do we make it intuitive to use? Hopefully naming the argument that controls this behaviour `query` makes it clearer:

(A) By default, the API will just return the `id` of each result

(B) If you want to query more fields (including virtual fields or related items) you pass the `query: 'fields'` argument, with GraphQL Query syntax for the fields you want to query from the graph

(C) If you want the raw / unresolved list item data, you pass `query: false` which bypasses the field resolvers


packages-next/keystone/src/lib/itemAPI.ts
```diff
@@ -34,27 +34,35 @@ export function getArgsFactory(list: BaseKeystoneList, schema: GraphQLSchema) {
   };
 }
 
+function defaultQueryParam(query?: string | false, resolveFields?: string | false) {
+  if (resolveFields !== undefined) return resolveFields;
+  if (query !== undefined) return query;
```
mitchellhamilton, line 39: We should prefer query or throw an error if both are passed

JedWatson: Good point.

Real q: what would be the right way of throwing an error in this part of the codebase?

mitchellhamilton: `throw new Error('')`?

JedWatson: I have updated it but lmk if we should be throwing errors differently

mitchellhamilton: I think you forgot to push?

mitchellhamilton (approved):

packages-next/keystone/src/lib/itemAPI.ts
```diff
@@ -34,27 +34,38 @@ export function getArgsFactory(list: BaseKeystoneList, schema: GraphQLSchema) {
   };
 }
 
+function defaultQueryParam(query?: string | false, resolveFields?: string | false) {
+  if (query !== undefined && resolveFields !== undefined) {
+    throw new Error('query and resolveFields cannot both be passed to an Items API query');
+  }
+  if (query !== undefined) return query;
+  if (resolveFields !== undefined) return resolveFields;
+  return 'id';
```
rohan-deshpande, line 43: I would have defaulted it to `false`

PR #5456, opened 2021-04-13
Update resolveFields and query behaviour in the Items API

JedWatson:
Followup to #5451 after further discussion

**NOTE:** This has been changed again, see comment below. Leaving this previous description for context.

I think this is even clearer, and the API now "does what it says"

You have two options:

(1) Query specific fields by passing the `query` option, as a string:

```js
const [post] = await lists.Post.findMany({
  where: { slug },
  query: `
    title
    content
    image {
      src
      width
      height
    }`,
});
```

(2) Return the raw data stored in the database, by passing `resolveFields: false`

```js
const [post] = await lists.Post.findMany({
  where: { slug },
  resolveFields: false,
});
```

For **backwards compatibility** the types and logic still accept a string value to `resolveFields`, I propose we leave it like that for now and narrow it to only accept `false` in a future release.


JedWatson:
Note I've updated the changeset from #5451 so the release notes are clearer, rather than adding a new one

rohan-deshpande (approved):
I like this. It makes the API a lot clearer.

mitchellhamilton (changes requested):
Just want to explicitly block this on further discussion

JedWatson:
Another change after *more* discussion:

`resolveFields` has been deprecated entirely (again) in favour of a new API on `context.db.lists.{List}`

The updated changelog:

---

With the goal of making the Lists API (i.e `context.lists.{List}`) more intuitive to use, the `resolveFields` option has been deprecated in favor of two new methods:

(1) You can specify a string of fields to return with the new `query` option, when you want to query for resolved field values (including querying relationships and virtual fields). This replaces the `resolveFields: string` use case.

For example, to query a Post you would now write:

```js
const [post] = await context.lists.Post.findMany({
  where: { slug },
  query: `
    title
    content
    image {
      src
      width
      height
    }`,
});
```

(2) Alternatively, there is a new set of APIs on `context.db.lists.{List}` which will return the unresolved item data from the database (but with read hooks applied), which can then be referenced directly or returned from a custom mutation or query in the GraphQL API to be handled by the Field resolvers. This replaces the `resolveFields: boolean` use case.

For example, to query for the raw data stored in the database, you would write:

```js
const [post] = await context.db.lists.Post.findMany({
  where: { slug }
});
```

mitchellhamilton (approved):

timleslie (approved):
👍 LGTM

PR #5460, opened 2021-04-14
Move the core code into packages-next/keystone

timleslie:
This PR brings the code from `packages/keystone` into `packages-next/keystone` without making any functional changes. This will allow us to more easily continue the process of converting this code to TypeScript while removing legacy code without as much `major` version churn.


emmatown (changes requested):
All CommonJS things need to be converted to ESM otherwise this will be broken when built.

emmatown (approved):

PR #5465, opened 2021-04-15
Use webpack 5

emmatown:
See https://nextjs.org/docs/messages/webpack5 for the what and why of this. Given we don't do super fancy webpack things, there's not much that can break.


mitchellhamilton:
Ughhhh, this is gonna be harder than I though.

PR #5467, opened 2021-04-15
Remove context.executeGraphQL

timleslie:
This was only being kept around because our tests depended on it. The test have all been updated, so this can now go away. 👋

### DIFF
--- a/.changeset/cyan-rules-smoke.md
+++ b/.changeset/cyan-rules-smoke.md
@@ -1,0 +1,7 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+'@keystone-next/website': patch
+---
+
+Removed the deprecated `context.executeGraphQL`. Identical functionality is available via `context.graphql.run`.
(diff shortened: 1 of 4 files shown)


.changeset/cyan-rules-smoke.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+'@keystone-next/website': patch
+---
+
+Removed the deprecated `context.executeGraphQL`. Identical functionality is available via `context.graphql.run`.
```
mitchellhamilton, line 7:
```suggestion
Removed the deprecated `context.executeGraphQL`. Identical functionality is available via `context.graphql.raw`.
```

emmatown (approved):

.changeset/cyan-rules-smoke.md
timleslie: Oops, yep, nice catch 😅

PR #5470, opened 2021-04-15
Fix example syntax for cloudinary image field

MichaelZaporozhets:


timleslie (approved):
Nice catch, thanks for the fix!

PR #5475, opened 2021-04-15
add a note about homebrew

kud:


timleslie (dismissed):
Holding off on merging pending larger website changes 👍

JedWatson (approved):
The website changes @timleslie mentioned have landed so I think we're good to merge this now.

Thanks for the contrib @kud!

kud:
My pleasure! I'm a newbie with keystone and if it can help, i'll be glad :)

bladey (dismissed):
I'm quite sure this should be `postgres://my-username@localhost/my-app`, I run into errors if I use the format `postgres://my-username:localhost/my-app`, is this right?

kud:
> I'm quite sure this should be `postgres://my-username@localhost/my-app`, I run into errors if I use the format `postgres://my-username:localhost/my-app`, is this right?

It seems that I was tired when I patched it. It's indeed `@` and not `:`, thank you. :)

JedWatson (approved):

PR #5476, opened 2021-04-15
Moving to emotion in the docs site and unifying the theme

dominikwilkowski:


JedWatson:
There's a bug with image max size that I assume is just something missing from the TW reset:

image

JedWatson:
Also on the getting started page, not sure if it's related to the change but there's a weird dangling comma:

image

JedWatson:
While you're playing with the styling, it seems like the main content area is slightly _thinner_ than in the current site; not a blocker, but tbh I think it should actually get a bit _wider_ at the default break point. The content is feeling pretty cramped since we added the right "in page" nav

docs-next/components/CopyToClipboard.tsx
```diff
@@ -1,36 +1,39 @@
...
     } else {
       // Fallback to a library that leverages document.execCommand
-      // for browser versions that dont' support the navigator object.
+      // for browser versions that don't support the navigator object.
       // As document.execCommand
       try {
         copy(text);
       } catch (e) {
-        addToast({ title: 'Faild to oopy to clipboard', tone: 'negative' });
+        addToast({ title: 'Failed to copy to clipboard', tone: 'negative' });
```
JedWatson, line 36: 🙏

docs-next/components/CopyToClipboard.tsx
```diff
@@ -42,7 +45,7 @@ export const CopyToClipboard = ({ value }: { value: string }) => {
       href={`#${value}`}
       css={{
         alignItems: 'center',
-        color: '#97A0AF', // colors.n40
+        color: 'var(--gray-400)',
```
JedWatson, line 48: so glad to be getting rid of these magic styles

docs-next/components/DocumentEditorDemo.tsx
```diff
@@ -334,9 +331,9 @@ export const DocumentEditorDemo = () => {
       />
       <div
         css={{
-          marginTop: theme.spacing.xlarge,
-          marginBottom: theme.spacing.xlarge,
-          borderBottom: `1px ${theme.colors.border} solid`,
+          marginTop: 'var(--space-large)',
```
JedWatson, line 334: Standardising on a single theme 💖

docs-next/components/Header.tsx
```diff
@@ -0,0 +1,168 @@
...
+                  verticalAlign: 'middle',
+                }}
+              >
+                <defs>
+                  <linearGradient id="logo-a" x1="0%" x2="50%" y1="0%" y2="71.9%">
+                    <stop offset="0%" stopColor="#5AE8FA" />
+                    <stop offset="100%" stopColor="#2684FF" />
+                  </linearGradient>
+                </defs>
+                <path
```
JedWatson, line 53: Not a blocker, but I think this would be easier to read if the svgs were put in their own components (even if not in separate files, just naming them rather than inlining them here would make the component easier to scan)

docs-next/components/Navigation.tsx
```diff
@@ -1,39 +1,61 @@
...
 type SectionProps = { label: string; children: ReactNode };
-const Section = ({ label, children }: SectionProps) => {
+function Section({ label, children }: SectionProps) {
   return (
     <Fragment>
-      <h3 className="uppercase mt-6 mb-2 text-gray-700 font-bold">{label}</h3>
+      <h3
+        css={{
+          textTransform: 'uppercase',
+          margin: '1.5rem 0 0.5rem 0',
```
JedWatson, line 14: Not a blocker, but it feels like a step backwards replacing spacing tokens with hard coded values 🙁 

We'll take a step forward again in a future PR?

docs-next/components/Header.tsx
JedWatson: **edit** noticed later in the review that these were inlined previously as well, worth scouting anyway

JedWatson (changes requested):
Left a few comments, really only blocking this on the image issue though

docs-next/components/Header.tsx
dominikwilkowski: Yeah I agree. I didn't want to fix too much in this already huge PR. There are other thing here as well that I want to do including aligning the blue with our theme.

docs-next/components/Navigation.tsx
dominikwilkowski: Yeah was thinking the same thing.
I did add some spacing tokens later but not applied them consistently ...
Needs an overall system PR at some point with font sizes as well

dominikwilkowski:
Great finds @JedWatson 
I was going over the site before and after page by page and at some point you just go blind for the biggest issues 😬 
Fixed image and dangling semicolon now

JedWatson (approved):
Looks good @dominikwilkowski! Thanks for fixing those issue so quickly 🎉

PR #5478, opened 2021-04-16
Convert the List class to TypeScript

timleslie:


packages-next/fields/src/types/relationship/Implementation.ts
```diff
@@ -194,7 +194,6 @@ export class Relationship<P extends string> extends Implementation<P> {
     operations: RelationshipOperation,
     item: any,
     context: KeystoneContext,
-    getItem: any,
```
mitchellhamilton, line 197: This change requires a major for fields and keystone

emmatown (changes requested):

emmatown (approved):

PR #5497, opened 2021-04-20
Add docs for the image field type

timleslie:
Adds API docs for the image field type. A full guide on how to use this feature still needs to be written.

### DIFF
--- a/docs-next/pages/apis/fields.mdx
+++ b/docs-next/pages/apis/fields.mdx
@@ -28,6 +28,9 @@ import {
 
   // Virtual type
   virtual,
+
+  // File types
+  image,
 } from '@keystone-next/fields';
 
 // Complex types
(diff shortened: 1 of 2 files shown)


docs-next/pages/apis/fields.mdx
```diff
@@ -510,6 +513,43 @@ export default config({
...
+
+### file
+
+(coming soon)
+
+### image
+
+An `image` field represents an image file, i.e. `.jpg`, `.png`, `.webp`, or `.gif`.
+
+See [`config.images`](./config#images) for details on how to configure your Keystone system with support for the `image` field type.
```
gautamsi, line 528: is this going to restrict us from saving images to different target (local/s3/etc) in different list? I see that config.images is keystone wide and I may want to use different target in different list. like avatar in local and gallery in s3 (when available)

timleslie: At the moment the intent is to have a single target location to begin with and give that some time to settle while we observe usage patterns. If there's enough demand we'll investigate expanding the API to support more complex configurations.

gwyneplaine (approved):

PR #5505, opened 2021-04-21
Dom/coc

dominikwilkowski:
Fixed COC email and tested it to be going to the entire team making sure that any and all emails going to that email are never forgotten or able to be hidden


bladey (approved):
👍🏻

PR #5506, opened 2021-04-21
Fix Syntax Error in example code block

MichaelZaporozhets:
Fix `Unterminated string constant` syntax error in example code


bladey (approved):
Thanks @MichaelZaporozhets!

dominikwilkowski (approved):
Nice find

PR #5508, opened 2021-04-21
Image field type tests

timleslie:
Add API tests for the image field type.


packages-next/fields/src/types/image/tests/test-fixtures.ts
```diff
@@ -0,0 +1,271 @@
+import path from 'path';
+import fs from 'fs-extra';
+// @ts-ignore
```
mitchellhamilton, line 3: Why is this ts-ignored?

timleslie: Otherwise we end up with:

```
Module '"graphql-upload"' has no exported member 'Upload'.
```

The code works, so I'm not keen on tracking down why typescript is complaining here, but if you know a better solution I'm all ears 👍

emmatown (approved):

PR #5511, opened 2021-04-21
Add guides on how to use relationship fields

timleslie:
Add docs on the core concepts of relationships, how to decide on relationship configuration, and API docs for the relationship field itself.

### DIFF
--- a/packages-next/fields/src/types/relationship/index.ts
+++ b/packages-next/fields/src/types/relationship/index.ts
@@ -44,6 +44,7 @@ export type RelationshipFieldConfig<
     hideCreate?: boolean;
   };
   defaultValue?: FieldDefaultValue<Record<string, unknown>>;
+  // FIXME: NOT IMPLEMENTED
   isUnique?: boolean;
 } & (SelectDisplayConfig | CardsDisplayConfig);
 
--- a/docs/components/Navigation.tsx
+++ b/docs/components/Navigation.tsx
@@ -69,6 +69,8 @@ export function Navigation() {
       <Section label="Guides">
         <NavItem href="/guides/keystone-5-vs-keystone-next">Keystone 5 vs Next</NavItem>
         <NavItem href="/guides/cli">Command Line</NavItem>
+        <NavItem href="/guides/understanding-relationships">Understanding Relationships</NavItem>
+        <NavItem href="/guides/configuring-relationships">Configuring Relationships</NavItem>
         <NavItem href="/guides/access-control" isPlaceholder>
           Access Control
         </NavItem>
--- a/docs/pages/apis/fields.mdx
+++ b/docs/pages/apis/fields.mdx
@@ -410,8 +410,26 @@ export default config({
 ### relationship
 
 A `relationship` field represents a relationship between two lists.
+See the guides on [understanding relationships](../guides/understanding-relationships) and [configuring relationships](../guides/configuring-relationships) for details of Keystone's relationship model and how to configure your project's relationships.
 
-(coming soon)
+- `ref` (required): A string of the form `<listKey>` or `<listKey>.<fieldPath>`.
+- `many` (default: `false`): Configures the cardinality of the relationship.
+- `defaultValue` (default: `undefined`): Can be either a relationship input value or an async function which takes an argument `({ context, originalItem })` and returns a relationship input value.
+  This value will be used for the field when creating items if no explicit value is set.
+  `context` is a [`KeystoneContext`](./context) object.
+  `originalItem` is an object containing the data passed in to the `create` mutation.
+- `ui` (default: `{ hideCreate: false, displayMode: 'select' }`): Configures the display mode of the field in the Admin UI.
+  - `hideCreate` (default: `false`). If `true`, the "Create related item" button is not shown in the item view.
+  - `displayMode` (default: `'select'`): Controls the mode used to display the field in the item view. The mode `'select'` displays related items in a select component, while `'cards'` displays the related items in a card layout. Each display mode supports further configuration.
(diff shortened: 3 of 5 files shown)


packages-next/fields/src/types/relationship/index.ts
```diff
@@ -44,6 +44,7 @@ export type RelationshipFieldConfig<
     hideCreate?: boolean;
   };
   defaultValue?: FieldDefaultValue<Record<string, unknown>>;
+  // FIXME: NOT IMPLEMENTED
```
timleslie, line 47: This feature isn't actually implemented, so I've not included it in the API docs. I'm going to follow up on what to do with the types themselves in the context of the upcoming field changes separate to this PR.

docs-next/pages/apis/fields.mdx
```diff
@@ -424,9 +442,21 @@ export default config({
...
           ref: '...',
           many: false,
-          ui: { /* ... */ },
           defaultValue: '...',
-          isUnique: '...',
+          ui: {
+            hideCreate: false,
+            // Display mode: 'select'
+            displayMode: 'select',
+            labelField: 'name,
```
dominikwilkowski, line 450: missing a quote
```suggestion
            labelField: 'name',
```

docs-next/pages/guides/configuring-relationships.mdx
```diff
@@ -0,0 +1,243 @@
...
+    User: list({
+      fields: {
+        name: text()
+      }
+    }),
+    Post: list({
+      fields: {
+        title: text(),
+        content: text(),
+        author: relationship({ ref: 'User', many: false }),
```
dominikwilkowski, line 128: This one has a trailing comma and the other bits don't ... not a blocker... just pointing out

timleslie: Ah yeah, I need to go and manually lint this 👍

docs-next/pages/guides/configuring-relationships.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+```typescript
+import { config, createSchema, list } from '@keystone-next/keystone/schema';
+import { text, relationship } from '@keystone-next/fields';
+
+export default config({
+  lists: createSchema({
+    User: list({
+      fields: {
+        name: text()
```
dominikwilkowski, line 121: This one doesn't have a trailing comma while the other bits do... not a blocker... just pointing out.

docs-next/pages/guides/configuring-relationships.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+## One-sided vs two-sided
+
+A relationship in Keystone exists _between_ two lists.
+In our blog, the concept of _authorship_ (who wrote a post) can be represented as a relationship between the `User` and the `Post` lists.
+
+The first question you need to consider is which list do you want to be able to access the relationship from in your GraphQL API?
+In our blog we might want to be able to ask about a `user's posts`, a `post's author`, or possibly both.
+If you only need to access one side of the relationship then you want to configure a _one-sided_ relationship. If you need both, then you want to configure a _two-sided_ relationship.
+
```
molomby, line 19:
```suggestion
Note that, in either case, the data stored is the same.
This distinction solely effects how the relationship is surfaces by the GraphQL API.

```

(This is correct, right?)

docs-next/pages/guides/understanding-relationships.mdx
```diff
@@ -0,0 +1,301 @@
...
+        title: text(),
+        content: text(),
+        authors: relationship({ ref: 'User', many: true }),
+      }
+    })
+  }),
+});
+```
+
+### Twos-sided
```
JedWatson, line 221:
```suggestion
### Two-sided
```

JedWatson (changes requested):

docs-next/pages/guides/configuring-relationships.mdx
```diff
@@ -0,0 +1,243 @@
...
+});
+```
+
+Note that we have used `many: false` in both the `authors` and `posts` fields.
+
+## Summary
+
+When configuring a relationship in Keystone there are two key questions you need to answer:
+
+- Do I want a one-sided or two-sided relationship?
```
molomby, line 238:
```suggestion
- Do I want to access the relationship from only one-side or both?
```

Or something similar; just feel we should be emphasising this is purely around access/API.

molomby (approved):
Generally looks great.

I do think we could do a better job of calling out that the one/two-sided distinction only effects the API. It's not about repeating ourselves, just about phrasing. See inline suggestions.

As mentioned, I thought some of your link paths were incorrect but apparently it's the [style guide](https://github.com/keystonejs/keystone/blob/master/STYLE_GUIDE.md#links) that needs fixing. skinner-could-it-be.jpg

Also, not one for now, but being all the "many: true | false" and one/two sided stuff explained really reinforces to me that our method of configuring this stuff is bad. It would be so much more natural to talk in terms of "one-to-many", "many-to-many", etc. rather than having to decode those categories to multiple bits of config that exist in different places. Still hope to resolve this one day.

docs-next/pages/guides/configuring-relationships.mdx
timleslie: This language is specifically goal oriented, not technically oriented, around what they want to be able to do, which will lead them to the technical decision of one-sided or two-sided.

timleslie: Oh wait, I'm reading this backwards, and it's my prose that's very wrong. yes, good suggestion indeed!

PR #5516, opened 2021-04-22
Use terminal-link command rather than link

timleslie:
The correct command to use is `terminal-link`, not `link`. This appears to have changed in `v3` (though wasn't mentioned in the release notes, and isn't run on CI, so we missed it when updating that package).


dominikwilkowski (approved):
LGTM

PR #5517, opened 2021-04-22
Only coerce items API input when using resolveFields: false

timleslie:
The argument coercion applied in the items API should only be done in the code path where we're calling the mutations directly. In the case where we go through the GraphQL API itself we don't want to do this. For most field types it doesn't make a different but for `Upload` types it will convert an object of the form 

```
    {
      upload: Upload {
        resolve: [Function],
        reject: [Function],
        promise: Promise { [Object] },
        file: {
          createReadStream: [Function: createReadStream],
          filename: 'keystone.jpeg',
          mimetype: 'image/jpeg',
          encoding: 'utf-8'
        }
      }
    }
```

into 

```
    {
      upload: Promise {
        {
          createReadStream: [Function: createReadStream],
          filename: 'keystone.jpeg',
          mimetype: 'image/jpeg',
          encoding: 'utf-8'
        }
      }
    }
```

which doesn't work as an input with the `graphql()` function which we use to execute server side GraphQL

### DIFF
--- a/packages-next/fields/src/types/image/tests/test-fixtures.ts
+++ b/packages-next/fields/src/types/image/tests/test-fixtures.ts
@@ -68,22 +68,20 @@ export const crudTests = (keystoneTestWrapper: any) => {
       keystoneTestWrapper(async ({ context }: { context: any }) => {
         const filenames = ['keystone.jpeg', 'keystone.jpg', 'keystone'];
         for (const filename of filenames) {
-          const data = await createItem({
-            context,
-            listKey: 'Test',
-            item: { avatar: prepareFile(filename) },
-            returnFields: `
-            avatar {
-              id
-              mode
-              filesize
-              width
-              height
-              extension
-              ref
-              src
-            }
-        `,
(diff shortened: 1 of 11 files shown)


packages-next/fields/src/types/image/tests/test-fixtures.ts
```diff
@@ -68,22 +68,20 @@ export const crudTests = (keystoneTestWrapper: any) => {
...
-              mode
-              filesize
-              width
-              height
-              extension
-              ref
-              src
-            }
-        `,
+          const data = await context.lists.Test.createOne({
```
timleslie, line 71: The changes in this PR allow this test to be re-written with the items API, where previously it would fail.

mitchellhamilton (changes requested):
I believe this is incorrect, I'm going to look into it more.

mitchellhamilton (approved):
I was incorrect, I interpreted this as "stop coercing the inputs completely" when really this is "let the `graphql` function coerce the inputs"

Issue #5485, opened 2021-04-19
Update API tests to use items API


timleslie:
**Contributions Welcome**

This issue contains work to be done which we would love community contributions on! If you are looking to contribute to the project then this is a great place to start. Please open a PR and reference this issue 🙏 

**Task description**

The GraphQL API tests in `tests/api-tests` currently make extensive use of the `@keystone-next/server-side-graphql-client-legacy` package. This package is being deprecated in favour of the [list item API](https://next.keystonejs.com/apis/context#list-item-api).

As part of this deprecation process we need to port all the GraphQL API tests to use this new API. 

To do this we simply need to replace all instances of the `server-side-graphql-client` functions with their list item equivalent. For example, the code

```javascript
const data = await getItem({
  context,
  listKey,
  itemId: items[0].id,
  returnFields,
});
```

becomes

```javascript
const data = context.lists[listKey].findOne({
  where: { id: items[0].id },
  query: returnFields,
});
```

In general there should be a a one to one correspondence between these APIs, so the changes should be reasonably simple, however there are a very large number of function calls to replace.

**Notes**

* We would prefer to review a series of small PRs rather than a single enormous PR. If you only have time to port a single test file to the new API then such a PR would be welcomed 🙏 
* We are still working on the documentation for the lists API. Please consult the code itself to find the appropriate replacements to use.
* We have an open internal task to fully document how to run all of the different tests in a local dev environment. For now, you should be able to run individual test files with the following command:

```shell
TEST_ADAPTER=sqlite DATABASE_URL=file:./test.db yarn test:unit tests/api-tests/fields/filter.test.ts 
```

timleslie:
This task is now complete 👍

PR #5518, opened 2021-04-22
Add API docs for the list items API

timleslie:
Adds docs for the lists item API. Preview here: https://keystone-next-docs-git-items-api-docs-keystonejs.vercel.app/apis/list-items

Updates the `findMany()` and `count()` functions to have a default arg of `{}` so that `.count()` and `.findMany()` work as expected when no argument is provided.

Updates an arbitrary handful of server-side-graphql-client calls in the tests to use the list items API to verify that the documents can be followed successfully. This is the first set of changes for #5485.

Does not document the DB items API, that will be done separately.

### DIFF
--- a/.changeset/fair-shrimps-love.md
+++ b/.changeset/fair-shrimps-love.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/keystone': patch
+'@keystone-next/types': patch
+---
+
+Updated the list item and db item APIs to include an empty default for `.findMany()` and `.count()`.
--- a/docs/components/Navigation.tsx
+++ b/docs/components/Navigation.tsx
@@ -98,9 +98,7 @@ export function Navigation() {
         <NavItem href="/apis/auth">Authentication API</NavItem>
         <NavItem href="/apis/context">Context API</NavItem>
         <NavItem href="/apis/graphql">GraphQL API</NavItem>
-        <NavItem href="/apis/list-items" isPlaceholder>
-          List Item API
-        </NavItem>
+        <NavItem href="/apis/list-items">List Item API</NavItem>
       </Section>
     </nav>
   );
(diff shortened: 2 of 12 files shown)


JedWatson (approved):
Great start @timleslie 👏 

Not required for this PR (good to publish this immediately imo) but a couple of next steps stood out when I was reading:

* You're demo'ing the API for relationship creation, I don't think we've actually documented this anywhere else and it's a bit like "oh wait, what's that doing? how do I learn more? what else can I do?" and it would be good to add a link out when there's more docs for that
* For the create/update mutation APIs, would be nice to add some explanation for what the data argument actually is; this subtly doesn't match the output types people may expect, again I think the API reference for this is missing though and people will probably wonder about that here, it's a bit mysterious
* We're not explaining anything about how permissions (/ session) interact here. That's kind of covered in the [context API](https://keystone-next-docs-git-items-api-docs-keystonejs.vercel.app/apis/context#new-context-creators) docs, but would probably be good to remind readers about here because it's usually going to be related and important to think about when using this API

timleslie:
> Great start @timleslie 👏
> 
> Not required for this PR (good to publish this immediately imo) but a couple of next steps stood out when I was reading:
> 
> * You're demo'ing the API for relationship creation, I don't think we've actually documented this anywhere else and it's a bit like "oh wait, what's that doing? how do I learn more? what else can I do?" and it would be good to add a link out when there's more docs for that

Yep, that's on my hit list 👍 

> * For the create/update mutation APIs, would be nice to add some explanation for what the data argument actually is; this subtly doesn't match the output types people may expect, again I think the API reference for this is missing though and people will probably wonder about that here, it's a bit mysterious

We have rudimentary API docs for the GraphQL operations here: https://next.keystonejs.com/apis/graphql#create-user I might go ahead and link do these docs.

> * We're not explaining anything about how permissions (/ session) interact here. That's kind of covered in the [context API](https://keystone-next-docs-git-items-api-docs-keystonejs.vercel.app/apis/context#new-context-creators) docs, but would probably be good to remind readers about here because it's usually going to be related and important to think about when using this API

Yep, will loop back on that once those guides and APIs are more fleshed out 👍

PR #5529, opened 2021-04-23
File field type

gwyneplaine:
Initial implementation of the file-field type
Essentially a tweaked port of the image field type with key differences: 
* we don't include width and height 
* there's no filetype validation 
* We store a filename instead of an id, the filename is the slugified original filename (provided by the user) with a uuid appended to it like so `${slugify(originalFilename}_${uuid()}`


examples-next/basic/schema.prisma
```diff
@@ -0,0 +1,51 @@
+datasource postgresql {
```
timleslie, line 1: This file needs to be moved into the `examples` directory.

examples-next/basic/schema.graphql
```diff
@@ -0,0 +1,777 @@
+enum FileMode {
```
timleslie, line 1: This file needs to be moved into the `examples` directory.

packages-next/fields/src/index.ts
```diff
@@ -12,3 +12,4 @@ export { virtual } from './types/virtual';
 export { Implementation } from './Implementation';
 export type { FieldConfigArgs, FieldExtraArgs } from './Implementation';
 export { image } from './types/image';
+export { file } from './types/file';
```
timleslie, line 15: Can we move `file` and `image` above `Implementation` so that all these exports are grouped more logically?

packages-next/fields/src/types/file/Implementation.ts
```diff
@@ -1,164 +1,94 @@
...
-        return {
-          publicUrl: this.fileAdapter.publicUrl(itemValues),
-          ...itemValues,
-        };
+  gqlAuxFieldResolvers() {
+    return {
+      FileFieldOutput: {
+        src(data: FileData, _args: any, context: KeystoneContext) {
+          if (!context.files) {
+            throw new Error('File context is undefined');
```
timleslie, line 40: Is this the error message we expect to see if we forget to include `config.files`? If so, we should probably make it more explicit about how to fix it.

gwyneplaine: Replaced with "File context is undefined, this most likely means that you havent configurd keystone with a file config, see https://next.keystonejs.com/apis/config#files for details"

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  return `${urlSafeName}_${uuid()}`;
+};
+
+const getFileMetadataFromBuffer = async (buffer: Buffer) => {
+  const filesize = buffer.length;
+  const fileType = fromBuffer(buffer);
+  if (!fileType) {
+    throw new Error('File type not found');
+  }
```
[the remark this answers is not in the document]
gwyneplaine: removed, derp moment

packages-next/types/src/config/index.ts
```diff
@@ -161,6 +162,24 @@ export type ImagesConfig = {
   };
 };
 
+// config.files
+export type FilesConfig = {
+  upload: FileMode;
+  maxSize?: number;
```
mitchellhamilton, line 168: Could you look into how the upload size limit works in graphql-upload compared with with this, I feel like we might want to align across all things that use the `Upload` scalar?

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,82 @@
...
+const generateSafeFilename = (filename: string) => {
+  //   /*
+  //     This regex lazily matches for any characters that aren't a new line
+  //     it then optionally matches the last instance of a "." symbol
+  //     followed by any alphabetical character before the end of the string
+  //    */
+  const [, name, ext] = filename.match(/^([^:\n].*?)(\.[A-Za-z]+)?$/) as RegExpMatchArray;
+
+  const id = uuid();
+  const urlSafeName = slugify(name).toLowerCase();
```
mitchellhamilton, line 23: Let's use https://github.com/sindresorhus/filenamify instead of @sindresorhus/slugify. (make the seperator `_` instead of the default though)

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,82 @@
...
+    getSrc: (mode, filename) => {
+      return `${baseUrl}/${filename}`;
+    },
+    getDataFromRef: async (ref: string) => {
+      const fileRef = parseFileRef(ref);
+      if (!fileRef) {
+        throw new Error('Invalid file reference');
+      }
+      const buffer = await fs.readFile(path.join(storagePath, `${fileRef.filename}`));
+      const metadata = await getFileMetadataFromBuffer(buffer);
```
mitchellhamilton, line 56: Can you use `fs.stat` to read the filesize rather than reading the whole thing into memory?

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,82 @@
...
+      const buffer = Buffer.concat(chunks);
+      const metadata = await getFileMetadataFromBuffer(buffer);
+      // name conflict strategy function goes here.
+      // const validFilename = resolveName({ name: filename, storagePath });
+      if (!isValidFileSize(buffer.length, maxSize)) {
+        throw new Error(`Filesize of ${buffer.length} exceeds max size of ${maxSize}`);
+      }
+
+      const pseudoSafeFilename = generateSafeFilename(filename);
+      await fs.writeFile(path.join(storagePath, pseudoSafeFilename), buffer);
```
mitchellhamilton, line 77: Can you stream the file to disk rather than collecting it in memory first? (and then probably use `fs.stat` to get the size)

gwyneplaine: done

gautamsi:
is there possibility to use `cuid` rather than `uuid`? `cuid` tends to be shorter.

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (!fileRef) {
+        throw new Error('Invalid file reference');
+      }
+      const { size: filesize } = await fs.stat(path.join(storagePath, `${fileRef.filename}`));
+      return { filesize, ...fileRef };
+    },
+    getDataFromStream: async (stream, filename) => {
+      const { upload: mode } = config;
+
+      const pseudoSafeFilename = generateSafeFilename(filename);
```
timleslie, line 52: Why is this variable called `pseudoSafeFilename`? If it's not actually safe for some reason we should capture that in a commend in the `generateSafeFilename` function explaining what might go wrong. If nothing can go wrong then we should rename this variable to `safeFilename`.

docs/pages/apis/config.mdx
```diff
@@ -167,12 +167,14 @@ Options:
 - `cors` (default: `undefined`): Allows you to configure the [cors middleware](https://www.npmjs.com/package/cors#configuration-options) for your Express server.
   If left undefined `cors` will not be used.
 - `port` (default: `3000` ): The port your Express server will listen on.
+- `maxFileSize` (default: `200 * 1024 * 1024`): The maximum allowed file size for file uploads. Defaults to 200mb.
```
timleslie, line 170: Let's break this out into a separate PR.

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,63 @@
...
+      const writeStream = fs.createWriteStream(path.join(storagePath, pseudoSafeFilename));
```
mitchellhamilton, line 53: Can you try uploading a file that is above the max file size and see what happens and then probably handle whatever error that results in?

gwyneplaine: yup, validated this yesterday we get a "payloadtoolarge" error type like so 
payload-too-large

Of course it's a nested error, so its obfuscated in the admin-ui, but that's no better or worse than what we had previously 😂. Ideally though we'd be able to append a better error message though, something like 
`file exceeds max file size specified in config, see config documentation for details`

gwyneplaine: Either way pulling this out of the PR for now, as per review with @timleslie

mitchellhamilton: This code needs to delete the file when an error occurs during the reading correctly regardless of how we configure the max file size and correctly handling that should happen in this PR imo.

gwyneplaine: > Do we delete the file once that happens?

No, we'll need to add logic to delete the file if the error occurs

docs/pages/apis/fields.mdx
```diff
@@ -520,7 +520,33 @@ File types allow you to upload different types of files to your Keystone system.
 
 ### file
 
-(coming soon)
+An `file` field represents a file of any type.
```
timleslie, line 523:
```suggestion
A `file` field represents a file of any type.
```

packages-next/types/src/config/index.ts
```diff
@@ -3,7 +3,7 @@ import { CorsOptions } from 'cors';
 import type { GraphQLSchema } from 'graphql';
 import type { Config } from 'apollo-server-express';
 
-import type { ImageMode, KeystoneContext } from '..';
+import type { ImageMode, FileMode, KeystoneContext } from '..';
```
mitchellhamilton, line 6: Probably a topic for discussion but ImageMode and FileMode are the same thing, does it makes sense for them to be the same thing and potentially configure the upload strategy in the same place for both? (this question is not intended to be answered in a GitHub comment now, just raising it for later)

packages-next/types/src/config/index.ts
```diff
@@ -143,6 +144,25 @@ export type GraphQLConfig = {
 
 export type ExtendGraphqlSchema = (schema: GraphQLSchema, keystone: BaseKeystone) => GraphQLSchema;
 
+// config.files
+
+export type FilesConfig = {
+  upload: FileMode;
+  maxSize?: number;
```
mitchellhamilton, line 151: This isn't used anymore, right?

packages-next/types/src/config/index.ts
```diff
@@ -143,6 +144,25 @@ export type GraphQLConfig = {
...
+  local?: {
+    /**
+     * The path local images are uploaded to.
+     * @default 'public/images'
+     */
+    storagePath?: string;
+    /**
+     * The base of the URL local images will be served from, outside of keystone.
+     * @default '/images'
```
mitchellhamilton, line 160: I assume?
```suggestion
     * @default '/files'
```

packages-next/types/src/config/index.ts
```diff
@@ -143,6 +144,25 @@ export type GraphQLConfig = {
...
+     */
+    baseUrl?: string;
```
mitchellhamilton, line 162: These comments need to be updated to the appropriate things for files.

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,80 @@
...
+        writeStream.write(chunk);
+      }
+
+      try {
+        await observeStreamErrors;
+        const { size: filesize } = await fs.stat(path.join(storagePath, safeFilename));
+        return { mode, filesize, filename: safeFilename };
+      } catch (e) {
+        fs.removeSync(path.join(storagePath, safeFilename));
+        throw new Error(e);
```
mitchellhamilton, line 76: Why are you wrapping the error in a new error? that will just stringify it and therefore potentially lose information?

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,80 @@
...
+      for await (let chunk of stream) {
+        writeStream.write(chunk);
+      }
+
+      try {
+        await observeStreamErrors;
+        const { size: filesize } = await fs.stat(path.join(storagePath, safeFilename));
+        return { mode, filesize, filename: safeFilename };
+      } catch (e) {
+        fs.removeSync(path.join(storagePath, safeFilename));
```
mitchellhamilton, line 75:
```suggestion
        await fs.remove(path.join(storagePath, safeFilename));
```

packages-next/admin-ui/src/system/generateAdminUI.ts
```diff
@@ -52,15 +52,24 @@ export const generateAdminUI = async (
...
   if (config.images) {
-    const publicDirectory = Path.join(projectAdminPath, 'public');
     await fs.mkdir(publicDirectory, { recursive: true });
     const storagePath = Path.resolve(config.images.local?.storagePath ?? './public/images');
     await fs.mkdir(storagePath, { recursive: true });
     await fs.symlink(storagePath, Path.join(publicDirectory, 'images'), 'junction');
   }
 
+  if (config.files) {
+    if (!fs.existsSync(Path.join(projectAdminPath, 'public'))) {
```
mitchellhamilton, line 65: Checking if something exists in the file system and then immediately reading it or etc. is generally bad because it might might at the time of the exists call and not at a read or write or something (not a concern here really but more generally so it's best to avoid it) and mkdir will not error if it already exists so the exists check is just adding another file system operation, could you change this so there is one call to mkdir like 

```ts
if (config.images || config.files) {
      await fs.mkdir(publicDirectory, { recursive: true });
}
```

gwyneplaine: ah good to know, will do thxx Mitchell 🎉

packages-next/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -0,0 +1,80 @@
...
+  return {
+    getSrc: (mode, filename) => {
+      return `${baseUrl}/${filename}`;
+    },
+    getDataFromRef: async (ref: string) => {
+      const fileRef = parseFileRef(ref);
+      if (!fileRef) {
+        throw new Error('Invalid file reference');
+      }
+      const { size: filesize } = await fs.stat(path.join(storagePath, `${fileRef.filename}`));
```
mitchellhamilton, line 46:
```suggestion
      const { size: filesize } = await fs.stat(path.join(storagePath, fileRef.filename));
```

emmatown (changes requested):
I personally have no more code comments that I can see but would like to explicitly block this on going through it as a group

packages-next/types/src/config/index.ts
JedWatson: We think they may diverge in the future so we'll leave them separate

packages/utils/src/index.ts
```diff
@@ -258,17 +258,31 @@ export const humanize = (str: string) => {
     .join(' ');
 };
 
-const REFREGEX = /^(local):([^:\n]+)\.(gif|jpg|png|webp)$/;
+const IMAGEREGEX = /^(local):([^:\n]+)\.(gif|jpg|png|webp)$/;
+const FILEREGEX = /^(local):file:([^:\n]+)/;
```
mitchellhamilton, line 262: This needs to disallow path separators

emmatown (approved):

timleslie (changes requested):
Pending

gwyneplaine:
> Pending

Explicitly thrown errors are now crashing the keystone process, as opposed to being bubbled up into displayed errors in the admin-ui. This PR is blocked on resolving this issue.

gwyneplaine:
> > Pending
> 
> Explicitly thrown errors are now crashing the keystone process, as opposed to being bubbled up into displayed errors in the admin-ui. This PR is blocked on resolving this issue.

Tim was unable to replicate this on his machine, we're releasing this now, if this issue persists we can resolve post release

timleslie (approved):
🚢

PR #5538, opened 2021-04-26
DB item API docs

timleslie:
Add docs for the DB item API at `context.db.lists`.

### DIFF
--- a/docs/pages/apis/db-items.mdx
+++ b/docs/pages/apis/db-items.mdx
@@ -3,7 +3,7 @@ import { Markdown } from '../../components/Page';
 # Database Items API
 
 The database items API provides a programmatic API for running CRUD operations against the internal GraphQL resolvers in your system.
-Importantly, this API bypasses the Apollo server itself, instead invoking the resolver functions directly.
+Importantly, this API bypasses the GraphQL Server itself, instead invoking the resolver functions directly.
 The return values of this API are **internal item** objects, which are suitable to be returned from GraphQL resolvers.
 
 Refer to the [internal items guide](../guides/internal-items) for details on how to work with internal items in Keystone.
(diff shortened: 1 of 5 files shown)


docs/pages/apis/db-items.mdx
```diff
@@ -1,8 +1,149 @@
 import { Markdown } from '../../components/Page';
-import { ComingSoon } from '../../components/ComingSoon';
 
 # Database Items API
 
-<ComingSoon/>
+The database items API provides a programmatic API for running CRUD operations against the internal GraphQL resolvers in your system.
+Importantly, this API bypasses the Apollo server itself, instead invoking the resolver functions directly.
```
JedWatson, line 6:
```suggestion
Importantly, this API bypasses the GraphQL Server itself, instead invoking the resolver functions directly.
```

docs/pages/apis/db-items.mdx
```diff
@@ -1,8 +1,149 @@
...
+  createOne({ data }),
+  createMany({ data }),
+  updateOne({ id, data }),
+  updateMany({ data }),
+  deleteOne({ id }),
+  deleteMany({ ids }),
+}
+```
+
+The arguments to these functions closely correspond to their equivalent GraphQL APIs, making it easy to switch between the programmatic API and the GraphQL API.
```
JedWatson, line 30: This paragraph feels left over from the List Item API docs. The DB Item API is really not a thing you'd switch out directly for the GraphQL API, you could "between the List Items API and the DB Items API" maybe?

JedWatson (changes requested):
Close

If you find a place / way to word it, would be good to address where the inputs come from (i.e that the arguments come from the GraphQL API, and match what is documented there)

JedWatson (approved):

PR #5542, opened 2021-04-26
Replace createItem with createOne

timleslie:
See #5485


JedWatson (approved):
Need to fix the lint error for import rules but otherwise 👍

PR #5552, opened 2021-04-26
Use the GraphQL API for lists.<list>.count

timleslie:
This makes the `.count()` method consistent with the rest of the API, which execute GraphQL queries via `context.graphql.run` to get their results.

### DIFF
--- a/.changeset/modern-pears-exist.md
+++ b/.changeset/modern-pears-exist.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Update `context.lists.<list>.count()` to use the GraphQL API rather than directly calling the resolver.
--- a/.changeset/moody-oranges-shout.md
+++ b/.changeset/moody-oranges-shout.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/adapter-prisma-legacy': patch
+---
+
+Improved handling of null filter inputs.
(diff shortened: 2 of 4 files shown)


emmatown (changes requested):
Don't conditionally add the variables to the graphql query, if the variable values are null or undefined, graphql will handle it

timleslie:
> Don't conditionally add the variables to the graphql query, if the variable values are null or undefined, graphql will handle it

See, that's what I thought, but in practice it barfed out, so I had to add this big pile of conditionals :-(

mitchellhamilton:
What was the error?

timleslie:
It complained about not being able to pass undefined to a field of type `Int!`. I'll see if I can recreate it.

mitchellhamilton:
You should make the type in the variable `Int`, not `Int!` then

timleslie:
```
  console.log
    {
      query: 'query ($first: Int!, $skip: Int!, $where: UserWhereInput) { _allUsersMeta(first: $first, skip: $skip, where: $where) { count }  }',
      variables: { first: undefined, skip: undefined, where: undefined }
    }

    GraphQLError: Variable "$first" got invalid value undefined; Expected non-nullable type "Int!" not to be null.
```

timleslie:
🤦 I'm glad someone's awake this morning.

emmatown (approved):

PR #5562, opened 2021-04-28
Remove unused enableSessionItem

timleslie:
AFAICT this feature isn't used. I'm not sure if there was an intent to keep it around for an upcoming feature, but I suspect it's legacy and should go away.

### DIFF
--- a/.changeset/silent-dogs-run.md
+++ b/.changeset/silent-dogs-run.md
@@ -1,0 +1,7 @@
+---
+'@keystone-next/admin-ui': major
+'@keystone-next/auth': major
+'@keystone-next/types': major
+---
+
+Removed the `enableSessionItem` value from `config.ui` and the GraphQL type `KeystoneAdminMeta`.
--- a/docs/pages/apis/config.mdx
+++ b/docs/pages/apis/config.mdx
@@ -144,7 +144,6 @@ export default config({
     isDisabled: false,
     isAccessAllowed: async context => true,
     // Optional advanced configuration
-    enableSessionItem: true,
     publicPages: [],
     getAdditionalFiles: async () => [],
     pageMiddleware: async () = {},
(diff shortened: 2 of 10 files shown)


molomby:
I think we want that left in. @JedWatson?

JedWatson:
ummm yes. I'm surprised it doesn't seem to be referenced from the React code in the Admin UI, it's the thing that tells the Admin UI to actually show the session item UI (i.e the current name / signout / etc that appears above the nav when session is enabled)

are we just checking for session there at the moment? if so, not safe to assume, and we should fix that instead of removing this

cc/ @mitchellhamilton in case you know more (or we ended up doing this differently)

mitchellhamilton:
It looks like I skipped checking enableSessionItem and instead determined whether session items were being used based on the existence of an `authenticatedItem` field on the query type.
https://github.com/keystonejs/keystone/blob/4bc07cdfbaf434b3158c4d834d1dde78d108e103/packages-next/admin-ui/src/templates/app.ts#L97

That wasn't an intentional thing by me but it raises some questions:
- Do we want to make it possible for people to have a field named `authenticatedItem` on the query type that doesn't conform to what the Admin UI expects? If yes, we need `enableSessionItem` and we should use it correctly.
- If we don't care users being able to have `authenticatedItem` field on the query type that doesn't conform to what the Admin UI expects, would we still like an explicit opt-in to the Admin UI session item things?

and related: it looks like enableSignout is also not used correctly, we don't show a sign out button the main page if enableSessionItem is not enabled. Also, enableSessionItem is kind of the wrong name, enableAuthenticatedItem would make more sense since it's about having an `authenticatedItem` field on the query type, not about sessions.

JedWatson:
Our decision (as discussed with @timleslie and @mitchellhamilton) is that we implement `enableSessionItem` correctly, so lets close this and do that

PR #5563, opened 2021-04-28
Update pageMiddleware args

timleslie:
Trying to document this config option made it clear we had the wrong arguments. Passing in an actual `context` argument makes much more sense than `req, session, createContext`.


emmatown (changes requested):
Same as #5565

PR #5565, opened 2021-04-28
[RFC] Rename pageMiddleware to redirects

timleslie:
The config option `config.ui.pageMiddleware` is something of a misnomer, which is making it difficult to document. It's not a middleware in the traditional sense, it's more of a function which can indicate the need for a redirect, which in turn will be consumed by the actual middleware to perform the redirect.

Furthermore, the current use case of the `withAuth` decorator makes it impossible for user defined config to conditionally disable the redirect . For example, if the user wanted to always skip the auth redirects when setting up custom pages via `getAdditionalFiles` and `publicPages`, the current API doesn't allow for it.

The modified API allows a return value of `{ never: true }`, which indicates that no further redirects should be considered.

It also changes the signature to an array of functions rather than a single function, which means we don't need to handle function composition when writing helper wrappers like `withAuth`.

### DIFF
--- a/.changeset/twelve-sloths-yell.md
+++ b/.changeset/twelve-sloths-yell.md
@@ -1,0 +1,7 @@
+---
+'@keystone-next/admin-ui': major
+'@keystone-next/auth': major
+'@keystone-next/types': major
+---
+
+Updated the arguments of `config.ui.pageMiddleware()` to accept `{ isValidSession: boolean; context: KeystoneContext; }`.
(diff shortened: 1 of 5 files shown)


mitchellhamilton (changes requested):
I don't think we should do this because we're likely to make another change to how this works some time soon to support statically hosting the admin ui so changing this would just create churn

JedWatson:
Leaving this for discussion later, there's more to it than meets the eye and we're not focusing on this area right now

PR #5569, opened 2021-04-29
Implement db lists API in terms of GraphQLSchema

emmatown:
There aren't any functional changes here, this just makes the new fields stuff easier


timleslie (approved):
Looks good 👍

PR #5571, opened 2021-04-29
Explicitly kill the prisma process before exit

### LINKED ISSUE #5477: Prisma client instance keep active in background when Keystone crashes
# Bug report

I catch the problem on debugging keystone app, that postgresql connections limit is full after some time of restarting keystone.
The source of problem is that when keystone-next crashes, the prisma client process stays active, so if keystone crashes 10 times, we will see 10 active prisma client processes.

## To Reproduce

1. Add the simple `onConnect` function that throws the exception in `keystone.ts`:
```js
export default auth.withAuth(
  config({
    db: process.env.DATABASE_URL
      ? { adapter: 'prisma_postgresql', url: process.env.DATABASE_URL, onConnect: (context) => {
          console.log('commected');
          throw 'test'
      } }
      : {
          adapter: 'mongoose',
          url: 'mongodb://localhost/keystone-examples-next-basic',
        },
```
2. Starts the `keystone-next` in dev mode 10 times with waiting when it throws the error and exits.
3. List the active prisma client process instances in memory - you will see about 10 active processes, that eats postgresql connections:
```
$ ps ax | grep -P '\.?prisma\/.+/query-engine' 
4075251 pts/0    Sl     0:00 /app/src/keystone/examples-next/basic/node_modules/.prisma/client/query-engine-debian-openssl-1.1.x --enable-raw-queries --port 44697
4075603 pts/0    Sl     0:00 /app/src/keystone/examples-next/basic/node_modules/.prisma/client/query-engine-debian-openssl-1.1.x --enable-raw-queries --port 42897
```
List the active `keystone-next` instances and see no one:
```
$ ps ax | grep '[k]eystone-next'
```
Also, if I stop the `keystone-next` process via <kbd>Ctrl+C</kbd> - prisma/client/query-engine exits well.

## Expected behaviour

Prisma client instance must die together with keystone-next.

timleslie:
Calling `process.exit()` was terminating the server process without triggering the appropriate cleanup of the child process spawned by the Prisma client. It's not clear why exactly Prisma isn't able to clean itself up, but sending a `SIGINT` signal in the `beforeExit` handler seems to solve the problem.

Fixes #5477.

### DIFF
--- a/.changeset/large-kiwis-smoke.md
+++ b/.changeset/large-kiwis-smoke.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/adapter-prisma-legacy': patch
+---
+
+Added a `beforeExit` handler to explicitly terminate the prisma child process to avoid zombie processes when the server crashes.
--- a/packages/adapter-prisma/src/adapter-prisma.ts
+++ b/packages/adapter-prisma/src/adapter-prisma.ts
@@ -91,6 +91,13 @@ class PrismaAdapter {
       log: this.enableLogging && ['query'],
       datasources: { [this.provider]: { url: this.url } },
     });
+    this.prisma.$on('beforeExit', async () => {
+      // Prisma is failing to properly clean up its child processes
+      // https://github.com/keystonejs/keystone/issues/5477
+      // We explicitly send a SIGINT signal to the prisma child process on exit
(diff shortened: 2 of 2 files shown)


molomby (approved):
Does seem like there should be a better way but fine for now

PR #5573, opened 2021-04-29
API Config docs

timleslie:
Updates to the API docs for both completeness and consistency.


docs/pages/apis/fields.mdx
```diff
@@ -65,11 +65,19 @@ Options:
   See the [Hooks API](./hooks) for full details on the available hook options.
 - `label`: The label displayed for this field in the Admin UI. Defaults to a human readable version of the field name.
 - `ui`: Controls how the field is displayed in the Admin UI.
-  - `views`: (coming soon)
-  - `description`: (coming soon)
-  - `createView`: (coming soon)
-  - `listView`: (coming soon)
-  - `itemView`: (coming soon)
+  - `views`: A [resolved](https://nodejs.org/api/modules.html#modules_require_resolve_request_options) path to a module containing code to replace or extend the default Admin UI components for this field. See the [Custom Field Views](../guides/custom-field-views) guide for details on how to use this option.
```
JedWatson, line 68: Are we 100% this has to be a resolved path? I remember talking about it with @mitchellhamilton for a while but just wanted to check this is where we landed

(e.g we could make it a relative path from the project root and resolve it ourselves, but I'm not sure if there were concerns with doing that)

docs/pages/apis/schema.mdx
```diff
@@ -144,18 +144,18 @@ export default config({
...
         createView: {
           defaultFieldMode: ({ session }) => 'edit',
         },
         itemView: {
           defaultFieldMode: ({ session, item }) => 'edit',
         },
+        listView: {
+          defaultFieldMode: ({ session }) => 'read',
+          initialColumns: ['name', /* ... */],
+          initialSort: { field: 'name', direction: 'ASC' },
```
JedWatson, line 156: unrelated to this pr, @mitchellhamilton can you take consistency here into account with the graphql spec you're looking at right now?

JedWatson (changes requested):
Looks great @timleslie 

Only change actually _needed_ to unblock this is the callout for `getAdditionalFiles` re: why it's there

docs/pages/apis/fields.mdx
mitchellhamilton: It doesn't necessarily have to be resolved fully(resolved in the sense that it points directly to a file). The thing that we require(or at least should require IMO) but it can't be relative which basically means `some-package` is bad (or at least should be) since it's relative, `/path/to/project/node_modules/some-package` is okay, note the second one points to a directory, not a file so that it can load different things for browsers, ESM, CJS, etc.

timleslie: So do we want to change anything here? If best practice is to provide a `resolved` path then let's just tell people to do that rather than trying to explain a laundry list of edge cases.

mitchellhamilton: No, we shouldn't change it.

JedWatson (approved):

PR #5578, opened 2021-04-29
Replace mode on GraphQL output type of image field with an interface

emmatown:
This means we can add different implementations of the `ImageFieldOutput` with potentially extra fields.


timleslie (approved):
👍

PR #5580, opened 2021-04-30
Understanding relationships -> relationships

JedWatson:
I think this is clearer and simpler, you both OK with this @raveling and @timleslie?

(didn't feel the need to add a redirect since this has been live for about 5 minutes)


raveling (approved):
LGTM!

PR #5586, opened 2021-05-02
fix custom views setter issue

gautamsi:
found issue when trying to use custom views, it gives this error `Cannot set property Field of #<Object> which has only a getter`

image

I was able to fix this using this change, not sure if this is best way to do this. I could not find any other way.


packages-next/admin-ui/src/utils/useAdminMeta.tsx
```diff
@@ -93,7 +93,7 @@ export function useAdminMeta(adminMetaHash: string, fieldViews: FieldViews) {
             );
           }
         });
-        const views = fieldViews[field.viewsIndex];
+        const views = { ...fieldViews[field.viewsIndex] };
```
gautamsi, line 96: I could not trace why this `views`  has Field as getter only, best way would be to make sure this allows setter, but this is quick fix.

timleslie: Can you show an example of how you defined your custom field views? I'd like to try to recreate this so we can get to the root cause of the bug.

JedWatson (approved):
Turns out this is happening because the module exports are special, and we need to turn it into a plain object to overwrite the built-in views with custom ones.

This fix may not be the most elegant solution (cc/ @mitchellhamilton) but verified it works for now.

Glad you picked it up @gautamsi and thanks for the PR! the fix will go out with the release today.

I've added back in an example (otherwise commented out) so this can be more easily tested, and will open a separate PR later with an actual example project demoing and testing this functionality.

PR #5592, opened 2021-05-03
Fix file stream

gwyneplaine:
Don't manually pipe your stream like an idiot. 
Use stream.pipeline instead. 

Node 15 has a native promisified version of this function, but since we're not on 15, doing it manually.

### DIFF
--- a/.changeset/quick-shirts-cry.md
+++ b/.changeset/quick-shirts-cry.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fix uncaught exception in file stream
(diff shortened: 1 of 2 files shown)


.changeset/quick-shirts-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fix uncaught exception in file stream
```
gwyneplaine, line 5: `
```suggestion
Fixed uncaught exception in file stream.
```

timleslie (approved):
Nice fix!

PR #5607, opened 2021-05-04
Update Prisma/MongoDB Connector to Early Access

ajcwebdev:
The MongoDB connector for Prisma is still under development and classified as "Early Access." But unlike the [last year and half of it being under discussion](https://github.com/prisma/prisma/issues/1277), today it is actually live and can be used by anyone, so it's not really accurate to say it's still in the "planning" phase.

* [MongoDB provider for Prisma 2 and Keystone Next](https://gautamsi.medium.com/mongodb-provider-for-prisma-2-and-keystone-next-1618418c8937)
* [Can I use MongoDB with Prisma Yet?](https://dev.to/ajcwebdev/can-i-use-mongodb-with-prisma-yet-50go)
* [RedwoodJS Example Blog with MongoDB](https://redwoodjs-prisma-mongo.netlify.app/)

For explanation of Early Access versus General Availability see [Releases and maturity levels](https://www.prisma.io/docs/about/releases) from the Prisma docs.

### DIFF
--- a/docs/pages/guides/keystone-5-vs-keystone-next.mdx
+++ b/docs/pages/guides/keystone-5-vs-keystone-next.mdx
@@ -22,7 +22,7 @@ There are a few major things that will change between 5 and 6 which we’re curr
 
 ### DB adapters out. Prisma in.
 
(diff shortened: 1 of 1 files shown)


timleslie (approved):
Yep, this is a good update on the state of Mongo 👍

PR #5612, opened 2021-05-04
withAuth example

timleslie:
This adds an example of the most basic usage of the `withAuth` function from the `auth` package.

### DIFF
--- a/examples/with-auth/README.md
+++ b/examples/with-auth/README.md
@@ -1,0 +1,82 @@
+## Feature Example - Authentication
+
+This project demonstrates how to add password based authentication to your Keystone system.
+It builds on the [Task Manager](../todo) starter project.
+
+## Instructions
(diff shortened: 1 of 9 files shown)


examples/with-auth/README.md
```diff
@@ -0,0 +1,82 @@
...
+  listKey: 'Person',
+  identityField: 'email',
+  secretField: 'password',
+  initFirstItem: { fields: ['name', 'email', 'password'] },
+});
+```
+
+### Session
+
+We use a basic `statelessSessions` from `@keystone-next/keystone/session` for session handling.
```
gwyneplaine, line 58: @timleslie  are there existing docs we can link to here that explains what `statelessSession` does?

gwyneplaine (approved):
lgtm

PR #5613, opened 2021-05-05
Improve statelessSessions types

timleslie:
The type parameter of this function needs to be more specific to work with `withItemData`.

### DIFF
--- a/packages-next/keystone/src/session/index.ts
+++ b/packages-next/keystone/src/session/index.ts
@@ -74,10 +74,10 @@ type FieldSelections = {
   - [ ] We could support additional where input to validate item sessions (e.g an isEnabled boolean)
 */
 
-export function withItemData<U extends Record<string, any>>(
-  createSession: () => SessionStrategy<U>,
+export function withItemData(
+  createSession: () => SessionStrategy<Record<string, any>>,
   fieldSelections: FieldSelections = {}
-): () => SessionStrategy<U & { listKey: string; itemId: string; data: any }> {
+): () => SessionStrategy<{ listKey: string; itemId: string; data: any }> {
   return (): SessionStrategy<any> => {
     const { get, ...sessionStrategy } = createSession();
     return {
@@ -115,7 +115,7 @@ export function withItemData<U extends Record<string, any>>(
   };
 }
 
-export function statelessSessions<T extends Record<string, any>>({
+export function statelessSessions<T>({
   secret,
   maxAge = MAX_AGE,
   path = '/',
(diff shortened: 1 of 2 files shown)


emmatown (changes requested):
This isn't right, statelessSessions should not require listKey and itemId

timleslie:
Ah yeah, you're right, I'm going to make a different change which will fix the type errors I'm getting. Will change this PR.

packages-next/keystone/src/session/index.ts
```diff
@@ -115,7 +115,7 @@ export function withItemData<T extends { listKey: string; itemId: string }>(
   };
 }
 
-export function statelessSessions<T>({
+export function statelessSessions<T extends Record<string, any>>({
```
mitchellhamilton, line 118: This restriction seems unnecessary. Iron supports anything that can be json stringified.

packages-next/keystone/src/session/index.ts
```diff
@@ -74,10 +74,10 @@ type FieldSelections = {
   - [ ] We could support additional where input to validate item sessions (e.g an isEnabled boolean)
 */
 
-export function withItemData<T extends { listKey: string; itemId: string }>(
-  createSession: () => SessionStrategy<T>,
+export function withItemData<U extends Record<string, any>>(
+  createSession: () => SessionStrategy<U>,
   fieldSelections: FieldSelections = {}
-): () => SessionStrategy<T & { data: any }> {
+): () => SessionStrategy<U & { listKey: string; itemId: string; data: any }> {
```
mitchellhamilton, line 80: This function being generic seems unnecessary. You could replace U and and the `U & { ... }` with passing in some type which is `{ listKey: string; itemId: string; data: any }`

packages-next/keystone/src/session/index.ts
timleslie: Right, but `withItemData` can't use it unless it's an object.

packages-next/keystone/src/session/index.ts
timleslie: I'm not sure what you mean by this 🤔

packages-next/keystone/src/session/index.ts
mitchellhamilton: Sure withItemData requires an object, why should statelessSessions require it though?

packages-next/keystone/src/session/index.ts
mitchellhamilton: I'm saying remove the `<U extends Record<string, any>>` generic

emmatown (approved):

PR #5615, opened 2021-05-05
Update Document Fields Docs to resolve error

MichaelZaporozhets:
the views key should be nested within an object on the `ui` key for the document config


timleslie (approved):

PR #5616, opened 2021-05-05
Add initial support for isIndexed

timleslie:
Add an `isIndexed` config option to the `text`, `integer`, `float`, `select`, and `timestamp` field types.

### DIFF
--- a/packages-next/fields/src/Implementation.ts
+++ b/packages-next/fields/src/Implementation.ts
@@ -31,6 +31,7 @@ class Field<P extends string> {
   isRequired: boolean;
   defaultValue?: any;
   isOrderable: boolean;
+  isIndexed: boolean;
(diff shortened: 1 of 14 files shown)


mitchellhamilton (approved):

packages-next/fields/src/Implementation.ts
```diff
@@ -58,6 +60,7 @@ class Field<P extends string> {
     this.schemaDoc = schemaDoc;
     this.config = config;
     this.isRequired = !!isRequired;
+    this.isIndexed = !!isIndexed;
```
mitchellhamilton, line 63: Aren't you reading isIndexed from the adapter, not the implementation class? So this is doing nothing? but it still works because isIndexed is already set on the adapter class?

PR #5618, opened 2021-05-05
 Add placeholder to document editor block menu

gwyneplaine:
Description: 
Add a placeholder to the document editor `insertBlockMenu` component, previously this would display an empty inline dialog. Now we have an empty state placeholder. Imo this is better than not rendering the toolbar item at all, since it makes it obvious that this functionality exists, rather than hiding it completely, but I'm non-committal on this.

--- EDIT ---
Updated now to omit the toolbar item completely if no block items or relationship items exist. 

Before: 
Screen Shot 2021-05-05 at 2 35 26 pm

After: 
Screen Shot 2021-05-05 at 3 35 42 pm

### DIFF
--- a/packages-next/fields-document/src/DocumentEditor/Toolbar.tsx
+++ b/packages-next/fields-document/src/DocumentEditor/Toolbar.tsx
@@ -1,6 +1,14 @@
 /** @jsx jsx */
 
-import { Fragment, ReactNode, forwardRef, useState, HTMLAttributes, useMemo } from 'react';
+import {
+  Fragment,
+  ReactNode,
+  forwardRef,
+  useState,
+  HTMLAttributes,
+  useMemo,
+  useContext,
+} from 'react';
 import { Editor, Transforms } from 'slate';
 import { applyRefs } from 'apply-ref';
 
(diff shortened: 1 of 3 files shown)


mitchellhamilton:
I feel like hiding it completely would be better. The people who will be primarily using this (content editors) are not the people who will action the fact that there aren't component blocks or relationships. Hiding it would also align with the way disabling other things work in the editor.

packages-next/fields-document/src/DocumentEditor/Toolbar.tsx
```diff
@@ -44,6 +52,9 @@ export function Toolbar({
   viewState?: { expanded: boolean; toggle: () => void };
 }) {
   const ExpandIcon = viewState?.expanded ? Minimize2Icon : Maximize2Icon;
+  const relationship = useContext(DocumentFieldRelationshipsContext);
+  const blockComponent = useContext(ComponentBlockContext);
+  const hasBlockItems = Object.entries(relationship).length || Object.keys(blockComponent).length;
```
gwyneplaine, line 57: this is kinda ick, since we're reaching into context and running object.entries and object.keys here *just* to check for items. 

If there's a better way I'm here for it

gautamsi:
@mitchellhamilton Would this be more helpful to have the button there for dev build and hide it in production. We can have a dropdown with placeholder text in dev build to remind devs that they can add custom blocks.

I have not checked, would I be able to add a top level menu item for custom component block?

packages-next/fields-document/src/DocumentEditor/Toolbar.tsx
mitchellhamilton: This isn't really a big deal, the things in those contexts won't change at runtime while a user is using the editor so getting them isn't a problem.

emmatown (commented):
Needs a changeset

emmatown (approved):

PR #5622, opened 2021-05-05
Disable sorting on many relationships

timleslie:
Sorting a list by a many relationship doesn't have a meaningful interpretation. In practice, attempting to do this would return errors. This change prevents the Admin UI from throwing errors from the list view when a user attempts to sort by a man relationship.


gwyneplaine (approved):
lgtm

PR #5631, opened 2021-05-06
5604/fix query caching

### LINKED ISSUE #5604: local-storage query caching in admin-ui not working as intended
# Bug report

## Describe the bug

Currently the useQueryParamsFromLocalStorage method doesn't update cached values correctly.

## To Reproduce

- Start up a keystone-next example 
- populate a list page with a few entries 
- Change the pageSize query once (this will work fine)
- Change the pageSize query again, this will be automatically replaced by the effect.

## Expected behaviour: 

only retrieve cached queries when no queries are specified
don't cache page or pageSize params

## Actual behaviour: 

Initial query will be saved, all subsequent query changes in the url will be replaced. 
caches page and pageSize params

gwyneplaine:
Fixes #5604: 
* pageSize query no longer cached
* page query no longer cached 
* sortBy query now cached 
* field filters remain cached 
* column queries remain cached 
* caching effect now checks for whether the router is ready or not. 
* Added a 'reset to defaults' button**

This needs better design and thought, but gets the job done for now. 

https://user-images.githubusercontent.com/12119389/117245314-ff29b300-ae7d-11eb-8051-e82cc351dbbb.mov

### DIFF
--- a/packages-next/admin-ui/src/pages/ListPage/index.tsx
+++ b/packages-next/admin-ui/src/pages/ListPage/index.tsx
@@ -239,7 +239,6 @@ const ListPage = ({ listKey }: ListPageProps) => {
 
   const theme = useTheme();
   const showCreate = !(metaQuery.data?.keystone.adminMeta.list?.hideCreate ?? true) || null;
-  console.log(query, Object.keys(query).length);
 
   return (
     <PageContainer header={<ListPageHeader listKey={listKey} />}>
--- a/packages-next/admin-ui/src/pages/ListPage/index.tsx
+++ b/packages-next/admin-ui/src/pages/ListPage/index.tsx
@@ -77,7 +77,7 @@ let listMetaGraphqlQuery: TypedDocumentNode<
 function useQueryParamsFromLocalStorage(listKey: string) {
   const router = useRouter();
   const localStorageKey = `keystone.list.${listKey}.list.page.info`;
-  const cacheableQueries = ['sortBy', 'fields'];
+  const storeableQueries = ['sortBy', 'fields'];
 
   const resetToDefaults = () => {
     localStorage.removeItem(`keystone.list.${listKey}.list.page.info`);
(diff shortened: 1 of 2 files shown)


packages-next/admin-ui/src/pages/ListPage/index.tsx
```diff
@@ -77,11 +77,23 @@ let listMetaGraphqlQuery: TypedDocumentNode<
 function useQueryParamsFromLocalStorage(listKey: string) {
   const router = useRouter();
   const localStorageKey = `keystone.list.${listKey}.list.page.info`;
+  const cacheableQueries = ['sortBy', 'fields'];
```
mitchellhamilton, line 80: Can we not call this "cacheable"? When I think of a cache, I think storing things so we don't have to compute them again, not storing user stuff to help them

packages-next/admin-ui/src/pages/ListPage/index.tsx
```diff
@@ -77,11 +77,23 @@ let listMetaGraphqlQuery: TypedDocumentNode<
...
   useEffect(() => {
-    let hasSomeQueryParamsWhichAreAboutListPage = Object.keys(router.query).some(
-      x => x.startsWith('!') || x === 'page' || x === 'pageSize' || x === 'fields'
-    );
-    if (!hasSomeQueryParamsWhichAreAboutListPage) {
+    let hasSomeQueryParamsWhichAreAboutListPage = Object.keys(router.query).some(x => {
+      return x.startsWith('!') || cacheableQueries.includes(x);
+    });
+
+    if (!hasSomeQueryParamsWhichAreAboutListPage && router.isReady) {
```
mitchellhamilton, line 96: What does router.isReady do here?

gwyneplaine: as per the docs https://nextjs.org/docs/api-reference/next/router#router-object
it lets us know when the router fields are updated client side. 

The bug we were experience was due to the fact that the !hasSomeQueryParams... check was being run on an empty query object, despite them being present on the path when we run the effect. Most likely due to the fact that the check was happening before the router object had been updated

emmatown (approved):

PR #5633, opened 2021-05-06
Some fixes for the filters 

gwyneplaine:
- We now auto focus the first input in the dialog as per https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal
- We now also update the filter on 'enter' keydown.

### DIFF
--- a/packages-next/admin-ui/src/pages/ListPage/FilterAdd.tsx
+++ b/packages-next/admin-ui/src/pages/ListPage/FilterAdd.tsx
@@ -123,11 +123,6 @@ function FilterAddPopoverContent({ onClose, listKey }: { onClose: () => void; li
     });
     return filtersByFieldThenType;
   }, [router.query, fieldsWithFilters]);
-  // useEffect(() => {
-  //   const listener = () => console.log('SOME KEY IS PRESSED');
-  //   document.addEventListener('keydown', listener);
-  //   return document.removeEventListener('keydown', listener);
-  // });
   const [state, setState] = useState<State>({ kind: 'selecting-field' });
 
   const updateFilters = () => {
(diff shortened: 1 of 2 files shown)


gautamsi:
We were missing a filter `is_null: true ` for all nullable field.

Can you add it if not already.
@timleslie @JedWatson

JedWatson:
@gautamsi that's on our list, we're in the middle of a big change to fields that I've just written up here: #5635

As part of that, we're going to be updating filters across all our field types

gautamsi:
I see big changes, thanks for heads up

packages-next/admin-ui/src/pages/ListPage/FilterAdd.tsx
```diff
@@ -125,22 +125,32 @@ function FilterAddPopoverContent({ onClose, listKey }: { onClose: () => void; li
...
+      onClose();
+    }
+  };
+
   return (
     <Stack
       padding="medium"
       as="form"
+      onKeyDown={(event: React.KeyboardEvent<HTMLFormElement>) => {
+        if (event.code === 'Enter') {
```
mitchellhamilton, line 145: Why does an enter not trigger the submit event?

emmatown (commented):
Needs a changeset

timleslie (approved):
This all looks good 👍

packages-next/admin-ui/src/pages/ListPage/FilterAdd.tsx
gwyneplaine: Ah... I thought it was react-select hijinx, but turns out its #justbuttonthings. 
Button didn't have an explicit type, so its hijacking form events in the onClick (which is also triggered by an enter key press) 

Fixed and pushed

emmatown (approved):

PR #5641, opened 2021-05-07
fixed tokens

dominikwilkowski:
- applied type scale everywhere (prose)
- added all tokens to ds page
- separated `TYPE` tokens into `TYPE` and `TYPESCALE`


bladey:
This looks good.

bladey (approved):

PR #5644, opened 2021-05-08
Give public access to the BaseAccessArgs type

DreamplaySE:
Not only is BaseAccessArgs a very useful thing to have access to but it's already used in one of the example projects (ecommerce) where it's rewritten in a types document. Exporting this directly not only allows for easier creation of base authentication methods but also makes sure that those that do don't have to copy the existing code and might miss out on a change in the original type.


timleslie:
Hi @DreamplaySE, thanks for the PR 🙏 . At the moment I don't see a specific use case where this type might be wanted. If this is still something you need please open a new PR with some more details and an example use case where this type is needed. Thanks!

PR #5649, opened 2021-05-08
export more ui components required to create full admin ui page

gautamsi:
I was struggling to customize the pages and could not find a way to add hooks in the pages. 

I tried to copy the default admin-ui pages and reconstruct the page for one list. I found there are several exports missing. 

This PR adds those missing exports which are needed to re-create most of the UI with customization. I also made little changes to copied files in basic demo, which should go away when it is decided to merge.

If this is merged (only the required part and not example), I can  re-compose the pages (Home/Item/List) from the base code and then add ui hooks which I need.

### A big reason to use customized admin-ui is to overcome the limitation of Document Editor not able to render outside admin-ui properly.

Ideally the page should be composed like below, this will allow simple customizations easy by easy page composition. Page state, filters etc with data and errors must flow through the `ListProvider` and `ItemProvider` context.

```jsx
<PageContainer>
    {metaQuery.error ? (
        <ErrorSummary error={metaQuery.error} />
    ) : (
        <Fragment>
            <PageHeader />
            <FilterSummary />
            <Pagination />
            <ListTable />
        </Fragment>
    )
</PageContainer>
```

I also found this warning when I create custom page which replaces the default List and Item pages.

```log
warn  - Duplicate page detected. pages\index.tsx and pages\index.js both resolve to /index.
warn  - Duplicate page detected. pages\posts\index.tsx and pages\posts\index.js both resolve to /posts/index.
warn  - Duplicate page detected. pages\posts\[id].tsx and pages\posts\[id].js both resolve to /posts/[id].
```

----
I also figured that I could reuse the page like this, but that does not solve the customization problem
```jsx
/* @jsx jsx */

import { getListPage } from '@keystone-next/admin-ui/pages/ListPage';
export default getListPage({ listKey: 'Post' });
```


examples/basic/admin/pages/index.tsx
```diff
@@ -0,0 +1,205 @@
+/* @jsx jsx */
```
gautamsi, line 1: This file does not need to be merged and should be removed before approval

examples/basic/admin/pages/posts/index.tsx
```diff
@@ -0,0 +1,766 @@
+/* @jsx jsx */
```
gautamsi, line 1: This file does not need to be merged and should be removed before approval

examples/basic/package.json
```diff
@@ -12,19 +12,26 @@
     "@babel/runtime": "^7.14.0",
```
gautamsi, line 12: This file changes does not need to be merged and should be removed before approval

packages-next/admin-ui/src/components/index.ts
```diff
@@ -1,8 +1,11 @@
 export { CellContainer } from './CellContainer';
```
gautamsi, line 1: this file could have exports sorted

gautamsi:
stale

PR #5655, opened 2021-05-10
Added top margin to nav when session isn't configured

JedWatson:
When you have a Keystone system without session configured, the session item UI is missing from above the navigation, and the first nav item sits right up against the top border.

This fixes that:

image


timleslie (approved):
👍

PR #5656, opened 2021-05-10
Updated generated schemas in example projects

JedWatson:
The generated schema files (prisma + graphql) in each example project haven't all been added/updated inline with recent changes to Keystone's CLI and the SQLite provider.

I created this PR by running `yarn dev` in each example project, so they're all up to date and won't show up as changes in git for anyone running the examples in the keystone monorepo.


timleslie (approved):
👍

Issue #5635, opened 2021-05-06
Upcoming changes to the GraphQL Engine, API, and Fields


JedWatson:
The purpose of this issue is to surface some work we're doing on the internals of Keystone Next in preparation for its launch as Keystone 6.

Now that we've completed our transition to Prisma and flattened out a lot of the internal complexity related to supporting multiple database adapters, **we are re-working how Fields work internally** to make them more streamlined and consistent.

Because of how Fields are at the heart of Keystone, this also effectively means re-working (and simplifying, and making more consistent) how we generate Keystone's **GraphQL Schema**.

We're also going to be making some **breaking changes to the GraphQL API** that Keystone generates - for very good reasons, but this is a big deal, and we've held off doing anything like this for years.

So it's important for us to start broadcasting what's coming, why, and how to manage the transition. Here's what's happening:

### 1. We're building a new Field Type system that will dramatically simplify fields

Right now, many of Keystone's internals work _by coincidence_, _by design_. This means that you have to express the filters a field supports; then implement those same filters in the query resolver. You have to express the GraphQL schema a field implements; then write input and output resolvers that implement behaviour to handle those same properties.

Here's an example of what I mean:

```ts
  gqlOutputFields() {
    return [`${this.path}: ${this.getTypeName()}`];
  }
  gqlOutputFieldResolvers() {
    return { [`${this.path}`]: (item: Record<P, any>) => item[this.path] };
  }
```

That's a small example; many fields are much more complex, which has a cost in consistency and effort required to ship new features. We also have a lot of magic for dealing with Relationships, or things like the `_supportsUnique` getter. **This is all going away**.

In the new system, you'll only have to express the structure of a Field (similar to how you'd define a type in a type system) and Keystone itself will do the implementation for you.

Here's a (current) example of what this looks like: the new implementation of a `text` field

```ts
export const text = <TGeneratedListTypes extends BaseGeneratedListTypes>({
  index,
  ...config
}: TextFieldConfig<TGeneratedListTypes> = {}): FieldTypeFunc => () =>
  fieldType({ kind: 'scalar', mode: 'optional', scalar: 'String', index })({
    ...config,
    input: {
      uniqueWhere: index === 'unique' ? { arg: types.arg({ type: types.String }) } : undefined,
      create: { arg: types.arg({ type: types.String }) },
      update: { arg: types.arg({ type: types.String }) },
      orderBy: { arg: types.arg({ type: sortDirectionEnum }) },
    },
    output: types.field({
[...]

PR #5665, opened 2021-05-10
New core

mitchellhamilton:
This is a re-implemention of the core of Keystone and all the server-side parts of field types. The original intent was to land this along with all the changes mentioned in #5635 but to make this easier to land, we've instead made this only contain architectural changes (+ some minimal coupled breaking changes for practicality). We'll land the other changes mentioned #5635 in future PRs.


.changeset/mean-chairs-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/fields': major
+---
+
+The API to configure `virtual` fields has changed to accept a `Field` using the `schema` API exported from `@keystone-next/types` rather than GraphQL SDL.
```
timleslie, line 5:
```suggestion
The API to configure `virtual` fields has changed to accept a `field` using the `schema` API exported from `@keystone-next/types` rather than GraphQL SDL. 
```

packages-next/types/src/config/lists.ts
```diff
@@ -46,29 +49,29 @@ export type ListConfig<
...
-  // Not currently supported
-  // plugins?: any[]; // array of plugins that can modify the list config
-
   /**
    * The label used for the list
    * @default listKey.replace(/([a-z])([A-Z])/g, '$1 $2').split(/\s|_|\-/).filter(i => i).map(upcase).join(' ');
    */
-  // Not currently supported
-  // label?: string;
+  label?: string;
```
timleslie, line 56: Do we support these new config options now? If so we should make sure these are documented in `schema.mdx` now before we forget about them.

packages-next/types/src/config/access-control.ts
```diff
@@ -5,50 +5,40 @@ type BaseAccessArgs = {
...
   /**
-   * The id being updated if a single item is being updated
+   * The id being updated
    */
-  itemId?: string;
-  /**
-   * The ids being updated if many items are being updated
-   */
-  itemIds?: string[];
+  itemId: string;
```
timleslie, line 33: References to `itemIds` in `apis/access-control.mdx` need to be removed to match this change.

packages-next/types/src/config/access-control.ts
```diff
@@ -5,50 +5,40 @@ type BaseAccessArgs = {
   session: any;
   listKey: string;
   context: KeystoneContext;
```
timleslie, line 7: `gqlName` is also referenced in `access-control.mdx` and should be removed there.

packages-next/types/src/config/lists.ts
mitchellhamilton: I've just removed support for them for now so we can make sure they're in the place we want them in a future PR.

timleslie (approved):
I think we're good to go! We've got a week until we ship, so we can clean up anything we find during the week. 🚀

PR #5667, opened 2021-05-11
Ssgc updates

timleslie:
Update the server-side-graphql-library to use the updated GraphQL APIs, unblocking tests which use this API indirectly via the items API.

### DIFF
--- a/packages-next/keystone/src/lib/context/server-side-graphql-client.ts
+++ b/packages-next/keystone/src/lib/context/server-side-graphql-client.ts
@@ -119,7 +119,7 @@ const getItems = async ({
 }: {
   listKey: string;
   where?: Record<string, any> | null;
-  orderBy?: readonly Record<string, 'asc' | 'desc'>[] | null;
+  orderBy?: readonly Record<string, any>[] | null;
   first?: number | null;
   skip?: number | null;
   pageSize?: number;
--- a/packages-next/keystone/src/lib/context/server-side-graphql-client.ts
+++ b/packages-next/keystone/src/lib/context/server-side-graphql-client.ts
@@ -76,7 +76,7 @@ const createItems = async ({
 }) => {
   const { createManyMutationName, createManyInputName } = context.gqlNames(listKey);
 
-  const query = `mutation ($items: [${createManyInputName}]){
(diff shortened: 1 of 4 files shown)


packages-next/keystone/src/lib/context/server-side-graphql-client.ts
```diff
@@ -119,7 +119,7 @@ const getItems = async ({
 }: {
   listKey: string;
   where?: Record<string, any> | null;
-  sortBy?: readonly string[] | null;
+  orderBy?: readonly Record<string, 'asc' | 'desc'>[] | null;
```
mitchellhamilton, line 122: This isn't right, the value can be anything, it will often be asc or desc but that is not guaranteed and there are definitely things that will deviate from it

packages-next/types/src/utils.ts
```diff
@@ -9,8 +9,7 @@ export type BaseGeneratedListTypes = {
       readonly search?: string | null;
       readonly first?: number | null;
       readonly skip?: number | null;
-      readonly orderBy?: string;
-      readonly sortBy?: ReadonlyArray<string> | null;
+      readonly orderBy?: ReadonlyArray<Record<string, 'asc' | 'desc'>> | null;
```
mitchellhamilton, line 12: Same as above

mitchellhamilton (commented):
just FYI, I haven't made all the GraphQL API changes that we discussed so this will need more changing in the future

mitchellhamilton (changes requested):
clicked the wrong button

packages-next/keystone/src/lib/context/server-side-graphql-client.ts
timleslie: Ah, interesting, what's the correct type here then, is it `string`, or `any`?

mitchellhamilton: any

timleslie: > just FYI, I haven't made all the GraphQL API changes that we discussed so this will need more changing in the future

Yep, just want to get this close enough to run for now 👍

PR #5669, opened 2021-05-11
Remove legacy tests

timleslie:
These tests of the legacy Keystone/List/Field classes are no longer required.


mitchellhamilton (approved):
Fine with this though we could PR this to master, no? Is there anything that these tests test that isn't covered elsewhere?

PR #5670, opened 2021-05-11
add focus-trap to popover

gwyneplaine:
Our popovers aren't keyboard friendly.
Previously we circumvented this in the admin-ui filter popover by autofocusing the select component. 

This is a more robust fix, which adds focus-trap to our popover component. 
I've opted for `focus-trap` over `focus-trap-react` here, as it:
* gives us more granular control 
* doesn't activate the focus-trap automatically on mount 
* doesn't pull in an unnecessary react-wrapper

Porting this over to use focus-trap-react later is arbitrary. 

Before: 

https://user-images.githubusercontent.com/12119389/117753502-eb09fb00-b25b-11eb-886c-b0902e0c9aa2.mov

After: 

https://user-images.githubusercontent.com/12119389/117753511-ef361880-b25b-11eb-933e-5db68cbc26bb.mov


design-system/packages/popover/src/Popover.tsx
```diff
@@ -194,6 +196,24 @@ type DialogProps = {
 export const PopoverDialog = forwardRef<HTMLDivElement, DialogProps>(
   ({ isVisible, children, arrow, ...props }, consumerRef) => {
     const { elevation, radii, shadow, colors } = useTheme();
+    const focusTrapRef = useRef<HTMLDivElement | null>(null);
```
gwyneplaine, line 199: popper's provided ref is a function, and not a refObject, we create a refObject here for us to attach our focustrap to

design-system/packages/popover/src/Popover.tsx
```diff
@@ -210,9 +230,10 @@ export const PopoverDialog = forwardRef<HTMLDivElement, DialogProps>(
           }}
           {...props}
         >
-          <div data-popper-arrow ref={arrow.ref} className="tooltipArrow" {...arrow.props} />
-
-          {children}
+          <div ref={focusTrapRef}>
```
gwyneplaine, line 233: Wrapping div to attach focus-trap to

design-system/packages/popover/src/Popover.tsx
```diff
@@ -194,6 +196,24 @@ type DialogProps = {
...
+
+    useEffect(() => {
+      if (focusTrapRef?.current) {
+        focusTrap.current = focusTrapModule.createFocusTrap(focusTrapRef.current);
+      }
+    }, [focusTrapRef]);
+
+    useEffect(() => {
+      if (focusTrap?.current) {
+        if (isVisible) {
```
gwyneplaine, line 210: hook into isVisible to activate / deactivate the focustrap

molomby (approved):
LGTM

PR #5677, opened 2021-05-11
Consolidate admin-ui package into keystone package

timleslie:


gautamsi:
Seems like completely opposite to request in #5650 

Ideally anyone would want following for customizing admin-ui

1.  Custom admin pages
2. write your own list/item pages
3. able to customize generated `_document.js` and `_app.js`
     1. able to add some global css, headers and other features using custom  `_document`
     2. able to wrap my custom provider, like modified Theme provider
4. Able to add components to pages like `ui-hooks` from v5
     1. not necessarily the same way, I may be able to use custom composed admin ui pages from `keystonejs-contrib`

with this change, it may become harder to get the customization without in built ui hooks.


I was working on an [example here](https://github.com/k6js/admin-ui/compare/use-theme-for-pages) on how to make theming changes without the need of change in core keystone code. some changes are made to admin ui fork in that repo which is already requested in #5649 . in future I would want to add more features to it like ui-hooks

What was felt needed and missing right now
* A way to customize `_app.js` with which I could have used my custom theme provider without the need of dummy pages in admin/pages/<list> folder
* A way to customize the `_document` to add custom header or font links,
* wish if admin-ui could be used independently from the keystone package.


I have no problem with this being merged in keystone, I need some kind of strategy to generate all the list page customization using one `index.tsx` and one `[id].tsx` instead of creating stub pages for each of the list in pages folder.

gautamsi:
@JedWatson This is going to create the dependency chain which will not be easy for customizations.

all fields currently depends on the `admin-ui` package, now they depends on keystone package. I have no way to fork just `admin-ui` or just `keystone`, I will need to fork the full repo and publish each package now if I want to customize `_app.js` generated by `admin-ui`. I started working on customizing the `admin-ui` in a separate fork but seems dependency on keystone is hinderance in this process.

packages-next/admin-ui/src/apollo.tsx
```diff
@@ -1,6 +1,5 @@
-/**
- * This file is exposed by the /apollo entrypoint, and helps ensure that other
- * packages import the same instance of apollo.
- */
+throw new Error(
+  'The @keystone-next/admin-ui package has been deprecated. Please import from @keystone-next/keystone/admin-ui instead.'
```
mitchellhamilton, line 2: deprecated to me means "you shouldn't use this, you still can right now but in the future, you probably won't be able to" which isn't this, could you change all these messages to say removed or merged into @keystone-next/keystone?

packages-next/keystone/package.json
```diff
@@ -99,11 +123,18 @@
...
       "migrations.ts",
       "schema/index.ts",
       "session/index.ts",
-      "scripts/index.ts"
+      "scripts/index.ts",
+      "admin-ui/components/index.ts",
+      "admin-ui/apollo.tsx",
+      "admin-ui/context.tsx",
+      "admin-ui/router.tsx",
+      "admin-ui/next-config.ts",
```
mitchellhamilton, line 136: next-config should be in `___internal-do-not-use-will-break-in-patch`

packages-next/keystone/src/admin-ui/static/next.config.js
```diff
@@ -1,3 +1,3 @@
 // this file is copied somewhere else so this lint rule is incorrect
```
mitchellhamilton, line 1: This and the tsconfig.json in the same directory need to the at the root of the package (outside of src)

mitchellhamilton (changes requested):
Could you add something to the admin ui readme to say this has been merged into @keystone-next/keystone so that people who come across it on npm know what's going on?

packages-next/keystone/src/admin-ui/static/next.config.js
mitchellhamilton: When i said root of the package, I meant in the static directory at the root of the repo

mitchellhamilton (changes requested):
Could you add something to the admin ui readme to say this has been merged into @keystone-next/keystone so that people who come across it on npm know what's going on?

mitchellhamilton (approved):
After this is released, could you `npm deprecate` the new major version of admin-ui? (but only the new major, not the previous versions)

JedWatson:
@gautamsi to answer the points you raise here:

Firstly acknowledging that we know Admin UI customisation is important, and we're planning to move our attention to that once the current round of refactoring is complete (that I outlined in #5635)

Regarding the reasoning for the change in this PR, maybe we haven't explained it sufficiently, but we're looking to consolidate the "minimum viable keystone" code into a single package. This will _probably_ also include the `fields` and `auth` packages as well.

We have a few reasons for this including:

* We currently have more Public API in our packages than we should, to support them working independently
* You really have to, at minimum, install three packages (often four) to get a Keystone system running; then you have to have them at compatible versions, which has already in the last month or so caused a number of problems for users. Less core packages means less compatibility issues, and version churn for Keystone users (as well as a simpler setup process)
* Our hope is that having one "main" package version will make changelogs easier to publish and understand

It's a trade-off, but technically these packages aren't easily separateable so managing them as separate dependencies on the consumer end is a bad experience for everyone.

This is something that (among other changes) our shift to only using Prisma for our ORM makes possible, because with Keystone 5 we didn't want (for example) you to have to install and build Mongo drivers if you're using Postgres and Knex.

It will hopefully also address problems we've experienced with the move to a monorepo in Keystone 5.

On the other hand, it's all about striking a balance, so where we can we will still be using the monorepo and building separate packages (e.g the cloudinary integration, and likely the Document field, as well as all the Design System components will remain separate packages in the monorepo and you'll only need to depend on them when you need them)

Second, in terms of something that helps you right now, instead of forking the main Keystone package so you can modify the Admin UI more than we have so far made possible, I would actually recommend looking into [patch-package](https://github.com/ds300/patch-package)

It would let you apply changes directly; e.g to the source of `_app.tsx` without forking the entire package. As we make changes to that, you'll still need to validate that your patches are going to be applied correctly, but in the future (when we get back to Admin UI customisation) hopefully the need for this goes away entirely.

gautamsi:
Thanks @JedWatson for detailed answer on the direction. I really like the idea to keep things together which are needed to run the basic keystone app.

I was also wondering same about the prisma adapter package and now it is clear that it will be merged and that is very good thing. 

you are right that some packages are already not separable and they should be put together. 

I am thinking that why not do a reexport from keystone rather than keeping it together?  I see that admin-ui already did re-export lot of packages to make sure everything is using same version of emotion and apollo.


I will give `patch-package` a try to see if that helps in some way. Other than re-exporting, there should not be many changes I need at the moment.

PR #5686, opened 2021-05-12
Reduce dependence on keystone object in createContext

timleslie:
More work towards #5665. This PR reduces the explicit dependence on the Keystone object, which will no longer exist soon.

### DIFF
--- a/packages-next/keystone/src/lib/context/createContext.ts
+++ b/packages-next/keystone/src/lib/context/createContext.ts
@@ -6,7 +6,7 @@ import {
   KeystoneGraphQLAPI,
   BaseKeystone,
   KeystoneConfig,
-  getGqlNames,
+  GqlNames,
 } from '@keystone-next/types';
 
 import { getDbAPIFactory, itemAPIForList } from './itemAPI';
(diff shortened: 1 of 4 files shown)


packages-next/keystone/src/lib/context/createContext.ts
```diff
@@ -18,32 +18,35 @@ export function makeCreateContext({
...
+  lists,
 }: {
   graphQLSchema: GraphQLSchema;
   internalSchema: GraphQLSchema;
   keystone: BaseKeystone;
-  imagesConfig: ImagesConfig | undefined;
-  filesConfig: FilesConfig | undefined;
+  config: KeystoneConfig;
+  prismaClient: any;
+  lists: Record<string, { listKey: string; itemQueryName: string; listQueryName: string }>;
```
mitchellhamilton, line 30: Could you make this `Record<string, GqlNames>` since that's the only thing it's used for?

packages-next/types/src/core.ts
```diff
@@ -42,7 +42,7 @@ export type GraphQLSchemaExtension = {
 // TODO: don't duplicate this between here and packages/keystone/ListTypes/list.js
 export function getGqlNames({
   listKey,
-  itemQueryName: _itemQueryName,
+  itemQueryName: _itemQueryName = listKey,
```
mitchellhamilton, line 45: This change does nothing? it's still required in the argument?
```suggestion
  itemQueryName: _itemQueryName,
```

timleslie: Ah yep, that was part of something else I was trying, will remove 👍

packages-next/keystone/src/lib/context/createContext.ts
timleslie: Do you mean:
```
  lists: Record<string, Parameters<typeof getGqlNames>>[0];
```

mitchellhamilton: I mean `Record<string, ReturnType<typeof getGqlNames>>` (though don't actually use ReturnType because we export the type that it returns)

emmatown (approved):

PR #5691, opened 2021-05-12
Allow pass custom app url via APP_URL and support for .env file in basic example

MurzNN:
This PR adds ability to set custom app url from APP_URL environment variable, to output right url when Keystone works behind the reverse proxy (eg `https://myapp.org/api` redirects to `http://localhost:1337`) - this is very common usage.

This feature was already added via https://github.com/keystonejs/keystone/pull/4738 but seems somehow disappears now.

Also I added support for `.env` file into `examples/basic` with example how to use `APP_URL` env value, together with allowing pass database type (postgresql) via DATABASE_URL and adding generated `schema.graphql`, `schema.prisma` files to gitignore for not add them into git.

### DIFF
--- a/.changeset/yellow-plants-lie.md
+++ b/.changeset/yellow-plants-lie.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/app-basic': minor
+'@keystone-next/keystone': minor
+---
+
+Custom app url via APP_URL and .env file for basic example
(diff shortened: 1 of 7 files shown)


IMalaniak:
Is there any better solution, as for example setting basePath to next.config.js?

dcousens:
I don't know what the pull request is trying to do, having read the description, I think we wanted to change the debugging messages to not use `localhost`, we should strip the host out completely, not add an environment variable for it.

PR #5695, opened 2021-05-13
Add connect and disconnect to createSystem

timleslie:
The `keystone` property will be removed in #5665, so this provides updated APIs which can be used by scripts and tests so they no longer need to use the keystone object directly.

### DIFF
--- a/.changeset/eleven-fireants-drum.md
+++ b/.changeset/eleven-fireants-drum.md
@@ -1,0 +1,7 @@
+---
+'@keystone-next/admin-ui': major
+'@keystone-next/keystone': major
+'@keystone-next/test-utils-legacy': major
+---
+
+Added `connect` and `disconnect` functions to the object returned by `createSystem`.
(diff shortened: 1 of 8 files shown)


emmatown (changes requested):
Could you change this so that we return a function which accepts a prisma client and returns an object with connect, disconnect and context? and make it so createSystem doesn't accept a prisma client

timleslie:
> Could you change this so that we return a function which accepts a prisma client and returns an object with connect, disconnect and context? and make it so createSystem doesn't accept a prisma client

Hmm, not really just yet, there are still a bunch of things tangled up with the prisma adapter and keystone adapter which don't make that.

mitchellhamilton:
You'd have to create another keystone instance in the current implementation but I'm not sure what problems you'd really have doing that? I don't think merging this PR as it is would be helpful because it's not really what the final API should end up looking like so it's just API churn.

PR #5699, opened 2021-05-13
Fix .count(), update tests

timleslie:
Updates some relationship tests to use the new APIs. Fixes .count() method when called with no args.

### DIFF
--- a/tests/api-tests/relationships/crud/many-to-many-one-sided.test.ts
+++ b/tests/api-tests/relationships/crud/many-to-many-one-sided.test.ts
@@ -115,7 +115,7 @@ multiAdapterRunners().map(({ runner, provider }) =>
                 ['D', 0],
               ].map(async ([name, count]) => {
                 const companies = await context.lists.Company.findMany({
-                  where: { locations_some: { name } },
+                  where: { locations: { some: { name: { equals: name } } } },
                 });
                 expect(companies.length).toEqual(count);
               })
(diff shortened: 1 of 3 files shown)


packages-next/keystone/src/lib/context/itemAPI.ts
```diff
@@ -101,7 +101,7 @@ export function itemAPIForList(
       }
     },
     async count(args = {}) {
-      const { first, skip, where } = args;
+      const { first, skip = 0, where={} } = args;
       const { listQueryMetaName, whereInputName } = context.gqlNames(listKey);
       const query = `query ($first: Int, $skip: Int, $where: ${whereInputName}) { ${listQueryMetaName}(first: $first, skip: $skip, where: $where) { count }  }`;
```
mitchellhamilton, line 106:
```suggestion
      const query = `query ($first: Int, $skip: Int! = 0, $where: ${whereInputName}! = {}) { ${listQueryMetaName}(first: $first, skip: $skip, where: $where) { count }  }`;
```

packages-next/keystone/src/lib/context/itemAPI.ts
```diff
@@ -101,7 +101,7 @@ export function itemAPIForList(
       }
     },
     async count(args = {}) {
-      const { first, skip, where } = args;
+      const { first, skip = 0, where={} } = args;
```
mitchellhamilton, line 104:
```suggestion
      const { first, skip, where } = args;
```

mitchellhamilton (changes requested):
I don't think the count your count fix will actually fix it(though if you actually tried it, and it worked, that's interesting)

mitchellhamilton (approved):
I think I was totally wrong.

timleslie:
> I don't think the count your count fix will actually fix it(though if you actually tried it, and it worked, that's interesting)

Yeah, it turned the tests from red to green, pretty sure it's working now (or at least more working than it was :-) )

PR #5706, opened 2021-05-14
Add instructions for writing examples

timleslie:
This is a first pass at instructions for writing example projects. We'll update these as we gain more experience.

### DIFF
--- a/examples/NEW-EXAMPLES.md
+++ b/examples/NEW-EXAMPLES.md
@@ -8,7 +8,7 @@ If you're looking for instructions on how to use the examples, please consult th
 
 There are three types of example project:
 
-- Base Projects: We have two of these, a blog and a task manager. We don't need any more of thse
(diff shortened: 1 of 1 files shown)


examples/NEW-EXAMPLES.md
```diff
@@ -0,0 +1,31 @@
...
+
+These docs are intended for developers adding new examples to the Keystone repository.
+
+If you're looking for instructions on how to use the examples, please consult the [README](./README.md).
+
+### Types of examples
+
+There are three types of example project:
+
+- Base Projects: We have two of these, a blog and a task manager. We don't need any more of thse
```
gwyneplaine, line 11:
```suggestion
- Base Projects: We have two of these, a blog and a task manager. We don't need any more of these
```

gwyneplaine (changes requested):

gwyneplaine (approved):
lgtm

PR #5727, opened 2021-05-15
fix: storedSessions get returns object not string

cameronbraid:


packages-next/keystone/src/session/index.ts
```diff
@@ -188,7 +188,8 @@ export function storedSessions({
     let isConnected = false;
     return {
       async get({ req, createContext }) {
-        let sessionId = await get({ req, createContext });
+        let data = await get({ req, createContext }) as any;
+        let sessionId = data?.sessionId
```
timleslie, line 192:
```suggestion
        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
        const sessionId = data?.sessionId;
```

packages-next/keystone/src/session/index.ts
```diff
@@ -207,7 +208,8 @@ export function storedSessions({
         return start?.({ res, data: { sessionId }, createContext }) || '';
       },
       async end({ req, res, createContext }) {
-        let sessionId = await get({ req, createContext });
+        let data = await get({ req, createContext }) as any;
+        let sessionId = data?.sessionId
```
timleslie, line 212:
```suggestion
        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
        const sessionId = data?.sessionId;
```

packages-next/keystone/src/session/index.ts
```diff
@@ -188,7 +188,9 @@ export function storedSessions({
     let isConnected = false;
     return {
       async get({ req, createContext }) {
-        let sessionId = await get({ req, createContext });
+        let data = await get({ req, createContext }) as any;
+        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
+        const sessionId = data?.sessionId;
```
timleslie, line 193:
```suggestion
        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
        const sessionId = data?.sessionId;
```

packages-next/keystone/src/session/index.ts
```diff
@@ -207,7 +209,9 @@ export function storedSessions({
         return start?.({ res, data: { sessionId }, createContext }) || '';
       },
       async end({ req, res, createContext }) {
-        let sessionId = await get({ req, createContext });
+        let data = await get({ req, createContext }) as any;
+        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
+        const sessionId = data?.sessionId;
```
timleslie, line 214:
```suggestion
        const data = (await get({ req, createContext })) as { sessionId: string } | undefined;
        const sessionId = data?.sessionId;
```

timleslie (approved):
Good catch, thanks for the PR 🙏

cameronbraid:
Thanks :)

PR #5747, opened 2021-05-17
Allow null relationship on create

timleslie:
Our GraphQL types allow a `one` relationship to be explicitly set to `null` on create, but the code then throws an error. I think the GraphQL types are correct and the following should be valid. 

```
            const company = await context.lists.Company.createOne({
              data: { location: null },
              query: 'id location { id }',
            });
```

This PR removes the error case to make it so. Let me know if you have other thoughts 👍


mitchellhamilton (changes requested):
> Our GraphQL types allow a one relationship to be explicitly set to null on create

That's not really out of us wanting to allow that, it's just that we can't express "this is optional but can't be null and it doesn't have a default value"

---

null for to-one relationships means disconnect and you can't disconnect when you're creating something so I think this should stay disallowed.

PR #5749, opened 2021-05-18
Add changeset for final Admin UI release

timleslie:
We'll ship this release and then we can do #5677 which moves the admin-ui package into the keystone package.

### DIFF
--- a/.changeset/mean-rules-lie.md
+++ b/.changeset/mean-rules-lie.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/admin-ui': patch
+---
+
+This is the final release of this package. In the future the imports available here will be made available as `@keystone-next/keystone/admin-ui`.


mitchellhamilton:
Are you suggesting that we do a release where you can import from both @keystone-next/keystone/admin-ui and @keystone-next/admin-ui or we do a release that tells people "in the next release you won't be able to import this but you can't migrate until that release"?

timleslie:
I'm proposing we ship this now, as a working release of `admin-ui`, with instructions on where to go next, then we can land the PR that deletes the admin-ui package from the repo, then we ship the updated `@keystone-net/keystone` package that contains the admin ui code.

mitchellhamilton:
So the second one that I said or is what you're proposing different in some way that I don't understand?

timleslie:
Yeah, I guess it is the second one.

mitchellhamilton (approved):

timleslie:
Change of plans as discussed with @mitchellhamilton, not going to bother with this changeset 👍

PR #5757, opened 2021-05-18
Fix sorting by multiple values

timleslie:
Prisma no longer (if it ever? 🤔 ) accepts an object with multiple values for sorting. I've added tests to exercise this feature and to avoid any future regressions.

### DIFF
--- a/.changeset/curly-jobs-tie.md
+++ b/.changeset/curly-jobs-tie.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/adapter-prisma-legacy': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/curly-jobs-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/adapter-prisma-legacy': patch
+---
+
+Fixed faliures in GraphQL queries using multiple `sortBy` values.
```
timleslie, line 5:
```suggestion
Fixed failures in GraphQL queries using multiple `sortBy` values.
```

emmatown (approved):

PR #5758, opened 2021-05-19
Add SKIP_PROMPTS to shouldPrompt in CLI commands

timleslie:
This PR adds a few things:

1. Adds a `SKIP_PROMPTS` env var check to `postinstall`, which ensures it won't  ever prompt, it will just exit with `1` on failure.
2. Adds a `lint:examples` script which uses `SKIP_PROMPTS=1 postinstall` to make sure our example schemas are up to date. This will prevent PRs getting through with out of date schemas.
3. Adds a `postinstall-examples` script which applies `postinstall --fix` to all the example projects for an easy fix, particularly when changing our GraphQL or prisma schema output.

### DIFF
--- a/.changeset/wet-jeans-report.md
+++ b/.changeset/wet-jeans-report.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Added a `SKIP_PROMPTS` environment variable to explicitly disable prompts in the CLI.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -292,3 +292,5 @@ jobs:
         run: yarn build
       - name: Remark
         run: yarn lint:markdown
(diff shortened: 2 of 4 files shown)


emmatown (changes requested):
Could we call this `--skip-update-prompt-even-in-interactive` or something to indicate that the things are always checked regardless of this flag?

And for the use case in the repo, why don't you want the fix prompt when running locally? (since regardless of the flag you're proposing here, it won't run on CI)

timleslie:
> And for the use case in the repo, why don't you want the fix prompt when running locally? (since regardless of the flag you're proposing here, it won't run on CI)

I want to be able to get identical behaviour locally and on CI so that I can easily verify and debug the behaviour when things go wrong. Without this flag there's no way to make this happen.

emmatown (approved):

PR #5761, opened 2021-05-19
Add 'as' to FieldContainer

gwyneplaine:
Currently there are quite a few places where we use `legend` elements to denote field-groups. 
None of these are working because the FieldContainer parent is a div, not a `fieldset`. 

This is a minor patch to our ds component, in preparation for incoming a11y fixes in fields.

### DIFF
--- a/design-system/packages/fields/src/FieldContainer.tsx
+++ b/design-system/packages/fields/src/FieldContainer.tsx
@@ -1,13 +1,19 @@
 /* @jsx jsx */
 
 import { ReactNode } from 'react';
-import { jsx } from '@keystone-ui/core';
+import { jsx, forwardRefWithAs } from '@keystone-ui/core';
 
 type FieldContainerProps = {
   children: ReactNode;
   className?: string;
 };
 
-export const FieldContainer = ({ children, ...props }: FieldContainerProps) => {
-  return <div {...props}>{children}</div>;
-};
+export const FieldContainer = forwardRefWithAs<'div', FieldContainerProps>(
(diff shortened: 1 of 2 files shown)


design-system/packages/fields/src/FieldContainer.tsx
```diff
@@ -1,13 +1,19 @@
...
 
 type FieldContainerProps = {
   children: ReactNode;
   className?: string;
 };
 
-export const FieldContainer = ({ children, ...props }: FieldContainerProps) => {
-  return <div {...props}>{children}</div>;
-};
+export const FieldContainer = forwardRefWithAs<'div', FieldContainerProps>(
```
mitchellhamilton, line 11: The second type arg should be `{}`, and FieldContainerProps should be removed since the children and className will be added based on the `as` prop

emmatown (changes requested):

design-system/packages/fields/src/FieldContainer.tsx
gwyneplaine: Removed the classname from `FieldContainerProps` but looks like children type isn't inferred from the forwardWithRefAs fn.

mitchellhamilton: It should be, though the `children` prop won't always be passed to the FieldContainer component since not all components you could pass to `as` will accept children so the implementation of FieldContainer needs to not handle children specifically.

gwyneplaine: ah cool, reoslved

emmatown (approved):

PR #5763, opened 2021-05-19
[WIP] Make mutation return types non-null

timleslie:
Adds `!` to the return types of the generated mutations.


timleslie:
This isn't actually the API that we want right now, as it's not consistent with our error handling and reporting strategy.

PR #5766, opened 2021-05-20
Fix/signin page a11y

gwyneplaine:
Sign-in page and welcome page inputs don't have labels.
While its ok to rely on placeholders for `VoiceOver`, JAWS and NVDA don't read out placeholders as whole words 
(read letter by letter). 

This PR adds a visually hidden label to amend this.

### DIFF
--- a/packages-next/auth/src/pages/SigninPage.tsx
+++ b/packages-next/auth/src/pages/SigninPage.tsx
@@ -1,8 +1,8 @@
 /* @jsx jsx */
 
-import { useState, FormEvent, useRef, useEffect } from 'react';
+import { useState, Fragment, FormEvent, useRef, useEffect } from 'react';
 
-import { jsx, H1, Stack } from '@keystone-ui/core';
+import { jsx, H1, Stack, VisuallyHidden } from '@keystone-ui/core';
 import { Button } from '@keystone-ui/button';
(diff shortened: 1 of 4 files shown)


packages-next/auth/src/pages/SigninPage.tsx
```diff
@@ -101,21 +101,31 @@ export const SigninPage = ({
           </Notice>
         )}
         <Stack gap="medium">
+          <VisuallyHidden as="label" htmlFor="identity">
+            Email Address
```
timleslie, line 105: This won't always be an email address. We should use whatever the identify field is called here.

packages-next/auth/src/pages/SigninPage.tsx
```diff
@@ -101,21 +101,31 @@ export const SigninPage = ({
...
-            <TextInput
-              name="password"
-              value={state.secret}
-              onChange={e => setState({ ...state, secret: e.target.value })}
-              placeholder="password"
-              type="password"
-            />
+            <Fragment>
+              <VisuallyHidden as="label" htmlFor="password">
+                Password
```
timleslie, line 118: Ditto here.

timleslie (approved):
👍

PR #5767, opened 2021-05-20
Update the orderBy GraphQL API

timleslie:
Replaces the `orderBy` string with the more flexible list of order directions in the list query arguments. Deprecates `sortBy`, which is no longer required.

### DIFF
--- a/.changeset/rude-dots-join.md
+++ b/.changeset/rude-dots-join.md
@@ -1,0 +1,48 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/adapter-prisma-legacy': major
+'@keystone-next/website': patch
+'@keystone-next/example-auth': patch
+'@keystone-next/app-basic': patch
+'@keystone-next/example-blog': patch
+'@keystone-next/example-ecommerce': patch
+'keystone-next-app': patch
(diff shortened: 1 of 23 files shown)


timleslie:
I'm going to change `OrderUsersBy` to `UserOrderByInput`.

molomby (approved):
I like

mitchellhamilton (changes requested):
To explicitly block this on the discussed changes

.changeset/rude-dots-join.md
```diff
@@ -0,0 +1,48 @@
...
+'@keystone-next/example-sandbox': patch
+'@keystone-next/example-todo': patch
+'@keystone-next/example-with-auth': patch
+'@keystone-next/types': patch
+'@keystone-next/api-tests-legacy': patch
+---
+
+Deprecated the `sortBy` GraphQL filter. Updated the `orderBy` GraphQL filter with an improved API.
+
+Previously a `User` lists `allUsers` query would have the argument:
```
timleslie, line 21:
```suggestion
Previously a `User` list's `allUsers` query would have the argument:
```

mitchellhamilton (approved):

PR #5769, opened 2021-05-20
Add the new itemsCount query to GraphQL API

timleslie:
This change deprecates `_allItemsMeta { count }` in favour of a new query `itemsCount` which simply returns a count. This change will make it easier to perform counts and will move away from the potentially confusing leading underscore syntax.

### DIFF
--- a/packages-next/keystone/src/lib/core/tests/List.test.ts
+++ b/packages-next/keystone/src/lib/core/tests/List.test.ts
@@ -229,6 +229,7 @@ describe('new List()', () => {
       itemQueryName: 'Test',
       listQueryName: 'allTests',
       listQueryMetaName: '_allTestsMeta',
+      listQueryCountName: 'testsCount',
       listSortName: 'SortTestsBy',
       listOrderName: 'TestOrderByInput',
       deleteMutationName: 'deleteTest',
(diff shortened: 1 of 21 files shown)


packages-next/keystone/src/lib/core/tests/List.test.ts
```diff
@@ -530,7 +531,9 @@ describe(`getGqlQueries()`, () => {
           orderBy: String
           first: Int
           skip: Int
-        ): _QueryMeta`,
+        ): _QueryMeta @deprecated(reason: \"This query will be removed in a future version. Please use allTestsCount instead.\")`,
+        `""" Count the number of items that match the where clause """
```
mitchellhamilton, line 535: Let's not have this description, it just clutters up the schema, the naming makes it clear what it does

jesstelford:
Do we need some logic to ensure list names do not end in `Counts` now?

A contrived example: I have two lists `User` & `UserCounts`, then Keystone will try create 4 queries:

1. `allUsers`
2. `allUserCounts`
3. `allUserCounts`
4. `allUserCountsCounts`

Note that 2 & 3 are conflicting.

timleslie:
> Do we need some logic to ensure list names do not end in `Counts` now?
> 
> A contrived example: I have two lists `User` & `UserCounts`, then Keystone will try create 4 queries:
> 
> 1. `allUsers`
> 2. `allUserCounts`
> 3. `allUserCounts`
> 4. `allUserCountsCounts`
> 
> Note that 2 & 3 are conflicting.

Yeah I need to add a test case to make sure that this throws an appropriate error 👍

molomby:
Should we consider modelling the GraphQL for this stuff on what Prisma are doing? https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing

mitchellhamilton (approved):

PR #5772, opened 2021-05-21
Add tests for validation hooks in both single and multi mutations

timleslie:
This adds tests for validation hooks which in turn explores the fact that our behaviour for some many-mutations where previously not well defined.


mitchellhamilton (changes requested):
This will still check access control separately and fail everything if access control on any of updates/creates fail, that seems inconsistent with the other behavior here?

mitchellhamilton:
Also, this should be a major, it's quite a substantial change in behavior

timleslie:
> This will still check access control separately and fail everything if access control on any of updates/creates fail, that seems inconsistent with the other behavior here?

Yeah, I think you're right 🤔 I'm going to add a similar set of tests to do all the same stuff, but have the errors be from access control rather than validation, and we can verify the current behaviour and possibly update it 👍

mitchellhamilton:
> Update: If using static, or imperative returning boolean, then this will be true. The static/imperative->static value is applied to the entire operation, and a single access control error is returned, with null data. If using a filter or imperative returning filter, then the filter will be applied and only those items which pass the filter will be operated on. Those that are filtered out will return null and the errors array will contain errors indicating the reason for the error. This makes the declarative access control case consistent with the validation errors case.

What's the rationale in making returning a boolean vs a filter different? You're just exposing the fact that you happen to use or not use filters in access control which means the behavior will change if you switch to using filter access control which doesn't seem ideal? IMO whether someone uses filter access control or not should be an implementation detail that shouldn't be observable (ignoring timing things) by a consumer (and therefore changing it is fine without worrying about breaking consumers).

timleslie:
> What's the rationale in making returning a boolean vs a filter different? You're just exposing the fact that you happen to use or not use filters in access control which means the behavior will change if you switch to using filter access control which doesn't seem ideal? IMO whether someone uses filter access control or not should be an implementation detail that shouldn't be observable (ignoring timing things) by a consumer (and therefore changing it is fine without worrying about breaking consumers).

No specific rationale, that's just what the existing behaviour was, and that hasn't changed. I'm happy to make the change to make everything more consistent 👍

mitchellhamilton (changes requested):
Could you make field level access control consistent with list-level in returning the field level access control errors per item rather than throwing for the whole thing?

packages-next/keystone/src/lib/core/ListTypes/list.ts
```diff
@@ -846,38 +918,68 @@ export class List implements BaseKeystoneList {
...
     } else {
-      return Promise.all(
-        itemsToUpdate.map(({ existingItem, id, data }) =>
-          this._updateSingle(id, data, existingItem, context, mutationState)
-        )
+      results = await Promise.allSettled(
+        itemsToUpdate.map((item, i) => {
+          if (item === null || fieldAccess[i] !== null) {
+            // The item either didn't exist, or was filtered out by access control
+            return new AccessDeniedError({ data: { type: 'mutation' } });
```
mitchellhamilton, line 972: I know it presently works because graphql-js does magic things if you return an Error it seems but for clarity:
```suggestion
            return Promise.reject(new AccessDeniedError({ data: { type: 'mutation' } }));
```

mitchellhamilton (approved):

PR #5771, opened 2021-05-20
V2 of embedded mode tutorial (with supporting `modes` guide)

raveling:

PR #5773, opened 2021-05-21
Move next-lite example

JedWatson:
This example is really about demo'ing Keystone's new "Embedded in Next.js" mode, as documented by #5771


timleslie (approved):
👍

PR #5774, opened 2021-05-22
Add sameSite option to sessions

jonowu:
Adds a `sameSite` option to the session options for cookies. Can be one of `true`, `false`, `'strict'`, `'lax'` or `'none'` as per  [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#attributes). The option defaults to `'lax'` which is in line with existing behaviour.

This option will be useful for users who have deployed their keystone instance to a different site from their frontend, e.g. hosting keystone on `example.herokuapp.com` with the frontend on `example.vercel.app`.

### DIFF
--- a/.changeset/shiny-drinks-camp.md
+++ b/.changeset/shiny-drinks-camp.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/website': minor
+'@keystone-next/keystone': minor
+---
+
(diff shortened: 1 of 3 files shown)


JedWatson (approved):
Looks great, thanks @jonowu!

kilinkis:
thanks for this!

PR #5777, opened 2021-05-23
Make skip non-null with default

timleslie:


packages-next/keystone/src/scripts/tests/__snapshots__/artifacts.test.ts.snap
```diff
@@ -81,7 +81,7 @@ export type TodoListTypeInfo = {
       readonly where?: TodoWhereInput | null;
       readonly sortBy?: ReadonlyArray<SortTodosBy> | null;
       readonly first?: Scalars['Int'] | null;
-      readonly skip?: Scalars['Int'] | null;
+      readonly skip: Scalars['Int'];
```
mitchellhamilton, line 84: this needs to be fixed in the schema type printer, this should be optional because it has a default

mitchellhamilton (changes requested):

mitchellhamilton (approved):

PR #5778, opened 2021-05-23
Move TS implementation into `create next-app` command.

raveling:
Per feedback on Twitter, and at Dom's suggestion. TypeScript implementation now happens in Step 1:

`yarn create next-app --typescript my-project`


dominikwilkowski (approved):
LGTM

PR #5780, opened 2021-05-24
Fix schema type printer for default values

emmatown:
This is sorta wrong because we actually need to model the input and coerced inputs as different things but the main place where people are practically using these types today is the lists API so I think catering for the uncoerced input is better than catering for the coerced input types.

This is for #5767 and #5777


timleslie (approved):
👍

PR #5797, opened 2021-05-25
Update many-count GraphQL API

timleslie:
This change mirrors the changes made to the top level meta queries.

### DIFF
--- a/.changeset/few-readers-admire.md
+++ b/.changeset/few-readers-admire.md
@@ -16,5 +16,5 @@ The GraphQL field `_all<path>Meta { count }` generated for `many` relationships 
 A `posts` relationship field would have the following field added to the API:
 
 ```graphql
-postsCount(where: PostWhereInput! = {}): Int!
+postsCount(where: PostWhereInput! = {}): Int
 ```
--- a/.changeset/few-readers-admire.md
+++ b/.changeset/few-readers-admire.md
@@ -1,0 +1,20 @@
+---
+'@keystone-next/fields': major
+'@keystone-next/app-basic': patch
+'@keystone-next/example-blog': patch
+'@keystone-next/example-ecommerce': patch
+'keystone-next-app': patch
+'@keystone-next/example-json-field': patch
+'@keystone-next/example-roles': patch
+'@keystone-next/example-sandbox': patch
+'@keystone-next/example-todo': patch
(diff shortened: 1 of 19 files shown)


packages-next/fields/src/types/relationship/Implementation.ts
```diff
@@ -91,7 +91,12 @@ export class Relationship<P extends string> extends Implementation<P> {
       const filterArgs = refList.getGraphqlFilterFragment().join('\n');
       return [
         `${this.path}(${filterArgs}): [${refList.gqlNames.outputTypeName}!]!`,
-        this.withMeta ? `_${this.path}Meta(${filterArgs}): _QueryMeta` : '',
+        this.withMeta
+          ? `_${this.path}Meta(${filterArgs}): _QueryMeta @deprecated(reason: "This query will be removed in a future version. Please use ${this.path}Count instead.")`
+          : '',
+        this.withMeta
+          ? `${this.path}Count(${`where: ${refList.gqlNames.whereInputName}! = {}`}): Int!`
```
mitchellhamilton, line 98: This needs to be nullable (and the root one should be as well if it isn't currently)

mitchellhamilton, line 98: This needs to be nullable (and the root one should be as well if it isn't currently)

emmatown (changes requested):

.changeset/few-readers-admire.md
```diff
@@ -0,0 +1,20 @@
...
+'@keystone-next/example-todo': patch
+'@keystone-next/example-with-auth': patch
+---
+
+The GraphQL field `_all<path>Meta { count }` generated for `many` relationships has been deprecated in favour of a new field `<path>Count`, which directly returns the count.
+
+A `posts` relationship field would have the following field added to the API:
+
+```graphql
+postsCount(where: PostWhereInput! = {}): Int!
```
mitchellhamilton, line 19:
```suggestion
postsCount(where: PostWhereInput! = {}): Int
```

emmatown (approved):

.changeset/few-readers-admire.md
mitchellhamilton, line 19:
```suggestion
postsCount(where: PostWhereInput! = {}): Int
```

PR #5798, opened 2021-05-25
Updating the "Keystone + SQLite" intro text 

molomby:
The existing intro doesn't give many readers a compelling reason to proceed. It assume the audience already understands what Keystone is, does and the "old way" of working with it.

The new text tries to make less assumptions. It's aimed at people who may never have heard of Keystone so calls out the concrete outcomes people will be left with having completed the steps.


JedWatson (approved):

JedWatson:
Nice, thanks @molomby

PR #5800, opened 2021-05-26
A11y/scalar fields

gwyneplaine:
Fix the labels for the following fields: 
* decimal
* integer
* text - we were previously using field.label to associate the input with the label *
* float
* select
* password

These fields now leverage the field.path attribute as the id linking labels with inputs via `htmlFor` and `id` respectively. 
Fields that comprise of multiple interactive elements such as `password` and `segmented-control` view of the `select` field are now wrapped in a `fieldset` rather than just a `div` such that the field legend is read out properly. 

`timestamp` field requires more changes and will be in a separate PR.


timleslie (approved):
This all LGTM 👍

PR #5801, opened 2021-05-26
Don't let withItemData clear session data

timleslie:
At the moment if the `withItemData` wrapper can't find an item based on `listKey` and `itemId` it completely clears the session data by returning `undefined`. This makes it impossible to use `withItemData` in conjunction with code which might be setting values on the session independently from the auth package. This PR changes the behaviour to always return the original session from the wrapped `get` function, and only ever augment it with `data` where appropriate.

### DIFF
--- a/.changeset/shaggy-mayflies-fry.md
+++ b/.changeset/shaggy-mayflies-fry.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Updated `withItemData` to retain the original session data, even if item data was not found.
--- a/packages-next/keystone/src/session/index.ts
+++ b/packages-next/keystone/src/session/index.ts
@@ -97,7 +97,7 @@ export function withItemData(
           !session.itemId ||
           !sudoContext.lists[session.listKey]
         ) {
-          return;
+          return session;
         }
 
         // NOTE: This is wrapped in a try-catch block because a "not found" result will currently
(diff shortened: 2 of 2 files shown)


emmatown (changes requested):
This isn't right, I'm busy rn, will explain tomorrow

mitchellhamilton:
This PR essentially reverts #5168. This would bring back the problem of "phantom sessions" where the item that you say you're authenticating as doesn't really exist anymore and the Admin UI lets you in but you don't see a button sign out and it's extremely confusing for users. That could be solved by telling people to always explicitly add an isAccessAllowed but IMO, withItemData saying essentially "there's no session here" when the session doesn't point to an item is correct behavior and more expected than "here's a session which seems like it might be valid but actually it isn't".  

> This makes it impossible to use withItemData in conjunction with code which might be setting values on the session independently from the auth package

It's not impossible currently and this fix wouldn't really make it better IMO since if you want to have some other session thing, you'll also want to hydrate the session with some data, make sure it's valid, etc.

You could do this:

```js
const sessions = statelessSessions({ ...etc })();

const sessionsWithItemData = withItemData(() => sessions, fieldSelections)();

config({
  session: () => ({
    ...sessions,
    async get({ req, createContext }) {
      if (req.headers.authorization) {
        // check to see if some header is correct and return an appropriate thing
      }
      return sessionsWithItemData.get({ req, createContext });
    },
  }),
});
```

What we should probably do though is remove the function that returns a session strategy and just have the config accept a session strategy directly, the extra function just makes everything weird. The reasoning behind having the function didn't really pan out (it was about not erroring in stuff like build and postinstall where we didn't want to require that you pass in a valid session store or etc. but we always need to get the session strategy anyway so it's not providing any value and the problem that it was trying to solve can be solved in other ways).

PR #5805, opened 2021-05-27
Fix images and files baseUrl bug

gwyneplaine:
Thanks @bladey  for the fix. 
Uses the `baseUrl` specified in the images/files config in the symlink to the storage directory for both images and files.

### DIFF
--- a/.changeset/curvy-stingrays-smell.md
+++ b/.changeset/curvy-stingrays-smell.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fixed baseUrl bug for files and images config. BaseUrl now correctly referenced in uploaded image and file paths.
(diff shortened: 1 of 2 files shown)


mitchellhamilton (changes requested):
The current behavior was intentional, the baseUrl is about constructing the urls for the `src` field, the user is responsible for getting the images available at that url. The fact that the Admin UI happens to serve the files at the same url is really a coincidence(and maybe the Admin UI should put them on a more obscure place to make that clear) and the Admin UI probably shouldn't use the url from `src`.

@JedWatson do you think that the "it is completely the users responsibility to get the images/files available at that url"  is 100% the behavior we want? (not do you think we should merge this - we definitely shouldn't since the baseUrl might not be a valid path so this change wouldn't solve for the original things we were trying to solve for)

PR #5806, opened 2021-05-27
Add sessionData option to createAuth

timleslie:
This makes the common case of wanting to populate session data based on the authentication mutations generated by `createAuth()` much simpler. Most users will no longer need to extend the session strategies used 👍

### DIFF
--- a/packages-next/auth/src/types.ts
+++ b/packages-next/auth/src/types.ts
@@ -48,6 +48,8 @@ export type AuthConfig<GeneratedListTypes extends BaseGeneratedListTypes> = {
   passwordResetLink?: AuthTokenTypeConfig;
   /** "Magic link" functionality */
   magicAuthLink?: AuthTokenTypeConfig;
+  /** Session data population */
+  sessionData?: { query: string };
 };
 
 export type InitFirstItemConfig<GeneratedListTypes extends BaseGeneratedListTypes> = {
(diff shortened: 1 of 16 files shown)


packages-next/auth/src/types.ts
```diff
@@ -48,6 +48,8 @@ export type AuthConfig<GeneratedListTypes extends BaseGeneratedListTypes> = {
   passwordResetLink?: AuthTokenTypeConfig;
   /** "Magic link" functionality */
   magicAuthLink?: AuthTokenTypeConfig;
+  /** Session data population */
+  sessionData?: { query: string };
```
JedWatson, line 52: This seems like weird / unnecessary nesting.

I know it's clear, because this is the `query` that populates `session.data` and it has all those words in it, and I see the similarity to the items API, but I also can't imagine anything else that might go in this object... so why nest it?

I think I'd prefer just

```suggestion
  sessionData?: string;
```

JedWatson (commented):
Not sure this is a blocking comment, but I'm not sure I'm 💯 there with this change.

tbh it feels "weirder". The whole positioning of `withAuth` is that it's configuring the _authentication_ stuff... I know that it probably doesn't make sense to use the item data functionality unless you're also using the item authentication stuff but it just really breaks what previously felt very clear and clean in my head.

You're now configuring session with auth, not with session, and I'm not sure I'm a fan looking at it played out. Worth a chat maybe.

JedWatson (approved):

PR #5811, opened 2021-05-28
Fix/popover a11y

gwyneplaine:
Fix popover bug that meant we were able to navigate to the dialog even when its closed. 
Screen readers were also able to access the hidden dialog. 

Screen Shot 2021-05-28 at 10 24 32 am


timleslie (approved):
👍

Issue #5652, opened 2021-05-10
Playground is not disabled for production


mahyarkarimi:
GraphQL playround for production is not disabled and either the apollo config passed to the graphql in config API does not take effect. I also tried following config to disable it at all but not working. here is my config for graphql:
...
graphql:{
    apolloConfig :{
      playground: false,
    }
},
...

Thanks.

bladey:
Hi @mahyarkarimi, thanks for the report, I'm looking into this.

bladey:
@mahyarkarimi confirmed that this is a bug we'll look into.

bladey:
Completed.

PR #5817, opened 2021-05-28
Fix playground logic

bladey:
The graphql playground `/api/graphql` isn't setup properly based on the options passed in the Keystone config, including it being available by default in `production`. This hardens up the logic and makes it work in all test cases.

```
// graphql.apolloConfig.playground === false (playground not accessible in all cases)
// graphql.apolloConfig.playground === true (playground accessible in all cases)
// graphql.apolloConfig.playground === { settings: ... } (playground accessible in all cases with further customisation - https://www.apollographql.com/docs/apollo-server/testing/graphql-playground)
// process.env.NODE_ENV === 'production' (playground not accessible in production)
// N/A (playground uses defaults)
```


bladey:
#5652

examples/playground/README.md
```diff
@@ -0,0 +1 @@
+
```
timleslie, line 1: Could you move this example into `examples-staging` for now 👍

packages-next/keystone/src/lib/server/createApolloServer.ts
```diff
@@ -62,16 +62,26 @@ const _createApolloServerConfig = ({
   graphQLSchema: GraphQLSchema;
   apolloConfig?: Config;
 }) => {
-  // Playground config
+  // Playground config, is /api/graphql available?
   const pp = apolloConfig?.playground;
   let playground: Config['playground'];
   const settings = { 'request.credentials': 'same-origin' };
+
+  // graphql.apolloConfig.playground === false (playground not accessible in all cases)
```
timleslie, line 70: Could you move all the comments so they live inside the `if() {}` block that they correspond to, rather than above the `if` statement. It's a little confusing in the current state.

bladey: Done.

timleslie (approved):
This looks right to me now 👍

PR #5823, opened 2021-05-30
Fix typo in db items api docs

gabrielkuettel:
The sample code for `createMany` is using `createOne` instead of `createMany`.


timleslie (approved):
Good catch, thanks for the PR!

PR #5831, opened 2021-05-31
Remove use of context.keystone in document fields

timleslie:


emmatown (changes requested):
Can you fix the type in `KeystoneContext` rather than casting?

emmatown (approved):

PR #5833, opened 2021-05-31
Move getGqlNames out of the types package

timleslie:
This function doesn't belong in the `types` package. Also updates its signature to what's in the `next-fields` branch.


emmatown (changes requested):
You need to remove itemQueryType(and maybe rename listQueryName to pluralName) from the GraphQL config and make this a major for the keystone package, and pass the listKey to the base keystone class as the itemQueryName

JedWatson (approved):

PR #5842, opened 2021-06-01
Added status callout to NEW-EXAMPLES.md

raveling:


raveling:
I added this notice as @timleslie mentioned we're not ready for that yet. So wanted to make this obvious to the reader seeing as we're going to announce it.

JedWatson (approved):

PR #5845, opened 2021-06-01
Fixing a typo in the default-values example schema

JedWatson:
Small scout fix I noticed while looking through our new examples


timleslie (approved):
👍

PR #5851, opened 2021-06-02
Add a guide on using GraphQL filters

timleslie:


gautamsi:
Isn't this changing soon? I mean the filter syntaxes with new fields??

timleslie:
> Isn't this changing soon? I mean the filter syntaxes with new fields??

Yep, the filters will be changing, but the shape of these documents will remain largely the same.

JedWatson:
Looking good @timleslie

dominikwilkowski (approved):
Looks great. Good start

raveling (approved):
comments sent in Slack ❤️

PR #5852, opened 2021-06-02
A11y/local storage fields

gwyneplaine:
Fixing labels for screen-reader access in the `file` and `image` fields. 
Similar to previous fixes, in this PR we:
* Ensure that the field is wrapped in a `fieldset` element, 
*  provided a `legend` for labelling. 
* Given visually hidden prompts where appropriate.


packages-next/fields/src/types/file/views/Field.tsx
```diff
@@ -22,44 +22,53 @@ export function validateRef({ ref }: { ref: string }) {
...
+      <VisuallyHidden as="label" htmlFor={`${field.path}--ref-input`}>
+        Paste the file ref here
+      </VisuallyHidden>
+      <Stack
+        gap="small"
+        across
         css={{
           width: '100%',
+          justifyContent: 'space-between',
+          'div:first-of-type': {
```
mitchellhamilton, line 46: which element is this targeting(and the one in the image field) and can you have the style the closer to where it is?

gwyneplaine: @mitchellhamilton not a code addition from this PR, (highlighted because we moved some code around above), however I'd hazard a guess that this is here because our stack component wraps all children in a partitioning div (since we couldn't rely on `gap` for all browsers at the time of creating that component), *looks* like its targeting the textinput

mitchellhamilton: oh ok, nvm then.

emmatown (approved):

PR #5854, opened 2021-06-03
Cloud images

rohan-deshpande:
### Added 

* Cloud config and cloud module for integrating images in Keystone projects with Keystone Cloud 
* Examples for both local and cloud assets (currently just showing images, but will add files in a subsequent PR)


packages-next/keystone/src/lib/cloud/assets.ts
```diff
@@ -0,0 +1,92 @@
+import { Readable } from 'stream';
+import fetch from 'node-fetch';
+import FormData from 'form-data';
+import { ImageMetadata } from '@keystone-next/types';
+
+const CLOUD_IMAGES_DOMAIN = 'imgix.net';
+const CLOUD_GRAPHQL_API_ENDPOINT = 'http://localhost:4000/api/graphql';
+const CLOUD_REST_API_ENDPOINT = 'http://localhost:4000/api/rest';
```
mitchellhamilton, line 8: These should point to live deployed versions of these before we merge this and ideally we shouldn't include imgix in the domain, it should go through some domain of ours.

mitchellhamilton (changes requested):

packages-next/fields/src/types/image/views/Field.tsx
```diff
@@ -204,12 +202,7 @@ function ImgView({
       {errorMessage === undefined ? (
         value.kind === 'from-server' ? (
           <ImageWrapper>
-            <NextImage
-              height={value.data.height}
-              width={value.data.width}
-              src={value.data.src}
-              alt={field.path}
-            />
+            <img css={{ width: '100%' }} src={value.data.src} alt={field.path} />
```
timleslie, line 205: This change looks valid (?), but orthogonal to the cloud stuff. Can we put it up as a separate PR for review with its own changeset.

examples/assets-cloud/.env.example
```diff
@@ -0,0 +1 @@
+KEYSTONE_CLOUD_API_KEY=KEYSTONE_CLOUD_API_KEY
```
timleslie, line 1: Can we move the examples into the `examples-staging` directory for now? The `examples` directory is for our "blessed" examples that have been developed according to the details of `NEW-EXAMPLES.md`. Also, let's delete the `yarn.lock` files

packages-next/types/src/config/index.ts
```diff
@@ -22,7 +23,8 @@ export type KeystoneConfig = {
   db: DatabaseConfig;
   ui?: AdminUIConfig;
   server?: ServerConfig;
-  session?: SessionStrategy<any>;
+  session?: () => SessionStrategy<any>;
+  cloud?: CloudConfig;
```
timleslie, line 27:
```suggestion
  session?: SessionStrategy<any>;
  cloud?: CloudConfig;
```

packages-next/fields/src/types/image/views/Field.tsx
rohan-deshpande: This needs to be done in order for the cloud images to work in the admin UI

examples/assets-cloud/.env.example
rohan-deshpande: Okay!

packages-next/types/src/config/index.ts
rohan-deshpande: Weird I don't remember doing that change... will do as suggested

timleslie:
Let's move the cloud config into the experimental section of the config until it's ready for prime-time. That way we can make things like the endpoints, etc, configurable for now to facilitate dev until the server is ready to go:

```
  experimental: {
    cloud: {
      apiKey: KEYSTONE_CLOUD_API_KEY,
      endpoint: ...,
      ...,
    },
  }
```

timleslie (approved):
👍

packages-next/keystone/package.json
```diff
@@ -30,6 +30,7 @@
     "@graphql-tools/merge": "^6.2.14",
     "@graphql-tools/schema": "^7.1.5",
     "@hapi/iron": "^6.0.0",
+    "@imgix/js-core": "3.1.3",
```
mitchellhamilton, line 33: (and the lockfile changes)
```suggestion
```

mitchellhamilton (approved):

PR #5855, opened 2021-06-03
assorted relationship field fixes

gwyneplaine:
* Replaced `div` FieldContainer with `fieldset` element
* Added flex styling to inline-connect such that the select has more space to sit. 
  * before Screen Shot 2021-06-03 at 4 03 10 pm
  * after Screen Shot 2021-06-03 at 4 02 58 pm
* Card view is now an unordered list for better SR experience (and semantics) 
* Added VisuallyHidden headings to mark the beginning of each new list item.
* Added role="status" to CardContainer to help with announcing dom changes when a card exits and enters `edit` mode. 

New screen reader experience below:

Uploading relationship-field-updates.mov…


packages-next/fields/src/types/relationship/views/cards/index.tsx
```diff
@@ -123,125 +133,150 @@ export function Cards({
...
+      </Stack>
       {onChange === undefined ? null : field.display.inlineConnect && showConnectItems ? (
         <CardContainer mode="edit">
-          <Stack gap="small" marginY="medium" across>
+          <Stack
+            gap="small"
+            marginY="medium"
+            across
+            css={{
+              width: '100%',
```
gwyneplaine, line 273: Same fix as in the images and file field. 
We don't have a good way to individually alter the flex styles of children since each child is programatically wrapped in an element, so we have to do it here.

packages-next/fields/src/types/relationship/views/index.tsx
```diff
@@ -113,7 +113,7 @@ export const Field = ({
 
   if (value.kind === 'cards-view') {
     return (
-      <Stack as="fieldset" gap="medium">
+      <FieldContainer as="fieldset">
```
gwyneplaine, line 116: the `legend` element expects to have a `fieldset` element as its immediate parent. 
the `Stack` component negates this via its inherent behaviour (wrapping each child in a pre-styled element)

packages-next/fields/src/types/relationship/views/index.tsx
```diff
@@ -143,7 +143,7 @@ export const Field = ({
   }
 
   return (
-    <FieldContainer>
+    <FieldContainer as="fieldset">
```
gwyneplaine, line 146: see above

packages-next/fields/src/types/relationship/views/cards/index.tsx
```diff
@@ -123,125 +133,150 @@ export function Cards({
...
+                  onChange({
+                    ...value,
+                    itemsBeingEdited,
+                  });
+                }}
+              />
+            </CardContainer>
+          ) : (
+            <CardContainer mode="view" key={id}>
+              <VisuallyHidden as="h2">{`${field.label} ${index + 1}`}</VisuallyHidden>
```
gwyneplaine, line 182: Adding a visually hidden heading to demarcate and differentiate list item content to screen readers.

packages-next/fields/src/types/relationship/views/cards/index.tsx
```diff
@@ -123,125 +133,150 @@ export function Cards({
...
-                  View {foreignList.singular} details
-                </Button>
-              )}
-            </Stack>
-          </CardContainer>
-        );
-      })}
+          return value.itemsBeingEdited.has(id) && onChange !== undefined ? (
+            <CardContainer mode="edit" key={id}>
+              <VisuallyHidden as="h2">{`${field.label} ${index + 1}: Edit mode`}</VisuallyHidden>
```
gwyneplaine, line 152: Adding a visually hidden heading to demarcate and differentiate list item content to screen readers. 
specifically we're denoting that an element has entered 'edit' mode here.

packages-next/fields/src/types/relationship/views/cards/index.tsx
```diff
@@ -123,125 +133,150 @@ export function Cards({
...
-                    return (
-                      <FieldContainer>
-                        <FieldLabel>{field.label}</FieldLabel>
-                        {errorMessage}
-                      </FieldContainer>
-                    );
-                  }
-                  itemForField[graphqlField] = fieldGetter.data;
-                }
+      <Stack
```
gwyneplaine, line 136: We only want to wrap the list items in a `ul`, the actions for creating a new item / inline connection to an existing item should sit outside of this list.

packages-next/fields/src/types/relationship/views/cards/index.tsx
```diff
@@ -123,125 +141,153 @@ export function Cards({
...
+          li: {
+            listStyle: 'none',
+          },
+        }}
+      >
+        {[...value.currentIds].map((id, index) => {
+          const itemGetter = items[id];
+          const isEditMode = !!(onChange !== undefined) && value.itemsBeingEdited.has(id);
+          return (
+            <CardContainer role="status" mode={isEditMode ? 'edit' : 'view'}>
```
gwyneplaine, line 159: persistent card container now so we can append a role status, and announce when a card enters edit mode and vice versa.

emmatown (approved):

PR #5857, opened 2021-06-04
Display image in relationship UI

bladey:
Local images aren't showing up when displayed as part of a relationship preview, only when you edit that item.

It seems to only expect a Cloudinary (?) `publicUrlTransformed`.

Example of issue:

Kapture 2021-06-04 at 14 43 52

Changing it to `src` OR `publicUrlTransformed` fixes the issue, but it also needed a styled container as well that exists in `packages-next/fields/src/types/image/views/Field.tsx`, which has been copied, otherwise the image is full width with no styling or size restrictions.

Preview of changes:

Kapture 2021-06-04 at 14 42 53

### DIFF
--- a/packages-next/fields/src/types/image/views/index.tsx
+++ b/packages-next/fields/src/types/image/views/index.tsx
@@ -37,11 +37,7 @@ export const CardValue: CardValueComponent = ({ item, field }) => {
       <FieldLabel>{field.label}</FieldLabel>
       {data && (
         <ImageWrapper>
-          <img
-            css={{ width: '100%' }}
-            alt={data.filename}
-            src={data.src}
-          />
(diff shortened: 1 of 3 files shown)


gwyneplaine (approved):

packages-next/fields/src/types/image/views/index.tsx
```diff
@@ -125,3 +131,29 @@ export const controller = (config: FieldControllerConfig): ImageController => {
     },
   };
 };
+
+// ==============================
+// Styled Components
+// ==============================
+
+const ImageWrapper = ({ children }: { children: ReactNode }) => {
```
mitchellhamilton, line 139: Could you export this from the Field file rather than duplicating it here?

emmatown (changes requested):

emmatown (approved):

PR #5865, opened 2021-06-07
Update readmes for better rendering on the web

timleslie:
The wording in our package READMEs is a bit out of date and doesn't render well on aggregation sites. For example:

https://changelogs.xyz/@keystone-next/keystone

image

https://deps.dev/npm/%40keystone-next%2Fkeystone

image

This PR changes the content to have a complete URL rather than a link in the opening paragraph, and updates the content to be more consistent with the present reality.


raveling (approved):
Nice one!

PR #5874, opened 2021-06-08
Remove unused utils

emmatown:
We don't need these, they're mostly weird abstractions that can be replaced with using Object.entries/fromEntries/assign directly


timleslie (approved):
👍

PR #5884, opened 2021-06-08
Fix password field things

emmatown:
No changeset since this isn't new behavior, it's just adding back behavior that I missed in #5665

### DIFF
--- a/docs/pages/apis/fields.mdx
+++ b/docs/pages/apis/fields.mdx
@@ -296,6 +296,7 @@ Options:
 
 - `isRequired` (default: `false`): If `true` then this field can never be set to `null`.
 - `minLength` (default: `8`): The minimum length password accepted.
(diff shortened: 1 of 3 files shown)


molomby (approved):
Hell yes

docs/pages/apis/fields.mdx
```diff
@@ -310,6 +311,7 @@ export default config({
         fieldName: password({
           isRequired: true,
           minLength: 10,
```
molomby, line 313: Maybe we should drop this [back to 8](https://github.com/keystonejs/keystone-5/blob/main/packages/fields/src/types/Password/Implementation.js#L25) now that `rejectCommon` back and defaults to being enabled? Length is a lot less critical if we're blocking dumb stuff.

```suggestion
          minLength: 8,
```

mitchellhamilton: The default is 8, that is showing an example of configuring it differently

PR #5885, opened 2021-06-09
Fix orderBy in generated TS types

emmatown:
Also made one of the things in examples-staging use it so that it's tested and we don't break it again (which requires we run keystone-next postinstall for that example so tsc passes)


timleslie (approved):
👍

PR #5886, opened 2021-06-09
Fix/segmented controls

gwyneplaine:
* add a *visible* focus style 
* initialIndex is now -1 so that the first item doesn't appear selected  when no items are selected. 

https://user-images.githubusercontent.com/12119389/121443720-61238e00-c9d1-11eb-819a-d158682ebd9d.mov


design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -89,12 +90,18 @@ export const SegmentedControl = ({
   useEffect(() => {
     if (animate && rootRef.current instanceof HTMLElement) {
       let nodes = Array.from(rootRef.current.children);
-      let selected = nodes[selectedIndex];
+      let selected = selectedIndex !== undefined && nodes[selectedIndex];
```
gwyneplaine, line 93: now that -1 is a potential value for selectedIndex, ensure we account for when selected is `undefined`

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -107,8 +114,26 @@ export const SegmentedControl = ({
...
+      css={css`
+        outline: 0;
+        box-sizing: border-box;
+      `}
+      {...props}
+    >
+      <Root
+        css={css`
+          border: 1px solid #e1e5e9;
+          &:focus-within {
```
gwyneplaine, line 127: This will work for most browsers, exception being `opera` certain `android` mobile browsers and `ie11`. 
I'm fine to not have a work around for these atm, but happy to revisit this decision if need be.

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -199,16 +227,22 @@ const Item = (props: ItemProps) => {
         textAlign: 'center',
         position: 'relative',
         zIndex: 2,
-
+        border: '1px solid transparent',
+        ':focus-within': {
+          boxShadow: '0 0 0 2px #bfdbfe;',
```
gwyneplaine, line 232: ideally all these new values should be tokenized, but now is not the time.

emmatown (changes requested):
Could you make the initialIndex default to undefined rather than -1?

Actually, let's chat for a sec after standup

gwyneplaine:
> Could you make the initialIndex default to undefined rather than -1?

@mitchellhamilton It was initially when i made the change, but the onChange usage would change from 
```
        <SegmentedControl
          segments={field.options.map(x => x.label)}
          selectedIndex={value ? field.options.findIndex(x => x.value === value.value) : undefined}
          onChange={index => {
            onChange?.(field.options[index]);
          }}
        />
```

to 
```
        <SegmentedControl
          segments={field.options.map(x => x.label)}
          selectedIndex={value ? field.options.findIndex(x => x.value === value.value) : undefined}
          onChange={index => {
             if (index) {
                onChange?.(field.options[index]);
             } else {
                onChange?.(null)
             }
          }}
        />
```

Retaining the former experience seems better, and it's also relatively sensible in the context of an array. 
I don't feel strongly about it though, if the handler change is something we're ok with I'm happy to change it back to `undefined`. 

Additionally changing initialIndex to `undefined` also means that our `useManagedState` hook, will need to be able to accept a conditional `defaultValue`, is this fine (unsure on the context behind that hook)?

emmatown (approved):

PR #5891, opened 2021-06-11
Support uniquely filtering by text and integer fields

mitchellhamilton:
This is another part of #5635. This will allow you to get a single post if it has a unique slug for example:

```graphql
query {
  Post(where: { slug: "something-something-something" }) {
    id
    title
    content
  }
}
```


.changeset/twelve-guests-flow.md
```diff
@@ -0,0 +1,17 @@
+---
+'@keystone-next/fields': major
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+Added support for filtering uniquely by text and integer fields that have `isUnique: true` like this:
```
timleslie, line 7:
```suggestion
Added support for filtering uniquely by `text` and `integer` fields that have `isUnique: true` like this:
```

timleslie (approved):
Looks good 👍

PR #5897, opened 2021-06-14
Fix error in document guide

borisno2:
Add `NotEditable` import to component-blocks.tsx in document guide to fix "NotEditable" not found when following document guide


timleslie (approved):
Good catch, thanks for the PR 🙏

PR #5906, opened 2021-06-15
Virtual field docs

emmatown:
Not ideal I don't think but want @timleslie's thoughts

### DIFF
--- a/docs/pages/guides/virtual-fields.mdx
+++ b/docs/pages/guides/virtual-fields.mdx
@@ -42,7 +42,7 @@ We can now run a GraphQL query and request the `hello` field on one of our `Exam
 
 ```graphql
 {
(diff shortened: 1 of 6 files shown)


JedWatson (approved):
Looks great! 🚢

docs/pages/guides/virtual-fields.mdx
```diff
@@ -1,8 +1,318 @@
...
+    }),
+  }),
+});
+```
+
+We can now run a GraphQL query and request the `hello` field on one of our `Example` items,
+
+```graphql
+{
+  Example(id: "1") {
```
mitchellhamilton, line 45:
```suggestion
  Example(where: { id: "1" }) {
```

PR #5907, opened 2021-06-15
Fix types of `lists` and `db.lists` APIs on `KeystoneContext`

emmatown:
TypeScript doesn't constrain the type passed into a type param to the constraint of the type param so the `any` currently gets propagated and the usages of the type params become `any` so before this PR, the arg to `context.lists.Something.findMany` was just any. This change of course doesn't make it the specific where and orderBy and etc. but it means you know that it accepts first, skip, where and etc.


timleslie (approved):
👍

PR #5910, opened 2021-06-15
Fix generated list types to allow passing a value directly when a GraphQL list of the value is expected

mitchellhamilton:
This is fix so that #5907 can happen


timleslie (approved):
👍

PR #5920, opened 2021-06-17
Fix Admin UI generation to allow files returned from `getAdditionalFiles` to overwrite the files generated by Keystone

emmatown:
The original intent of this was to allow overwriting the files that Keystone generates but the comparison was broken because it was comparing unresolved paths with resolved ones.


timleslie (approved):
👍

gautamsi:
@mitchellhamilton does this also overwrite the file with typescript? I have noticed that f you have same file name created as custom pages in typescript, it does copy and use Typescript file but the js file from template is still there.

PR #5921, opened 2021-06-17
More popover and filter dialog fixes

gwyneplaine:
* Popover now completely obscures navigation to other dom elements via SR when visible. 
* Filter popover now has additional more informative `aria-label` and `aria-description`.

Before:

https://user-images.githubusercontent.com/12119389/122331803-c1d63c00-cf78-11eb-9ce9-af64c2d92e8f.mov

After:

https://user-images.githubusercontent.com/12119389/122331791-bbe05b00-cf78-11eb-9206-a2937948b721.mov

Screen Shot 2021-06-18 at 2 20 14 pm


packages-next/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/SortSelection.tsx
```diff
@@ -43,7 +43,14 @@ export function SortSelection({ list }: { list: ListMeta }) {
         </span>
       </Button>
 
-      <PopoverDialog arrow={arrow} isVisible={isOpen} {...dialog.props} ref={dialog.ref}>
+      <PopoverDialog
+        aria-label="Sort options dialog"
```
mitchellhamilton, line 47: q: when a screen reader will announce this, won't they already say that this is a dialog because of other aria attributes or no?

gwyneplaine: that's a good point,
```suggestion
        aria-label="Sort options"
```

gwyneplaine:
@mitchellhamilton  i've updated the` aria-descriptions` and `aria-labels` to address your concern, let me know what you think

emmatown (approved):

PR #5922, opened 2021-06-17
fix reset to default behaviour

gwyneplaine:
This was broken by a patch update of next from 10.2.0 to 10.2.1 likely a side effect of the following PR 
https://github.com/vercel/next.js/pull/24199 

The code has now been updated to include the pathname to nullify the query, rather than instantiating null as the query object (as was previously allowed)

https://user-images.githubusercontent.com/12119389/122338183-b38d1d80-cf82-11eb-8186-1a06182c17e0.mov


packages-next/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -80,9 +80,11 @@ function useQueryParamsFromLocalStorage(listKey: string) {
   const storeableQueries = ['sortBy', 'fields'];
 
   const resetToDefaults = () => {
+    console.log('we are being called here', `keystone.list.${listKey}.list.page.info`);
+    console.log(localStorage.getItem(`keystone.list.${listKey}.list.page.info`));
```
mitchellhamilton, line 84:
```suggestion
```

packages-next/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -80,9 +80,11 @@ function useQueryParamsFromLocalStorage(listKey: string) {
...
     localStorage.removeItem(`keystone.list.${listKey}.list.page.info`);
     router.replace({
-      query: null,
+      pathname: router.pathname,
```
mitchellhamilton, line 87: Can you try this and see if it works?
```suggestion
      query: {},
```

emmatown (changes requested):

packages-next/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
gwyneplaine: @mitchellhamilton it does not :(

mitchellhamilton: That's interesting, I'm guessing that's what caused not being able to remove a filter sometimes.

emmatown (approved):

JedWatson:
Wow, wicked regression! good sleuthing @gwyneplaine

PR #5926, opened 2021-06-17
Add a @keystone-next/testing package

timleslie:
This PR adds the @keystone-next/testing package and removes the legacy @keystone-next/test-utils-legacy package.

The core ideas of these packages are quite similar, but the API of the new package has been greatly simplified with external consumers in mind. The new package should give a good out of the box experience to a developer of a Keystone project, while still being flexible enough to support all of our own testing use cases.

Documentation including both a guide and an example project are being completed in PR #5887 and will be published when the package is released so as not to have docs published before the package is available.


packages-next/testing/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  const isolateRuns = config.testing?.isolateRuns || false;
+  let artifactPath = path.resolve(config.testing?.artifactPath || '.keystone/tests');
+
+  const artifacts = await getCommittedArtifacts(graphQLSchema, config);
+  const hash = _hashPrismaSchema(artifacts.prisma);
+
+  if (isolateRuns) {
+    if (config.db.provider === 'postgresql') {
+      config.db.url = `${config.db.url}?schema=keystone-test-${hash.toString()}`;
+    }
```
mitchellhamilton, line 57: What's the purpose of this? It doesn't allow running tests concurrently since two tests might have prisma schemas with the same hash and we drop the whole database anyway so this just seems like it'll fill up your local database with a bunch of stuff rather than just filling up a single postgres schema with the way we have it currently?

emmatown (changes requested):

packages-next/types/src/config/index.ts
```diff
@@ -177,6 +178,13 @@ export type ImagesConfig = {
   };
 };
 
+// config.testing
+
+export type TestingConfig = {
+  isolateRuns?: Boolean;
+  artifactPath?: string;
+};
```
mitchellhamilton, line 186: What's the reasoning behind having these as part of the keystone config vs arguments to the test stuff?

packages-next/testing/src/index.ts
timleslie: This is largely an optimisation that might no longer be needed. When working with `knex` the step of recreating the schema itself was extremely slow, so recreating new schemas for each different test made local testing of the keystone project really painful when running locally. I haven't done any experiments since we moved to prisma, so this may no longer be such an issue. Would need to test it out.

mitchellhamilton: Oh, sorry, I thought this wasn't here before, nvm then.

packages-next/types/src/config/index.ts
```diff
@@ -177,6 +178,13 @@ export type ImagesConfig = {
   };
 };
 
+// config.testing
+
+export type TestingConfig = {
+  isolateRuns?: Boolean;
```
mitchellhamilton, line 184: This name feels misleading since it's really "isolate unique prisma schemas" not isolating every run

packages-next/types/src/config/index.ts
```diff
@@ -177,6 +178,13 @@ export type ImagesConfig = {
...
+  artifactPath?: string;
```
mitchellhamilton, line 185: When would someone want to use or not use these options?

packages-next/types/src/config/index.ts
mitchellhamilton: Also, 
```suggestion
  isolateRuns?: boolean;
```

packages-next/types/src/config/index.ts
timleslie: To frame this a bit. I think this is a relatively marginal decision, I could go either way on it. I prefer this as it makes it feel like `testing` is a core part of Keystone, rather than a bolt on (even if that's the case from a technical point of view). This will nudge people to think more about testing when they see it's one of the things which is configurable in their Keystone system (even though almost everyone will be able to use the defaults).

packages-next/types/src/config/index.ts
timleslie: Yep, fair call, open to suggestions 🤔

packages-next/types/src/config/index.ts
timleslie: So this is captured in the docs (see other PR), but artifactPath is useful if you want to use a different config to your primary config when running your tests, and isolateRuns is useful if you have more than one config which you want to run tests against. I should include this info in the typescript doc-strings.

packages-next/types/src/config/index.ts
timleslie: `isolateSchemas`? `isolateConfigs`? I think I like `isolateConfigs`.

packages-next/types/src/config/index.ts
mitchellhamilton: What would the harm be in having isolateRuns always be true and it not being an option?

> artifactPath is useful if you want to use a different config to your primary config when running your tests

The place that it generates the stuff to defaults to `.keystone/tests` right so how does the config you use when running your tests affect anything?

timleslie: The motivation for `isolateRuns` as an option is that for the vast majority of users, they're going to want this option to be false. Setting it to true for those users would result in tests landing in a slightly unexpected location (`{path}/{hash}`), and will lead to us manipulating their database URL with the schema trick. I'd prefer to avoid having this behaviour unless it's actually necessary, such as in our own tests which use a bunch of different configs.

Good point re: `artifactPath` :-) I had it landing in the primary config location originally, forgot that I'd moved it. We hard-code a bunch of different paths into `.keystone` at the moment, don't we? If that's the case perhaps we can remove this as an option 🤔

mitchellhamilton: > tests landing in a slightly unexpected location ({path}/{hash}),

I don't really get why this matters, when will people need to see the files that are generated for their tests? imo we should just write the stuff inside the os temp dir so it gets cleaned up on restarts(and we could make it try to cleanup on process exit as well) which would also mean that artifactPath is unnecessary. And from #5928, it seems like the schema thing is unnecessary so we could also remove isolateRuns and just always generate the stuff into a temp dir(which would be different across prisma schemas) and we wouldn't need to modify their database URL

timleslie: I've done some local testing too and it looks like the schema modification is no longer getting us a win. It looks like the problem this was fixing (paying DB setup overhead when re-running tests locally) has returned in the form of Prisma client setup. A test run now looks something like this:

```
 PASS  tests/api-tests/relationships/crud/one-to-one.test.ts (40.605 s, 136 MB heap size)
  One-to-one relationships
    Read
      ✓ Where A (19182 ms)
      ✓ Where B (797 ms)
...
```
That 15-20s overhead in the first test is coming from the call to `getGenerator`. Ideally this function would recognise if the prisma client had already been generated and act as a no-op, so that subsequent runs of the test didn't have to wait for this to happen, however this doesn't seem to be the case. We should look into this, because fixing it will make running local tests way less painful.

For now I'm going to remove the schema modification stuff.

timleslie: With that gone, it makes setting `isolateRuns: true` as a fixed value less strange, as it doesn't come with the schema hack. Based on the discussion above, this also tips me in favour of hard-coding the location of the generated output to `.keystone/tests/{hash}`, since having a config block with only one option is not very appealing, and we can easily explain this one instance of the hash in the docs.

I'm going to keep the directory in `.keystone` rather than using a temporary directory. When debugging failing tests I find it very helpful to have them living locally, rather than in a system tmp file. I don't think we should be attempting to clean these up for people.

packages-next/testing/src/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export type GraphQLRequest = (arg: {
+  query: string;
+  variables?: Record<string, any>;
+  operationName?: string;
+}) => Test;
+
+export type TestArgs = {
+  context: KeystoneContext;
+  config: KeystoneConfig;
```
mitchellhamilton, line 24: Is this used and if so, why is this better than assigning the config to a variable and just using that?

timleslie: This is used to make the final config object used by the test runner available to the test, since we make changes to the value that's passed in. This was more useful when we were doing more manipulations to the config (e.g. iterating over providers, database schema, etc), but I think it's still helpful as is for cases where the original config object might not be easily available in the place where the test function is run.

That said I don't think we explicitly depend on it anywhere any more, so we could remove it if needed. I'm only marginally in favour of keeping it.

mitchellhamilton: Could we remove it then? And if we need it in the future, we can add it back?

emmatown (approved):

PR #5927, opened 2021-06-17
[POC] Remove the utils-legacy package

timleslie:
At this point this package only contains a couple of helpers for images and files. I don't think that justifies an entire package. I've moved the code into places that make the most sense to me as a POC. @mitchellhamilton perhaps you have some thoughts on how we should be sharing code across the `keystone` and `fields` packages?


mitchellhamilton:
imo leaving this package as it is is better than exporting some of these things from the `keystone` package since then breaking changes to these mean a major for the keystone package. the "share code but the sharing is just implementation details which aren't part of our public API of our main things" use case here is actually fine imo and this separate package isn't harming anything

timleslie:
Yep, that's fair. I think it also serves well as a holding place for other code that needs to be shared across packages in the future. In that case I think we should promote this package from `@keystone-next/utils-legacy` to `@keystone-next/utils` and move it into the `packages-next` directory.

mitchellhamilton:
Yeah, sgtm

PR #5940, opened 2021-06-21
Document field example

mitchellhamilton:
This doesn't show off all the ways you can use component blocks but we can add more component blocks or another example or something in the future.


gautamsi:
Can you add "New Post" page to showcase how to render the Document Field for getting input from user outside of admin ui.

PR #5947, opened 2021-06-22
New id fields

mitchellhamilton:
This replaces the idField option accepting regular field types to an object that can specify cuids, uuids or autoincrements. The default is now `cuid` instead of `autoincrement`


gautamsi:
Can we have id field config globally in keystone config?

I am thinking we would also want to map it for provider, once mongodb is there we can have mongoid or dbgenerated option.

packages-next/keystone/src/lib/id-field.ts
```diff
@@ -0,0 +1,79 @@
...
+        uniqueWhere: { arg: schema.arg({ type: schema.ID }), resolve: parseVal },
+        orderBy: { arg: schema.arg({ type: orderDirectionEnum }) },
+      },
+      output: schema.field({
+        type: schema.nonNull(schema.ID),
+        resolve({ value }) {
+          return value.toString();
+        },
+      }),
+      views: '@keystone-next/fields/types/integer/views',
```
timleslie, line 56: FIXME

packages-next/keystone/src/lib/id-field.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (Number.isInteger(parsed)) {
+      return parsed;
+    }
+    throw new Error('Only an integer can be passed to id filters');
+  },
+  cuid(val: string | null) {
+    // isCuid is just "it's a string and it starts with c"
+    // https://github.com/ericelliott/cuid/blob/215b27bdb78d3400d4225a4eeecb3b71891a5f6f/index.js#L69-L73
+    if (typeof val === 'string' && isCuid(val)) {
+      return val;
```
mitchellhamilton, line 34: Note not lower-casing here is a difference from what we discussed. My thinking is that isCuid only accepts strings that with `c`(as in  only lower-case `c`) so treating it as case insensitive seems strange.

docs/pages/docs/apis/config.mdx
```diff
@@ -71,6 +71,8 @@ Advanced configuration:
 
 - `enableLogging` (default: `false`): Enable logging from the Prisma client.
 - `useMigrations` (default: `false`): Determines whether to use migrations or automatically force-update the database with the latest schema and potentially lose data.
+- `idField` (default: `{ kind: "cuid" }`): The kind of id field to use, it can be one of: `cuid`, `uuid` or `autoincrement`
```
timleslie, line 74:
```suggestion
- `idField` (default: `{ kind: "cuid" }`): The kind of id field to use, it can be one of: `cuid`, `uuid` or `autoincrement`.
```

docs/pages/docs/apis/config.mdx
```diff
@@ -92,6 +95,8 @@ Advanced configuration:
...
```
timleslie, line 98:
```suggestion
- `idField` (default: `{ kind: "cuid" }`): The kind of id field to use, it can be one of: `cuid`, `uuid` or `autoincrement`.
```

.changeset/eight-lions-stare.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+Replaced the `idField` option with a more constrained option that an object with a kind property that is one of `cuid`, `uuid` or `autoincrement`. It's also inside of the `db` config, instead of at the root of the list config. The default is now to use cuids. To keep the current behaviour, you should set `{ kind: 'autoincrement' }` at `db.idField` at the root of your config.
```
timleslie, line 6:
```suggestion
Replaced the `idField` list configuration option with a more constrained option, `db.idField`, that accepts an object with a `kind` property with a value of `cuid`, `uuid` or `autoincrement`. `db.idField` can be set on either the top level `config` object, or on individual lists.

The default behaviour has changed from using `autoincrement` to using cuids. To keep the current behaviour, you should set `{ kind: 'autoincrement' }` at `db.idField` in your top level config.
```

timleslie (approved):
👍 pending changeset update.

PR #5948, opened 2021-06-22
first draft for /docs/updates page

raveling:
First draft ready for review.

Notes:

Now that i've written this first draft, I feel that "what's new in 6" is totally redundant, and  should have any relevant points merged into this timeline for efficiency's sake.  Past facing content like: 
 
> **All New Admin UI**
> Our Admin UI has been completely rewritten with Next.js behind the scenes and a new Design System. It's a lot faster, and more customisable than ever before.
> Among other improvements, the Keystone Authoring UX is now aware of your application permissions, so users get the right field interfaces based on whether they can view or edit a field.

Feels right at home in this timeline.

---

I reduced the lede content quite a bit. This kinda stuff:

> In Keystone 6, we've focused on improving Keystone's interfaces, including the way you configure and run Keystone projects, our GraphQL backend, and a whole new Admin UI. See Whats New in v6 to learn more.
> We're also adding powerful new features to make Keystone the best headless content management system around, especially when you're using a component-based front-end like React and Vue.

Feels like it would be more effective if future facing (in roadmap?

So `updates` shows you how we're implementing on the vision, and `roadmap` describes the vision? 

---

I'm a bit torn between using the "read more" link that's baked into the `<box>` component, and inline links. Feels like the latter is more effective without being able to modify the text within the "read more" link.

---
I've bundled a few things outside of their chronological order for efficiency, and to tell a more cohesive picture overall. People can still rely on content in /releases for a chronological accuracy?

Thx!

### DIFF
--- a/docs/pages/updates/index.tsx
+++ b/docs/pages/updates/index.tsx
@@ -1,6 +1,8 @@
 /** @jsx jsx  */
 import { jsx } from '@emotion/react';
 import Link from 'next/link';
+import { Button } from '../../components/primitives/Button';
+import { ArrowR } from '../../components/icons/ArrowR';
 import { ComponentProps, HTMLAttributes, ReactNode } from 'react';
 
 import { getServerSideProps } from '../../components/Markdown';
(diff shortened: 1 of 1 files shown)


raveling:
Resolved in #6003

PR #5949, opened 2021-06-22
Fix readonly property set in format-error of createApolloServer

### LINKED ISSUE #5898: Exception when formatting GraphQL error
Keystone attempts to write the following readonly property:

https://github.com/keystonejs/keystone/blob/master/packages-next/keystone/src/lib/server/format-error.ts#L67

Original declaration:

https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-errors/src/index.ts#L7

This causes the following error, which has appeared on Slack a few times:

    Cannot assign to read only property 'path' of object '#<Object>'

I haven't submitted a PR since it's not immediately clear why the `path` property needs to be set.

tjbp:
Fixes #5898 - simply removed the setting of a readonly property on the ApolloError object since I couldn't ascertain what its purpose was.

### DIFF
--- a/.changeset/clean-countries-drop.md
+++ b/.changeset/clean-countries-drop.md
@@ -1,0 +1,5 @@
+---
+"@keystone-next/keystone": patch
+---
+
+Fixed readonly property set in `format-error` of `createApolloServer`.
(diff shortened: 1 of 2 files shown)


timleslie (approved):
Thanks for the PR @tjbp. I think this is a good solution for now. We are planning to do an overhaul of our error handling system in the near future, but this is a good solve for the immediate problem 👍

PR #5958, opened 2021-06-23
Site TypeScript fixes

JedWatson:


docs/components/SkipLinks.tsx
```diff
@@ -24,36 +23,37 @@ function SkipLink({ href, children, ...props }) {
...
+          display: ['none', undefined, 'block'],
         }}
       >
         Skip to Page Navigation
       </SkipLink>
       <SkipLink
         href="#skip-link-navigation-btn"
         css={mq({
-          display: ['block', null, 'none'],
+          display: ['block', undefined, 'none'],
```
dominikwilkowski, line 54: uhm ... why not `null`? That's the pattern we use everywhere where we want to keep the same value for that breakpoint?

docs/components/marketing/EndCta.tsx
```diff
@@ -33,7 +33,6 @@ export function EndCta({ grad = 'grad1', ...props }: EndCtaProps) {
           display: 'flex',
           gap: '1rem',
           justifyContent: 'center',
-          gap: '1rem',
```
dominikwilkowski, line 36: I think you accidentally removed the gap here. Need it or it looks like this: 
image

docs/pages/index.tsx
```diff
@@ -48,7 +48,7 @@ export default function IndexPage() {
       <MWrapper>
         <IntroWrapper>
           <IntroHeading>
-            The <Highlight grad="grad1">superpowered</Highlight> CMS for developers
+            The <Highlight look="grad1">superpowered</Highlight> CMS for developers
```
dominikwilkowski, line 51: I fixed that already in my branch... why is it not picking that up?

docs/pages/index.tsx
```diff
@@ -650,7 +650,6 @@ export const lists = createSchema({
               display: 'flex',
               gap: '1rem',
               justifyContent: 'center',
-              gap: '1rem',
```
dominikwilkowski, line 653: What's with you removing gaps? Is Safari somehow being silly?

dominikwilkowski (changes requested):
There is something wrong with your branch. There are some fixes in there that I have already in my branch

docs/pages/index.tsx
dominikwilkowski: See here: https://github.com/keystonejs/keystone/blob/dom/site/docs/pages/index.tsx#L51

docs/components/icons/FrontEndLogos.tsx
```diff
@@ -25,7 +25,7 @@ export function FrontEndLogos({ grad, ...props }: IconProps) {
           <path fill="var(--app-bg)" d="M0 0h44.7v44.7H0z" transform="translate(504.3 .6)" />
         </clipPath>
       </defs>
-      <g clip-path="url(#FrontEndLogos-clip1)">
+      <g clipPath="url(#FrontEndLogos-clip1)">
```
dominikwilkowski, line 28: This was something I had fixed too but not pushed... We got it double fixed now :D

docs/pages/index.tsx
dominikwilkowski: No gap here means it looks like this:
image

JedWatson: It's a double property, see line 651

JedWatson:
Had to manually merge the changes made on your branch since this was created, should be up to date now. No conflicts 👍

docs/components/SkipLinks.tsx
JedWatson: Technically (according to emotion's types) `null` isn't a valid value for this property.

Guessing it's one of those "browsers are forgiving" things, if you throw it a value it doesn't recognise it'll ignore it?

@mitchellhamilton might have an insight I don't on this

docs/components/marketing/EndCta.tsx
JedWatson: That's really weird, I removed this because it's repeated (see line 34)

docs/components/SkipLinks.tsx
dominikwilkowski: Right. `null` isn't passed to the browser here. It just means the media query for this breakpoint is skipped. If `null` is not right we have to fix them all which should be a couple more places (I would estimate 2342348792834629346 places) ;)

docs/components/marketing/EndCta.tsx
dominikwilkowski: Oh didn't see the duplication. Was a long day :)

docs/pages/index.tsx
dominikwilkowski: nice find

docs/components/SkipLinks.tsx
JedWatson: OK, good catch, mitchell pointed out the problem here is actually the missing `mq` calls which I missed. Fixed now.

PR #5966, opened 2021-06-24
added svg versions of home page images

raveling:
proof of concept

if the SVGs are stable enough across all ouyr browser needs I'd replace everything in this branch

(these SVGS have had height/width params removed, and been optimized using SVGOMG)


JedWatson:
Closing this for further discussion later... SVGs are promising but have some complexities / downsides that we need to work through.

PR #5971, opened 2021-06-25
Update the document field guide

timleslie:
Updates the document field guide to split out the demonstration from the guide itself, and improves the usability of both the demo and the guide.

Having the demo in a separate place lets us link directly to the demo itself, giving people a chance to play with it without being overwhelmed by all the details of the guide. We CTA to the demo in a lot of places, so we expect a very different audience mix to the readers of the guide who are going to be devs who are ready to start using the document field in their project.

I've updated the content of the guide up to the `Component Blocks` section. I'd like to separate that out into its own guide, since it's an advanced use case, but that can happen in subsequent PRs.


bladey (approved):
👍

PR #5978, opened 2021-06-25
Docs: fixed why-keystone heading and first content block on small screen

mikehazell:
On the "Why keystone" page the heading was being clipped and the first content block was in 2 very narrow columns on smaller screens.

| Before | After |
|:----|:----|
| image |image|

### DIFF
--- a/docs/pages/why-keystone.tsx
+++ b/docs/pages/why-keystone.tsx
@@ -31,6 +31,7 @@ import { Shield } from '../components/icons/Shield';
 import { Watch } from '../components/icons/Watch';
 import { Cli } from '../components/icons/Cli';
 import { Page } from '../components/Page';
+import { IntroHeading } from '../components/content/Intro';
 
 import adminUi from '../public/assets/admin-ui.png';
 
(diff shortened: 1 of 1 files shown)


docs/pages/why-keystone.tsx
```diff
@@ -40,17 +40,17 @@ export default function WhyKeystonePage() {
   return (
     <Page>
       <MWrapper>
-        <Type as="h1" look="heading92">
+        <Type as="h1" look="heading92" fontSize={['4rem', '5.75rem']}>
```
mikehazell, line 43: I copied these sizes from `docs/components/primitives/Type.tsx`

Not sure if there's a more Design-Token-y way of writing this.

JedWatson: Arguably we should possibly do this for all the headings with this `look`?

It's defined here: https://github.com/keystonejs/keystone/blob/master/docs/components/primitives/Type.tsx#L68-L74

Or do you think this fix should be local to this page?

raveling (commented):
looks good!

docs/pages/why-keystone.tsx
raveling: I opted to use the `<IntroHeading>` component Dominik setup on other pages which has all the responsive behaviour baked in.
hope that's cool

raveling (approved):

PR #5984, opened 2021-06-25
Docs: responsive fixes on For Organisations page + grammar nits

mikehazell:
Two columns of text here feels like too many on mobile (For Organisations page)
Also, I picked some grammar nits.

| Before | After |
|:---|:---|
| image | image |


JedWatson (approved):
Thanks @mikehazell

PR #5987, opened 2021-06-25
assorted filter pill fixes

### LINKED ISSUE #5939: Filter items are not very SR friendly
# Bug report

## Describe the bug

At the moment screen readers announcing the filter item "pills" at the top of the list page only announce the contents of the pill. There's no context behind what they're used for, which makes it difficult for a screen reader user to understand what they are in the context of the rest of the page. 

## To Reproduce

1. Start up the basic example `yarn workspace @keystone-next/examples-app-basic dev` 
2. Open up the admin-ui on `http://localhost:3000`
2. Go to the User list page
3. Add a filter 
4. Open Voice over
5. Navigate to the newly added Filter pill
6. See error

## Expected behaviour

Screen reader should announce out the context behind the pill, something like `filter item: [name of filter]  [value]` when hovering over each filter item pill.

## Screenshots

https://user-images.githubusercontent.com/12119389/122710394-805cce00-d2a3-11eb-8784-377b5507c549.mov

gwyneplaine:
* resolves #5939 by adding aria-label and aria-description to pill 
* fixes autoFocus always being on in text field Filter component 
* revert redundant status="role" on editDialog 
* Added `containerProps` prop to Pill component in design-system that are spread onto the container element, props are now spread onto the PillButton element by default. 
* Ensured trigger.ref is applied to PillButton ref and not the pill container ref (which previously broke focus) 
* Fixed issue with filters not always being removable (due to undocumented breaking change with Next)


JedWatson (approved):
👍 Nice fixes

JedWatson (dismissed):
Just realised, this needs a changeset!

timleslie (approved):
👍

PR #5989, opened 2021-06-25
Site: enable Dark Mode defaulting based on OS settings

JedWatson:
With the improvements in #5988 I think we're good to enable defaulting to dark mode based on the OS settings.

🌃 

(leaving the feature flag for now in case we want to toggle this later)


mattbx (approved):
super awesome!

PR #5991, opened 2021-06-27
updated community CTA

mattbx:
Tweaked layout (mostly spacing and padding) of the Community CTA panel
also updated GLOBAL heading lineheights - making them slightly larger


timleslie (approved):
👍

PR #5993, opened 2021-06-28
removed "updates" links until content is differentiated from "releases

raveling:
Short term fix until we get the right content on `/updates`.

Roadmap is a child of /updates per breadcrumb and location in updates folder 👇 

CleanShot 2021-06-28 at 12 30 27

Maybe we can live with this for a few days?

### DIFF
--- a/docs/components/Footer.tsx
+++ b/docs/components/Footer.tsx
@@ -157,11 +157,12 @@ export function Footer() {
               Updates
             </Type>
             <List>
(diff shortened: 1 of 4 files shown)


JedWatson:
This is solved by #6003 and #6007 right?

raveling:
> This is solved by #6003 and #6007 right?

That's right.  👍

PR #5995, opened 2021-06-28
color contrast updates

mattbx:
fixing color contrasts on type

- darkening of our `muted` text color
- revising of light color themed font color snippets


timleslie (approved):
👍

PR #6011, opened 2021-06-28
Fix website page titles

JedWatson:
I just realised that most of the content pages on our new website don't have unique page titles - this fixes that

(Should do a better review than this, including the `- Keystone 6` that's automatically appended, but it's a start)


raveling:
great catch!

PR #6018, opened 2021-06-29
Secondary buttons

mattbx:
addition of new secondary buttons - sizing and coloring - currently only used in the footer for the mail sign up and on the docs homepage in the slack community CTA


JedWatson:
Visually I like it, but also for some reason it comes across as the button looking kind of disabled when all the other actions are so bright

mattbx:
> In isolation, visually I like it, but in context it comes across as the button looking kind of disabled when all the other actions are so bright

ok let me have a look at alternatives

dominikwilkowski (approved):

PR #6025, opened 2021-06-29
Docs pages fixes pt1

mattbx:
updates to docs pages:

- Increased spacing on left nav items
- indent of list items in left nav
- fixing body copy color
- a few general docs typography tweaks


docs/lib/prose-lite.ts
```diff
@@ -52,6 +53,7 @@ export const proseStyles = {
     borderRadius: '50%',
     width: '0.375em',
     height: '0.375em',
+
```
timleslie, line 56:
```suggestion
```

docs/components/Breadcrumbs.tsx
```diff
@@ -76,9 +76,10 @@ export function Breadcrumbs() {
             <Link href={href}>
               <a>
                 <Type
-                  look="body12"
+                  look="body14"
                   css={{
-                    textTransform: 'uppercase',
+                    textTransform: 'capitalize',
```
timleslie, line 81: I think we might need to be a bit more specific about how we generate our breadcrumbs. I don't think `capitalize` is quite what we want. Some of our titles have specific casing requirements, so we might be better off passing through explicit values in these cases

image
image

timleslie: I think our best bet here is to leave this as `uppercase` for now and revisit this in a separate PR.

timleslie (changes requested):

docs/components/Breadcrumbs.tsx
mattbx: ok damn didnt think about that! I was thinking about just leaving it as written in the markdown, but its not consistent there either... `uppercase` it is then

timleslie (approved):
👍

PR #6040, opened 2021-07-01
Fix behaviour of hydrateRelationships with dangling references

timleslie:
If an item referenced in a document field is deleted from Keystone, the document field should handle this gracefully. The current behaviour will return an error when querying the field, which prevents the user from being able to manage the issue. This PR fixes this to simply return `null` for the missing item. This also covers the behaviour for when a user doesn't have read access for an item.

We also need to make a distinction between `data.data === null` (object not found/access denied) and `data.data === {}` (no selection fields in the config) in the document renderer.

### DIFF
--- a/.changeset/rare-ducks-suffer.md
+++ b/.changeset/rare-ducks-suffer.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/document-renderer': patch
+---
+
+Inline relationship data which is null or undefined is now explicitly set to `null`, rather than `{}`.
--- a/.changeset/rare-ducks-suffer.md
+++ b/.changeset/rare-ducks-suffer.md
@@ -1,5 +1,6 @@
 ---
-'@keystone-next/document-renderer': patch
+'@keystone-next/document-renderer': major
 ---
 
-Inline relationship data which is null or undefined is now explicitly set to `null`, rather than `{}`.
+The value of `data` passed to the inline relationship renderer now matches the data returned by the GraphQL query.
+Falsey values of `data.label` are no longer set to `data.id` and falsey values of `data.data` are no longer set to `{}`.
--- a/packages-next/document-renderer/src/index.tsx
+++ b/packages-next/document-renderer/src/index.tsx
@@ -33,7 +33,7 @@ interface Renderers {
     link: Component<{ children: ReactNode; href: string }> | 'a';
     relationship: Component<{
       relationship: string;
-      data: { id: string; label: string; data: Record<string, any> | null };
+      data: { id: string; label: string; data: Record<string, any> | null } | null;
     }>;
   } & MarkRenderers;
   block: {
@@ -64,7 +64,7 @@ export const defaultRenderers: Renderers = {
     superscript: 'sup',
     underline: 'u',
     relationship: ({ data }) => {
-      return <span>{data.label}</span>;
+      return <span>{data?.label}</span>;
     },
   },
   block: {
@@ -187,11 +187,7 @@ function DocumentNode({
       return (
         <renderers.inline.relationship
           relationship={node.relationship as string}
-          data={{
-            id: data.id,
-            label: data.label || data.id,
-            data: data.data || null,
-          }}
+          data={data ? { id: data.id, label: data.label, data: data.data } : null}
         />
       );
     }
--- a/packages-next/document-renderer/src/index.tsx
+++ b/packages-next/document-renderer/src/index.tsx
@@ -33,7 +33,7 @@ interface Renderers {
     link: Component<{ children: ReactNode; href: string }> | 'a';
     relationship: Component<{
       relationship: string;
-      data: { id: string; label: string; data: Record<string, any> } | null;
+      data: { id: string; label: string; data: Record<string, any> | null };
     }>;
   } & MarkRenderers;
   block: {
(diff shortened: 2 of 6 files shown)


packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,16 +41,27 @@ export function addRelationshipData(
...
+            }(where: {id:$id}) {${labelFieldAlias}:${labelField}\n${
+              relationship.selection || ''
+            }}}`,
+            variables: { id },
+          });
+        } catch (err) {
+          // If we're unable to find the user (e.g. we have a dangling reference), or access was denied
+          // then simply return null.
+          console.error(`Unable to fetch relationship data: relationship: ${JSON.stringify(relationship)}, id: ${id} `);
+          console.error(err);
```
timleslie, line 62: I'm undecided as to whether these console error messages are helpful, or just noise 🤔

packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,16 +41,31 @@ export function addRelationshipData(
       }
       return [];
     } else {
+      // Inline
```
mitchellhamilton, line 44: This isn't right, this case also applies for non-many relationships in component block props.

packages-next/fields-document/src/relationship-data.tsx
mitchellhamilton: Yeah, i don't think they're helpful though could you look at what it looks like to only do this on access control/does not exist? (the easiest way might be to do a many query and just get the first one from the array and handle if the array is empty?)

packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,16 +41,31 @@ export function addRelationshipData(
...
+        } catch (err) {
+          // If we're unable to find the user (e.g. we have a dangling reference), or access was denied
+          // then simply return null.
+          console.error(
+            `Unable to fetch relationship data: relationship: ${JSON.stringify(
+              relationship
+            )}, id: ${id} `
+          );
+          console.error(err);
+          return null;
```
mitchellhamilton, line 67: Returning null here is a problem imo, if a user doesn't have access to the related item that would mean that loading the document and then saving it would remove the related item, could you return `{ id }` instead?

emmatown (changes requested):

packages-next/document-renderer/src/index.tsx
```diff
@@ -33,7 +33,7 @@ interface Renderers {
     link: Component<{ children: ReactNode; href: string }> | 'a';
     relationship: Component<{
       relationship: string;
-      data: { id: string; label: string; data: Record<string, any> } | null;
+      data: { id: string; label: string; data: Record<string, any> | null } | null;
```
mitchellhamilton, line 36:
```suggestion
      data: { id: string; label: string | undefined; data: Record<string, any> | undefined } | null;
```

.changeset/rare-ducks-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/document-renderer': patch
```
mitchellhamilton, line 2:
```suggestion
'@keystone-next/document-renderer': major
```

packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,30 +41,36 @@ export function addRelationshipData(
...
+            }}}`,
+            variables: { id },
+          });
+        } catch (err) {
+          // If we're unable to find the user (e.g. we have a dangling reference), or access was denied
+          // then simply return null.
+          const r = JSON.stringify(relationship);
+          console.error(`Unable to fetch relationship data: relationship: ${r}, id: ${id} `);
+          console.error(err);
+          return { id, data: null };
```
mitchellhamilton, line 64:
```suggestion
          return { id };
```

packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,30 +41,36 @@ export function addRelationshipData(
...
+        try {
+          val = await graphQLAPI.run({
+            query: `query($id: ID!) {item:${
+              relationship.listKey
+            }(where: {id:$id}) {${labelFieldAlias}:${labelField}\n${
+              relationship.selection || ''
+            }}}`,
+            variables: { id },
+          });
+        } catch (err) {
```
mitchellhamilton, line 58: The only ignore access control error thing we discussed

packages-next/fields-document/src/relationship-data.tsx
```diff
@@ -41,30 +41,36 @@ export function addRelationshipData(
       }
       return [];
```
mitchellhamilton, line 42: Need to handle the many case

.changeset/rare-ducks-suffer.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Inline relationship data which is null or undefined is now explicitly set to `null`, rather than `{}`.
```
mitchellhamilton, line 5: Needs to mention the label thing

emmatown (changes requested):

mitchellhamilton:
(oh sorry, missed that the time that the review was requested was before we talked)

timleslie:
> (oh sorry, missed that the time that the review was requested was before we talked)

All good, just working through things now. I'm gonna save the `many` changes until I tackle the component blocks as a whole. Just adding final tests for the non-expected error cases and then will clean up the PR.

timleslie:
@mitchellhamilton OK, Ready for re-review. The only thing I haven't covered here is the `many` case, but I plan to cover that in a follow up which covers the overall behaviour of the `prop` kind.

emmatown (approved):

PR #6043, opened 2021-07-01
Change default of hydrateRelationships from false to true

timleslie:
The common use case when querying a document field is to query the document and then hand it off to a `DocumentRenderer`. For inline relationships to work as expected with the document renderer we need to have `hydrateRelationships: true`. Having a default of `false` makes this more complicated than it needs to be. It also makes it more difficult to explain in the supporting documentation. Having a default of `true` makes the common case easy, and makes it easier to document how to use inline relationships in general.


mitchellhamilton (changes requested):
(just explicitly blocking this given our discussion)

PR #6045, opened 2021-07-02
Fix confusing error when providing fields that don't exist to `ui.cardFields`, `ui.inlineCreate.fields` and `ui.inlineEdit.fields` in relationship field

emmatown:


timleslie (approved):
👍

PR #6047, opened 2021-07-02
Dev startup performance experimenting

mitchellhamilton:


mitchellhamilton:
It seems as though the performance issue in Prisma no longer exists

PR #6048, opened 2021-07-02
Add ability to customize idField properties

MurzNN:
Very often users need to use custom id values for List items, but Keystone provides only auto-generated types at now. More detailed description about needing this feature - here: https://github.com/keystonejs/keystone/discussions/6034

This PR implements customization ability to idField:
- Override default settings of `db.idField` to change type of database field
- Set custom function for default value
- Set custom function for parse value
- Change access control to allow
- Change views to set custom widget for id field in Admin UI

Here is example of usage implemented features, based on "Country" list, that uses the `ISO 3166-1 alpha-2` code as id:
```js
import { createSchema, list } from '@keystone-next/keystone/schema';
import { text } from '@keystone-next/fields';
import { schema } from '@keystone-next/types';
import { resolveView } from '@keystone-next/fields/src/resolve-view';

  Country: list({
    fields: {
      name: text({ isRequired: true }),
    },
    db: {
      idField: {
        kind: 'cuid',
        experimental: { enableOverrides: true },
        override:
          parseVal: function (val: string | null) {
            return val;
          },
          dbField: {
            scalar: 'String',
            default: { kind: 'literal', value: 'ZZ' },
          },
          field: {
            access: {
              create: true,
            },
            getAdminMeta: () => ({ displayMode: 'input' }),
            input: {
              create: {
                arg: schema.arg({ type: schema.ID }),
                resolve: function (val: string | null) {
                  return val;
                },
              },
            },
            views: resolveView('text/views'),
            ui: {
              createView: {
                fieldMode: 'edit',
              },
            },
          },
        },
      }
    }
  }),
});
```
Screenshot from Admin UI:
image

Second simple example - using BigInt instead of Int as idField:
Example of using BigInt field as idField:
```js
    db: {
      idField: {
        kind: 'autoincrement',
        override: {
          dbField: {
            scalar: 'BigInt',
          },
        },
      },
    },
```


packages-next/keystone/src/lib/core/field-assertions.ts
```diff
@@ -13,9 +13,9 @@ export function assertFieldsValid(list: ListForValidation) {
 function assertUniqueWhereInputsValid(list: ListForValidation) {
   for (const [fieldKey, { dbField, input }] of Object.entries(list.fields)) {
     if (input?.uniqueWhere) {
-      if (dbField.kind !== 'scalar' || (dbField.scalar !== 'String' && dbField.scalar !== 'Int')) {
+      if (dbField.kind !== 'scalar') {
```
gautamsi, line 16: you may not need this

gautamsi:
We can add some experimental flag to have this work in case we do not want to change default behavior

this will definitely a good addition.

MurzNN:
> We can add some experimental flag to have this work in case we do not want to change default behavior

Yes, this is good idea, but where is the best place to add it? Maybe in the `idField` as boolean, something like `experimentalOverrides: true`?

packages-next/keystone/src/lib/core/field-assertions.ts
MurzNN: This fails, when dbField is, for example, BigInt, that is scalar too, but throws the error.

MurzNN:
I added `enableIdFieldOverrieds` config flag in `experimental` section, but can't find the way to read it in `packages-next/keystone/src/lib/id-field.ts` file funciton.

rachmanov:
I use Keystone in addition to other framework that solves specific tasks. In cases like this (when several systems shares one dataset) it would be great to have a feature like this, it helps to sync data management

MurzNN:
Another common pattern is to store user metadata in different Lists, but with same id as user.

For example, one user have profile as "student" (20 fields), second as "teacher" (30 fields), third as "admin" (15 fields):

Instead of extending List "user" with 65 fields, much better to create separate lists like userProfileStudent, userProfileTeacher, userProfileAdmin.

And very convenient when id of such profile items are exactly same as id of user item, so we can load profile directly from needed List, without more heavy query using relation to user.

molomby:
Thanks for the work here @MurzNN but this isn't a path we're interested in going down currently. We're aware of it though and it's possibly we'll revisit the functionality in future.

I've put a [more detailed response](https://github.com/keystonejs/keystone/discussions/6034#discussioncomment-1372545) in your related issue.

MurzNN:
Meanwhile I reworked this feature for new Keystone version and add an example with it, that implements BigInt id field type: https://github.com/MurzNN/keystone/commit/6204e87d3d32339d6aef6865605c0fe7a55ace67

PR #6053, opened 2021-07-05
pages.goto should be awaited

gwyneplaine:
Not sure how this was working previously, but this was causing errors in the admin-ui tests branch, so a little bit of boyscouting.


timleslie (approved):
👍

PR #6059, opened 2021-07-06
 Allow access to Lists fields and other metadata from Backend API

MurzNN:
After Keystone's core was reworked by #5665, we miss the ability to access Lists metadata at backend side. 

For example, earlier we can get list of fields from needed List via something like this:
```js
const myListFields = context.lists[listName].fieldsByPath;
const myListAuthorFieldType = context.lists[listName].fieldsByPath['author'].adapter.fieldName; // eg 'Relationship'
```

This PR is returns ability to get list of fields for all Lists, with all metadata such as field type (kind), access, hooks, views, etc:
```js
const myListAuthorFieldType = context.initialisedLists[listName].fields['author'].dbField.kind; // 'relation'
```

Please review id and describe if something wrong... Or if this info is already accessible at backend via some API call, please show some examples. Thanks!

### DIFF
--- a/.changeset/yellow-bees-know.md
+++ b/.changeset/yellow-bees-know.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Allow access to Lists fields and other metadata from Backend API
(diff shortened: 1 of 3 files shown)


timleslie:
Implemented in #6403.

PR #6061, opened 2021-07-07
Fix TypeScript errors

emmatown:


packages-next/types/src/legacy-filters.ts
```diff
@@ -3,7 +3,7 @@ import { schema } from '.';
 const identity = (x: any) => x;
 
 export const impls = {
-  equalityConditions<T>(fieldKey: string, f: (a: any) => any = identity) {
+  equalityConditions<T>(fieldKey: string, f: (a: any) => any = identity): any {
```
mitchellhamilton, line 6: These things are because TypeScript generates declarations that are broken but having actual types for the returns types here doesn't matter.

design-system/packages/core/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@babel/runtime": "^7.14.6",
     "@emotion/react": "^11.4.0",
+    "@types/facepaint": "1.2.1",
```
mitchellhamilton, line 20: This is because the `@types` packages make the things work globally so we didn't catch that the type is used in the generated declarations do this dep is needed

nathsimpson (approved):

PR #6086, opened 2021-07-12
Cleanup todos

JedWatson:
I found these hanging around, they're now outdated and/or captured elsewhere.

Also removed unused frontmatter from the styleguide docs.


timleslie (approved):
👍

PR #6088, opened 2021-07-12
Add React dependencies to reduce peerDep noise

JedWatson:
This is probably strictly incorrect, but resolves _most_ of the "unmet peer dependency" warnings when you run `yarn` in the project root.

I don't think it does any harm, and getting screen full of warnings is (imo) worse, so I _think_ adding these deps to the root is a thing worth doing.

Happy to defer to your wisdom @mitchellhamilton


mitchellhamilton (approved):
The description is correct

PR #6095, opened 2021-07-13
Nested filters

emmatown:
Note the old filters are still in the field types so that they're there if we want them for some kind of compatibility mode.


yarn.lock
```diff
@@ -2097,6 +2097,14 @@
...
     "@prisma/engines-version" "2.28.0-17.89facabd0366f63911d089156a7a70125bfbcd27"
 
+"@prisma/debug@2.26.0":
+  version "2.26.0"
+  resolved "https://registry.yarnpkg.com/@prisma/debug/-/debug-2.26.0.tgz#a6e7b6e752aacc793211d50ed30d815576e3132d"
+  integrity sha512-dBZPmxGKXP0VhDpKgpsFM3NhjFZZmP09MKRgV3oWfZmU+X1uA+crVscrajfZsOP0T6lbP1FCDLkFEEGyNCyuFA==
+  dependencies:
+    debug "4.3.2"
+    ms "^2.1.3"
+
```
timleslie, line 2107:
```suggestion
```

yarn.lock
```diff
@@ -2164,6 +2172,16 @@
...
+"@prisma/generator-helper@2.26.0":
+  version "2.26.0"
+  resolved "https://registry.yarnpkg.com/@prisma/generator-helper/-/generator-helper-2.26.0.tgz#c7ef7be8d560ceffa2acded2ca599d930ac3baac"
+  integrity sha512-w8OwQntKje/mvK/yQV4mo/Ce65/xFcEPN8wybOu/2U/kfN9Hmpm2yVjMP+tqPwngwWCG0yxxsyNTfAOS6zoYMw==
+  dependencies:
+    "@prisma/debug" "2.26.0"
+    "@types/cross-spawn" "^6.0.1"
+    chalk "^4.0.0"
+    cross-spawn "^7.0.2"
+
```
timleslie, line 2184:
```suggestion
```

timleslie (changes requested):

gautamsi:
Can this new filter support `null` or `not null` value in UI? I guess we would need it for most nullable fields

mitchellhamilton:
> Can this new filter support null or not null value in UI? I guess we would need it for most nullable fields

With this PR, the filters in the Admin UI will behave the same as they do currently, we're planning to revamp the filter UI which will handle null properly among other things but that's not happening in this PR

timleslie (approved):
Looking good 👍

PR #6104, opened 2021-07-14
Cosmetic core changes before error handling

timleslie:
These changes in the `core` directory are purely cosmetic, but they set the code up to make it easier to see where the changes to the error handling are going to land.

### DIFF
--- a/packages/keystone/src/lib/core/mutations/create-update.ts
+++ b/packages/keystone/src/lib/core/mutations/create-update.ts
@@ -116,7 +116,7 @@ export function updateMany(
     const updatedItem = await writeLimit(() =>
       getPrismaModelForList(context.prisma, list.listKey).update({ where: { id: item.id }, data })
     );
-    afterChange(updatedItem);
+    await afterChange(updatedItem);
     return updatedItem;
   });
 }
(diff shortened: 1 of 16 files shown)


packages/keystone/src/lib/core/mutations/create-update.ts
```diff
@@ -127,6 +103,24 @@ export async function updateOne(
...
+  provider: DatabaseProvider
+) {
+  const writeLimit = pLimit(provider === 'sqlite' ? 1 : Infinity);
+  return data.map(async ({ data: rawData, where: rawUniqueWhere }) => {
+    const item = await getAccessControlledItemForUpdate(list, context, rawUniqueWhere, rawData);
+    const { afterChange, data } = await resolveInputForCreateOrUpdate(list, context, rawData, item);
+    const updatedItem = await writeLimit(() =>
+      getPrismaModelForList(context.prisma, list.listKey).update({ where: { id: item.id }, data })
+    );
+    afterChange(updatedItem);
```
mitchellhamilton, line 119: (i know this isn't a change with this PR)

```suggestion
    await afterChange(updatedItem);
```

emmatown (approved):

packages/keystone/src/lib/core/mutations/create-update.ts
timleslie: Oh, nice catch!

PR #6105, opened 2021-07-15
Remove unnecessary GraphQL descriptions

emmatown:
These descriptions just repeat what's in the name and don't actually add any new information so having a smaller schema that's easier to read at a glance is better imo

note: major because a build that would succeed with the current version would fail since the schema.graphql file would have to be updated.


timleslie (approved):
Fantastic!

PR #6106, opened 2021-07-15
Check unique input format before using it

timleslie:
The function `resolveUniqueWhereInput` is used to check that the user hasn't provided an invalid unique filter, e.g. providing two different keys, or trying to filter on `null`. This PR ensures that this function is always called before we attempt to use the filter itself.

### DIFF
--- a/packages/keystone/src/lib/core/mutations/access-control.ts
+++ b/packages/keystone/src/lib/core/mutations/access-control.ts
@@ -13,7 +13,6 @@ import {
   PrismaFilter,
   resolveUniqueWhereInput,
   resolveWhereInput,
-  UniquePrismaFilter,
 } from '../where-inputs';
 
 export async function getAccessControlledItemForDelete(
@@ -22,7 +21,7 @@ export async function getAccessControlledItemForDelete(
   uniqueInput: UniqueInputFilter
 ): Promise<ItemRootValue> {
   const uniqueWhere = await resolveUniqueWhereInput(uniqueInput, list.fields, context);
-  const itemId = await getStringifiedItemIdFromUniqueWhereInput(uniqueWhere, list.listKey, context);
+  const itemId = await getStringifiedItemIdFromUniqueWhereInput(uniqueInput, list.listKey, context);
 
   // List access: pass 1
   const access = await validateNonCreateListAccessControl({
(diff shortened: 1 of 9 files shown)


packages/keystone/src/lib/core/mutations/access-control.ts
```diff
@@ -13,15 +13,16 @@ import {
...
 export async function getAccessControlledItemForDelete(
   list: InitialisedList,
   context: KeystoneContext,
-  filter: UniqueInputFilter,
-  inputFilter: UniqueInputFilter
+  uniqueInput: UniqueInputFilter
 ): Promise<ItemRootValue> {
-  const itemId = await getStringifiedItemIdFromUniqueWhereInput(filter, list.listKey, context);
+  const uniqueWhere = await resolveUniqueWhereInput(uniqueInput, list.fields, context);
+  const itemId = await getStringifiedItemIdFromUniqueWhereInput(uniqueWhere, list.listKey, context);
```
mitchellhamilton, line 25: This change isn't right. `getStringifiedItemIdFromUniqueWhereInput` getting the value directly from GraphQL input is intentional because e.g. it will be in the string version of the id which `getStringifiedItemIdFromUniqueWhereInput` needs to be return.

packages/keystone/src/lib/core/mutations/nested-mutation-many-input-resolvers.ts
```diff
@@ -15,33 +15,35 @@ type _UpdateValueType = schema.InferValueFromArg<
...
   return (
     await Promise.all(
-      uniqueWheres.map(async filter => {
-        if (filter === null) return [];
+      uniqueInputs.map(async uniqueInput => {
+        if (uniqueInput === null) return [];
+        const uniqueWhere = await resolveUniqueWhereInput(uniqueInput, foreignList.fields, context);
         try {
-          await context.sudo().db.lists[foreignList.listKey].findOne({ where: filter });
+          await context.sudo().db.lists[foreignList.listKey].findOne({ where: uniqueWhere });
```
mitchellhamilton, line 28: This isn't right, `context.sudo().db.lists[foreignList.listKey].findOne` should have the values before they go through the unique where input resolvers. (arguably even using `db.lists` is also wrong because the value will be validated and coerced through GraphQL again but it's less wrong since for the types that you would want to use here since they should be able to round-trip as GraphQL inputs (though fixing it would still be good imo))

PR #6107, opened 2021-07-15
Validate unique inputs before use

timleslie:
This replaces #6106 with an explicit decoupling of the validation step from the resolver step.


emmatown (changes requested):
I'm not a fan of separating these, feels like a recipe for "oh whoops, i forgot to validate it but everything was working so I didn't notice". Mind if I try some things and then I'll open a PR?

PR #6108, opened 2021-07-15
Alternative uniqueWhere resolving changes

mitchellhamilton:
@timleslie I'm assuming the reason behind the changes was that you didn't want the errors to be wrapped in another `GraphQLError`, this will achieve the same thing.


packages/keystone/src/lib/core/mutations/nested-mutation-many-input-resolvers.ts
```diff
@@ -24,7 +25,7 @@ async function getDisconnects(
       uniqueWheres.map(async filter => {
         if (filter === null) return [];
         try {
-          await context.sudo().db.lists[foreignList.listKey].findOne({ where: filter });
+          await findOne({ where: filter }, foreignList, context);
```
mitchellhamilton, line 28: This isn't going through GraphQL here so when `resolveUniqueWhereInput` throws an error, it won't be wrapped.

(and this is also more correct because the values won't go through GraphQL coercion again)

PR #6110, opened 2021-07-15
Fix virtual field rendering of false & 0

ChuckJonas:
`0` and `false` currently don't render when they are returned by virtual fields.  Seems like this component could be more "aware" if we were passing in the `schema.field type`.

### DIFF
--- a/.changeset/shy-bobcats-float.md
+++ b/.changeset/shy-bobcats-float.md
@@ -2,4 +2,4 @@
 '@keystone-next/fields': patch
 ---
 
-Fix virtual field rendering of false & 0 values
+Fixed virtual field rendering of false & 0 values.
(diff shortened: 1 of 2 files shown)


ChuckJonas:
I just tested this using the `built libraries` above and it seems to be working for both `false` and `0` now 👍

.changeset/shy-bobcats-float.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/fields': patch
+---
+
+Fix virtual field rendering of false & 0 values
```
timleslie, line 5:
```suggestion
Fixed virtual field rendering of false & 0 values.
```

timleslie: We should write this down somewhere obvious, but our convention is to write our changeset descriptions in past tense, so that they make sense when a person in the future is reading the CHANGELOG file.

timleslie (approved):
This all look good, thanks for the PR 🙏

PR #6111, opened 2021-07-15
Customisable Navigation

gwyneplaine:
This PR does the following: 

- [x]  Adds the Navigation component as a customisable property in the `AdminConfig` 
- [x] Decomposes the Navigation component into smaller components to make providing custom Navigation easier. 

We expose the following components now besides the core Navigation component: 
* NavItem - thin a11y and styling wrapper around the `Link` component from Next.js 
* ListNavItem - thin wrapper on top of NavItem, that takes a singe list item (of type `ListMeta`)  and renders a NavItem. **
* ListNavItems - this component takes an array of lists `ListMeta[]` and an optional `include` array. If an include array is specified only list items with keys that match the includes array are rendered as NavItems. Otherwise all lists are rendered as NavItems
* NavContainer - Component encapsulates rendering logic for the AuthenticatedItem, as well as keystones a11y and styling decisions for the layout of the side nav. 

** The impetus behind `ListNavItem`  is to have a higher-level abstraction that can allow users to not have to care about the contents and structure of our lists. This is especially relevant for future feature additions to the nav. (i.e. if we add icon support, that would work out of the box from a dependency bump for users of ListNavItem, whereas you'd have to add that icon prop manually leveraging a lower level component.) 

**The impetus behind `ListNavItems` is similar, but for rendering a list of NavItems. This is also a pre-emptive DX optimisation for creating custom navigation groups and subsections in the future. 

Guide and example split into #6185 for expedited functional release.

Preview: 
Screen Shot 2021-07-16 at 9 47 37 am


.changeset/fluffy-schools-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Added the ability to customise the Navigation component in the Admin-UI, and provided helper components to do so.
```
timleslie, line 5:
```suggestion
Added the ability to customise the Navigation component in the Admin UI, and provided helper components to do so.
```

.changeset/rare-carrots-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/types': minor
+---
+
+Add admin-ui specific types AuthenticatedItem, VisibleLists, CreateViewFieldModes and NavigationProps to exports.
```
timleslie, line 5:
```suggestion
Add Admin UI specific types AuthenticatedItem, VisibleLists, CreateViewFieldModes and NavigationProps to exports.
```

timleslie (dismissed):
yarn lock changes should be reverted

packages/types/src/admin-meta.ts
```diff
@@ -1,7 +1,34 @@
+import { GraphQLError } from 'graphql';
 import type { ReactElement } from 'react';
 import { GqlNames, JSONValue } from './utils';
 
-export type AdminConfig = { components?: { Logo?: (props: {}) => ReactElement } };
+export type NavigationProps = {
+  authenticatedItem: AuthenticatedItem;
+  lists: ListMeta[];
```
mitchellhamilton, line 7: What's the rationale behind providing these as props over users using `useKeystone`?

packages/types/src/admin-meta.ts
```diff
@@ -1,7 +1,34 @@
+import { GraphQLError } from 'graphql';
 import type { ReactElement } from 'react';
 import { GqlNames, JSONValue } from './utils';
 
-export type AdminConfig = { components?: { Logo?: (props: {}) => ReactElement } };
+export type NavigationProps = {
```
gwyneplaine, line 5: @mitchellhamilton  impetus here is to avoid circular dependency of `@keystone-next/types` on `@keystone-next/keystone` 

Chatted with @JedWatson  about this beforehand, but happy to revert or change as necessary. 
Also overall keen to understand the boundaries around @keystone-next/types, what should go in here and what shouldn't etc.

packages/types/src/admin-meta.ts
gwyneplaine: useKeystone provides lists as a record, and visibleLists as an array of strings. 
The rationale is mostly DX, we already do the work to create an array of visible ListMeta items in the Navigation component, surfacing this to the user cuts out the boilerplate of 

```
Object.keys(lists).map(list => ...filter out non visible lists and create a new array of lists) 
```

timleslie (commented):
I'm 👍 on this. I'll let @mitchellhamilton have a final review and give it the ✅

packages/keystone/src/admin-ui/components/index.ts
```diff
@@ -1,8 +1,18 @@
+// FIELD VIEW SPECIFIC COMPONENTS
 export { CellContainer } from './CellContainer';
 export { CellLink } from './CellLink';
+
 export { ErrorBoundary, ErrorContainer } from './Errors';
+
+// ADMIN-UI CUSTOM COMPONENTS
 export { Logo } from './Logo';
-export { Navigation } from './Navigation';
+export { Navigation, NavigationContainer, NavItem, ListNavItems, ListNavItem } from './Navigation';
```
mitchellhamilton, line 9: When would someone want to use `Navigation`? Also, it is a different thing to the thing that accepts `NavigationProps`, right?

gwyneplaine: Correct, the wrapping `Navigation` component would only be used for custom pages. 
For example if someone were to not want to opt into the entire layout provided by `PageContainer` but did want to have a sidebar nav, they'd leverage that Navigation export

emmatown (approved):

PR #6119, opened 2021-07-16
Split default value loop from field-type resolver loop

timleslie:
Decouples these two loops so that we can provide better error messages moving forward.

This also lifts the `null` input case out of the individual relationship resolver functions so we can have the discussion around what the correct behaviour for this input should be (currently a no-op).

### DIFF
--- a/packages/keystone/src/lib/core/mutations/create-update.ts
+++ b/packages/keystone/src/lib/core/mutations/create-update.ts
@@ -145,72 +145,73 @@ async function resolveInputForCreateOrUpdate(
 ) {
   const operation: 'create' | 'update' = existingItem === undefined ? 'create' : 'update';
   const nestedMutationState = new NestedMutationState(context);
-  let resolvedData = Object.fromEntries(
+
+  let resolvedData = originalInput;
+
+  // Apply default values
+  // We don't expect any errors from here, so we can wrap all these operations
+  // in a generic catch-all error handler.
+  if (operation === 'create') {
+    resolvedData = Object.fromEntries(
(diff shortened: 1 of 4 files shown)


packages/keystone/src/lib/core/mutations/create-update.ts
```diff
@@ -145,72 +145,73 @@ async function resolveInputForCreateOrUpdate(
...
+                return undefined;
+              }
+              if (input === undefined) {
+                // No-op: This is what we want
+                return () => undefined;
+              }
+              if (input === null) {
+                // No-op: Should this be UserInputError?
+                return () => undefined;
+              }
```
mitchellhamilton, line 193: Why is this being here better?

timleslie: It's just about visibility. I want to have a single place to point at for this discussion, and it frees up the four other resolvers to be simpler.

mitchellhamilton: Could you not do this or write it in such a way where the value used to decide what to return is based on the value passed to the function rather than the input from GraphQL? The resolve function passed to relationship field input resolvers currently accepts some input that the field's input resolver calls it with and decides what to do based on that rather than using the GraphQL input directly and this breaks that.

timleslie: I don't want to bike shed this too much because what I really want is for all this code to evaporate even further, so as long as this works I want to move forward and have these specific lines go away as part of broader changes I'm working on.

emmatown (approved):

PR #6123, opened 2021-07-16
Simplify NestedMutationState

timleslie:


packages/keystone/src/lib/core/mutations/nested-mutation-many-input-resolvers.ts
```diff
@@ -135,15 +127,11 @@ export function resolveRelateToManyForUpdateInput(
       foreignList
     );
 
-    const [disconnect, connectAndCreates] = await Promise.all([
+    const [disconnect, connect] = await Promise.all([
```
mitchellhamilton, line 130: This still includes connect as well
```suggestion
    const [disconnect, connectAndCreates] = await Promise.all([
```

emmatown (approved):

PR #6146, opened 2021-07-19
Add ability to customise `AuthenticatedItem`

borisno2:
Add `AuthenticatedItem` to `adminConfig.components` and return `adminConfig.components.AuthenticatedItem` in `Navigation` if configured to allow for custom `AuthenticatedItem`

I would love to ability to edit the `AuthenticatedItem` component this will enable a custom logout, as well as custom profile links. So this one works exactly the same as the `Logo` Component.

Thanks!


borisno2:
Is there any feedback or a better way to do this? I notice PR #6111 however it doesn't look like this will allow for a customised `AuthenticatedItem`.

borisno2:
PR #6111 should cover what is needed by this PR

PR #6153, opened 2021-07-21
Replace implementation of the lists API

emmatown:
This makes the implementation of the lists API be determined by the GraphQL schema rather than hand crafted so it won't need to change for the GraphQL API changes(besides count because it's Special). The external impact of this is that the implicit chunking is no longer a thing which we vaguely discussed removing a while ago.


timleslie (approved):
👍

PR #6159, opened 2021-07-22
updated guide, redirects, and links

raveling:
The `Keystone 5 vs Next` guide needed updating to reflect the fact we're now in the `6 Preview` release.


docs/redirects.js
```diff
@@ -102,4 +102,13 @@ const ORIGINAL_NEXT = [
   { source: '/whats-new', destination: '/updates/whats-new-in-v6', permanent: true },
 ];
 
-module.exports = [...ORIGINAL_NEXT, ...KEYSTONE_5, ...KEYSTONE_4];
+/* URLs from the current website */
+const CURRENT = [
+  {
+    source: '/docs/guides/keystone-5-vs-keystone-6-preview.mdx',
```
bladey, line 108: Should not have the `.mdx` extension as that's the internal file

bladey: Should be just `/docs/guides/keystone-5-vs-keystone-6-preview`

bladey (changes requested):
See comment about redirects

docs/redirects.js
```diff
@@ -102,4 +102,13 @@ const ORIGINAL_NEXT = [
...
+    destination: '/docs/guides/keystone-5-vs-keystone-6-preview.mdx',
+    permanent: true,
+  },
+];
+
+module.exports = [...ORIGINAL_NEXT, ...KEYSTONE_5, ...KEYSTONE_4, ...CURRENT];
```
bladey, line 114: Really small thing, but it would be great to have `...CURRENT` at the start of the array, so it goes in order from newest to oldest (sites).

raveling: fixed in f777c2d thanks for spotting!

docs/redirects.js
raveling: good catch! fixed in fed8412

bladey (approved):

PR #6161, opened 2021-07-22
Move run-time checks to startup for unique fields

timleslie:
Moves run time check of types from query time to startup time. Removes run time checks for input values, which will be surface as errors when running queries if there's a problem.

### DIFF
--- a/.changeset/thirty-windows-dance.md
+++ b/.changeset/thirty-windows-dance.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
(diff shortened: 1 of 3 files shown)


emmatown (changes requested):
The first check already exists at startup startup time, it's there to avoid casting for the checks after basically
https://github.com/keystonejs/keystone/blob/88f3ddf1b0a0805dbf65cdc9aa8526be3bbd26f3/packages/keystone/src/lib/core/field-assertions.ts#L15-L20

> Removes run time checks for input values, which will be surface as errors when running queries if there's a problem.

They exist to specifically disallow returning arbitrary regular filters which wouldn't otherwise be surfaced as an error

PR #6162, opened 2021-07-22
Use queries.count in relationship count resolver

timleslie:
No functional changes.


emmatown (approved):

emmatown (changes requested):
one sec, i want to check something

emmatown (approved):

PR #6165, opened 2021-07-23
Add `ui.searchFields` option

emmatown:
This is primarily to remove the usage of the `search` argument in the Admin UI because we're planning on removing it soon because people may think that it's proper full-text search which it is _not_ and we recommend people use `contains` filters directly(which is what the `search` argument does) until Prisma supports full-text search and we expose it in the GraphQL API.


examples-staging/assets-cloud/schema.graphql
```diff
@@ -426,6 +427,11 @@ enum KeystoneAdminUIFieldMetaItemViewFieldMode {
   hidden
 }
 
+enum QueryMode {
+  default
+  insensitive
+}
```
mitchellhamilton, line 433: Note the naming and enum values here align with the `mode` field that will be in the new filters for strings (and since the new filters align with Prisma, so does this).

.changeset/serious-pigs-attack.md
```diff
@@ -0,0 +1,7 @@
+---
+'@keystone-next/fields': major
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+Added `ui.searchFields` option to lists to allow searching by multiple fields in the Admin UI(the only current usage of this is in the select used in relationship fields) and to replace the usage of the `search` GraphQL argument which will be removed soon and should be replaced by using contains filters directly.
```
timleslie, line 7:
```suggestion
Added `ui.searchFields` option to lists to allow searching by multiple fields in the Admin UI (the only current usage of this is in the select used in relationship fields) and to replace the usage of the `search` GraphQL argument which will be removed soon and should be replaced by using `contains` filters directly.
```

timleslie (approved):
👍

PR #6179, opened 2021-07-23
Require the `identityField` passed to `createAuth` has `isUnique: true`

emmatown:
This prevents the `MULTIPLE_IDENTITY_MATCHES` error from ever occurring.

Note I didn't remove the `MULTIPLE_IDENTITY_MATCHES` error from the GraphQL enum and etc. because we need to look at those errors generally given only ever actually return `FAILURE`.


gautamsi:
We do need to override the default behaviour, I have a need to have emails unique within the organisation (a field on each user) and need not be unique in database. seems like I must use custom authentication.

is this going to break the scenario? if it is checked at startup, it will definitely break this.

mitchellhamilton:
That use case is already not supported by the auth package since we throw an error when multiple items are returned for a given identity so you already would have to use a custom solution and you would need some way to provide the organisation as well when signing in.

timleslie (approved):
👍

PR #6180, opened 2021-07-23
Add react-hooks/exhaustive-deps ESLint rule

emmatown:
When I was doing #6165, I noticed that a dependency array was very wrong and this rule would have stopped it (the fix for that particular case is done in #6165 so that must be merged before this one)


packages/auth/src/pages/InitPage.tsx
```diff
@@ -211,7 +211,7 @@ const InitPage = ({ fieldPaths, listKey, enableWelcome }: InitPageProps) => {
         router.push((router.query.from as string | undefined) || '/');
       }
     }
-  }, [rawKeystone.authenticatedItem, router.query.from]);
+  }, [rawKeystone.authenticatedItem, router.query.from, enableWelcome, router]);
```
mitchellhamilton, line 214:
```suggestion
  }, [rawKeystone.authenticatedItem, enableWelcome, router]);
```

gwyneplaine (approved):

Issue #1047, opened 2019-04-19
GraphQL Access denied error when empty result


gautamsi:
I see that graphql queries return "Access denied" error instead of empty results, is this by design?

Should we have this configurable to retuen empty  or 'access denied' error?

timleslie:
@gautamsi I'm not sure I understand the context of where you're getting this. Could you give some more details, including the query used and the full response obtained?

gautamsi:
for example, run the blog demo project with this query in GraphiQL

```gql
query {
    User(where: {id: "5c77a50a939aa93073dabe11"}) {
    id, isAdmin, _label_ name email password_is_set avatar { filename publicUrl}
  }
}
```

returns this when there are no usrs by above id. 

```gql
{
  "errors": [
    {
      "message": "You do not have access to this resource",
      "name": "AccessDeniedError",
      "time_thrown": "2019-04-23T10:10:32.812Z",
      "data": {
        "type": "query",
        "target": "User"
      },
      "path": [
        "User"
      ],
      "uid": "cjutmp629002i3dw18otugefp"
    }
  ],
  "data": {
    "User": null
  },
  "extensions": {}
}
```

dcousens:
Cannot reproduce

```
{ 
  allUsers (where: { id: "PUT_THE_USER_ID_HERE" }) {
    id
    isAdmin 
    _label_ 
    name 
    email 
    password_is_set 
    avatar {
      filename publicUrl
    }
  }
}
```

bookyo:
i have same issue , have idea resolve it?

Temkit:
same here, trying to debug !

PR #6181, opened 2021-07-23
Throw NotFoundError when findOne failed to find an item

tjbp:
The following query (obviously an ID shouldn't ever be zero):

```gql
{
  User(where: {id: 0}) {
    id
  }
}
```

misleadingly throws an AccessDeniedError when it fails to locate a relevant entry. This pull request creates a differently-named NotFoundError so that this can be differentiated from a true access denial in a try/catch block.

Happy to create a changeset in accordance with guidelines and fix tests if this patch is wanted (not sure if it's an access denial by design and I'm missing something).

Incidentally this ~~is likely~~ might be a fix for #1047.

### DIFF
--- a/packages/keystone/src/lib/core/graphql-errors.ts
+++ b/packages/keystone/src/lib/core/graphql-errors.ts
@@ -4,6 +4,10 @@ export const AccessDeniedError = createError('AccessDeniedError', {
   message: 'You do not have access to this resource',
   options: { showPath: true },
 });
+export const NotFoundError = createError('NotFoundError', {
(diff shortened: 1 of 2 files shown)


timleslie:
Hi @tjbp, thanks for the PR! We're actively working on all of our error handling at the moment in #6050, and we're going to be changing the behaviour of this code path to return `null`, rather than throwing an access denied error.

Having the query return `null` for both the `not found` and the `access denied` case will give a more intuitive result for users of the API without leaking details about which IDs are valid, which can be a security risk.

PR #6192, opened 2021-07-27
Add health check endpoint to Keystone's express server

JedWatson:
This adds a configurable health check endpoint to Keystone's express server, avoiding the need to use a custom admin page to implement your own.

It will respond on `/_healthcheck` with JSON: `{ status: 'pass', timestamp: Date.now() }`

You can change the path and data with the `config.server.healthCheck` option (documentation to be added in a separate PR)

This PR also adds a response to the dev server while Keystone is initialising, which strictly shouldn't be necessary since you shouldn't be using the dev server in production, but it seems prudent to cover this there anyway just in case.


packages/types/src/config/index.ts
```diff
@@ -113,13 +113,20 @@ export type AdminFileToWrite =
 
 // config.server
 
+type HealthCheckConfig = {
+  path?: string;
+  data?: Record<string, any>;
```
timleslie:
```suggestion
  data?: () => Record<string, any> | Record<string, any>;
```

timleslie (changes requested):
Needs changeset

packages/types/src/config/index.ts
JedWatson: Got me while I was writing a changeset 😜 added now

JedWatson:
@timleslie you got me while I was writing a changeset 😜

Done now, fixed the type too

timleslie (approved):
👍

gautamsi:
Not that you have added something for express server, I hope other express server needs will be re evalutated and open soon.

One of the pressing one is bringing back `configureExpress` feature

PR #6200, opened 2021-07-27
Switch from apollo-errors to apollo-server-errors

timleslie:
The only functional difference is that `apollo-server-errors` injects `extensions: { code: undefined }` into the error, but we're not using this field just yet. This change highlights that our error functions were passing through a bunch of extra data which was never being surfaced in our errors. That's fine, since we want re-assess what we're exposing to people anyway. This gives us a nice clean slate to work from.

### DIFF
--- a/packages/keystone/src/lib/core/graphql-errors.ts
+++ b/packages/keystone/src/lib/core/graphql-errors.ts
@@ -1,12 +1,12 @@
 import { ApolloError } from 'apollo-server-errors';
 
-export const AccessDeniedError = () => new ApolloError('You do not have access to this resource');
+export const accessDeniedError = () => new ApolloError('You do not have access to this resource');
 
-export const ValidationFailureError = () =>
+export const validationFailureError = () =>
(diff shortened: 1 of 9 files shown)


packages/keystone/src/lib/core/graphql-errors.ts
```diff
@@ -1,23 +1,12 @@
...
-});
-export const ValidationFailureError = createError('ValidationFailureError', {
-  message: 'You attempted to perform an invalid mutation',
-  options: { showPath: true },
-});
-export const LimitsExceededError = createError('LimitsExceededError', {
-  message: 'Your request exceeded server limits',
-  options: { showPath: true },
-});
+export const AccessDeniedError = () => new ApolloError('You do not have access to this resource');
```
mitchellhamilton, line 3: Could you either make all these classes or make the first character lowercase?

emmatown (changes requested):

emmatown (approved):

PR #6206, opened 2021-07-28
Unique where inputs in deletes

emmatown:
This is part of #5635

### DIFF
--- a/.changeset/wise-pianos-flash.md
+++ b/.changeset/wise-pianos-flash.md
@@ -1,0 +1,8 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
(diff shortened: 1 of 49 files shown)


.changeset/wise-pianos-flash.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+The delete mutations now accept where unique inputs instead of only an id.
```
timleslie, line 6:
```suggestion
The delete mutations now accept `where` unique inputs instead of only an `id`.
```

timleslie (approved):
👍

PR #6209, opened 2021-07-28
Rework meta tags for website

bladey:


raveling:
Note about the meta descriptions. The `coming soon` pages have placeholder descriptions to match the placeholder callout ([example](https://github.com/keystonejs/keystone/pull/6209/commits/f7b2004995cf29c1d152579a5483d6e90f39b841#diff-0149fa55079a1be36999b0b1e028017778ac362638bae75fbb4219fdc083fd6c)). When these guides get filled in with content author will need to ensure they replace the value in `description`. As it's not empty it will pass CI in the current state, but we don't want that placeholder content.

raveling (approved):
LGTM! 🚀

PR #6220, opened 2021-07-29
Preserve whitespace in errors

timleslie:
The aim here is to preserve white-space from the error messages returned by the GraphQL API.

@gwyneplaine Is this the right approach? Are there are places these errors are surfaced that might also need this change?


gwyneplaine (approved):
This seems fine, let's get this in so we can validate it against the inflight graphql error changes

gwyneplaine:
Newline works as expected for: 
* Create.
* Edit.
* Inline create vie card view.
* Inline create via nested modal.

PR #6224, opened 2021-07-30
To-many relationship input updates

emmatown:
This is part of #5635


.changeset/rich-tomatoes-own.md
```diff
@@ -0,0 +1,22 @@
+---
+'@keystone-next/fields': major
+'@keystone-next/keystone': major
+'@keystone-next/types': major
+---
+
+`disconnectAll` has been replaced by `set` in to-many relationship inputs, the equivalent to `disconnectAll: true` is now `set: []`. There are also seperate input types for create and update where the input for create doesn't have `disconnect` or `set` and also the items. The inputs in the lists in the input field are now also non-null.
```
timleslie, line 7:
```suggestion
`disconnectAll` has been replaced by `set` in to-many relationship inputs, the equivalent to `disconnectAll: true` is now `set: []`. There are also seperate input types for create and update where the input for create doesn't have `disconnect` or `set`. The inputs in the lists in the input field are now also non-null.
```

timleslie (approved):
👍

PR #6226, opened 2021-07-30
Tweak navigation for `Recent Releases`

bladey:
The `Recent Releases` heading is displayed even if there are no releases.

image

There are releases once you go into the actual page as that is when data is passed in:

image

Thus it will now work as:

image

Additionally adds description to latest release, which is NOT captured by the linter, separate issue that I'm looking into.


timleslie (commented):
👍

timleslie (approved):
Yes!

PR #6228, opened 2021-07-30
Fix Markdown type checking for website

bladey:
This PR adds runtime checks to make sure `title` and `description` exist for every `.mdx` file in the website.

Has been tested for each individual use case.

---

**Original issue:**

The file https://github.com/keystonejs/keystone/blob/master/docs/pages/docs/guides/custom-admin-ui-navigation.mdx does not have a `description` prop passed in, it should be failing tests as `description` is a required prop inside https://github.com/keystonejs/keystone/blob/master/docs/components/Markdown.tsx, but it is not being caught.

The above `.mdx` file will ultimately have this `description` prop when this PR is resolved, but is left out for testing:

```
export default ({ children }) => <Markdown description="In this guide, we'll show you how to create a custom Navigation component to be rendered in the Admin UI.">{children}</Markdown>
```

Required prop in `master` that should be throwing errors - https://github.com/keystonejs/keystone/blob/91d7b2cc41994559831848185cdcd988dd351e4b/docs/components/Markdown.tsx#L28

Error throw I had to add in order for the missing description to be picked up in tests in this PR - https://github.com/keystonejs/keystone/pull/6228/files#diff-b8a2b9ec87db10bc7036a032dfafbb4048f8e70b5faa544cee1a31630bef38c1R33

Without this throw all tests pass.

`master` should currently be failing as https://github.com/keystonejs/keystone/blob/master/docs/pages/docs/guides/custom-admin-ui-navigation.mdx does not have a description.

Does it have something to do with Next.js not compiling `.mdx` files until runtime so it doesn't know that the description is missing?

Note this affects all `.mdx` files.


timleslie (approved):
👍

PR #6232, opened 2021-07-31
invalid TSX/JSX example

larrybahr:
When you copy/paste the example you would get the error
```
Syntax error: Adjacent JSX elements must be wrapped in an enclosing tag. Did you want a JSX fragment <>...</>?
```

This change adds the suggested JSX fragment so it compiles.


JedWatson (approved):

JedWatson:
Thanks @larrybahr!

PR #6256, opened 2021-08-04
Use Next 11 in the website

mitchellhamilton:
Mainly so that we don't have the think about it when upgrading the Admin UI


docs/package.json
```diff
@@ -34,7 +34,7 @@
     "cypress": "^5.6.0",
     "date-fns": "^2.23.0",
     "facepaint": "^1.2.1",
-    "next": "^10.2.3",
+    "next": "npm:next@^11.0.1",
```
bladey, line 37: @mitchellhamilton do we need `npm:next`? I'm finding it hard to find more info on what this does, currently the public version is `11.0.1` on npm - https://www.npmjs.com/package/next

mitchellhamilton: manypkg treats `npm:` specially and doesn't force all the versions of the package to be the same, other than that, it's practically equivalent to just writing `^11.0.1`

bladey: awesome, thank you!

bladey (approved):

PR #6260, opened 2021-08-05
Integration tests admin UI

gwyneplaine:
This PR contains:
* Integration tests for admin-ui, centred mostly around Navigation
* Name change to the `seedData` helper, and moving it into the `utils` file

### DIFF
--- a/tests/admin-ui-tests/utils.ts
+++ b/tests/admin-ui-tests/utils.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 import { promisify } from 'util';
+import fetch from 'node-fetch';
 import execa from 'execa';
 import _treeKill from 'tree-kill';
 import * as playwright from 'playwright';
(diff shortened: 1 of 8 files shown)


tests/admin-ui-tests/utils.ts
```diff
@@ -21,6 +22,26 @@ const promiseSignal = (): Promise<void> & { resolve: () => void } => {
...
+        query,
+        variables,
+      }),
+    }).then(res => res.json());
+    if (errors) {
+      throw errors;
+    }
+  } catch (e) {
+    throw e;
+  }
```
mitchellhamilton, line 42: this is unnecessary?

tests/admin-ui-tests/utils.ts
```diff
@@ -21,6 +22,26 @@ const promiseSignal = (): Promise<void> & { resolve: () => void } => {
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        query,
+        variables,
+      }),
+    }).then(res => res.json());
+    if (errors) {
+      throw errors;
```
mitchellhamilton, line 38: You might want to make an actual `Error` here with the message being something like ``graphql blah: ${graphql errors.map(x=>x.message).join('\n')}`` so that you get a stacktrace

tests/admin-ui-tests/utils.ts
gwyneplaine: I maybe misunderstanding how error handling in async/await works, but wouldn't we miss out on network errors, or other non-gql related errors if this was omitted? 

I assumed this was the equivalent to: 
```typescript
const { errors } = await fetch('...').catch(e => {
  throw e
})

if (errors) { 
 throw errors 
} 
```

emmatown (approved):

PR #6267, opened 2021-08-06
Include stacktrace flag

timleslie:
Adds a `config.graphql.debug` flag to control whether stacktraces are included in GraphQL API error messages.

Also groups the various error generating functions into an object that can be configured and passed around, as other error functions will be needing similar functionality very soon.

### DIFF
--- a/.changeset/real-sheep-end.md
+++ b/.changeset/real-sheep-end.md
@@ -1,0 +1,6 @@
+---
+'@keystone-next/keystone': minor
+'@keystone-next/types': minor
+---
+
+Add a `config.graphql.debug` option, which can be used to control with debug information such as stack traces are include in the errors returned by the GraphQL API.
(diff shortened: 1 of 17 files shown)


emmatown (changes requested):
Could you change this so that the errors always add the debug stuff(and then also make it so the errors aren't passed around everywhere) and then use formatError to remove them if debug is disabled? (and if formatError is provided by the user, do the removing the debug stuff and then pass it to the user's formatError)

.changeset/real-sheep-end.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-next/keystone': minor
+'@keystone-next/types': minor
+---
+
+Add a `config.graphql.debug` option, which can be used to control with debug information such as stack traces are include in the errors returned by the GraphQL API.
```
mitchellhamilton, line 6:
```suggestion
Added `config.graphql.debug` option, which can be used to control whether debug information such as stack traces are included in the errors returned by the GraphQL API.
```

emmatown (approved):

PR #6270, opened 2021-08-06
Adjusted copy in website banner.

raveling:
The current copy feels a little long/dense for a banner.

This PR simplifies the copy into 2 clauses with the same reference links.

CleanShot 2021-08-06 at 11 49 05

CleanShot 2021-08-06 at 11 53 53


JedWatson (approved):

raveling:
fixed in #6290

PR #6271, opened 2021-08-06
Check exceptions returned from GraphQL

timleslie:
Stop throwing away `extensions.exception` and actually check its value. This becomes relevant now that we're trying to provide more control over precisely when stacktrace information is exposed to the GraphQL API.

### DIFF
--- a/tests/api-tests/utils.ts
+++ b/tests/api-tests/utils.ts
@@ -17,21 +17,24 @@ export const apiTestConfig = (
 });
 
 const unpackErrors = (errors: readonly any[] | undefined) =>
-  (errors || []).map(({ locations, extensions: { exception, ...extensions }, ...unpacked }) => ({
-    extensions,
-    ...unpacked,
-  }));
+  (errors || []).map(({ locations, ...unpacked }) => unpacked);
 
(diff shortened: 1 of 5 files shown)


tests/api-tests/utils.ts
```diff
@@ -17,21 +17,24 @@ export const apiTestConfig = (
 });
 
 const unpackErrors = (errors: readonly any[] | undefined) =>
-  (errors || []).map(({ locations, extensions: { exception, ...extensions }, ...unpacked }) => ({
-    extensions,
-    ...unpacked,
-  }));
+  (errors || []).map(({ locations, ...unpacked }) => unpacked);
```
timleslie, line 20: This is the actual change. Everything else in this PR is in response to this change.

tests/api-tests/utils.ts
```diff
@@ -17,21 +17,24 @@ export const apiTestConfig = (
...
 
 const j = (messages: string[]) => messages.map(m => `  - ${m}`).join('\n');
 
+// FIXME: It's not clear to me right now why sometimes
+// we get an expcetion, and other times we don't - TL
```
timleslie, line 25: Will loop back to this in a future PR. Bigger fish to fry first.

emmatown (approved):

PR #6280, opened 2021-08-09
Remove `gqlType` option on `autoIncrement` field type

emmatown:
This was intended for when this field type could be used as an id field but the field type can no longer be used for that so this is unnecessary


gautamsi:
Will this impact people who have large number of items given that there is limit on max integer number graphql can support?

mitchellhamilton:
This is stored as Postgres's `integer` type which stores 32 bit signed integers which is the same as GraphQL's `Int` so no

JedWatson (approved):

Issue #6261, opened 2021-08-05
showing both `successful` and `not allowed` alert notification for unauthorized delete action.


jakegiri:
## Describe the bug

The access permission is working fine till now with ListKey items. 
But for non-authorized DELETE action, alert is showing inappropriately.
I have a product list. which has a access permission set to it for delete .
When an unauthorized user tries to delete it by clicking delete button, it prevents the deletion as expected but the card-notification is shown inappropriately  (as shown in image). The notification shows both "not allowed" and "successfully deleted". 

## Expected behaviour

Only one alert notification expected to be shown "Failed to delete item. You do not have access to this resource". 

## Screenshots

![image](https://user-images.githubusercontent.com/24270559/128281678-79e2e9fb-171e-42f1-b00d-38d31ca27a34.png)

mitchellhamilton:
Fixed by #6296

PR #6296, opened 2021-08-11
6261/fix delete alert

gwyneplaine:
See loom https://www.loom.com/share/959712942c834ccba88ee269af8b4bdb for break down of OLD behaviour. 

To sample the behaviour locally:
* run `yarn` at the root of your project 
* run `yarn workspace @keystone-next/test-projects-crud-notifications dev --reset-db` 
* navigate to `http://localhost:3000/tasks` 
* try and delete items (do not delete, do not destroy, do not kill should fail to delete, the rest will succeed)

-------
Refactored solution in list-view based on feedback. At the moment I believe this to be the minimum viable solution to unblock #6261. There are still a few issues with this implementation that I've outlined in #6307 , which we can resolve later. 

* Success now lists the deleted item 
* Failure now lists how many deletions failed, as well as a reduced list of reasons for failure. (if 50 items failed for the same reason, that reason is listed only once)  

Before: 
https://user-images.githubusercontent.com/12119389/129154218-36eb8275-7066-420e-9860-e2270e7bb719.mov

After:
https://user-images.githubusercontent.com/12119389/129154330-f3f55279-6c1c-43c9-aa89-ceb7156fc413.mov

------
Moved deletion tests to a separate PR, so we don't block release.


packages/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ItemPage/index.tsx
```diff
@@ -253,15 +253,17 @@ function DeleteButton({
           confirm: {
             label: 'Delete',
             action: async () => {
-              await deleteItem().catch(err => {
-                toasts.addToast({
+              try {
+                await deleteItem();
+              } catch (err) {
+                return toasts.addToast({
```
gwyneplaine, line 259: This error was also occurring on the item view. 
Early return here to stop the route changing and success toast from being rendered.

packages/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -436,14 +438,18 @@ function DeleteManyButton({
...
+            id
+            ${list.labelField}
+          }
+        }
 `,
       [list]
-    )
+    ),
+    {
+      errorPolicy: 'all',
```
gwyneplaine, line 449: [errorPolicy set to all](https://www.apollographql.com/docs/react/data/error-handling/#graphql-error-policies), so we can get handle scenarios where partial deletion success occurs.

package.json
```diff
@@ -35,7 +35,7 @@
...
-    "lint:examples": "for d in `find examples -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../..; done; for d in `find examples-staging -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../..; done; for d in `find tests/test-projects -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../..; done",
+    "lint:examples": "for d in `find examples -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../..; done; for d in `find examples-staging -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../..; done; for d in `find tests/test-projects -type d -maxdepth 1 -mindepth 1`; do cd $d; echo $d; SKIP_PROMPTS=1 yarn keystone-next postinstall; if [ $? -ne 0 ]; then exit 1; fi; cd ../../..; done",
```
gwyneplaine, line 38: Updated `lint:examples` to not break on test-projects repo.

timleslie (approved):
🚢

PR #6297, opened 2021-08-11
GraphQL API docs changes

emmatown:
This includes updates to the docs to account for the GraphQL API changes(including the filter changes) minus the Query Filter API page changes which is included in #6095


timleslie (approved):
👍

PR #6301, opened 2021-08-11
Added top margin to docs page component

raveling:
Aligns whitespace to original design intent.

docs-whitespace-fix


bladey (changes requested):
I don't think we should have this margin on mobile, otherwise it looks like this - 
image

raveling:
Do you mean the margin is too large, or too small?

Matt's design specs a larger margin on mobile (64px). 

> I don't think we should have this margin on mobile, otherwise it looks like this -
> image

bladey (approved):

PR #6302, opened 2021-08-11
[RFC] config.graphql.extendContext

timleslie:
This config option provides a general escape hatch to let developers extend/modify the context object however they want. In general it should be used for extension purposes.

### DIFF
--- a/examples/blog/keystone.ts
+++ b/examples/blog/keystone.ts
@@ -1,6 +1,9 @@
 import { config } from '@keystone-next/keystone';
+import { KeystoneContext } from '../../packages/types/src';
 import { lists } from './schema';
 import { insertSeedData } from './seed-data';
+
+type Context = KeystoneContext & { foo: string };
 
 export default config({
   db: {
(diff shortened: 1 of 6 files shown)


timleslie:
We've decided not to move forward with this particular solution at this time due to a lack of consensus on what specific problems we're trying to solve for. If anyone was hoping to have this feature, please open a discussion outlining the problem you're trying to solve and we will consider what an appropriate solution might look like.

gautamsi:
I was hoping for this to release

one strong use case is that we can have central logic for each request and extend context with custom data or even functions which runs per request.  This enables writing extensions, or plugins which will add stuff to context and can be used to easily wire up hooks logic without loading same item in each hook for each entity. Like settings value for organization which

another thing I had in mind for multi tenancy feature. I could add tenant information to context which could be used by different hooks and access control. What is happening now is that you can only add this information in `session.data` which is strictly for user object.

I think a bare minimum would be useful for community, run function when the context is being created or cloned (for sudo) and extend the context. have this extra information available where context is used

PR #6303, opened 2021-08-11
[WIP] Return null from findOne for not found/not allowed

timleslie:


timleslie:
This will be handled in the full access control update.

PR #6311, opened 2021-08-12
Upgrade Next to 11.1.0 for the website

timleslie:
This brings the website versions of Next (and also `@next/mdx`) up to latest. This version bump closes a CVE which doesn't impact us. See https://github.com/vercel/next.js/releases/tag/v11.1.0


mitchellhamilton (approved):
Looks like `next-images` was interfering with the Next's built-in image stuff but it doesn't look like `next-images` was actually doing anything since the images seem fine? You should probably go through the site in case I missed something

PR #6315, opened 2021-08-13
Added styles to table

raveling:
Attempts to make the visual appearance of  markdown generated `<table>` elements perform better. Comes with some tradeoffs. Let's discuss before merging this.

(may want to open the original for the image below in a separate tab)

table-poc-branch


bladey (approved):

bladey:
@raveling needed 2 x `as const`

PR #6317, opened 2021-08-13
Split create/update field input resolvers for relationship fields

timleslie:
Relationships can have nested errors which require special handling, so this just sets us up to work on that, and the handling of other resolver errors, separately.


emmatown (approved):
this makes me uncomfy but ok

PR #6327, opened 2021-08-16
Fix head titles from not populating on the website

bladey:
You can't nest tags in a `<Fragment>` or similar when using `next/head`, this resolves the issue of website titles not displaying on the live website (they are ALL blank right now).


timleslie (approved):
👍

PR #6330, opened 2021-08-16
Update text filter API table

timleslie:
## Before

Screen Shot 2021-08-16 at 4 46 43 pm

## After

Screen Shot 2021-08-16 at 4 47 06 pm


raveling (approved):
yaassss! great idea @timleslie ! 💡

PR #6334, opened 2021-08-17
Fix issue with VisuallyHidden checkbox interactions in table 

gwyneplaine:
Bug only affects chromium browsers.
My assumption is that either:
1. something funky is going on with how `position:sticky` elements interact with the rest of the DOM in chromium
2. something funky is going on with tables in chromium
3. combination of 1 + 2 

Either way, this fix resolves the bug by more explicitly specifying the `position: relative` parent that the `position: absolute` visually hidden element is attached to

Before:
https://user-images.githubusercontent.com/12119389/129643712-48e7cfdd-d9cc-4e9f-bde8-a6dd9bfbbfa7.mov

After:
https://user-images.githubusercontent.com/12119389/129643715-0273594c-b8d6-4cda-9d5b-00312fe9ff19.mov


packages/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -742,6 +740,7 @@ const TableHeaderCell = (props: HTMLAttributes<HTMLElement>) => {
         padding: spacing.small,
         textAlign: 'left',
         position: 'sticky',
+        zIndex: 1,
```
gwyneplaine, line 743: zIndex manually applied on sticky `thead` element here to maintain the layering order of DOM elements withiin the table, as position: 'relative'  on elements within tbody changes the zindex to be relative to the body

timleslie (approved):
LGTM 👍

PR #6343, opened 2021-08-17
Tweaking for local dev using Docker

aaronpowell:
I came to contribute to the repo only to find that I didn't have the right yarn installed, I needed to do some local env setup, etc. which just made it a little tricky to poke around in the codebase.

This PR adds a [VS Code remote containers](https://code.visualstudio.com/docs/remote/containers?WT.mc_id=javascript-00000-aapowell) definition file that, when used, will create a Dockerised dev environment that you can use from VS Code in a seamless fashion.

If you're not using VS Code, or not using the remote containers extension, these files will have no impact on you.


timleslie (approved):
This looks good, thanks for the PR @aaronpowell 🙏

PR #6355, opened 2021-08-17
[WIP] File adapters

aaronpowell:
This is a prototype of how Keystone v6 could support File Adapters in an abstract way, allowing for custom adapters to be written to store files in cloud services like S3, Azure Storage, etc., rather than on local disk.

I've done the refactor to have the two existing types, local file and keystone-cloud, implemented in this pattern.

Edit - added image adapter too

Edit 2 - Linting/etc. all looks fixed. Test's are failing on the stuff I changed but I can't figure out how to run/debug tests locally to fix them

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,2 @@
+ARG VARIANT="14-buster"
+FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:0-${VARIANT}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,25 @@
+{
+  "name": "Keystone",
+  "build": {
+    "dockerfile": "Dockerfile",
+    "args": {
+      "VARIANT": "14"
+    }
+  },
+
+  "settings": {},
+
+  "extensions": [
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    "gamunu.vscode-yarn",
+    "prisma.prisma"
+  ],
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  // "forwardPorts": [],
+
+  "postCreateCommand": "yarn install",
+
+  "remoteUser": "node"
+}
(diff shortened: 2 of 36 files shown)


raveling:
@aaronpowell I sent a reply to your related discussion:
https://github.com/keystonejs/keystone/discussions/6344
The best we can do for now is acknowledge that we've seen this, and return to it after we reach the General Availability release around the end of Q3. Thank you!

bjufre:
This would simplify allowing the users to expand the system without needing to duplicate or create custom views for fields that are only changing the way a file/image gets stored.

Really looking forward to something like this landing on Keystone!!

dcousens:
Hi @aaronpowell,

This is an area we are actively working on,  including in https://github.com/keystonejs/keystone/pull/7070 (but that pull request may yet be re-opened).
It covers a broad part of our problem space so we are spending our time to try and get this right.

For now,  I am going to close this pull request,  but no doubt we'll be taking parts of this into consideration in the ongoing work.

aaronpowell:
> Hi @aaronpowell,
> 
> This is an area we are actively working on, including in #7070 (but that pull request may yet be closed/re-opened).
> 
> For now, I will close this pull request to de-duplicate the work stream, but no doubt we'll be taking parts of this pull request into consideration in ongoing work.

No problems @dcousens, I'll keep an eye on #7070 and see what I can provide on the discussion front. And let me know if you want/need any help following the patterns of this PR

dcousens:
Thanks @aaronpowell!

One of the key considerations we're taking into account compared to this pull request is we want to offer a limited opening into new storage functionality for now,  by _not_ supporting custom adapters.

The intent is to allow us to iterate on the idiomatic parts of how this should be implemented first before opening up those parts of API to the community.  We don't want to end up with recurrent breaking changes on something the community might have built 'adapters' around in the short term.

aaronpowell:
The only problem I see with this approach @dcousens is that it keeps a barrier up for those who aren't using the _supported_ file providers, forcing them to rely on workarounds such as custom field types (which is the current solution).

Now yes, I have an _overly_ vested interest in this (compared to the average person) but what it shows is that it is possible to have an abstract approach to file storage but it's being gated for "some reason" (putting the perspective of someone who wouldn't have context).

I don't think Keystone should have to ship adapters for all the different cloud platforms, that's what community can support (and be a measure of direction on what would make the right places to invest by seeing what the community is most interested in), but it sort of feels like until the Keystone team has the bandwidth to tackle <insert cloud provider> support, you'll have to find workarounds that means you can't leverage a core feature of Keystone.

PR #6362, opened 2021-08-18
Move fields package into main package

emmatown:
To reduce problems with mismatching versions of Keystone packages, we're merging some packages into the main Keystone package.


timleslie (approved):
👍

PR #6375, opened 2021-08-19
Fix website typo

bladey:


raveling (approved):
`syntex`... good name for a band!

PR #6376, opened 2021-08-19
Add graphql.isEnabled config option at the list and field levels

timleslie:
This PR proposes a config option to allow solution designers to control what is included in the GraphQL schema. This allows us to decouple this from the access control functionality, and also adds more fine-grained control for things like `filter`, and `orderBy`.


molomby:
Yeah looks great. Two questions...

* Should we have `filter?: boolean` and `orderBy?: boolean` config on the list level too, to default values for the rest of the list? It's just sugar but, if I have a list with 40 fields and I never need to order it, nice to turn that functionality off in one place.
* Confirming we don't need to ever disable the single-node field, right? There's are some use cases when auto inc Ids are used where a dev might want to do this but they're pretty obscure and presumably it'd break a lot of core admin UI stuff.

molomby (approved):
👍🏼

timleslie:
> * Should we have `filter?: boolean` and `orderBy?: boolean` config on the list level too, to default values for the rest of the list? It's just sugar but, if I have a list with 40 fields and I never need to order it, nice to turn that functionality off in one place.

Definitely maybe. Let me build out the base pattern and then we can loop back and see if this is a worthwhile addition 👍 

> * Confirming we don't need to ever disable the single-node field, right? There's are some use cases when auto inc Ids are used where a dev might want to do this but they're pretty obscure and presumably it'd break a lot of core admin UI stuff.

I don't know what you mean by `single-node field` 🤔

gautamsi:
From code it feels like you are replacing the access control with new feature in generating the mutation and queries for the list. Should this be combination of two?

timleslie:
> From code it feels like you are replacing the access control with new feature in generating the mutation and queries for the list. Should this be combination of two?

The aim here is to separate these two concerns, as they're actually very different things, and historically it's been confusing to explain why particular usage of the Access Control API would make changes to the generated GraphQL API. These changes separate these two concerns and give the developer better control over how much of the autogenerated GraphQL API they want to include in their system.

packages/keystone/src/lib/core/field-assertions.ts
```diff
@@ -87,12 +87,15 @@ function assertIdFieldGraphQLTypesCorrect(list: ListForValidation) {
...
-  if (idField.access.read === false) {
+  if (idField.graphql.isEnabled.read === false) {
     throw new Error(
-      `The idField on a list must not have access.read be set to false but ${list.listKey} does`
+      `The idField on a list must not have graphql.isEnabled.read be set to false but ${list.listKey} does`
     );
   }
-  if (idField.output.type.kind !== 'non-null' || idField.output.type.of !== graphql.ID) {
+  if (
+    idField.output &&
```
mitchellhamilton, line 96:
```suggestion
    !idField.output ||
```

packages/keystone/src/lib/core/types-for-lists.ts
```diff
@@ -262,7 +312,37 @@ export function initialiseLists(
...
+            }
+            if (isEnabled?.read === false) {
+              delete f.output;
+            }
+            if (isEnabled?.update === false) {
+              delete f.input?.update;
+            }
+            if (isEnabled?.create === false) {
+              delete f.input?.create;
+            }
```
mitchellhamilton, line 344: Can you copy this stuff to avoid mutating what's returned from fields?

packages/keystone/src/lib/core/types-for-lists.ts
```diff
@@ -288,10 +368,30 @@ export function initialiseLists(
...
+              hooks: field.hooks ?? {},
+              graphql: field.graphql as {
+                isEnabled: {
+                  read: boolean;
+                  create: boolean;
+                  update: boolean;
+                  filter: boolean;
+                  orderBy: boolean;
+                };
+              },
```
mitchellhamilton, line 386: After using a spread for the field stuff above, this cast should be unnecessary

packages/keystone/src/lib/core/where-inputs.ts
```diff
@@ -62,8 +62,10 @@ export async function resolveWhereInput(
           };
         }
         const field = list.fields[fieldKey];
-        // we know if there are filters in the input object with the key of a field, the field must have defined a where input so this non null assertion is okay
-        const where = field.input!.where!;
+        const where = field.input!.where;
+        if (!where) {
+          throw new Error(`The field "${list.listKey}.${fieldKey}" does not support filtering.`);
+        }
```
mitchellhamilton, line 68: When will this happen?

timleslie: This happens when someone is using the items API and they use a filter which doesn't exist.

mitchellhamilton: But if the filter doesn't exist, they'll get a validation error from GraphQL and never get there?

timleslie: Let me pull up an example, I was able to trigger this exact code path while running tests that hadn't been updated to enable filters.

timleslie: Ah yeah, it's in access control, if you return declarative access control (e.g. a filter) it will send the returned object through this code path, at which point we catch the error.

mitchellhamilton: We run GraphQL validation on filters returned from access control though so this still shouldn't be hit, do you have a test I could have a look at?

packages/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/FilterAdd.tsx
```diff
@@ -105,7 +105,7 @@ function FilterAddPopoverContent({ onClose, listKey }: { onClose: () => void; li
     > = {};
     Object.keys(list.fields).forEach(fieldPath => {
       const field = list.fields[fieldPath];
-      if (field.controller.filter) {
+      if (field.isFilterable) {
```
mitchellhamilton, line 108:
```suggestion
      if (field.isFilterable && field.controller.filter) {
```

packages/keystone/src/lib/core/types-for-lists.ts
```diff
@@ -262,7 +312,34 @@ export function initialiseLists(
...
+            let f = fieldFunc({ fieldKey, listKey, lists: listInfos, provider });
+
+            const isEnabled = {
+              read: f.graphql?.isEnabled?.read ?? true,
+              update: f.graphql?.isEnabled?.update ?? true,
+              create: f.graphql?.isEnabled?.create ?? true,
+              filter: f.graphql?.isEnabled?.filter ?? false,
+              orderBy: f.graphql?.isEnabled?.orderBy ?? false,
+            };
+            const field = { ...f, graphql: { ...f.graphql, isEnabled } };
```
mitchellhamilton, line 324: (I forgot this in the diff I sent before I edited it)
```suggestion
            const field = { ...f, graphql: { ...f.graphql, isEnabled }, input: { ...f.input } };
```

packages/keystone/src/types/config/lists.ts
```diff
@@ -194,6 +194,23 @@ export type ListGraphQLConfig = {
...
+    | {
+        query?: boolean; // Does item()/items() exist? Can I read via a related item? (?). Will disable entirely in the admin UI if this is selected, because otherwise... pain.
+        // Mutations
+        create?: boolean; // Does createItem/createItems exist? Does `create` exist on the RelationshipInput types?
+        update?: boolean; // Does updateItem/updateItems exist?
+        delete?: boolean; // Does deleteItem/deleteItems exist?
+
+        // Defaults to apply to all fields. Default: false
+        filter?: boolean; // The default value to use for graphql.isEnabled.filter on all fields for this list
+        orderBy?: boolean; // The default value to use for graphql.isEnabled.orderBy on all fields for this list
```
mitchellhamilton, line 212: We should discuss this(not today), you'd intuitively expect that since this defaults to false that you can't filter or orderBy on this list but that's not true, it's only about disabling all the filters except the id filters (I am _not_ saying that we should let people disable filtering or ordering by ids, just that the way this is presented is confusing)

packages/keystone/src/lib/core/types-for-lists.ts
```diff
@@ -152,7 +233,13 @@ export function initialiseLists(
         const { fields } = lists[listKey];
         return Object.fromEntries(
           Object.entries(fields).flatMap(([key, field]) => {
-            if (!field.input?.orderBy?.arg || field.access.read === false) return [];
+            if (
+              !field.input?.orderBy?.arg ||
+              !field.graphql.isEnabled.read ||
+              !field.graphql.isEnabled.orderBy
```
mitchellhamilton, line 239: I'm not sure about whether `field.graphql.isEnabled.read === false` should disable filtering and ordering, I can see arguments both ways but we should discuss

packages/keystone/src/lib/core/field-assertions.ts
mitchellhamilton: Seems like this change was reverted?
```suggestion
    !idField.output ||
```

timleslie: Ah yep, though this new check isn't actually needed anymore, since the change to output has been reverted, so I'm just gonna set this back to what it was originally.

packages/keystone/src/lib/core/queries/output-field.ts
```diff
@@ -74,7 +74,7 @@ function getValueForDBField(
 }
 
 export function outputTypeField(
-  output: NextFieldType['output'],
+  output: Exclude<NextFieldType['output'], undefined>,
```
mitchellhamilton, line 77: This is required again right?
```suggestion
  output: NextFieldType['output'],
```

timleslie:
Merging this so as to unblock the work on access control. Some of the decisions implemented in this PR may be adjusted before release.

PR #6377, opened 2021-08-20
Move the exports of `@keystone-next/keystone` to `@keystone-next/keystone/system`

emmatown:
Using the current exports of `@keystone-next/keystone` is way less common than using the exports of`@keystone-next/keystone/schema`so we're moving the current exports of `@keystone-next/keystone` to `@keystone-next/keystone/system` and we'll be moving `@keystone-next/keystone/schema` to `@keystone-next/keystone` in a future PR.


timleslie (approved):
👍

PR #6382, opened 2021-08-20
6261/delete alert tests

gwyneplaine:
Minor changes to the test utils in the admin-ui-tests folder, and added tests for deletion success and failure alerts.


tests/admin-ui-tests/utils.ts
```diff
@@ -115,7 +124,7 @@ export const adminUITests = (
       beforeAll(async () => {
         await deleteAllData(projectDir);
       });
-      tests(playwright[browserName]);
+      tests(playwright[browserName], deleteAllData);
```
gwyneplaine, line 127: pass `deleteAllData` to our test fn so we can call it as necessary from within each test file.

tests/admin-ui-tests/list-view-crud.test.ts
```diff
@@ -0,0 +1,126 @@
+import { Browser, Page } from 'playwright';
+import fetch from 'node-fetch';
+import { adminUITests } from './utils';
+
+adminUITests('./tests/test-projects/crud-notifications', (browserType, deleteAllData) => {
+  let browser: Browser = undefined as any;
+  let page: Page = undefined as any;
+  const seedData = async (page: Page, query: string, variables?: Record<string, any>) => {
+    try {
+      const { errors } = await fetch('http://localhost:3000/api/graphql', {
```
gwyneplaine, line 10: we were doing things like this previously 
```
await page.evaluate((query, variable) => {
   return fetch('http://localhost:3000/api/graphql', {
        method: 'POST',
        headers: {
          'Content-Type': 'application/json',
        },
        body: JSON.stringify({
          query,
          variables,
        }),
      }).then(res => res.json());
}, { query, variable)
```
but I ripped that out in favor of node-fetch, as we were getting a few rather non-deterministic network issues with this method, (most likely due to shifting network contexts from where page.evaluate is called) 

Running requests to api from node seems like the better more deterministic choice here, but I'm happy to re-evaluate this choice if i'm missing something here.

yarn.lock
```diff
@@ -4025,11 +4025,6 @@ babel-plugin-polyfill-regenerator@^0.2.2:
   dependencies:
     "@babel/helper-define-polyfill-provider" "^0.2.2"
 
-babel-plugin-remove-graphql-queries@2.7.2:
```
timleslie, line 4028: Is this removal intentional?

gwyneplaine: definitely not, let me revert from master and see what's going on

timleslie (approved):
👍

PR #6385, opened 2021-08-22
fixed count in relationship field count mode

gautamsi:
All my lists were broken as i decided to use `displayMode: 'count'`.

this count was using old Meta count instead of fieldCount field


gautamsi:
@timleslie I am not sure if this qualifies for quick patch release as the next release is 2 weeks out

timleslie:
@gautamsi Thanks for the PR, we're just weighing up whether this warrants a hot-fix release or if it can wait for the next release cycle, but the fix itself looks good 👍

timleslie (approved):
Let's get this merged 👍

PR #6391, opened 2021-08-23
Add `introspection` support to Apollo Server Config

bladey:
Adds support for `introspection` in the Apollo Server config. Introspection enables you to query a GraphQL server for information about the underlying schema. If the playground is enabled then introspection is automatically enabled - unless specifically disabled.


timleslie (approved):
Nice one 👍

PR #6392, opened 2021-08-23
Fix negative `take` values above the list's `maxResults` not causing an early error

emmatown:


packages/keystone/src/lib/core/queries/resolvers.ts
```diff
@@ -200,7 +200,7 @@ function applyMaxResults(results: unknown[], list: InitialisedList, context: Key
     throw limitsExceededError({ list: list.listKey, type: 'maxResults', limit: list.maxResults });
   }
   if (context) {
-    context.totalResults += Array.isArray(results) ? results.length : 1;
+    context.totalResults += results.length;
```
mitchellhamilton, line 203: (this is unrelated to the `take` thing, i just saw that this was always an array now)

timleslie (approved):
👍

PR #6393, opened 2021-08-23
Update @graphql-ts/schema, rename `schema` export to `graphql` and move it to main entrypoint

emmatown:
Given this needs to be used when using a virtual field(and it'll likely have an alternative schema extension API in the future), it makes sense to export it from a more prominent main entrypoint than `/types`. The rename to `graphql` makes it clear it's about GraphQL where just seeing the word `schema` in a Keystone config could mean any number of things.


timleslie (approved):
👍

PR #6403, opened 2021-08-25
Add config.experimental.contextInitialisedLists

timleslie:
This provides the full implementation of the idea suggested in #6059. This API is temporary until the `context.config` API is finalised.


emmatown (changes requested):
Could we put this under an experimental key on context as well and make it optional rather than passing in an empty object if it's disabled?

Also, what stability guarantees are you suggesting for this?

timleslie:
> Could we put this under an experimental key on context as well and make it optional rather than passing in an empty object if it's disabled?

Yep, good call 👍 
 
> Also, what stability guarantees are you suggesting for this?

None at all.

mitchellhamilton:
> None at all.

As in, we'll break it in a patch version, yes?

mitchellhamilton:
And if so, could you make that clear in the changeset and the JSDoc comments around the places?

timleslie:
> > None at all.
> 
> As in, we'll break it in a patch version, yes?

Yep, precisely.

emmatown (approved):

dcousens:
@MurzNN, @gautamsi could you provide context on why this was helpful?

gautamsi:
I was working on some plugin/hooks which could be taking advantage of same to create generic hooks independent of list names but still works with other list. it took a while to get thig thing in but we moved to different solution at the time.

MurzNN:
@dcousens  In our projects we use the `context.experimental.initialisedLists` in many backend functions, to get the detailed information about lists directly, without querying it as separate GraphQL query.

For example, sometimes we need to get the `label` field for specific list, so we have the function:
```js

export const getLabelFieldForLists = (context, listName) => {
  const list: any = context.experimental?.initialisedLists[listName];
  return (
    (list.ui?.labelField as string | undefined) ??
    (list.fields.label ? 'label' : list.fields.name ? 'name' : list.fields.title ? 'title' : 'id')
  );
};
```

Also we have a function that should go through all lists to find all relationship fields and do some business logic with them, something like this:
```js
async function getDirectDependencies(
  existingItem: BaseItem,
  context: KeystoneContext,
  listKey: string,
  deleteRules: RelatedItems,
  limitByType: number,
  addRules: boolean
) {
  const initialisedLists: any = context.experimental?.initialisedLists ?? [];
  const result = [];

  const listFields = [];
  const relatedListsFields: any[] = [];
  for (const fieldName in deleteRules.direct) {
    const relatedListKey = initialisedLists[listKey].fields[fieldName].dbField.list;
    const allListFieldNames: any = Object.keys(initialisedLists[relatedListKey].fields).reduce(
      (prev, cur) => ({ ...prev, [cur]: cur }),
      {}
    );
    const relatedListFields = {
      fieldName,
      labelName: getItemLabel(allListFieldNames),
      isMany: !!initialisedLists[listKey].fields[fieldName].many,
      relatedListKey,
      deleteRule: deleteRules.direct[fieldName],
    };
```

So the common usage is to retrieve the information about list fields, and detailed information about field (type, dbField, isMany, etc).

PR #6412, opened 2021-08-26
merge aria-description text into aria-label

### LINKED ISSUE #6274: aria-description in popovers
Presently aria-description is used to give additional usage context for popover dialogs  in the ListPage. 
This is an ARIA 1.3 spec inclusion https://w3c.github.io/aria/#aria-description, so browser support is only at around 71%.
This manifests itself not only in [React console errors](https://github.com/facebook/react/issues/21035), but also the descriptive text not being read out in unsupported browsers. https://caniuse.com/?search=aria-description

Ideally we should either have a more universally supported solution (aria-describedby or visually hidden element), or fallback to one if the attribute is not supported.

gwyneplaine:
resolves #6274 
PR merges aria-description text into aria-label attribute for:
* Filter options dialog 
* Sort options dialog
* Column options dialog   

not leaning on `aria-describedby` here as we don't really want the description text to be visible in the DOM, just additional context for screen reader users.

### DIFF
--- a/.changeset/spotty-ducks-wonder.md
+++ b/.changeset/spotty-ducks-wonder.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Merged aria-description text for dialogs on list view in Admin UI into their respective aria-labels.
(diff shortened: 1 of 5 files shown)


dominikwilkowski (approved):
Nice solve!

PR #6426, opened 2021-08-30
Update access control APIs

timleslie:


docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,244 +20,171 @@ export default config({
...
-Access control is applied to the generated CRUD (**c**reate, **r**ead, **u**pdate, **d**elete) queries and mutations in the [GraphQL API](./graphql).
-Access control is applied before any [hooks](./hooks) are executed.
+Keystone lets you apply access control on a per-list basis.
+By default it allows all operations for all users.
+Access control is applied to the generated CRUD (**c**reate, **r**ead, **u**pdate, **d**elete) operations and mutations in the [GraphQL API](./graphql), and is applied before any [hooks](./hooks) are executed.
 
-You can specify access control using either **concise** or **verbose** syntax.
-Verbose syntax provides a separate access control rule for each operation type; `create`, `read`, `update`, and `delete`.
-Concise syntax provides a single access control rule which is used for all operation types.
+List-level access control can be specified in three ways, of increasing power.
```
Noviny, line 32: Probs a nitpick, but 'increasing power' doesn't explain what this means to me.

docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,244 +20,171 @@ export default config({
...
+If access is **denied** due to any of the access control methods, the following responses will be returned from the GraphQL API:
 
-### Declarative (list)
+- **Mutations**
+  - **Single** operations return `null` and an access denied error.
+  - **Multi** operations return a data array with `null` values for the items which have access denied.
+    Error responses are returned for each `null` item.
+- **Queries**
+  - **Single** item queries return `null` with no errors.
+  - **Many** item queries filter out items which have access denied. No error responses are returned.
```
Noviny, line 48: So unlike multi mutations this will be an array without nulls? Probs not relevant to these docs, but is there reasons for this?

docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,244 +20,171 @@ export default config({
...
-      access: args => true,
+      fields: {
+        isReadable: checkbox(),
+        isUpdatable: checkbox(),
+        isDeletable: checkbox(),
+      }
+      access: {
+        filter: {
+          query: ({ session, context, listKey, operation }) => {
+            return { isReadable: { equals: true } };
```
Noviny, line 105: I'm assuming that the reason that this is not the `true | false` dichotomy of the others is that we add this to the graphql query. Can we call this out a bit more explicitly?

docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,244 +20,171 @@ export default config({
...
-        itemId,
-        context,
-      }) => {
-        return true;
-      },
-    }),
-  }),
-});
-```
+### Item (mutations only)
```
Noviny, line 122: Is why this is mutation only documented somewhere?

docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,244 +20,171 @@ export default config({
...
-          // Verbose access control definition
-          access: {
-            create: true,
-            read: true,
-            update: true,
-          },
-        }),
+        isReadable: checkbox(),
+        isUpdatable: checkbox(),
+        isDeletable: checkbox(),
```
Noviny, line 142: Since we're not using these fields in this example should we remove them for decluttering?

(or is this intended as a progressive build up from the last one...)

docs/pages/docs/apis/access-control.mdx
```diff
@@ -268,55 +195,32 @@ export default config({
     ListKey: list({
       fields: {
         fieldName: text({
-          // Imperative access control definition
-          access: args => true,
+          access: {
+            read: ({ session, context, listKey, fieldKey, operation, item }) => true,
+            create: ({ session, context, listKey, fieldKey, operation, originalInput }) => true,
```
Noviny, line 200: Wait, `create` gets `originalInput` with no item? Consistency matching bikeshed conversation?

Noviny (commented):
Opening some bikesheds most likely.

My biggest comment is that there's a bunch of context (you can do this, you can't do this) that I'm trying to infer from this. I'd love a document on the design of this API and why these choices separate to this.

docs/pages/docs/apis/access-control.mdx
timleslie: Oh good catch, definitely should be removed!

docs/pages/docs/apis/access-control.mdx
timleslie: That's correct. When you `create` an object, it doesn't already exist, so `item` has no meaning for that operation.

docs/pages/docs/apis/access-control.mdx
timleslie: That's correct. The general principle here is "queries never have errors. access denied should look identical to objects not existing".

So with that principle in mind, a `many` query where some items are access denied, the expected result is an array where those items simply don't show up at all.

It's probably worth us adding that general principle somewhere up the top of the document, as it helps inform each of the specific behaviours described in the details.

docs/pages/docs/apis/access-control.mdx
timleslie: Sentence removed 👍

docs/pages/docs/apis/access-control.mdx
timleslie: Updated to read as such:
```
- For **mutations**, the access control filter will be combined with the unique identifier provided to the operation, and access will be denied if no item is found with this combined filter.
- For **queries**, the access control filter will be combined with the query filter and only items which match both filters will be returned.
```

docs/pages/docs/apis/access-control.mdx
Noviny: Pls add this principle. Would be super helpful

docs/pages/docs/apis/access-control.mdx
timleslie:
```
?> Item-level access control is not available for `query` operations.
Applying access control after fetching items would lead to inconsistent pagination behaviour and incorrect `count` results.
```

docs/pages/docs/apis/access-control.mdx
```diff
@@ -2,7 +2,7 @@ import { Markdown } from '../../../components/Markdown';
 
 # Access Control API
 
-The `access` property of the [list configuration](./schema) and [field configuration](./fields) objects configures who can read, create, update, and delete items in your Keystone system.
+The `access` property of the [list configuration](./schema) and [field configuration](./fields) objects configures who can query, create, update, and delete items in your Keystone system.
```
mitchellhamilton, line 5: > field configuration

> who can query, create, update, and delete **items**

field configuration doesn't configure the access for __items__?

docs/pages/docs/apis/access-control.mdx
```diff
@@ -20,81 +20,74 @@ export default config({
...
+        }
+      },
     }),
   }),
 });
 ```
 
-### Declarative (list)
+?> The `query` access control rule is applied when running GraphQL query operations.
+It does not prevent a user **reading** the data returned by the mutation operations.
```
mitchellhamilton, line 80: might be good to explicitly say how someone _can_ prevent a user reading data: to prevent a user from reading an item, you need to have access control rules for create, update and delete as well as query

tests/api-tests/utils.ts
```diff
@@ -69,11 +69,13 @@ export const expectAccessDenied = (
...
   //   - graphql.debug is true or
   //   - graphql.debug is undefined and mode !== production or
-  const expectDebug =
-    _debug === true || (_debug === undefined && mode !== 'production') || !httpQuery;
+  // const expectDebug =
+  //   _debug === true || (_debug === undefined && mode !== 'production') || !httpQuery;
   // We expect to see the Apollo exception under the same conditions, but only if
   // httpQuery is also true.
-  const expectException = httpQuery && expectDebug;
+  const expectException = false;
```
mitchellhamilton, line 76: What do the changes here mean?

packages/keystone/src/types/config/access-control.ts
```diff
@@ -7,102 +7,137 @@ type BaseAccessArgs = {
...
 
-type CreateAccessArgs<GeneratedListTypes extends BaseGeneratedListTypes> = BaseAccessArgs & {
+// List Filter Access
+
+type FilterOutput<GeneratedListTypes extends BaseGeneratedListTypes> =
+  | boolean
+  | GeneratedListTypes['inputs']['where'];
+
+export type ListFilterAccessControl<
+  Operation extends 'create' | 'query' | 'update' | 'delete',
```
mitchellhamilton, line 17:
```suggestion
  Operation extends 'query' | 'update' | 'delete',
```

packages/keystone/src/types/config/access-control.ts
```diff
@@ -7,102 +7,137 @@ type BaseAccessArgs = {
...
+  GeneratedListTypes extends BaseGeneratedListTypes
+> =
+  | FilterOutput<GeneratedListTypes>
```
mitchellhamilton, line 20: It feels inconsistent to allow the value directly here but not for the others. (i _think_ we should require a function here as well)

packages/keystone/src/lib/core/queries/resolvers.ts
```diff
@@ -89,10 +87,21 @@ export async function findMany(
...
+  if (!operationAccess) {
+    return [];
+  }
+
   applyEarlyMaxResults(take, list);
 
+  const accessFilters = await getAccessFilters(list, context, 'query');
+  if (accessFilters === false) {
+    return [];
+  }
```
mitchellhamilton, line 101: Why do `applyEarlyMaxResults(take, list)` in the middle of these two? (and the same question for the stuff between the unique where above)

packages/keystone/src/lib/core/mutations/nested-mutation-many-input-resolvers.ts
```diff
@@ -27,7 +27,10 @@ function getResolvedUniqueWheres(
     // Validate and resolve the input filter
     const uniqueWhere = await resolveUniqueWhereInput(uniqueInput, foreignList.fields, context);
     // Check whether the item exists
-    await context.db.lists[foreignList.listKey].findOne({ where: uniqueInput });
+    const item = await context.db.lists[foreignList.listKey].findOne({ where: uniqueInput });
+    if (item === null) {
+      throw new Error('Sorry but nope');
```
mitchellhamilton, line 32: Is this error visible publicly?

mitchellhamilton (changes requested):

packages/keystone/src/lib/core/mutations/nested-mutation-many-input-resolvers.ts
timleslie: Yep, will change 👍

packages/keystone/src/lib/core/queries/resolvers.ts
timleslie: It's pretty arbitrary, no specific motivation. I'll update it to do all the access control early updates first 👍

.changeset/many-beans-crash.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystone-next/fields-document': major
+'@keystone-next/keystone': major
```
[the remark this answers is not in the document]
timleslie: Yeah, it hasn't had any direct changes, but it should get a changeset 👍

tests/api-tests/utils.ts
timleslie: The previous code had the wrong logic, we just never exercised it. This will be cleaned up properly once the Apollo 3 stuff lands.

mitchellhamilton (approved):

PR #6429, opened 2021-08-30
Respect graphql.omit in fieldMode resolvers

timleslie:


emmatown (changes requested):
We need to either have a static fieldMode that signals "never fetch this" for the itemView or fetch the fieldModes and _then_ do the item query otherwise the whole query will be invalid (i would prefer the former)

timleslie:
> We need to either have a static fieldMode that signals "never fetch this" for the itemView or fetch the fieldModes and _then_ do the item query otherwise the whole query will be invalid (i would prefer the former)

Yeah, so that's definitely on the todo list for today, but I want to do it separate to this particular change, as it's going to require a more complex solution than these low-hanging fixes which need to go through.

packages/keystone/src/admin-ui/system/createAdminMeta.ts
```diff
@@ -98,7 +98,6 @@ export function createAdminMeta(
     }
 
     for (const [fieldKey, field] of Object.entries(list.fields)) {
-      if (field.access.read === false) continue;
```
mitchellhamilton, line 101: Could you add this back(but based on isEnabled) until we deal with this properly then?

packages/keystone/src/admin-ui/system/createAdminMeta.ts
```diff
@@ -98,6 +98,7 @@ export function createAdminMeta(
     }
 
     for (const [fieldKey, field] of Object.entries(list.fields)) {
+      if (field.access.read === false) continue;
```
mitchellhamilton: I meant check `graphql.isEnabled` here and exclude it if read is disabled

emmatown (approved):

PR #6437, opened 2021-08-31
Change isUnique: true to isIndexed: 'unique' in fields

emmatown:
This means trying to configure both a regular and unique index will be disallowed by TS instead of throwing a runtime error.


.changeset/quiet-tables-hunt.md
```diff
@@ -0,0 +1,8 @@
+---
+'@keystone-next/keystone': major
+'@keystone-next/auth': patch
+'@keystone-next/cloudinary': patch
+'@keystone-next/fields-document': patch
```
mitchellhamilton, line 5: These are patch because they're only changes to error messages

timleslie (approved):
👍

PR #6453, opened 2021-09-02
Add getChildTag fn to Inline component

gwyneplaine:
Same fix as #5858 
Move the common `getChildTag` utility to `utils` file, and apply to the `Inline` component.


packages/keystone/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/HomePage/index.tsx
```diff
@@ -158,7 +162,15 @@ export const HomePage = () => {
           <LoadingDots label="Loading lists" size="large" tone="passive" />
         </Center>
       ) : (
-        <Inline gap="large" paddingY="xlarge">
+        <Inline
+          as="ul"
+          gap="large"
+          paddingY="xlarge"
+          css={{
+            paddingInlineStart: '0px',
```
gwyneplaine, line 170: negating default ul styles

timleslie (approved):
👍

PR #6458, opened 2021-09-02
Add config.graphql.path option

timleslie:


packages/keystone/src/admin-ui/system/getAdminMetaSchema.ts
```diff
@@ -257,6 +257,9 @@ export function getAdminMetaSchema({
   const adminMeta = graphql.object<AdminMetaRootVal>()({
     name: 'KeystoneAdminMeta',
     fields: {
+      apiPath: graphql.field({
+        type: graphql.nonNull(graphql.String),
+      }),
```
mitchellhamilton, line 262: Why provide this when you need to know it to query it?

packages/keystone/src/admin-ui/templates/index.ts
```diff
@@ -41,7 +46,7 @@ export const writeAdminFiles = (
         {
           mode: 'write' as const,
           src: apiTemplate,
-          outputPath: 'pages/api/graphql.js',
+          outputPath: `pages/${config.graphql?.path || '/api/path'}.js`,
```
mitchellhamilton, line 49: In addition to the change below, this won't work if the path isn't within `/api` so we should either throw an error if `enableNextJsGraphqlApiEndpoint` is enabled and the path isn't within `/api` or remove `enableNextJsGraphqlApiEndpoint`.

```suggestion
          outputPath: `pages/${config.graphql?.path || '/api/graphql'}.js`,
```

emmatown (approved):

PR #6459, opened 2021-09-02
Admin UI menu available for unauthenticated users

timleslie:
The aim of this PR is to make the triple-dot menu available in the Admin UI even if authentication isn't being used.

Screen Shot 2021-09-03 at 9 53 10 am

### DIFF
--- a/.changeset/swift-feet-heal.md
+++ b/.changeset/swift-feet-heal.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Update Navigation component to show docs and playground links irrespective of authentication.
--- a/.changeset/swift-feet-heal.md
+++ b/.changeset/swift-feet-heal.md
@@ -2,4 +2,4 @@
 '@keystone-next/keystone': minor
 ---
 
-Update Navigation component to show docs and playground links irrespective of authentication.
+Updated Navigation component to show docs and playground links irrespective of authentication.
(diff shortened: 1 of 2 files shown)


gwyneplaine:
Minor update, so we don't have dangly white space (and people have a sense of what's in the popover) 

Signed in user:
Screen Shot 2021-09-03 at 10 24 39 am

Unauthed:
Screen Shot 2021-09-03 at 10 24 35 am

packages/keystone/src/admin-ui/components/Navigation.tsx
```diff
@@ -73,9 +73,14 @@ const AuthenticatedItem = ({ item }: { item: { id: string; label: string } }) =>
         marginBottom: 0,
       }}
     >
-      <div css={{ fontSize: typography.fontSize.small }}>
-        Signed in as <strong>{item.label}</strong>
-      </div>
+      {item && item.state === 'authenticated' ? (
```
gwyneplaine, line 76: I'd love to be able to pull this up into a single variable, but typescript can't infer the state otherwise so leaving it for now.

gwyneplaine (approved):

.changeset/swift-feet-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
+Update Navigation component to show docs and playground links irrespective of authentication.
```
timleslie, line 5:
```suggestion
Updated Navigation component to show docs and playground links irrespective of authentication.
```

timleslie:
Looks good, let's 🚢

PR #6476, opened 2021-09-06
Update intermittently failing nav test

gwyneplaine:
Looking at both failing and successful tests on CI, as well as interrogating via the playwright inspector. The failing test appears to be doing the following: 

1. Assert some stuff on '/'
2. Open create drawer for Task list 
3. Create a task item via the inline create drawer 
4. Wait for navigation after task creation
5. Make assertions on the nav once navigated to the item.

Step 4 is failing intermittently on CI, likely because the create operation takes too long and the test times out. (the avg time for successful run on this particular test is around 5k ms, which is not great) 

This PR refactors the test to create the task item via a direct `graphql` request beforehand, instead of creating it through the Admin UI flow. This roughly reduces the avg time to run the test by 25%, and should also reduce the intermittent failure rate for this test to close to 0 🤞 (as we're no longer waiting on the create operation via `waitForNavigation()`). 

We'll eventually have to look into why the create item flow is so slow in `dev`  though, since we'll need to write test cases explicitly to test this flow. 

In this PR we also: 
* Return the result of the executed query in `makeGqlRequest` 
* Remove a duplicative test that made it through to `master`


timleslie (approved):
👍

PR #6477, opened 2021-09-06
Fix windows issues in view resolver

gautamsi:
Fixed issues with Windows due to view path resolving changes in #6414 

this is similar to fix in #4636

I would insist to release a fix soon, possibly club few more such issues without waiting for next release cycle

### DIFF
--- a/.changeset/orange-rice-return.md
+++ b/.changeset/orange-rice-return.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fixed windows issues with new view resolver from #6414.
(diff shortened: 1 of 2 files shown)


mitchellhamilton (changes requested):
Could you explain how exactly it breaks, the paths are escaped, so it should work already?

gautamsi:
Generated _app.ts Import comes back as `..\..\..\..\xxx\yyy`

mitchellhamilton:
Okay, could you move the replace to `serializePathForImport`?

gautamsi:
@mitchellhamilton Done

PR #6494, opened 2021-09-07
Make all docs pages editable via Github

raveling:
Exposes `edit on github` functionality across all docs pages.

CleanShot 2021-09-08 at 09 08 35
CleanShot 2021-09-08 at 09 08 24


timleslie (approved):
👍

PR #6501, opened 2021-09-08
fix: the tabbable error when you click the popover in the keystone-ui website

Oluwasetemi:
I noticed in the keystone-ui website, there was a section showing an example of popover. The example has a bug and the error was something around this line `Your focus-trap must have at least one container with at least one tabbable node in it at all times`. I attempt to fix this with this PR. 

Add tabIndex={0} will make the element in question tabbable.

### DIFF
--- a/design-system/website/pages/components/popover.tsx
+++ b/design-system/website/pages/components/popover.tsx
@@ -1,8 +1,8 @@
 /** @jsxRuntime classic */
 /** @jsx jsx */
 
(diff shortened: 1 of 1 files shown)


bladey:
Thanks @Oluwasetemi!

bladey (approved):

Oluwasetemi:
Thanks @bladey! Do you think there is any plan to make the keystone-ui design guide website? Or is it an internal tool made open.

PR #6510, opened 2021-09-09
Fix explanation of allowed Field-level operations

loklaan:
* The "delete" operation was mentioned as being settable for fields - fixed
* The "create" operation was mentioned as being controlled by "delete" operation functions - fixed
* The order of information in "delete" operation tip was backwards in my opinion, so I reversed it to mention "List-level" as soon as possible - this one is subjective


loklaan:
Looks like an unrelated test timed-out

timleslie (approved):
Great catches, thanks for the PR 🙏

PR #6512, opened 2021-09-09
Remove unqueryable lists from admin meta

gwyneplaine:
update createAdminMeta to not include lists with graphql querying disabled. also apply this to relationships to those omitted lists

https://user-images.githubusercontent.com/12119389/132619815-a33f6139-df7d-4d0d-8897-0230969b2c61.mov

**PLEASE FINE TOOTH COMB THIS PR**
As I've operated here with half context at best.

### DIFF
--- a/.changeset/twelve-apples-happen.md
+++ b/.changeset/twelve-apples-happen.md
@@ -2,4 +2,4 @@
 '@keystone-next/keystone': patch
 ---
 
-Fixed unqueryable lists being accessible via the Admin UI.
+Fixed lists with `graphql.omit: ['query']` causing issues in the Admin UI.
--- a/.changeset/twelve-apples-happen.md
+++ b/.changeset/twelve-apples-happen.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fixed unqueryable lists being accessible via the Admin UI.
--- a/packages/keystone/src/admin-ui/system/createAdminMeta.ts
+++ b/packages/keystone/src/admin-ui/system/createAdminMeta.ts
@@ -16,8 +16,17 @@ export function createAdminMeta(
     views: [],
   };
 
+  const omittedLists: string[] = [];
+
   for (const [key, list] of Object.entries(initialisedLists)) {
     const listConfig = lists[key];
+    if (list.graphql.isEnabled.query === false) {
+      // If graphql querying is disabled on the list,
+      // push the key into the ommittedLists array for use further down in the procedure and skip.
+      omittedLists.push(key);
+
+      continue;
+    }
     // Default the labelField to `name`, `label`, or `title` if they exist; otherwise fall back to `id`
     const labelField =
       (listConfig.ui?.labelField as string | undefined) ??
@@ -80,6 +89,7 @@ export function createAdminMeta(
   }
   // Populate .fields array
   for (const [key, list] of Object.entries(initialisedLists)) {
+    if (omittedLists.includes(key)) continue;
     const searchFields = new Set(config.lists[key].ui?.searchFields ?? []);
     if (searchFields.has('id')) {
       throw new Error(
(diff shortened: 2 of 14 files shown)


packages/keystone/src/admin-ui/system/createAdminMeta.ts
```diff
@@ -98,6 +108,8 @@ export function createAdminMeta(
     }
 
     for (const [fieldKey, field] of Object.entries(list.fields)) {
+      // If the field is a relationship field and is related to an omitted list, skip.
+      if (field.dbField.kind === 'relation' && omittedLists.includes(field.dbField.list)) continue;
```
gwyneplaine, line 112: We want to check if the field is a relationship to an omitted list, is interrogating the dbField on said list the right way to make this check? @timleslie

packages/keystone/src/admin-ui/system/createAdminMeta.ts
```diff
@@ -16,8 +16,17 @@ export function createAdminMeta(
     views: [],
   };
 
+  const omittedLists: string[] = [];
+
   for (const [key, list] of Object.entries(initialisedLists)) {
     const listConfig = lists[key];
+    if (list.graphql.isEnabled.query === false) {
```
gwyneplaine, line 23: is `isEnabled.query` the only flag we want to skip on here?

mitchellhamilton (dismissed):
I need to look at things

packages/keystone/src/admin-ui/system/createAdminMeta.ts
mitchellhamilton: As a separate thing, if `list.graphql.isEnabled.create`/`list.graphql.isEnabled.delete` is false, we should force hideCreate/hideDelete to return false https://github.com/keystonejs/keystone/blob/3cfc2a3839142dd3ccdbf1dd86768257e9acc0dc/packages/keystone/src/admin-ui/system/getAdminMetaSchema.ts#L204-L227. We should also force the itemView fieldMode to read for all the fields of a list when the list has update omitted so you can't get a confusing error by editing and then trying to save (doing that is in the same file as above)

While we could technically make the creating in the Admin UI work for a list that has querying disabled so arguably the above check should be `isEnabled.type`, we definitely shouldn't do that now so `isEnabled.query` is correct imo.

timleslie: Yes.

packages/keystone/src/admin-ui/system/createAdminMeta.ts
timleslie: Yep

packages/keystone/src/admin-ui/system/createAdminMeta.ts
gwyneplaine: > As a separate thing, if `list.graphql.isEnabled.create`/`list.graphql.isEnabled.delete` is false, we should force hideCreate/hideDelete to return false
> 
> https://github.com/keystonejs/keystone/blob/3cfc2a3839142dd3ccdbf1dd86768257e9acc0dc/packages/keystone/src/admin-ui/system/getAdminMetaSchema.ts#L204-L227
> 
> . We should also force the itemView fieldMode to read for all the fields of a list when the list has update omitted so you can't get a confusing error by editing and then trying to save (doing that is in the same file as above)
> While we could technically make the creating in the Admin UI work for a list that has querying disabled so arguably the above check should be `isEnabled.type`, we definitely shouldn't do that now so `isEnabled.query` is correct imo.

As discussed will implement this in a separate PR

.changeset/twelve-apples-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+Fixed unqueryable lists being accessible via the Admin UI.
```
timleslie, line 5:
```suggestion
Fixed lists with `graphql.omit: ['query']` causing issues in the Admin UI.
```

mitchellhamilton (approved):

PR #6513, opened 2021-09-09
Remove isRequired and defaultValue from some fields

emmatown:
For these fields, these are the only changes needed in the field updates.


timleslie (approved):
👍

PR #6514, opened 2021-09-09
Relationship field updates

emmatown:


examples/default-values/schema.ts
```diff
@@ -27,15 +27,21 @@ export const lists = createSchema({
...
+                where: { name: { equals: 'Anonymous' } },
+              });
+              if (anonymous.length > 0) {
+                return { connect: { id: anonymous[0].id } };
+              }
+            }
+            // If we don't have an anonymous user we return the value
+            // that was passed in(which might be nothing) so as not to apply any default
+            return resolvedData.assignedTo;
+          },
```
mitchellhamilton, line 44: This whole example will need to be updated but that can happen later when all the field have been updated

examples-staging/ecommerce/schemas/Product.ts
```diff
@@ -44,8 +44,14 @@ export const Product = list({
     price: integer(),
     user: relationship({
       ref: 'User.products',
-      defaultValue: ({ context }) =>
-        context.session?.itemId ? { connect: { id: context.session?.itemId } } : null,
+      hooks: {
+        resolveInput({ operation, resolvedData, context }) {
```
timleslie, line 48:
```suggestion
        resolveInput({ operation, resolvedData, context }) {
          // Default to the currently logged in user on create.
```

Took my brain too long to work out the semantics of this, so adding a comment to save the next poor soul.

timleslie (approved):
👍

PR #6519, opened 2021-09-10
Skip fetching fields with static fieldMode: 'hidden' on the item view

emmatown:
The main reason behind this is to allow never fetching data for a virtual field if it's not needed in the Admin UI


timleslie (approved):
👍

gautamsi:
This could be helpful when the conditional rendering of admin ui fields feature added in near future.

PR #6525, opened 2021-09-10
Remove unused function coerceAndValidateForGraphQLInput

timleslie:


mitchellhamilton (changes requested):
This should be used for access control filters, I must have missed that it was removed when reviewing the access control changes

Issue #6465, opened 2021-09-05
Admin UI linking a tag to a post in default example throws an error


jrmyio:
## Describe the bug

Linking an existing tag to a post in the default keystone-next example throws a graphql error when selecting it in the tag from the select field:

**Request payload:**
{"variables":{"ids":["ckt7762ds10138bcjx4fibihyq"]},"query":"query ($ids: [ID!]!) {\n  items: tags(where: {id_in: $ids}) {\n    name\n    id\n    __typename\n  }\n}\n"}

**Error Response**
{"errors":[{"message":"Field \"id_in\" is not defined by type \"TagWhereInput\". Did you mean \"id\"?","locations":[{"line":2,"column":23}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED","exception":{"stacktrace":["GraphQLError: Field \"id_in\" is not defined by type \"TagWhereInput\". Did you mean \"id\"?","    at ObjectField (/home/test/keystone/node_modules/graphql/validation/rules/ValuesOfCorrectTypeRule.js:73:29)","    at Object.enter (/home/test/keystone/node_modules/graphql/language/visitor.js:323:29)","    at Object.enter (/home/test/keystone/node_modules/graphql/utilities/TypeInfo.js:370:25)","    at visit (/home/test/keystone/node_modules/graphql/language/visitor.js:243:26)","    at Object.validate (/home/test/keystone/node_modules/graphql/validation/validate.js:69:24)","    at validate (/home/test/keystone/node_modules/apollo-server-core/src/requestPipeline.ts:536:14)","    at Object.<anonymous> (/home/test/keystone/node_modules/apollo-server-core/src/requestPipeline.ts:302:32)","    at Generator.next (<anonymous>)","    at fulfilled (/home/test/keystone/node_modules/apollo-server-core/dist/requestPipeline.js:5:58)","    at runMicrotasks (<anonymous>)","    at processTicksAndRejections (node:internal/process/task_queues:96:5)"]}}}]}

## To Reproduce

- Install keystone and go to the admin UI. 
- Create a Post and Create a Tag, then save the post.
- Then remove the tag from the post and Save Changes.
- Now try to select the tag in the input field again. This will throw a graphql error as shown above.

## Expected behaviour
It should not throw an error and add the tag to the post.

## Package.json
```json
{
  "name": "keystone-app",
  "version": "2.0.12",
  "private": true,
  "scripts": {
    "dev": "keystone-next dev",
[...]

PR #6526, opened 2021-09-10
Fixes Relationship field inline connect fields throwing 400 errors

bladey:
Issue with all details: https://github.com/keystonejs/keystone/issues/6465


timleslie (approved):
👍

PR #6530, opened 2021-09-12
fix windows build issue for keystone-next build

gautamsi:
during build, this file is being generated wrong on windows, I traced this to the duplicate code for which I fixed one part. now removed duplicate code and this one should be gone.

`\.keystone\admin\pages\api\__keystone_api_build.js`
```
export { default as config } from '..\\..\\..\\..\\keystone';
export default function (req, res) {
  return res.status(500);
}

```

### DIFF
--- a/.changeset/strange-pianos-raise.md
+++ b/.changeset/strange-pianos-raise.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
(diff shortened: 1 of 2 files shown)


timleslie:
This should be fine once lint failures are fixed up 👍

timleslie (approved):
👍

PR #6533, opened 2021-09-12
`create-keystone-app` outputed URLs sometimes have hidden characters, cater for this

bladey:
Added redirect to cater for this issue, paraphrased from @Noviny:

>The links the keystone generator spits sometimes don’t work, specifically don’t work in some terminal environments.

>The error is that the library adds \u200B to the end of the link as which some terminals don't support links natively, but then you click or copy the link, it includes the space in the URL when copied, which is then not a valid URL.

### DIFF
--- a/docs/pages/docs/examples.tsx
+++ b/docs/pages/docs/examples.tsx
@@ -2,17 +2,17 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/react';
 
+import Link from 'next/link';
 import { GitHubExamplesCTA } from '../../components/docs/GitHubExamplesCTA';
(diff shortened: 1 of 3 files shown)


bladey:
Includes fixes for:

``` /home/runner/work/keystone/keystone/docs/pages/docs/examples.tsx
Error:   11:1   error  `next/link` import should occur before import of `../../components/docs/GitHubExamplesCTA`                                                              import/order
Error:   13:12  error  React Hook "useMediaQuery" cannot be called at the top level. React Hooks must be called in a React function component or a custom React Hook function  react-hooks/rules-of-hooks
```

raveling (approved):

PR #6535, opened 2021-09-13
[RFC] Rename `context.lists` to `context.query` and `context.db.lists` to `context.db`

timleslie:
(context provided by @JedWatson, this was my idea but @timleslie did the PR)

Based on my experience using and teaching the new Lists and DB APIs to developers picking up Keystone 6, they're both important and unique APIs that you really need to understand how to use well, but it's tricky to explain.

Once you get the mental model you're good but getting there isn't super intuitive. The hope is that this change will improve that for developers starting Keystone 6 or picking up Keystone project codebases in the future.

---

### TL;DR:

> If you want to query fields, use the **Query API**. It's functionally the same as the GraphQL API, available in a server-side context. If you want to retrieve data stored in the database, use the **DB API**. Input arguments, Access control and Hooks work identically in both APIs.

#### Query API

This lets you run queries and mutations against lists in the Keystone schema, as a shortcut for executing full GraphQL queries yourself. You provide the same input arguments and query the same output fields as in the GraphQL API (behind the scenes this is just sugar for executing queries against the internal schema)

Because it resolves return data through the GraphQL Schema you can query virtual fields, return deeply nested relational data, etc.

#### Database API

This lets you run the same queries and mutations against lists in the Keystone schema that the Query API does, but will return the backing data for each item directly from the database without going through the GraphQL field resolvers.

The only difference is in the return type; the input arguments (for where / data / etc) as well as access control and hooks functionality is identical between both APIs.

The primary use for this API is if you want to return a Keystone item type from a custom query or mutation in your GraphQL extensions.

Additionally, you get all the data stored in the database for the item (equivalent to `select * from {table}`) so you can access some data that fields do not make accessible through their GraphQL resolvers, like password hashes which you would need to query to compare with a candidate when validating a signing attempt (but are never available through the GraphQL API)

### DIFF
--- a/docs/components/docs/Navigation.tsx
+++ b/docs/components/docs/Navigation.tsx
@@ -192,7 +192,7 @@ export function DocsNavigation() {
 
         <SubHeading>Context</SubHeading>
         <NavItem href="/docs/apis/context">Context API</NavItem>
-        <NavItem href="/docs/apis/queries">Query API</NavItem>
+        <NavItem href="/docs/apis/query">Query API</NavItem>
         <NavItem href="/docs/apis/db-items">DB API</NavItem>
 
         <SubHeading>GraphQL</SubHeading>
--- a/docs/components/docs/Navigation.tsx
+++ b/docs/components/docs/Navigation.tsx
@@ -192,8 +192,8 @@ export function DocsNavigation() {
 
         <SubHeading>Context</SubHeading>
         <NavItem href="/docs/apis/context">Context API</NavItem>
-        <NavItem href="/docs/apis/list-items">List Item API</NavItem>
-        <NavItem href="/docs/apis/db-items">DB Item API</NavItem>
+        <NavItem href="/docs/apis/queries">Query API</NavItem>
+        <NavItem href="/docs/apis/db-items">DB API</NavItem>
 
         <SubHeading>GraphQL</SubHeading>
         <NavItem href="/docs/apis/graphql">
--- a/docs/pages/releases/2021-04-20.mdx
+++ b/docs/pages/releases/2021-04-20.mdx
@@ -47,12 +47,12 @@ const [post] = await context.lists.Post.findMany({
 
 ### 2. Return the unresolved item data with read hooks
 
-This **replaces the `resolveFields: boolean` use case**. We now have a new set of APIs on `context.db.lists.{List}` which return unresolved item data from your database (but with read hooks applied). They can be referenced directly or returned from a custom mutation or query in the GraphQL API to be handled by the Field resolvers.
+This **replaces the `resolveFields: boolean` use case**. We now have a new set of APIs on `context.db.{List}` which return unresolved item data from your database (but with read hooks applied). They can be referenced directly or returned from a custom mutation or query in the GraphQL API to be handled by the Field resolvers.
 
 For example, to query for the raw data stored in the database:
 
 ```js
-const [post] = await context.db.lists.Post.findMany({
+const [post] = await context.db.Post.findMany({
   where: { slug },
 });
 ```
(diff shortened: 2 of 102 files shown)


JedWatson:
@timleslie we could alias backwards compatibility for this pretty easily, might ease the upgrade process

(deprecate the TS types, add a single (per-process) warning log when first accessed using a getter)

then remove the aliases when we actually ship to `@keystone-6`

otoh we haven't done a lot of this for other breaking api changes recently so and I'm not sure it's been a problem

emmatown (changes requested):
Just want to explicitly block this for now, gonna raise some questions in slack

docs/components/docs/Navigation.tsx
```diff
@@ -192,8 +192,8 @@ export function DocsNavigation() {
 
         <SubHeading>Context</SubHeading>
         <NavItem href="/docs/apis/context">Context API</NavItem>
-        <NavItem href="/docs/apis/list-items">List Item API</NavItem>
-        <NavItem href="/docs/apis/db-items">DB Item API</NavItem>
+        <NavItem href="/docs/apis/queries">Query API</NavItem>
```
mitchellhamilton, line 195:
```suggestion
        <NavItem href="/docs/apis/query">Query API</NavItem>
```

gautamsi:
Is this slated for release this week?

docs/pages/releases/2021-04-20.mdx
```diff
@@ -47,12 +47,12 @@ const [post] = await context.lists.Post.findMany({
...
+This **replaces the `resolveFields: boolean` use case**. We now have a new set of APIs on `context.db.{List}` which return unresolved item data from your database (but with read hooks applied). They can be referenced directly or returned from a custom mutation or query in the GraphQL API to be handled by the Field resolvers.
 
 For example, to query for the raw data stored in the database:
 
 ```js
-const [post] = await context.db.lists.Post.findMany({
+const [post] = await context.db.Post.findMany({
   where: { slug },
 });
 ```
```
mitchellhamilton, line 58: Can you revert these changes? Having a page on a release be inaccurate about what it is true in that release is not good imo

timleslie: Oh, good catch, they weren't meant to go in 👍

timleslie:
> Is this slated for release this week?

It's in the final phase of review and will be out in the next release.

packages/keystone/src/___internal-do-not-use-will-break-in-patch/node-api.ts
```diff
@@ -6,5 +6,5 @@ export function createListsAPI(config: KeystoneConfig, prismaClient: any) {
   const { getKeystone } = createSystem(initConfig(config));
```
mitchellhamilton, line 6: `createListsAPI` -> `createQueryAPI` or something + the thing we export from the `.keystone/api` thing should be renamed too, no?

packages/keystone/src/artifacts.ts
```diff
@@ -123,10 +123,10 @@ const nodeAPIJS = (
   config: KeystoneConfig
 ) => `import keystoneConfig from '../../keystone';
 import { PrismaClient } from '.prisma/client';
-import { createListsAPI } from '@keystone-next/keystone/___internal-do-not-use-will-break-in-patch/node-api';
+import { createQueryAPI } from '@keystone-next/keystone/___internal-do-not-use-will-break-in-patch/node-api';
 ${makeVercelIncludeTheSQLiteDB(cwd, path.join(cwd, 'node_modules/.keystone/next'), config)}
 
-export const lists = createListsAPI(keystoneConfig, PrismaClient);
+export const lists = createQueryAPI(keystoneConfig, PrismaClient);
```
mitchellhamilton, line 129: And should mention this in a changeset
```suggestion
export const query = createQueryAPI(keystoneConfig, PrismaClient);
```

gautamsi:
I thought this is release week, but seems like it is next week :)

emmatown (approved):
Looks like some snapshots need updating but other than that, LGTM

PR #6542, opened 2021-09-13
Remove `createSchema`

emmatown:
Unlike `config` and `list` which actually aid in auto complete and type checking, `createSchema` doesn't help since you're gonna call `list` anyway.


timleslie (approved):
👍

PR #6565, opened 2021-09-15
Add userInputError error function

timleslie:


emmatown (changes requested):
Can you make it a class and export it publicly somewhere?

timleslie:
> Can you make it a class and export it publicly somewhere?

The plan is to do that consistently for all of the error functions once they're all in place, but there are still a few to go before we get to that.

emmatown (approved):

PR #6571, opened 2021-09-16
6306/fix deletion pagination bug

### LINKED ISSUE #6306: Deleting all items on the last page, breaks pagination
# Bug report

## Describe the bug

Deleting all items on  the last page will not redirect to the previous page, but instead display an empty table with no pagination (despite their being additional pages) 

## To Reproduce

1. Create a 100 items in any list 
2. Navigate to page 2 
3. Delete all items on this page using the batch delete functionality 
4. Observe the bug. 

## Expected behaviour

User should be redirected to the previous page after successful deletion. 

## Screenshots

https://user-images.githubusercontent.com/12119389/129151646-5c802469-4917-41c5-8bd3-1c4941df79c1.mov

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]

gwyneplaine:
resolves #6306 where deletions on the last page would break the ListPage pagination, by adding an effect that monitors for page anomalies when the `total` changes. 

Within the `Pagination` component seems to be a rational place to locate the effect, but not a strongly held preference. 

before: 
https://user-images.githubusercontent.com/12119389/133530376-a14fa68d-a6c7-47f4-8adc-ac2d0f5bc31d.mov

after: 
https://user-images.githubusercontent.com/12119389/133530390-d4eba6d7-b570-4acd-a8af-5631f6cc21bb.mov


packages/keystone/src/admin-ui/components/Pagination.tsx
```diff
@@ -43,6 +44,21 @@ export function Pagination({ currentPage, total, pageSize, list }: PaginationPro
...
+  useEffect(() => {
+    // Check if the current page is larger than
+    // the maximal page given the total and associated page size value.
+    // (This could happen due to a deletion event, in which case we want to reroute the user to a previous page).
+    if (currentPage > Math.ceil(total / pageSize)) {
+      push({
+        pathname,
+        query: {
+          ...query,
+          page: currentPage - 1,
```
mitchellhamilton, line 56: Without the below suggestion, if I add a filter that lowers the number of pages, this will decrease the page number one by one instead of jumping to the last available one.

```suggestion
          page: Math.ceil(total / pageSize),
```

emmatown (changes requested):
Needs a changeset

packages/keystone/src/admin-ui/components/Pagination.tsx
gwyneplaine: ah nice, good catch

emmatown (approved):

PR #6608, opened 2021-09-20
Feature/relationship filters

gwyneplaine:
Add Admin UI filter support for relationships. 
Refactor `RelationshipSelect` such that `portalMenu` has to be explicitly defined. 
* This is mostly so that we can add the `RelationshipSelect` to the AddFilter component and *not* have the focus trap block click events on the menu. 
* But also, not having the select magically portal the menu (which isn't documented) is probably a good idea.


packages/keystone/src/fields/types/relationship/views/index.tsx
```diff
@@ -554,3 +662,41 @@ export const controller = (
...
+  });
+
+  return {
+    filterValues:
+      data?.items?.map((item: any) => {
+        return {
+          id: item.id,
+          label: item[list.labelField],
+        };
+      }) || foreignIds,
```
mitchellhamilton, line 692: `foreignIds` is an array of strings when it's expected that the returned thing is `{ id: string; label: string; }[]`

mitchellhamilton: Also, can you `item[list.labelField] || item.id` since the user might not have access to the label or something

.changeset/lemon-suns-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/fields-document': patch
+---
+
+Updated package to accomodate changes to the RelationshipSelect.
```
mitchellhamilton, line 5:
```suggestion
Updated package to accommodate changes to the RelationshipSelect.
```

packages/keystone/src/fields/types/relationship/views/index.tsx
```diff
@@ -478,10 +504,87 @@ export const controller = (
...
+        } = {
+          kind: 'many',
+          value: filterValues,
+          onChange(newItems) {
+            onChange(newItems.map(item => item.id).join(','));
+          },
+        };
+
+        console.log(state);
+
```
mitchellhamilton, line 532:
```suggestion
```

emmatown (approved):

PR #6616, opened 2021-09-20
Add createContext argument to config.server.extendExpressApp

timleslie:
See discussions here: https://github.com/keystonejs/keystone/discussions/6577  and https://keystonejs.slack.com/archives/C01STDMEW3S/p1632158879313800

### DIFF
--- a/.changeset/thirty-poets-itch.md
+++ b/.changeset/thirty-poets-itch.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': minor
+---
+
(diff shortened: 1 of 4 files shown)


gautamsi:
So, the context returned by createcontext will have sudo and user level access, right?

timleslie:
> So, the context returned by createcontext will have sudo and user level access, right?

It will obey the boolean `sudo` argument passed to `createContext`.

emmatown (approved):

PR #6617, opened 2021-09-21
improved messaging on startup

Noviny:


packages/keystone/src/scripts/run/dev.ts
```diff
@@ -110,7 +110,8 @@ export const dev = async (cwd: string, shouldDropDatabase: boolean) => {
   });
   const server = app.listen(port, (err?: any) => {
     if (err) throw err;
-    console.log(`⭐️ Dev Server Ready on http://localhost:${port}`);
+    console.log(`⭐️ Dev Server Starting on http://localhost:${port}`);
+    console.log(`⭐️ GraphQL API Starting on http://localhost:${port}/api/graphql`);
```
timleslie, line 114: The path here should be `(config.graphql?.path || '/api/graphql')`

timleslie (changes requested):

timleslie (approved):
👍

PR #6618, opened 2021-09-21
Consolidate non-major changes in auth

timleslie:
Following on from the removal of the hard coded `protectIdentities = true`, this PR removes code paths which are no longer accessible. This significantly simplifies the code without introducing any breaking changes. This sets the ground work for the next steps here, which include:
* Adding comprehensive tests
* Adding support for hooks
* Adding support for logging
* Simplifying the GraphQL API itself by removing enum values which are never returned.


molomby (approved):
Yep

PR #6631, opened 2021-09-22
Add Splitbee Analytics to Website

JedWatson:
This sets up [splitbee](http://splitbee.io) analytics on the website so we can try it out (looks like a more privacy-friendly alternative to Google Analytics)


bladey:
@JedWatson needs to be merged into `website_live` instead of `master`.

bladey (changes requested):
@JedWatson needs to be merged into website_live instead of master.

bladey:
Re-created - https://github.com/keystonejs/keystone/pull/6636

PR #6632, opened 2021-09-22
Improve the createContext API in extendExpressApp

JedWatson:
This makes the `createContext` API we added in #6616 much easier to use, in particular avoiding the challenge involved in correctly binding the new context to the current request (which is almost always what you want)

It also brings the default usage inline with the behaviour in graphql extensions, making transitioning between REST APIs (and/or other custom route handlers in express) and GraphQL resolvers really simple.

I'm also working on an example of how to use the Keystone Schema to create a custom REST API using these features, which I'll open a separate PR for after this lands.

**Note** I haven't added a changeset, because this only modifies an API that hasn't been released yet... if we'd released, this would actually be a major, but it's not, and I think including another changeset would make the resulting changelog when we do release make less sense 🤷

### DIFF
--- a/docs/pages/docs/apis/config.mdx
+++ b/docs/pages/docs/apis/config.mdx
@@ -256,6 +256,11 @@ config({
 
 This lets you modify the express app that Keystone creates _before_ the Apollo Server and Admin UI Middleware are added to it (but after the `cors` and `healthcheck` options are applied).
 
+The function is passed two arguments:
+
+- `app`: The express app keystone has created
+- `async createContext(req, res)`: A function you can call to create a Keystone Context for the request
+
 For example, you could add your own request logging middleware:
 
 ```ts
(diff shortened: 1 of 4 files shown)


gautamsi:
Another pattern would be to add this piece of code as middleware and access context on req object in subsequent usage like api or something

This also brings interesting use case as automatic rest api based on context.db.lists

timleslie (approved):
👍

sidharthachatterjee:
This is just what I'm looking for! Are we releasing this anytime soon? I'm currently on `@keystone-next/keystone": "^25.0.3`

timleslie:
> This is just what I'm looking for! Are we releasing this anytime soon? I'm currently on `@keystone-next/keystone": "^25.0.3`

The next release will be early next week.

Issue #6650, opened 2021-09-24
Unhandled Runtime Error on fresh install


sagerio:
# Bug report

Hi, 
I wanted to give KeystoneJS a try and installed v6 as in the docs described.
On opening http://localhost:3000 with either browser, it throws an unhandled exception.
I removed everything and tried it a second time with different names etc, but same result.

## To Reproduce

```yarn create keystone-app```
success Installed "create-keystone-app@5.0.0" with binaries:
    - create-keystone-app
...
⠧ Installing dependencies with yarn. This may take a few minutes
...
Keystone created a starter project in: keystone-test
...
    ```cd keystone-test```
    ```yarn dev```
    ```(calling http://localhost:3000/ in browser)```

## Error message

```
Unhandled Runtime Error
TypeError: Cannot read properties of undefined (reading 'document')

Call Stack
Module.createFocusTrap
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/pages/_app.js (61799:25)
<unknown>
/_next/static/chunks/pages/_app.js (29159:67)
invokePassiveEffectCreate
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (41577:20)
HTMLUnknownElement.callCallback
/_next/static/chunks/main.js (22040:14)
Object.invokeGuardedCallbackDev
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (22089:16)
invokeGuardedCallback
/_next/static/chunks/main.js (22151:31)
flushPassiveEffectsImpl
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (41664:9)
unstable_runWithPriority
/_next/static/chunks/main.js (48357:12)
runWithPriority$1
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (29371:10)
flushPassiveEffects
/_next/static/chunks/main.js (41537:14)
performSyncWorkOnRoot
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (40359:3)
<unknown>
/_next/static/chunks/main.js (29422:26)
unstable_runWithPriority
file:///C:/Repository/keystone-test/.keystone/admin/.next/static/chunks/main.js (48357:12)
runWithPriority$1
/_next/static/chunks/main.js (29371:10)
flushSyncCallbackQueueImpl
[...]

PR #6654, opened 2021-09-27
Upgrade focus-trap

timleslie:
This is the better solution for #6650.


emmatown (approved):

sidharthachatterjee:
Just noticed this and came to create an issue and found that you already patched it. You’re a gentleman and a scholar. Cheers!

PR #6656, opened 2021-09-27
Float field updates

Noviny:
I think this covers the things.

Questions for me looking at this are:
- what are the base min and max values? Do floats have these?
- do we still need the 'legacy' field option?
- I stole a lot of code from integer - is this too much duplication?


packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -43,17 +161,40 @@ export const CardValue: CardValueComponent = ({ item, field }) => {
...
+      value: data[config.path],
+    }),
+    serialize: value => {
+      let vale = value.value;
+      let val;
+
+      if (vale === '' || vale === null) {
+        val = null;
+      } else if (typeof vale === 'string') {
+        val = parseFloat(vale);
```
gu-stav, line 190: I apologize in advance for the comment from the sidelines: this has an issue in some locales. In Germany e.g. `,` is used as a decimal seperator, which doesn't work well with `parseFloat`.

I've recently written a small function which basically deals with this case: https://gist.github.com/gu-stav/c8ab96802d2826939e3996ba351e1f65

After sharing that within my company, someone mentioned that the Adobe Spectrum system contains already a package for that: https://www.npmjs.com/package/@internationalized/number

I'd be really happy, if this logic could become part of the field, since it's often a source of confusion. Especially if you are moving into the direction of a translated admin ui, translation and localization should be thought about together :)

Noviny: Thanks for this! This is something we've been thinking about but haven't had a good solution for. I'll look into your snippet/the library tomorrow morning (sydney time) and see if we can just plug it in to help us out.

packages/keystone/src/fields/types/float/views/utils.tsx
```diff
@@ -0,0 +1,81 @@
+import { ChangeEvent, FocusEvent, useState } from 'react';
```
mitchellhamilton, line 1: Can you import this from integer rather than duplicating it?

packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -11,22 +12,139 @@ import {
...
+    return `${label} is required`;
+  }
+
+  if (typeof val === 'number') {
+    if (validation?.min && val < validation.min) {
+      return `${label} must be greater than or equal to ${validation.min}`;
+    }
+    if (validation?.max && val > validation?.max) {
+      return `${label} must be less than or equal to ${validation.max}`;
+    }
```
mitchellhamilton, line 52: 0 is a valid min or max so the `validation?.min` check should be `validation?.min !== undefined` or something

packages/keystone/src/fields/types/float/index.ts
```diff
@@ -1,47 +1,162 @@
...
+              ? graphql.nonNull(graphql.Float)
+              : graphql.Float,
+            defaultValue:
+              config.graphql?.create?.isNonNull && typeof defaultValue === 'number'
+                ? defaultValue
+                : undefined,
+          }),
+          resolve(value) {
+            if (value === undefined && typeof defaultValue === 'number') {
+              return defaultValue;
```
mitchellhamilton, line 138:
```suggestion
            if (value === undefined) {
              return defaultValue ?? null;
```

packages/keystone/src/fields/types/float/index.ts
```diff
@@ -1,47 +1,162 @@
...
       ...config,
+      hooks: {
+        ...config.hooks,
+        async validateInput(args) {
+          const value = args.resolvedData[meta.fieldKey];
+
+          if (
+            validation?.isRequired &&
+            (value === null || (args.operation === 'create' && value === undefined))
+          ) {
```
mitchellhamilton, line 100:
```suggestion
          if (validation?.isRequired && value === null) {
```

packages/keystone/src/fields/types/float/index.ts
```diff
@@ -1,47 +1,162 @@
...
+    validation?: {
+      min?: number;
+      max?: number;
+      isRequired?: boolean;
+    };
+    graphql?: {
+      create?: {
+        isNonNull?: boolean;
+      };
+      read?: {
```
mitchellhamilton, line 29: This(specifically the read one) should only be allowed when `isNullable: false`, have a look at the integer field

packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -11,22 +12,139 @@ import {
...
+
+  if (value.kind === 'create' && value.value === null) {
+    return undefined;
+  }
+
+  if (validation.isRequired && val === null) {
+    return `${label} is required`;
+  }
+
+  if (typeof val === 'number') {
```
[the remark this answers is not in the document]
Noviny: Do you mean as a new check, or adding to this check that does min and max checking?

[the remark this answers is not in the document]
mitchellhamilton: Not sure I understand, I mean like this:
```suggestion
  if (typeof val === 'number') {
    if (!val.isFinite()) {
      return `${label} must be finite`;
    }
```

packages/keystone/src/fields/types/float/index.ts
Noviny: Can we talk on this? Using the patterns for this in integer led to failing tests, or failing typescript. I did this to make it a resolution

packages/keystone/src/fields/types/float/tests/non-null/test-fixtures.ts
```diff
@@ -0,0 +1,35 @@
...
+});
+
+export const initItems = () => {
+  return [
+    { name: 'post1', testField: -21.5 },
+    { name: 'post2', testField: 0 },
+    { name: 'post3', testField: 1.2 },
+    { name: 'post4', testField: 2.3 },
+    { name: 'post5', testField: 3 },
+    { name: 'post7' },
```
mitchellhamilton, line 23:
```suggestion
```

mitchellhamilton (changes requested):
Needs a changeset + updates to the docs, it should largely be a copy of the integer ones but with a couple changes

packages/keystone/src/fields/types/float/tests/non-null/test-fixtures.ts
```diff
@@ -0,0 +1,34 @@
...
+  testField: float({ isFilterable: true, isNullable: false }),
+});
+
+export const initItems = () => {
+  return [
+    { name: 'post1', testField: -21.5 },
+    { name: 'post2', testField: 0 },
+    { name: 'post3', testField: 1.2 },
+    { name: 'post4', testField: 2.3 },
+    { name: 'post5', testField: 3 },
```
mitchellhamilton, line 22:
```suggestion
    { name: 'post5', testField: 3 },
    { name: 'post6', testField: 5 },
    { name: 'post7', testField: 20.8 },
```

packages/keystone/src/fields/types/float/tests/non-null/test-fixtures.ts
```diff
@@ -0,0 +1,34 @@
...
+  ];
+};
+
+export const storedValues = () => [
+  { name: 'post1', testField: -21.5 },
+  { name: 'post2', testField: 0 },
+  { name: 'post3', testField: 1.2 },
+  { name: 'post4', testField: 2.3 },
+  { name: 'post5', testField: 3 },
```
mitchellhamilton, line 31:
```suggestion
  { name: 'post5', testField: 3 },
  { name: 'post6', testField: 5 },
  { name: 'post7', testField: 20.8 },
```

packages/keystone/src/fields/types/float/index.ts
```diff
@@ -1,47 +1,166 @@
...
+      throw new Error(
+        `The float field at ${meta.listKey}.${meta.fieldKey} specifies validation.min: ${validation.min} but it must be a valid number`
+      );
+    }
+
+    if (validation?.max !== undefined && typeof validation.min !== 'number') {
+      throw new Error(
+        `The float field at ${meta.listKey}.${meta.fieldKey} specifies validation.max: ${validation.max} but it must be a valid number`
+      );
+    }
```
mitchellhamilton, line 67: Can you check that all these are finite?

packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -11,22 +12,142 @@ import {
...
+}: {
+  id: string;
+  autoFocus?: boolean;
+  value: number | string | null;
+  onChange: (value: number | string | null) => void;
+  forceValidation?: boolean;
+  validationMessage?: string;
+  placeholder?: string;
+}) {
+  console.log({ value, type: typeof value });
```
mitchellhamilton, line 78:
```suggestion
```

packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -11,22 +12,141 @@ import {
...
+  const [hasBlurred, setHasBlurred] = useState(false);
+  const props = useFormattedInput<number | null>(
+    {
+      format: value => (value === null ? '' : value.toString()),
+      parse: raw => {
+        raw = raw.trim();
+        if (raw === '') {
+          return null;
+        }
+        if (/^[+-]?\d+\.?\d?$/.test(raw)) {
```
mitchellhamilton, line 87: Can you remove regex check, there's other syntax that should work here like `5e200`?

packages/keystone/src/fields/types/float/views/index.tsx
```diff
@@ -43,17 +163,40 @@ export const CardValue: CardValueComponent = ({ item, field }) => {
...
+
+      if (vale === '' || vale === null) {
+        val = null;
+      } else if (typeof vale === 'string') {
+        val = parseFloat(vale);
+      } else {
+        val = vale;
+      }
+
+      return { [config.path]: val };
```
mitchellhamilton, line 197: It should already be parsed here
```suggestion
    serialize: value => {
      return { [config.path]: value.value };
```

docs/pages/docs/apis/fields.mdx
```diff
@@ -264,12 +266,24 @@ Options:
 
 - `defaultValue` (default: `undefined`): Can be either a float value or an async function which takes an argument `({ context, originalInput })` and returns a float value.
   This value will be used for the field when creating items if no explicit value is set.
```
mitchellhamilton, line 268:
```suggestion
- `defaultValue` (default: `undefined`): Can be a finite float value.
  This value will be used for the field when creating items if no explicit value is set.
```

mitchellhamilton (approved):

PR #6674, opened 2021-09-29
Timestamp field updates

emmatown:


tests/api-tests/fields/required.test.skip.ts
```diff
@@ -4,6 +4,8 @@ import { text } from '@keystone-next/keystone/fields';
 import { setupTestRunner } from '@keystone-next/keystone/testing';
 import { apiTestConfig, expectValidationError } from '../utils';
 
+// TODO: test the new validation.isRequired
```
mitchellhamilton, line 7: Had to rename this because otherwise Jest would complain about this file having no tests, I'll make this file test the new `validation.isRequired` on the fields that support it in a future PR

timleslie (approved):
👍

PR #6675, opened 2021-09-29
Change newsletter opt-in to unchecked on create first user page

bladey:
Newsletter checkbox is now unchecked by default (previously checked), we have also changed email subscriptions from double opt-in (requires clicking confirmation in a follow-up email for confirmation) to single opt-in (subscribed when submitted, no further action required).

image


timleslie (approved):
👍

PR #6676, opened 2021-09-29
Re-enable all field-type CRUD tests

timleslie:
Not sure when this filter slipped in, but it's definitely not what we want!

I believe all these changes are covered under existing changesets.

### DIFF
--- a/packages/fields-document/src/index.ts
+++ b/packages/fields-document/src/index.ts
@@ -7,6 +7,7 @@ import {
   graphql,
   JSONValue,
 } from '@keystone-next/keystone/types';
+import { userInputError } from '@keystone-next/keystone/src/lib/core/graphql-errors';
 import { Relationships } from './DocumentEditor/relationship';
 import { ComponentBlock } from './component-blocks';
 import { DocumentFeatures } from './views';
@@ -99,7 +100,7 @@ export const document =
 
     const inputResolver = (data: JSONValue | null | undefined): any => {
       if (data === null) {
-        throw new Error(`Document fields cannot be set to null`);
+        throw userInputError(`Document fields cannot be set to null`);
       }
       if (data === undefined) {
         return data;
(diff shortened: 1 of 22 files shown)


packages/fields-document/src/index.ts
```diff
@@ -7,6 +7,7 @@ import {
   graphql,
   JSONValue,
 } from '@keystone-next/keystone/types';
+import { userInputError } from '@keystone-next/keystone/src/lib/core/graphql-errors';
```
mitchellhamilton, line 10: This will break when shipped to npm so it needs to be exported publically and when you do so, please make it a class

packages/keystone/src/fields/types/decimal/index.ts
```diff
@@ -132,6 +132,9 @@ export const decimal =
           if (val === null && validation?.isRequired) {
             args.addValidationError(`${fieldLabel} is required`);
           }
+          if (val === null && config.isNullable === false) {
+            args.addValidationError(`${fieldLabel} cannot be set to 'null'`);
```
mitchellhamilton, line 136: Can you make these messages do `${fieldLabel} is required` (as in part of the same isRequired check as above and similarly in the other fields with isNullable)

packages/keystone/src/fields/types/integer/tests/test-fixtures.ts
```diff
@@ -4,6 +4,7 @@ export const name = 'Integer';
 export const typeFunction = integer;
```
mitchellhamilton, line 4: Integer should need updating so that should happen + given the tests aren't failing because of it, I'm guessing something is wrong with the tests

timleslie: Looks like it's missing the `non-null` test fixtures 👍

emmatown (approved):

PR #6682, opened 2021-09-30
Auth GraphQL API changes

emmatown:
This is part of (and likely the only thing) in a review of the GraphQL API


timleslie (approved):
SImple!

PR #6684, opened 2021-09-30
Consolidate before/after hooks

timleslie:
Consolidate the beforeChange/beforeDelete hooks into beforeOperation, and afterChange/afterDelete hooks into afterOperation.

### DIFF
--- a/packages/keystone/src/lib/core/mutations/create-update.ts
+++ b/packages/keystone/src/lib/core/mutations/create-update.ts
@@ -41,7 +41,7 @@ async function createSingle(
   //  Item access control. Will throw an accessDeniedError if not allowed.
   await applyAccessControlForCreate(list, context, rawData);
 
-  const { afterChange, data } = await resolveInputForCreateOrUpdate(
+  const { afterOperation, data } = await resolveInputForCreateOrUpdate(
     list,
     context,
     rawData,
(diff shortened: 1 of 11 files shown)


packages/keystone/src/lib/core/mutations/create-update.ts
```diff
@@ -52,11 +52,11 @@ async function createSingle(
     runWithPrisma(context, list, model => model.create({ data }))
   );
 
-  return { item, afterChange };
+  return { item, afterOperation };
 }
 
 export class NestedMutationState {
-  #afterChanges: (() => void | Promise<void>)[] = [];
+  #afterOpertions: (() => void | Promise<void>)[] = [];
```
mitchellhamilton, line 59:
```suggestion
  #afterOperations: (() => void | Promise<void>)[] = [];
```

packages/keystone/src/lib/core/mutations/create-update.ts
```diff
@@ -68,20 +68,20 @@ export class NestedMutationState {
...
+    const { item, afterOperation } = await createSingle(
       { data },
       list,
       context,
       operationAccess,
       writeLimit
     );
 
-    this.#afterChanges.push(() => afterChange(item));
+    this.#afterOpertions.push(() => afterOperation(item));
```
mitchellhamilton, line 79:
```suggestion
    this.#afterOperations.push(() => afterOperation(item));
```

packages/keystone/src/lib/core/mutations/create-update.ts
```diff
@@ -68,20 +68,20 @@ export class NestedMutationState {
...
     return { id: item.id as IdType };
   }
 
-  async afterChange() {
-    await promiseAllRejectWithAllErrors(this.#afterChanges.map(async x => x()));
+  async afterOperation() {
+    await promiseAllRejectWithAllErrors(this.#afterOpertions.map(async x => x()));
```
mitchellhamilton, line 84:
```suggestion
    await promiseAllRejectWithAllErrors(this.#afterOperations.map(async x => x()));
```

emmatown (changes requested):
Can you make the inputData and resolvedData `undefined` when it doesn't exist rather than `null`?

emmatown (approved):

PR #6695, opened 2021-10-01
Update virtual-fields.mdx

jakegiri:
There is a typo in docs of virtual-field page's Resolver-arguments section example.
type is suppose to be a property of graphql.field config object but it shown as a property of virtual config object.

image


timleslie (approved):

PR #6696, opened 2021-10-03
Fix typo in with-auth example readme

bjoluc:


timleslie (approved):
Nice catch, thanks for the PR! 🙏

PR #6688, opened 2021-09-30
Update `readme-header.png`

bladey:
This PR replaces the image at the top of README.md with a transparent PNG instead of an opaque PNG.

This is important for people who use dark mode on the GitHub website.

Reported by @dcousens who provided this reference image:

whatigot

PR #6698, opened 2021-10-03
Replaced README header image

raveling:
Also works in Github dark theme now.


raveling:
Already solved by @bladey in #6688 😌

PR #6707, opened 2021-10-05
Change default for isFilterable/isOrderable to true

timleslie:


packages/auth/src/schema.ts
```diff
@@ -68,7 +68,7 @@ export const getSchemaExtension =
       throw new Error(
         `createAuth was called with an identityField of ${identityField} on the list ${listKey} ` +
           `but that field doesn't allow being searched uniquely with a String or ID. ` +
-          `You should likely add \`isIndexed: 'unique'\, isFilterable: true\` ` +
```
JedWatson, line 71: This was a bad comment anyway tbh, you shouldn't set `isFilterable: true` on the identity field, it opens up the specific security issue this feature was designed to fix.

The actual implementation uses the `sudo` schema so `isFilterable` never made a difference here to the auth package

JedWatson (changes requested):

emmatown (approved):

JedWatson (approved):

PR #6714, opened 2021-10-05
Add support for extensions with numbers (eg mp3, woff2)

Zlitus:


gautamsi:
you want to add a changeset to this

packages/keystone/src/lib/context/createFilesContext.ts
```diff
@@ -27,7 +27,7 @@ const generateSafeFilename = (
   // This regex lazily matches for any characters that aren't a new line
   // it then optionally matches the last instance of a "." symbol
   // followed by any alphabetical character before the end of the string
```
dcousens, line 29:
```suggestion
  // followed by any alphanumerical character before the end of the string
```

dcousens (approved):
LGTM,  but I don't see why we splitting these strings at all,  I'll want to look into the fundamental output requirements for this function soon

PR #6715, opened 2021-10-05
Adding SVG in image types available for upload

Zlitus:


dcousens:
A few concerns in respect to this and security, that we need to reach consensus on what are safe defaults and reasonable escape hatches.

For example, we might want to explore using https://github.com/cloudflare/svg-hush in this particular scenario.

IdkMan2Usertive:
I have created a repository with some custom fields for Keystone.js, including `vectorImage` field which allows us to upload .svg files.

Have a look:
https://github.com/usertive/keystone-custom-fields

Zlitus:
That's definitely very good and useful to treat SVG files and make sure they are safe, but in my case, KeystoneJS is the admin interface for my website and only me and my team which I trust do have access to it and can upload SVG files. I'm a little scared than the protections you will put to treat SVG could have some side-effect and break some SVG features I might want to use.

Maybe this should be an option than we can enable/disable by config?

dcousens:
That's an interesting consideration @Zlitus, and honestly that might mean we up with this configuration being explicitly configurable by users with documentation warnings, instead of being specifically prescripted by us.

yassinrais:
📌 For me, I wish there was a variable in the Keystone configuration that gave us the ability to enable/disable svg, that might be the best solution for now.

MartinDawson:
Is this ever going to be merged? 3 years in and still can't upload a basic thing like SVG's.

I too do not care about security as we only have authorized users...

dcousens:
@MartinDawson I think the conclusion is that we want to change this to be configurable by users, instead of prescribed by us.
We are accepting of contributor pull requests :yellow_heart:

luwangshell:
@dcousens, could you please share the issue/PR for "we want to change this pull request to support the image types to be configurable by users, instead of prescribed by @keystone-6/*"? Thank you.

dcousens:
See feature request https://github.com/keystonejs/keystone/discussions/8913

PR #6722, opened 2021-10-05
Fix references to `graphql` export

mitchellhamilton:
Reverts keystonejs/keystone#6561

Now that this change has actually landed properly, this can be reverted


timleslie (approved):
👍

PR #6726, opened 2021-10-06
fix bad async flow in filters test

gwyneplaine:
we're not awaiting on certain playwright async methods,, this is causing more than the usual intermittent failure.


timleslie (approved):
👍

emmatown (approved):

PR #6731, opened 2021-10-06
[POC] Simplify relationship resolver code

timleslie:
This is an attempt to remove what appear to be unnecessary higher order functions which we go through in the process of resolving relationship fields. It seems to me that this simplifies things considerably (we'll need to clean up some types, but that's ok). I have a suspicion that I'm missing something either very obvious or very subtle, so throwing this up as a POC for now to see how it tracks on CI.


timleslie:
Closing this as it implies removing the ability to build custom relationship fields.

PR #6732, opened 2021-10-06
Improve error messages for user input errors

timleslie:


emmatown (changes requested):
Could you use either single or double quotes rather than both in the messages? (I would personally say double to be consistent with when we use JSON.stringify in messages but don't mind strongly)

emmatown (approved):

PR #6737, opened 2021-10-07
Performance improvements

emmatown:


timleslie (approved):
Running local tests on the `blog` project brings startup time for the Admin UI down from 6.4s to 2.9s. This is a great win 🚀

PR #6738, opened 2021-10-07
Improve error messages from to-many relationships

timleslie:


emmatown (changes requested):
Could you do something like this instead of passing relationshipErrors down?

```ts
class RelationshipErrors extends Error {
  errors: { error: Error; tag: string }[];
  constructor(errors: { error: Error; tag: string }[]) {
    super(asString(errors));
    this.errors = errors;
  }
}

throw new RelationshipErrors(errors.map((error: Error) => ({ error, tag })));

} catch (error: any) {
  if (error instanceof RelationshipErrors) {
    relationshipErrors.push(...error.errors)
  } else {
    relationshipErrors.push({ error, tag });
  }
}
```

emmatown (approved):

PR #6750, opened 2021-10-10
add code highlighting and collapsing

Noviny:


dominikwilkowski (approved):
We need to stage an intervention at some point about variable names but it works :D

Noviny:
I've updated the `relationship` guide page to show off the new features

PR #6752, opened 2021-10-10
Upgrade notices

raveling:
- added notices on website surfaces to reflect recent changes to `fields` API
- added `well` for the new REST API endpoint example on /examples
- added timeline item to /updates for REST example
- improved copy in a few places


raveling:
branch snafu

PR #6775, opened 2021-10-12
Remove obsolete integration tests

timleslie:
This also let's us get rid of our cypress dependency 🚀


timleslie:
> What have these been replaced with? Or did priorities change?

These tests were deprecated when we did a major overhaul of the website. We kept them around for possible future reference, but they're no longer of any value. We've also moved away from cypress as a test framework in favour of playwright, so any new tests will need to be written in that context.

PR #6789, opened 2021-10-13
Remove `resolveFields`

emmatown:
This was deprecated quite a while ago and we even renamed `context.lists` to `context.query` while this was deprecated so I'd be surprised if many people are still using it.


dcousens (dismissed):

timleslie (approved):
👍

PR #6792, opened 2021-10-14
Add an example using Nexus to extend the Keystone schema

JedWatson:
This came up on Slack, and I was interested in testing out how to use [Nexus](https://nexusjs.org) to extend Keystone's GraphQL API. So I wrote up this example.

Turns out it's pretty neat, although there are a couple of limitations that we could improve in the future:

* ~~Nexus **really** wants you to run a watch process while you're coding, to regenerate types as you work. Since Keystone doesn't do hot reloading at the moment, I've added a `dev-nexus` script which is completely hacky but gets the job done.~~ not a problem anymore with the live reloading in Keystone now
* You can't use Keystone's GraphQL types in Nexus, it should be totally possible but I haven't worked out how to make Nexus aware of the GraphQL schema Keystone generates yet
* Nexus also has a Prisma plugin, which simplifies type definition. We should be able to get Nexus in this example hooked up to the Prisma schema that Keystone generates


JedWatson:
Note on the status of the Prisma plugin:

I just read through graphql-nexus/nexus-plugin-prisma#1039 and it's being rewritten, the one you can use today only works with an older version of Prisma than the one that Keystone uses so we should definitely wait for the new one ([here](https://github.com/prisma/nexus-prisma)) is ready for use.

bladey:
Nice example.

Getting the following when running `yarn dev` and/or `yarn dev-nexus` (doesn't matter the combination), regardless it all _seems_ to work still:

Screen Shot 2021-10-20 at 12 30 43 pm

When attempting to request `nexusPosts`:

Pasted Graphic 2

Aren't dates supposed to be ISO Strings?

async resolve(root, ( authorid, days ), context)

Works if changed based on publish date being in the future:

Pasted Graphic

.changeset/three-beans-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@keystone-next/example-extend-graphql-schema-nexus": patch
+"@keystone-next/example-extend-graphql-schema": patch
```
mitchellhamilton, line 3:
```suggestion
```

mitchellhamilton (approved):

bladey (approved):
Works well.

PR #6798, opened 2021-10-20
New Auth and Access Control guide

JedWatson:
There's more to write, but I think this adds a lot of value over the current "docs coming soon" page.

Could also probably use a good edit, but (refer to previous point 😉)

Preview here: https://keystone-next-docs-git-auth-and-access-control-guide-keystonejs.vercel.app/docs/guides/auth-and-access-control


docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+```ts
+const { withAuth } = createAuth({
+  listKey: 'Person',
+  identityField: 'email',
+  secretField: 'password',
+});
+```
+
+The `createAuth` function returns another function called `withAuth` that will automatically extend Keystone's config to set up everything we need. Behind the scenes, the auth package is just using lower-level Keystone APIs to do everything, which means if you want to do something differently, you can fork our implementation of auth and build your own (this is true for session management as well)
```
gwyneplaine, line 52: nit: We make the assumption here that the reader knows where `createAuth` comes from, and go straight into explaining what it does. We should probably at the minimum mention that its an export from `@keystone-next/auth`. It's implicitly stated in the second paragraph, and that information is available in the Auth API Docs link below, but it'd be nice to have that information explicitly declared and not hidden behind a click, at this point in the guide.

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+Now that we have information about the current user, lets put it to good use and add some security to our API with **Access Control**.
+
+The way to think about access control is –
+
+- With **Session Data**, you know who the user is, and what they should have access to (whether that's permissions granted by a role, or items related to them)
+- For each **List** you can restrict which operations they can perform, filter the items they have access to, and control which fields they can read and update.
+
+!> In other words: **Step 1** load data about the current user; **Step 2** use that data to limit what they can do.
+
+You define access control for each List (and field) individually; the rules you specify will always be run, even when the list being queried or mutated through a relationship.
```
gwyneplaine, line 162:
```suggestion
You define access control for each List (and field) individually; the rules you specify will always be run, even when the list is being queried or mutated through a relationship.
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+### Posts Example
+
+Let's say we have a list of Blog Posts, which have a published status. We want Admin users to see all blog posts and be able to crete and update them, but anonymous (public) users can only see published posts and can't write to the API. Our Post list might look like this:
```
gwyneplaine, line 166:
```suggestion
Let's say we have a list of Blog Posts, which have a published status. We want Admin users to see all blog posts and be able to create and update them, but anonymous (public) users can only see published posts and can't write to the API. Our Post list might look like this:
```

gwyneplaine (changes requested):

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+```ts
+type Session = {
+  data: {
+    id: string;
+    isAdmin: boolean;
+  }
+}
+```
+
+We can now set up operational access control to restrict the **create**, **update** and **delete** operations to authenticated users with the `isAdmin` checkbox set:
```
gwyneplaine, line 191: we embolden `filter` access control below, but not here, this seems inconsistent.

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+      query: filterPosts,
+    }
+  },
+  fields: {
+    // see above
+  },
+});
+```
+
+See that object we return when the session data doesn't say there's a user with `isAdmin`? It's the same format as `where` arguments you can provide to the `posts` query through the GraphQL API:
```
gwyneplaine, line 235: Not a change, but tonality check here. "See that object we return when the session data doesn't say there's a user with `isAdmin`?" is a tone of voice we haven't used anywhere else in the docs iirc, is consistency of tone something we care about across docs? If yes, might be worthwhile rewriting this phrase.

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+  }
+}
+mutation {
+  updateUser(where: { id: 1 }) {
+    name
+  }
+}
+```
+
+Both of these operations will "read" the user's name. Having access control named "read" on the list might be confusing as to whether it affects just the query, or also the mutations, so we called it "query" to make it clearer that it only affects queries and **not** mutations.
```
gwyneplaine, line 309: nit: Moving the "Having access control named "read" on the list might be confusing as to whether it affects just the query, or also the mutations, so we called it "query" to make it clearer that it only affects queries and **not** mutations." to beginning of this section instead of the end makes reading easier. 

The rest seems like useful, but secondary information that I can choose to ignore once i grep the main point. 

That's the point you're making, the rest is nuance that ideally

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+Query filters are also applied to count operations, so you can paginate items predictably without revealing the existence of items the user does not have access to.
+
+Although Keystone doesn't implement database-level security, the design and effect of `filter` access control is similar to [Row Level Security](https://www.postgresql.org/docs/current/ddl-rowsecurity.html) in Postgres, with the benefit of being expressed declaratively in TypeScript and using the same filter syntax as the GraphQL API. This makes it easier to reason about in the context of your schema, session data, and other logic in your application.
+
+You can't specify filters for the `create` operation because you can't query records that don't exist yet.
+
+#### How `item` access control works
+
+When mutating items, item access control is checked with the mutation input (for `create` and `update`) and the item that has been retrieved from the database (for `update` and `delete`).
```
gwyneplaine, line 366:
```suggestion
When mutating items, item access control is checked against the mutation input (for `create` and `update`) and the item that has been retrieved from the database (for `update` and `delete`).
```

gwyneplaine (changes requested):

docs/pages/docs/guides/auth-and-access-control.mdx
bladey, line 52:
```suggestion
The `createAuth` function returns another function called `withAuth` that will automatically extend Keystone's config to set up everything we need. Behind the scenes, the `auth` package is just using lower-level Keystone APIs to do everything, which means if you want to do something differently, you can fork our implementation of `auth` and build your own (this is true for session management as well)
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+The result of the query is stored in `context.session`, and available to almost every function and hook in Keystone – including access control.
+
+In this initialisation phase, you'll want to load any data you'll need to work out what the user should be able to do (and what they should not be able to do).
+
+If no session cookie is present, or no matching item can be found, `context.session` will be `undefined`.
+
+### Add a sessionData query
+
+In our example above, we probably want to know the current user's ID and the value of the `isAdmin` checkbox. The auth package makes this simple by providing a `sessionData` option, which should contain the fields to query when a session is found.
```
bladey, line 127:
```suggestion
In our example above, we probably want to know the current user's ID and the value of the `isAdmin` checkbox. The `auth` package makes this simple by providing a `sessionData` option, which should contain the fields to query when a session is found.
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+  fields: {
+    name: text(),
+    email: text({ isIndexed: 'unique' }),
+    password: password(),
+    isAdmin: checkbox(),
+  },
+});
+```
+
+!> Read more about creating lists in the [schema](../apis/schema) and [fields](../apis/schema) API Docs
```
bladey, line 38:
```suggestion
!> Read more about creating lists in the [schema](../apis/schema) and [fields](../apis/schema) API Docs.
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+const { withAuth } = createAuth({
+  listKey: 'Person',
+  identityField: 'email',
+  secretField: 'password',
+});
+```
+
+The `createAuth` function returns another function called `withAuth` that will automatically extend Keystone's config to set up everything we need. Behind the scenes, the auth package is just using lower-level Keystone APIs to do everything, which means if you want to do something differently, you can fork our implementation of auth and build your own (this is true for session management as well)
+
+!> Read more about `createAuth` in the [Auth API Docs](../apis/auth)
```
bladey, line 54:
```suggestion
!> Read more about `createAuth` in the [Auth API Docs](../apis/auth).
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+
+```ts
+const session = statelessSessions({
+  secret: '-- EXAMPLE COOKIE SECRET; CHANGE ME --',
+});
+```
+
+Keystone also comes with a Redis session adapter, which uses a cookie to store a session ID that is looked up in a Redis database; or you can use your own session adapter (for example, if you are using OAuth sessions).
+
+!> Read more about [Session Stores in the Session API Docs](../apis/session#session-stores)
```
bladey, line 68:
```suggestion
!> Read more about [Session Stores in the Session API Docs](../apis/session#session-stores).
```

docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -0,0 +1,532 @@
...
+  };
+  item: {
+    create: ({ session, context, listKey, operation, inputData }) => boolean;
+    update: ({ session, context, listKey, operation, inputData, item }) => boolean;
+    delete: ({ session, context, listKey, operation, item }) => boolean;
+  };
+};
+```
+
+!> We'll cover field access control separately below
```
bladey, line 338:
```suggestion
!> We'll cover field access control separately below.
```

bladey (changes requested):

docs/pages/docs/guides/auth-and-access-control.mdx
JedWatson: Good catch @gwyneplaine 

I've added another paragraph to the section above, and put the import in the code snippet to make the import path clear.

For the rest I think it's safe to assume the reader already knows where to import `list` and field types from 🤞

docs/pages/docs/guides/auth-and-access-control.mdx
JedWatson: Ha thanks, I changed conventions halfway through and knew I'd miss one 🙂 

Fixed

docs/pages/docs/guides/auth-and-access-control.mdx
JedWatson: Fair call. I think I'm trying to make the guide generally more conversational, but tbh it's all feeling a bit wordy at the moment. I've re-written this more consistently for now 👍

bladey (approved):

docs/pages/docs/guides/auth-and-access-control.mdx
JedWatson: 👍 rewritten

gwyneplaine (approved):

PR #6799, opened 2021-10-20
Add update many options on list

gautamsi:
I needed this and thought it would be good idea to bring this feature back.

* I would still love to see how to make this on left side to distinguish a update many and create form

adding comments where it may need improvements.

I would appreciate this in next release.

NO TESTS written, I would need help with this.

keystone-update-many

### DIFF
--- a/.changeset/shaggy-months-design.md
+++ b/.changeset/shaggy-months-design.md
@@ -1,0 +1,32 @@
+---
+'@keystone-ui/fields': minor
+'@keystone-next/example-assets-cloud': minor
+'@keystone-next/example-assets-local': minor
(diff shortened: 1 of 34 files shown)


gautamsi:
I use my own fork https://github.com/k6js/k6js for my projects which includes this feature. That fork is updated whenever I want to update to latest keystone for feature or fixes. 

lot of conflicts now, will close this and let core team decide the course of action.

dcousens:
@gautamsi yes, I think we can close this - but please know that the feature is not being abandoned out of the gate.

We are aware that bulk management is a strong use case for Keystone, and do we want to support that use case.

PR #6836, opened 2021-10-27
fix documentation for itemView/listView

dcousens:
I think this was copy pasted from the `createView` field


bladey:
@dcousens could you please get this merged before the release?

PR #6846, opened 2021-10-29
Fix docs code size in Safari

JedWatson:
Our code examples are rendering in tiny ant font in Safari, this fixes it (at the cost of specifying pixels instead of using our rem tokens, which is a reasonable trade since Safari's about 10% of our traffic

Before:

image

After:

image


dominikwilkowski (approved):
LGTM

bladey (approved):

JedWatson:
Thanks for the assist @dominikwilkowski!

PR #6856, opened 2021-10-31
Graphql TS schema extensions

mitchellhamilton:


packages/auth/src/gql/getMagicAuthLinkSchema.ts
```diff
@@ -1,52 +1,67 @@
...
+    fields: {
+      token: graphql.field({ type: graphql.nonNull(graphql.String) }),
+      item: graphql.field({ type: graphql.nonNull(base.object(listKey)) }),
+    },
+  });
+  const RedeemItemMagicAuthTokenResult = graphql.union({
+    name: gqlNames.RedeemItemMagicAuthTokenResult,
+    types: [RedeemItemMagicAuthTokenSuccess, RedeemItemMagicAuthTokenFailure],
+    resolveType(val) {
+      return 'token' in val
```
dcousens, line 52: From @mitchellhamilton in a call,  this should never exist as a `null`, so this optimisation should be OK

dcousens (approved):

PR #6863, opened 2021-11-01
update context api docs to reflect present state

gwyneplaine:
This may have been missed during the context update a few releases ago.


JedWatson (approved):
Good catch @gwyneplaine

PR #6854, opened 2021-10-31
Release notes (2021-11-02)

bladey:

PR #6864, opened 2021-11-01
add null relationship filter example

dcousens:
There didn't seem to be any example of a `null` query,  as discussed in https://keystonejs.slack.com/archives/C01STDMEW3S/p1635766214367400


bladey (approved):

dcousens:
CI failed as a result of a typescript error,  that is fixed in https://github.com/keystonejs/keystone/pull/6854

PR #6866, opened 2021-11-02
Fix flakey test

emmatown:
We think that the failures happen because Playwright goes through the test really fast and a request is cancelled when trying to navigate to the custom page so it just doesn't navigate to the page? (this is very much a guess, it's quite hard to actually debug this since we haven't been able to reproduce it locally) So we're just gonna retry it. We've retried it three times and it hasn't failed yet so hopefully it's ok

### DIFF
--- a/tests/examples-smoke-tests/custom-admin-ui-pages.test.ts
+++ b/tests/examples-smoke-tests/custom-admin-ui-pages.test.ts
@@ -1,18 +1,7 @@
+import retry from 'async-retry';
 import { Browser, Page } from 'playwright';
-import { exampleProjectTests, loadIndex } from './utils';
 
-async function retry(cb: () => Promise<void>, maxTimes: number) {
-  while (true) {
-    try {
-      return await cb();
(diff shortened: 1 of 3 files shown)


gwyneplaine (approved):

JedWatson (approved):

dcousens (approved):
LGTM,  changed it to use the `async-retry` dependency we already have instead of DIY

tests/examples-smoke-tests/custom-admin-ui-pages.test.ts
```diff
@@ -26,7 +14,7 @@ exampleProjectTests('custom-admin-ui-pages', browserType => {
     retry(async () => {
       await page.goto('http://localhost:3000/custom-page');
       await page.waitForSelector('main h1:has-text("This is a custom Admin UI page")');
-    }, 5));
+    });
```
dcousens, line 17: default is 10,  this is fine

PR #6875, opened 2021-11-02
Update tutorial "Embedded Keystone in Next.js app"

j-corral:
Fix code example to avoid error "Cannot read property 'Post' of undefined"


JedWatson (approved):
🙌  thanks @j-corral!

PR #6883, opened 2021-11-03
Fix incorrect db key in timestamp documentation

davidpmccormick:
/s/createdAt/updatedAt


JedWatson (approved):
Nice catch, thanks @davidpmccormick

PR #6915, opened 2021-11-10
Support unique where filters where `isIndexed: 'unique'` is supported

### LINKED ISSUE #6809: Select field type does not define uniqueWhere for isIndexed = 'unique'
## Describe the bug

Select field can be set isIndexed= unique but the uniqueWhere is missing from graphql definition. 

## To Reproduce

1. set `isIndexed: 'unique'` for select field
2. try querying like this `query { listWithSelectField(where: {selectField: 'value'}){...fields} }`
3. See error

## Expected behaviour
it should allow to get single entry based on unique select field value.

## Additional context

This is a good use case to prevent creation of duplicate entry in list which is reserved for config/settings list. this enables me uniquely create configuration item for my app and query them without needing to get multiple item and select first.

emmatown:
Closes #6809

### DIFF
--- a/packages/keystone/src/lib/core/queries/resolvers.ts
+++ b/packages/keystone/src/lib/core/queries/resolvers.ts
@@ -14,30 +14,14 @@ import { InitialisedList } from '../types-for-lists';
 import { getDBFieldKeyForFieldOnMultiField, runWithPrisma } from '../utils';
 import { checkFilterOrderAccess } from '../filter-order-access';
 
-// doing this is a result of an optimisation to skip doing a findUnique and then a findFirst(where the second one is done with access control)
-// we want to do this explicit mapping because:
-// - we are passing the values into a normal where filter and we want to ensure that fields cannot do non-unique filters(we don't do validation on non-unique wheres because prisma will validate all that)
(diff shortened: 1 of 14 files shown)


packages/keystone/src/lib/core/queries/resolvers.ts
```diff
@@ -14,30 +14,14 @@ import { InitialisedList } from '../types-for-lists';
...
-// doing this is a result of an optimisation to skip doing a findUnique and then a findFirst(where the second one is done with access control)
-// we want to do this explicit mapping because:
-// - we are passing the values into a normal where filter and we want to ensure that fields cannot do non-unique filters(we don't do validation on non-unique wheres because prisma will validate all that)
-// - for multi-field unique indexes, we need to a mapping because iirc findFirst/findMany won't understand the syntax for filtering by multi-field unique indexes(which makes sense and is correct imo)
-export function mapUniqueWhereToWhere(
-  list: InitialisedList,
-  uniqueWhere: UniquePrismaFilter
-): PrismaFilter {
+// we want to put the value we get back from the field's unique where resolver into an equals
+// rather than directly passing the value as the filter even though Prisma supports directly passing in a value to do an equals
```
dcousens, line 18:
```suggestion
// rather than directly passing the value as the filter (even though Prisma supports that), we use equals
```

dcousens (approved):
LGTM

PR #6917, opened 2021-11-11
Shorter many to many relation names

### LINKED ISSUE #6812: Issues with long index names in migrations with Prisma and Postgres
When using migrations and complex data schemas with a Postgres database, Prisma is throwing errors as it is truncating long index names when running migrations.

```
EngineError: P3006
Migration `20211011130400_new_required_validation_defaults` failed to apply cleanly to the shadow database.
Error:
db error: ERROR: relation "_Organisation_categories_OrganisationCategory_organisations_AB_" already exists
   0: sql_migration_connector::flavour::postgres::sql_schema_from_migration_history
             at migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs:354
   1: sql_migration_connector::validate_migrations
             at migration-engine/connectors/sql-migration-connector/src/lib.rs:342
   2: migration_core::api::DevDiagnostic
             at migration-engine/core/src/api.rs:108
```

Prisma state the issue is fixed [here](https://github.com/prisma/prisma/issues/9415) but it is still an issue even when upgrading Prisma packages to 3.3.0 and above. It may be an underlying Postgres issue.

A workaround is to shorten field names so not exceeded index limits.

It is unknown when this will be fixed, needs further investigation and could be a core Postgres issue.
 
Related Prisma Issue - https://github.com/prisma/prisma/issues/9415

_Adapted and sourced from Ben Booth in the Keystone Community Slack - https://keystonejs.slack.com/archives/C01STDMEW3S/p1633988631173200_

emmatown:
Closes #6812

This makes two-sided many to many relationships only include one of the sides in the relation table (which Prisma will then include in the table name with an `_` in front). I've also removed `_many` from the end of one-sided many relationships for consistency.


dcousens:
Rebased

dcousens:
I think we have a few paths with this:

- Recommending a migration strategy in the release notes :heavy_check_mark:
- Adding a temporary option to opt-in to the previous behaviour of `config.experimental.useLegacyRelationNames`, or
- Adding a temporary option to opt-in to the previous behaviour of `config.experimental.useShortRelationNames`

The 3rd option helps with a blind major update,   and removes the automated migration in the default instance.
The automatic `drop` is worrying if users aren't being careful.

packages/keystone/src/lib/migrations.ts
```diff
@@ -252,7 +252,7 @@ We need to reset the ${credentials.type} database "${credentials.database}" at $
       console.log(`✨ A migration has been created at migrations/${generatedMigrationName}`);
 
       let shouldApplyMigration =
-        migrationCanBeApplied && (await confirmPrompt('Would you like to apply this migration?'));
```
dcousens, line 255: @mitchellhamilton I have defaulted this to `no`,  this should help prevent an unintentional `DROP`.

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
+---
+'@keystone-next/keystone': major
+---
+
+The names of two-sided many-many relationships now only include the list key and field key for one side of the relationship and one-sided many relationships(which are many-many) no longer have `_many` at the end of them for consistency with two-sided relationships. This allows having many-many relationships with long list and field keys without hitting Postgres's 63 character limit.
```
dcousens, line 5:
```suggestion
The names of two-sided many-many and one-to-one relationships has have been shortened to the left hand side only.
This reduces the probability that you will exceed [POSTGRESQL's 63 character limit for identifiers](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) with typical usage.

This is a breaking change. 

```

mitchellhamilton: > one-to-one relationships has have been shortened to the left hand side only

This isn't correct and you've removed the mention of one-sided many relationships.

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
...
+
+There are two ways to update:
+
+1. Enable the `experimental.useLegacyManyRelationNames` option at the root of your config. This will make Keystone revert to the old relation names. This option will be removed in a future release.
```
dcousens, line 9:
```suggestion
1. Set `experimental.useLegacyManyRelationNames: true` at the root of your config. This will make Keystone revert to the old relation names.  This feature will be removed in a future release.
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
...
+2. Apply a migration to rename your many relation tables
```
dcousens, line 10:
```suggestion
2. Or,  rename your many relation tables (see below)
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
...
+
+Given a schema like this:
```
dcousens, line 12:
```suggestion
For example,  given a schema like this:
```

.changeset/tasty-buses-jog.md
dcousens: Fixed to read `one-to-many`

mitchellhamilton: It's not about one to many, it's about one sided many relationships(which are many to many) - https://keystonejs.com/docs/guides/relationships

.changeset/tasty-buses-jog.md
dcousens: While this feature will be removed in the future,  we are looking at the capability to use custom relation names rather than these automatically generated names.  This should resolve the problem in a robust manner.

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
...
+  },
+}),
+Tag: list({
+  fields: {
+    posts: relationship({ ref: 'Post.tags', many: true }),
+  },
+}),
+```
+
+When you run Keystone after updating, Keystone will prompt you to create a migration, you should do this but **DO NOT APPLY IT**, it needs to be modified before being applied.
```
dcousens, line 27:
```suggestion
When updating to this change,   and running `yarn dev`, Keystone will prompt you to update your schema.

- If you are using `useMigrations: true`, Keystone will follow the typical migration flow offer to apply an automatically generated migration.

### DO NOT APPLY THE AUTOMATICALLY GENERATED MIGRATION
(unless you want to `DROP` your data)

- If you are using `useMigrations: false`, Keystone will follow the typical flow and offer to automatically migrate your schema.   Again,

### DO NOT AUTOMATICALLY MIGRATE
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,114 @@
...
+
+On Postgres, Prisma will generate a migration that looks something like this:
```
dcousens, line 29:
```suggestion
On PostgreSQL, Prisma will generate a migration that looks something like this:
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,127 @@
+---
+'@keystone-next/keystone': major
+---
+
+The names of one-sided and two-sided, many-many relationships has been shortened.  Two-sided many-many relationship names contain only the left-hand side names now; and the `_many` suffix has been dropped from one-sided many-many relationships.
```
dcousens, line 5:
```suggestion
The names of one-sided and two-sided, many-many relationships has been shortened. Two-sided many-many relationship names contain only the left-hand side names now; and the `_many` suffix has been dropped from one-sided many-many relationships.
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,127 @@
...
+
+This reduces the probability that you will exceed [PostgreSQL's 63 character limit for identifiers](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS) with typical usage.
+
+This is a breaking change. 
+
+There are two ways to update:
+
+1. Set `experimental.useLegacyManyRelationNames: true` at the root of your config. This will make Keystone revert to the old relation names.  This feature will be removed in a future release.
```
dcousens, line 13:
```suggestion
1. Set `experimental.useLegacyManyRelationNames: true` at the root of your config. This will make Keystone revert to the old relation names. This feature will be removed in a future release.
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,127 @@
...
+2. Or,  rename your many relation tables using a migration (see below)
```
dcousens, line 14:
```suggestion
2. Or, rename your many relation tables using a migration (see below)
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,127 @@
...
+
+For example,  given a schema like this:
```
dcousens, line 16:
```suggestion
For example, given a schema like this:
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,127 @@
...
+```sql
+ALTER TABLE "_Post_tags_Tag_posts" RENAME TO "_Post_tags";
+ALTER INDEX "_Post_tags_Tag_posts_AB_unique" RENAME TO "_Post_tags_AB_unique";
+ALTER INDEX "_Post_tags_Tag_posts_B_index" RENAME TO "_Post_tags_B_index";
+ALTER TABLE "_Post_tags" RENAME CONSTRAINT "_Post_tags_Tag_posts_A_fkey" TO "_Post_tags_A_fkey";
+ALTER TABLE "_Post_tags" RENAME CONSTRAINT "_Post_tags_Tag_posts_B_fkey" TO "_Post_tags_B_fkey";
+```
+
+On SQLite, Prisma will generate a migration that looks something like this:
+
```
dcousens, line 90:
```suggestion
```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,126 @@
...
+  },
+}),
+```
+
+When updating to this change, and running `yarn dev`, Keystone will prompt you to update your schema.
+
+- If you are using `useMigrations: true`, Keystone will follow the typical migration flow offer to apply an automatically generated migration.
+
+### DO NOT APPLY THE AUTOMATICALLY GENERATED MIGRATION
+(unless you want to `DROP` your data)
```
dcousens, line 36:
```suggestion

```

.changeset/tasty-buses-jog.md
```diff
@@ -0,0 +1,121 @@
...
+    posts: relationship({ ref: 'Post.tags', many: true }),
+  },
+}),
+```
+
+When updating to this change, and running `yarn dev`, Keystone will prompt you to update your schema.
+
+- If you are using `useMigrations: true`, Keystone will follow the typical migration flow offer to apply an automatically generated migration. **DO NOT APPLY THE AUTOMATICALLY GENERATED MIGRATION** - unless you want to `DROP` your data.
+
+- If you are using `useMigrations: false`, Keystone will follow the typical flow and offer to automatically migrate your schema.   Again, **DO NOT RUN THE AUTOMATIC MIGRATION** - unless you want to `DROP` your data.
```
dcousens, line 35:
```suggestion
- If you are using `useMigrations: false`, Keystone will follow the typical flow and offer to automatically migrate your schema.  Again, **DO NOT RUN THE AUTOMATIC MIGRATION** - unless you want to `DROP` your data.
```

molomby:
Linking to [relevant Slack thread](https://thinkmill.slack.com/archives/C0107LF95QX/p1636595990326700)

molomby:
If you were building the DB yourself, most people would name these adjacency tables for the two tables they related. Eg, given `User.reviewedPosts` <-> `Post.reviewedByUsers` you'd name the table `users_posts`, `User_Post` or something like that. Obviously the problem with this generally is when you have multiple many-to-many relationships, like, say you also have `User.contributedToPosts` <-> `Post.contributors`. In that case (again, assuming you were doing this "by hand") you'd just pick sensible names for the tables, eg: maybe `PostReviews` for the former and `PostContributions` for the latter. Problem solved.

Having multiple many-to-many relationships between two lists is going to be pretty rare for most systems but of course, we still need to solve for it in the general case. My "perfect world" solution would look something like this:

* By default, adjacency tables are named for `${LEFT_LIST}_${RIGHT_LIST}`, as above
* We add a config option to relationship fields that lets app builders specify a alternate "relationship key" if they want
* If you have multiple many-to-many relationships between two lists with no relationship keys Keystone throws an error on startup, directing devs to specify a value for one of the colliding relationships
* I suppose we'd also need to validate the dev hasn't set conflicting keys/names on both sides of the same relationship, etc.

This is more work that some other solutions we could use but the upsides are..

* It follows our general principal that the automatically generated stuff should, ideally, be "what you'd build yourself if you had time"
* It give us another option for people when upgrading to this release, eg. "if you don't want your DB structure to change, update the config for all your existing many-to-many relationships and specify a relationship key matching your the current table names" (It might be worth it just for this, no?)
* Letting app builders specify a "relationship key"/table name to use will sometimes be handy even if there's no naming conflict. Eg. to align with an existing DB structure or, really, by anyone who care about the details of their DB structure

dcousens:
I think `users_reviewed_posts` (approximately what this implements) neatly accounts for the multiple many-to-many scenario without issue and meets the canonical expectation.

The problem is we have `users_reviewedPosts_posts_reviewedByUsers` now,  and that's triggering PostgreSQL inbuilt limits quickly.

The primary concern for me is that by automatically generating migrations without enforcing user consideration could be an unnecessary foot gun.  Maybe I'm too conservative.

molomby:
> I think `users_reviewed_posts` (approximately what this implements) neatly accounts for the multiple many-to-many scenario without issue and meets the canonical expectation.

Yeah the current solution is fine – I like that it prevents collisions but I don't love the way it arbitrarily picks the name (ie. depending on which list is left and which is right). Like, for `User.reviewedPosts` <-> `Post.reviewedByUsers` my table's either going to be `User_reviewedPosts` or `Post_reviewedByUsers`, both of which are ok but, as a developer, I don't know which I'm going to get.

Also, if you're going to give people a way to effect the name of these tables (ie. `useLegacyManyRelationNames`) why not just let them pick it? Eg. instead of...

```
const relationName = `${leftRel.listKey}_${leftRel.fieldPath}${
    useLegacyManyRelationNames ? `_${rightRel.listKey}_${rightRel.fieldPath}` : ''
}`;
```

Why not...

```
const relationName = (leftRel.name || rightRel.name) || `${leftRel.listKey}_${rightRel.listKey}`;
```

> The primary concern for me is that by automatically generating migrations [with this change], without enforcing user consideration, could be an unnecessary foot gun. Maybe I'm too conservative.

Yeah, this is the more important problem really, I just don't have any helpful opinion on it.

.changeset/tasty-buses-jog.md
```diff
@@ -10,8 +10,70 @@ This is a breaking change.
 
 There are two ways to update:
 
-1. Set `experimental.useLegacyManyRelationNames: true` at the root of your config. This will make Keystone revert to the old relation names. This feature will be removed in a future release.
-2. Or, rename your many relation tables using a migration (see below)
+### Set `db.relationName` on many to many relation
+
+To avoid doing a migration, you can set `db.relationName` to the previous relation name on one side of a many to many relationship.
```
dcousens, line 15:
```suggestion
Rather than doing a migration, you can set the new field property `db.relationName` on any side of a many to many relationship.
If set to the existing relation name, e.g `Post_tags_Tag_posts`, you will not need to run any migration code.
```

.changeset/tasty-buses-jog.md
```diff
@@ -10,8 +10,70 @@ This is a breaking change.
...
+  tags Tag[]  @relation("Post_tags_Tag_posts")
+}
+
+model Tag {
+  id    String @id @default(cuid())
+  posts Post[] @relation("Post_tags_Tag_posts")
+}
+```
+
+You can add `db: { relationName: 'Post_tags_Tag_posts' }` to one side of the many to many relationship. It doesn't matter which side of the relationship it's on but it must only be on one side.
```
dcousens, line 60:
```suggestion
By adding `db: { relationName: 'Post_tags_Tag_posts' }` to one side of the many-to-many relationship; you can preclude yourself from a migration.

**Note:** It doesn't matter which side of the relationship you put this property,  but it should be only on one side; otherwise you will receive an error.
```

dcousens (approved):

PR #6918, opened 2021-11-11
Use the user's locale to determine the format of the date in the timestamp field's view

### LINKED ISSUE #6913: `timestamp` field uses conflicting date formats that don't match users locale
The `timestamp` field uses `dd/mm/yyyy` in list view, `dd/mm/yyyy` as the field placeholder, `mm/dd/yyyy` when a date is selected, and `dd/mm/yyyy` when in read mode.

It should use the users locale in all cases.

**List**

Screen Shot 2021-11-10 at 2 01 52 pm

**Placeholder**

Screen Shot 2021-11-10 at 1 59 25 pm

**Editing**

Screen Shot 2021-11-10 at 1 59 10 pm

**Read mode**

Screen Shot 2021-11-10 at 2 06 18 pm

emmatown:
Closes #6913

### DIFF
--- a/.changeset/dirty-lemons-wash.md
+++ b/.changeset/dirty-lemons-wash.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/keystone': patch
+---
+
+The format of the date shown in the `timestamp` field in the Admin UI now uses the user's locale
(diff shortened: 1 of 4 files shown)


JedWatson (approved):

gautamsi:
Why not provide config in timestamp field.

It was quiet good with datetime field in v5.


I would want to use the ate time dropdown and also humanize form of  entry into the field (today at 5 pm)

PR #6920, opened 2021-11-11
Use Prisma's Node-API query engine

emmatown:
Basically this should improve performance when using the Prisma client. See https://www.prisma.io/docs/concepts/components/prisma-engines/query-engine and the section mentioned Node-API on https://github.com/prisma/prisma/releases/tag/3.0.1 for more details. This doesn't look to have an impact on CI performance which is expected since we generally create lots of Prisma clients and do only a couple operations with each of them in tests.

Using this caused some tests to more consistently fail with doing multiple writes on SQLite at the same time so I've made it so we limit writes on SQLite to 1 at a time for a given Prisma client rather than the previous sorta limiting but only if you use *Many operations and not if you do a bunch of single writes, see https://github.com/keystonejs/keystone/pull/6920/files#diff-a6e92b0e6eb325370eba0e5e2be3a8afdb3e866d484db7af462670267dcfc070 for more details.


tests/admin-ui-tests/utils.ts
```diff
@@ -76,7 +76,11 @@ export const deleteAllData: (projectDir: string) => Promise<void> = async (proje
 
     let prisma = new PrismaClient();
 
-    await Promise.all(Object.values(prisma).map((x: any) => x?.deleteMany?.({})));
+    await prisma.$transaction(
+      Object.values(prisma)
+        .filter((x: any) => x?.deleteMany)
+        .map((x: any) => x?.deleteMany?.({}))
+    );
```
mitchellhamilton, line 83: Doing things in transactions also gets around the SQLite write timeout immediately thing so I've done that here

tests/api-tests/parallel.test.ts
```diff
@@ -0,0 +1,67 @@
+import { text } from '@keystone-next/keystone/fields';
```
mitchellhamilton, line 1: These tests fail (or rather the update and delete ones fail, the create one for some reason does not) on `writes-in-parallel-failure`(which is main + this file) for me locally. For some reason they passed on CI.

mitchellhamilton: Update: I got these tests to fail on CI on this branch without the limiting: https://github.com/keystonejs/keystone/runs/4205904123?check_suite_focus=true#step:8:17

dcousens:
LGTM, but let's wait until after the next release for this

dcousens (approved):

PR #6926, opened 2021-11-12
Writes in parallel failure

mitchellhamilton:
I'm just opening this PR to get CI to run to demonstrate a problem that is fixed in #6920.


mitchellhamilton:
This doesn't seem to fail on CI even though it fails 100% of the time locally 🙃

PR #6927, opened 2021-11-12
Add explanation for `KS_ACCESS_DENIED` errors vs "not found"

JedWatson:


docs/pages/docs/apis/graphql.mdx
```diff
@@ -432,6 +432,8 @@ The following error codes can be returned from the Keystone GraphQL API.
 - `KS_RELATIONSHIP_ERROR`: An error occured while resolving the input relationship field.
 - `KS_PRISMA_ERROR`: An error occured while running a Prisma client operation.
 
+> A note on `KS_ACCESS_DENIED`: If you return a "not found" error from a mutation like `updateUser({ where: { secretKey: 'abc' } })` but an "access denied" error from `updateUser({ where: { secretKey: 'def' } })`, it would reveal the existence of a user with the secret key "def". To prevent leaking private information in this way, Keystone will always say "access denied" when you try to perform a mutation on an item that can't be operated on, whether that is because there is no matching record in the database, or there was but the user performing the operation doesn't have access to it.
```
mitchellhamilton, line 435: > If you return a "not found" error from a mutation like

Using `you` sounds to me like I as a Keystone user could choose to do this

JedWatson: Good feedback, have simplified it back to just "Returning a ..."

docs/pages/docs/apis/graphql.mdx
```diff
@@ -432,6 +432,8 @@ The following error codes can be returned from the Keystone GraphQL API.
 - `KS_RELATIONSHIP_ERROR`: An error occured while resolving the input relationship field.
 - `KS_PRISMA_ERROR`: An error occured while running a Prisma client operation.
 
+> A note on `KS_ACCESS_DENIED`: Returning a "not found" error from a mutation like `updateUser({ where: { secretKey: 'abc' } })` but an "access denied" error from `updateUser({ where: { secretKey: 'def' } })`, it would reveal the existence of a user with the secret key "def". To prevent leaking private information in this way, Keystone will always say "access denied" when you try to perform a mutation on an item that can't be operated on, whether that is because there is no matching record in the database, or there was but the user performing the operation doesn't have access to it.
```
mitchellhamilton, line 435:
```suggestion
> A note on `KS_ACCESS_DENIED`: Returning a "not found" error from a mutation like `updateUser({ where: { secretKey: 'abc' } })` but an "access denied" error from `updateUser({ where: { secretKey: 'def' } })` would reveal the existence of a user with the secret key "def". To prevent leaking private information in this way, Keystone will always say "access denied" when you try to perform a mutation on an item that can't be operated on, whether that is because there is no matching record in the database, or there was but the user performing the operation doesn't have access to it.
```

mitchellhamilton (approved):

docs/pages/docs/apis/graphql.mdx
JedWatson: Thanks

Issue #6930, opened 2021-11-15
Move from jest to @playwright/test for admin UI integration tests


gwyneplaine:
At the moment integration tests and smoke tests are failing in CI non-deterministically at an unacceptable rate. 
The working short term solution is to apply `async-retry` to these tests to reduce the probability of failure, however the likely best response is to migrate to the @playwright/test test-runner which handles retries for these classes of errors natively.

dcousens:
@gwyneplaine these test failures started appearing en-mass with the bump to nextjs-12.   I think establishing the underlying reason why they are failing will be more useful than migrating test frameworks;  to ensure that switching is not a wasted effort.

gwyneplaine:
> @gwyneplaine these test failures started appearing en-mass with the bump to nextjs-12. I think establishing the underlying reason with a barebones reproduction will be more useful than migrating test frameworks; to ensure that switching is not a wasted effort.

yep, that's reasonable, let's talk about this when you're in, I'd like to pair on what that minimal viable reproduction could look like, since this is quite the annoying heisenbug

aaronpowell:
If this is still something on your radar, I'm happy to reach out to the Playwright team to see if they can offer any advice on migration.

PR #6931, opened 2021-11-15
add async-retry to non-deterministic filters test

gwyneplaine:
Same as https://github.com/keystonejs/keystone/pull/6866
Temporary application of async-retry to resolve deterministically failing tests. Likely to do with interactions between NextJS and playwright. 

The *best* solution for this is likely to migrate to `playwright test runner`, which is tracked in #6930


emmatown (changes requested):
Naively retrying this test like this will make it fail since it will create more tasks will be assigned to the author that is being filtered on, can you try only adding the retry to this bit:
```ts
const filteredElements = await page.$$('table tbody tr');
expect(filteredElements.length).toBe(1);
```

Or see if you can get playwright to wait for only 1 of those elements to exist

emmatown (approved):

PR #6942, opened 2021-11-15
add comment about text field default

dcousens:
A question was asked why the `text` field was `isNullable` where others use `getResolvedIsNullable`.
My suspicion is that Keystone 6 has prioritized a zero-length string as a reasonable default instead as it is additionally falsy,  compared to the null?
If no other reasoning present's itself, I'm happy to make that canon.


packages/keystone/src/fields/types/text/index.ts
```diff
@@ -70,6 +70,7 @@ export const text =
       },
     };
 
+    // defaulted to false as a zero length string is preferred
```
dcousens, line 73:
```suggestion
    // defaulted to false as a zero length string is preferred to null
```

bladey: // defaulted to false as a zero length string is preferred to null
```suggestion
    // defaulted to false as a zero length string is preferred
```

bladey (approved):

packages/keystone/src/fields/types/text/index.ts
```diff
@@ -70,6 +70,7 @@ export const text =
       },
     };
 
+    // defaulted to false as a zero length string is preferred to null
```
bladey, line 73:
```suggestion
    // defaulted to false as a zero length string is preferred to null
```

bladey (changes requested):

bladey (approved):

molomby:
There's now a discussion on [text field nullability defaults](https://github.com/keystonejs/keystone/discussions/7158) that relates to this.

PR #6954, opened 2021-11-16
Update Node Engine Versions (and update test)

bladey:
There are issues with Node 17 (Next.js, Prisma) and we shouldn't be using it for production regardless. since it won't ever be LTS.

Also, Node 12 is old enough to phase out now that we have Node 14 and Node 16 as LTS.

I feel we should support Node 14 and Node 16 in the ranges as per PR.

The Node 14 and Node 16 ranges are as per LTS releases here - https://nodejs.org/en/download/releases/

Node 16 LTS also matches our test suites.

This also needs to be rolled out to all other Keystone repos using Keystone 6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,6 @@
     "babel-plugin-remove-graphql-queries": "2.7.2"
   },
   "engines": {
-    "node": "^12.20 || >= 14.13"
+    "node": "^14.15 || ^16.13"
   }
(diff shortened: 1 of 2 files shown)


bladey:
Also adds retry for a common failing test.

dcousens (approved):

PR #6961, opened 2021-11-17
Fix default renderer of layout block

datner:
Hey, while developing some project I noticed a very annoying bug. They layout in the editor does not correlate to the layout rendered by the `<DocumentRenderer />` component.
for example:
## editor (Slate):
image
image
DOM is structured correctly
```
div.grid
├── div
│   └── h1
└── div
    ├── h4
    └── p
```

## app (DocumentRenderer):
image
image
DOM is structured incorrectly
```
div.grid
├── h1
├── h4
└── p
```
-------------------------
To be honest I have no idea how this wasn't reported yet, but the fix is simple (I think, it's hard to test within the repo)

Please tell me if I missed something I should do <3

### DIFF
--- a/packages/document-renderer/src/index.tsx
+++ b/packages/document-renderer/src/index.tsx
@@ -42,7 +42,6 @@ interface Renderers {
     blockquote: OnlyChildrenComponent;
     code: Component<{ children: string }> | keyof JSX.IntrinsicElements;
     layout: Component<{ layout: [number, ...number[]]; children: ReactElement[] }>;
-    layoutArea: OnlyChildrenComponent;
     divider: Component<{}> | keyof JSX.IntrinsicElements;
     heading: Component<{
       level: 1 | 2 | 3 | 4 | 5 | 6;
@@ -98,11 +97,12 @@ export const defaultRenderers: Renderers = {
             gridTemplateColumns: layout.map(x => `${x}fr`).join(' '),
           }}
         >
-          {children}
+          {children.map((element, i) => (
+            <div key={i}>{element}</div>
+          ))}
         </div>
       );
     },
-    layoutArea: 'div',
   },
 };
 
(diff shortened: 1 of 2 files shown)


mitchellhamilton (approved):

dcousens:
Thanks for reporting this @datner! :tada: 
Let us know if @mitchellhamilton's changes work for you

packages/document-renderer/src/index.tsx
```diff
@@ -42,6 +42,7 @@ interface Renderers {
     blockquote: OnlyChildrenComponent;
     code: Component<{ children: string }> | keyof JSX.IntrinsicElements;
     layout: Component<{ layout: [number, ...number[]]; children: ReactElement[] }>;
+    layoutArea: OnlyChildrenComponent;
```
mitchellhamilton: Note I've changed this so the layout renders the `div`s rather than add another renderer to be consistent with how the lists work and so that e.g. if you wanted to customize the rendering of the layout area based on the index that the layout area is at, you can easily do that because you're controlling the rendering of both the layout and the layout area in one place

datner: @mitchellhamilton but won't that make it inconsistent with the model slate provides? I didn't invent the `layout-area` key 🤔 
image

this essentially blocks the ability to provide a custom renderer for layout-areas

mitchellhamilton: Not sure how you mean it's inconsistent with the model slate provides, The layout-area nodes still exist, just because there's a node of a given type doesn't mean there has to a one to one mapping of node type to renderer.

The way you provide a custom renderer for layout areas is by providing the custom renderer for layout and mapping over children and rendering whatever you want for each layout area, that way you have more control over rendering like you could decide to render the first layout area differently than all the other ones or anything else like that because you control the mapping of the layout areas. This is the same way the list renderer works, there's no list item renderer, only list.

PR #6967, opened 2021-11-18
Fix relationship field not respecting ui.displayMode: cards in create view

mitchellhamilton:
I also ended up refactoring some things to avoid awkward type things because what the display mode is was represented in two places and the `Field` view needed to access both of them.

Note the relationship field doesn't respect `ui.displayMode: 'count'` on the create view and this doesn't change that, I think that's fine and good given that would just mean showing nothing, and `ui.displayMode: 'count'` just exists because our other UIs are bad at handling relationships with a lot of related items so we might as well give users a better UI when we know that it's okay.

### DIFF
--- a/packages/keystone/src/fields/types/relationship/views/cards/index.tsx
+++ b/packages/keystone/src/fields/types/relationship/views/cards/index.tsx
@@ -75,23 +75,23 @@ export function Cards({
 }: {
   foreignList: ListMeta;
   localList: ListMeta;
-  field: ReturnType<typeof controller> & { display: { mode: 'cards' } };
-  id: string;
+  id: string | null;
   forceValidation: boolean | undefined;
   value: { kind: 'cards-view' };
-} & Pick<FieldProps<typeof controller>, 'value' | 'onChange'>) {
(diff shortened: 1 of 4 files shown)


packages/keystone/src/fields/types/relationship/views/cards/index.tsx
```diff
@@ -75,23 +75,22 @@ export function Cards({
 }: {
   foreignList: ListMeta;
   localList: ListMeta;
-  field: ReturnType<typeof controller> & { display: { mode: 'cards' } };
-  id: string;
-  forceValidation: boolean | undefined;
+  id: string | null;
   value: { kind: 'cards-view' };
```
gwyneplaine, line 79: @mitchellhamilton curiosity: Is there value in providing a more explicit type here, or is this purely for type narrowing purposes?

mitchellhamilton: It's to narrow the type

gwyneplaine (approved):
lgtm

PR #6972, opened 2021-11-19
Fix item page crashing when failing to fetch an item

mitchellhamilton:
Before in prod:
Something went wrong text in the middle of a web page with a button to reload the page
Before in dev:
Error from Next.js in the middle of a web page pointing to an line in Keystone's source

In this PR:
Keystone Admin UI on the item page with the nav bar shown with no fields but an error notice saying the item could not be found or you don't have access to it


gwyneplaine (approved):
nice

PR #6982, opened 2021-11-23
feat: support prisma SHADOW_DATABASE_URL

jlarmstrongiv:
See slack discussions:
- https://keystonejs.slack.com/archives/C01STDMEW3S/p1637513388084600?thread_ts=1637285054.059900&cid=C01STDMEW3S

Relevant links:
- https://www.prisma.io/docs/concepts/components/prisma-migrate/shadow-database#cloud-hosted-shadow-databases-must-be-created-manually
- https://github.com/prisma/prisma-engines/pull/1672

### DIFF
--- a/examples-staging/assets-cloud/schema.prisma
+++ b/examples-staging/assets-cloud/schema.prisma
@@ -2,8 +2,9 @@
 // Modify your Keystone config when you want to change this.
 
 datasource sqlite {
-  url      = env("DATABASE_URL")
-  provider = "sqlite"
+  url               = env("DATABASE_URL")
+  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
+  provider          = "sqlite"
 }
 
 generator client {
--- a/examples-staging/assets-local/schema.prisma
+++ b/examples-staging/assets-local/schema.prisma
@@ -2,8 +2,9 @@
 // Modify your Keystone config when you want to change this.
 
 datasource sqlite {
-  url      = env("DATABASE_URL")
-  provider = "sqlite"
+  url               = env("DATABASE_URL")
+  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
+  provider          = "sqlite"
 }
 
 generator client {
--- a/examples-staging/auth/schema.prisma
+++ b/examples-staging/auth/schema.prisma
@@ -2,8 +2,9 @@
 // Modify your Keystone config when you want to change this.
 
 datasource sqlite {
-  url      = env("DATABASE_URL")
-  provider = "sqlite"
+  url               = env("DATABASE_URL")
+  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
+  provider          = "sqlite"
 }
 
 generator client {
--- a/examples-staging/basic/schema.prisma
+++ b/examples-staging/basic/schema.prisma
@@ -2,8 +2,9 @@
 // Modify your Keystone config when you want to change this.
 
 datasource sqlite {
-  url      = env("DATABASE_URL")
-  provider = "sqlite"
+  url               = env("DATABASE_URL")
+  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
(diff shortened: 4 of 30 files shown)


gwyneplaine:
Thanks for this contribution @jlarmstrongiv ! We're currently in feature freeze mode leading up to a general availability release in the coming weeks. We'll have a look at this in more detail after that 🎉 

Thanks again for your contribution and patience

jlarmstrongiv:
@gwyneplaine perhaps I mislabeled this as a feature?  This fixes the error:

```
EngineError: P3014

Prisma Migrate could not create the shadow database. Please make sure the database user has permission to create databases. Read more about the shadow database (and workarounds) at https://pris.ly/d/migrate-shadow

Original error:
db error: ERROR: permission denied to create database
   0: sql_migration_connector::flavour::postgres::sql_schema_from_migration_history
             at migration-engine/connectors/sql-migration-connector/src/flavour/postgres.rs:354
   1: migration_core::api::DevDiagnostic
             at migration-engine/core/src/api.rs:108
```

And is necessary developing with cloud databases.

gwyneplaine:
@jlarmstrongiv thanks for clarifying the situation, we’ve had a brief look into the space and there are ramifications across the system that we need to consider more deeply. However, we’re focused on GA for near term, that said we’ll discuss this more internally and get back to you as soon as we can 👍

jlarmstrongiv:
Thank you for taking a look @gwyneplaine !  

For system ramifications, are you considering including this setting in the keystone config like the `DatabaseConfig.url` property?  If so, I can look into adding `DatabaseConfig.shadowDatabaseUrl` to the config and protecting the environment variables like https://github.com/keystonejs/keystone/blob/04c54a4eb4aa6076cf87d441060eaa2091bc903b/packages/keystone/src/lib/migrations.ts#L20

Unfortunately, this issue prevents me from using keystone.  If you have guidance after internal discussions, please let me know!  I will gladly try to fix this PR as soon as possible.

The good news is that a blank `env("SHADOW_DATABASE_URL")` works with prisma https://github.com/prisma/prisma-engines/pull/1672

bladey:
Hey @jlarmstrongiv, could you run `yarn` in the root directory so the GraphQL schemas are updated? Due to this change they all need to be updated (~29 files). It's causing the tests to fail and I don't have permission to commit to this PR.

jlarmstrongiv:
@bladey done 👍

molomby:
@jlarmstrongiv 

> [...] are you considering including this setting in the keystone config like the `DatabaseConfig.url` property? If so, I can look into adding `DatabaseConfig.shadowDatabaseUrl` to the config [...]

Yep, this would be the preferred solution I think – we should expose the Prisma option in the Keystone config.

Additionally, as you note...

> The good news is that a blank `env("SHADOW_DATABASE_URL")` works with prisma [...]

Indeed, so if we started adding `shadowDatabaseUrl = env("SHADOW_DATABASE_URL")` to the generated `prisma.schema` files, we wouldn't _necessarily_ need to support it to the Keystone config object. Still probably nicer to support in directly, along side the other config.

Note though, the code changes currently in PR don't actually implement either these solutions.

jlarmstrongiv:
@molomby I have a local branch that adds the `shadowUrl` to the config and docs as described in the preferred solution.  But, I need to learn more about the monorepo works to build the changes locally and also find why those checks failed.

The relevant files are:

- packages/core/src/lib/migrations.ts
- packages/core/src/types/config/index.ts
- docs/pages/docs/apis/config.mdx

> Note though, the code changes currently in PR don't actually implement either these solutions.
I believe PR currently implements the `shadowDatabaseUrl = env("SHADOW_DATABASE_URL")`.  Not ideal, but simply put, it’s impossible use keystone without it—that’s why I love the codesandbox packages.  Both solutions are doable, as it’s a relatively small change.

jlarmstrongiv:
Followup in https://github.com/keystonejs/keystone/pull/7350

PR #6995, opened 2021-11-24
Header fixes

gwyneplaine:
* Documentation button now visible on all non small breakpoints
* Why Keystone and Updates now removed for small breakpoint. 

Before:
 

https://user-images.githubusercontent.com/12119389/143175470-dd51f9ba-d699-43f1-871d-a9aa45e830a2.mov

After:

https://user-images.githubusercontent.com/12119389/143175478-6b97929d-ae02-470b-b229-2760a0782fbc.mov

### DIFF
--- a/docs/components/Header.tsx
+++ b/docs/components/Header.tsx
@@ -213,7 +213,7 @@ export function Header() {
         css={mq({
           display: 'grid',
           gridTemplateColumns: [
-            'auto auto max-content max-content',
+            'auto max-content max-content max-content',
             'auto max-content max-content max-content max-content max-content max-content',
             'max-content auto max-content max-content max-content max-content max-content',
             'max-content auto max-content max-content max-content max-content max-content max-content',
@@ -242,7 +242,7 @@ export function Header() {
         <div
           id="search-field-container"
           css={mq({
-            display: 'block',
+            display: ['none', null, 'block'],
             width: ['100%', null, null, null, '80%'],
           })}
         >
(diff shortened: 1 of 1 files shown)


docs/components/Header.tsx
```diff
@@ -275,7 +286,7 @@ export function Header() {
           shadow
           css={mq({
             '&&': {
-              display: ['none', null, null, 'inline-flex'],
+              display: ['none', null, 'inline-flex', 'inline-flex'],
```
dominikwilkowski, line 289: To make it easier on yourself you don't need to declare the later break point:
```suggestion
              display: ['none', null, 'inline-flex'],
```

dominikwilkowski (changes requested):
I checked this out and it breaks the mobile view currently in both firefox and chrome:
image

docs/components/Header.tsx
gwyneplaine: yeah nice 👍

gwyneplaine:
> I checked this out and it breaks the mobile view currently in both firefox and chrome: image

Solved and promptly redacted when i realised the search dialog itself is broken in a non-trivial way in mobile. 
Going to submit a PR for a better search experience for mobile separately, redacting it for now so that mobile experience is as before (searchless)

dominikwilkowski (approved):
Works now. The two smallest breakpoints now don't show the documentation link which I think was intended.

PR #6996, opened 2021-11-25
add canonical Dockerfile to examples/basic

dcousens:
This is a draft.

I am open to community thoughts on where this example should live?
We want to encourage test deployments of our examples in different environments for users trial Keystonejs.
However,  as is,  we'll end up copying a `Dockerfile` to every example. 

We might also want to promote different environments, `arm64`, `x86_64` et al;  files for each of those (and other deployment examples) does not scale, and would add noise to the examples as documentation.

- [ ] Add `USER`
- [x] Fix the `SESSION_SECRET` `keystone build` problems
- [x] `migrate deploy` shouldn't need `keystone.ts`

### DIFF
--- a/examples/basic/Dockerfile
+++ b/examples/basic/Dockerfile
@@ -1,0 +1,18 @@
+# Build
+FROM node:16-alpine as build
+COPY . /app
+WORKDIR /app
+
+ENV NODE_ENV production
+ENV SESSION_SECRET 'temporary SESSION_SECRET for build'
+RUN yarn install
+RUN yarn build
+
+# Serve
+FROM node:16-alpine
+
+WORKDIR /app
+COPY --from=build /app .
+
+ENV NODE_ENV production
+CMD ["yarn", "keystone", "start"]
--- a/examples/basic/Dockerfile
+++ b/examples/basic/Dockerfile
@@ -1,4 +1,4 @@
-# Build
+# build
 FROM node:lts-alpine as build
 COPY . /app
 WORKDIR /app
@@ -8,7 +8,7 @@ RUN corepack enable
 RUN pnpm install
 RUN pnpm build
 
-# Serve
+# serve
 FROM node:lts-alpine
 
 WORKDIR /app
--- a/examples/basic/Dockerfile
+++ b/examples/basic/Dockerfile
@@ -1,18 +1,18 @@
 # Build
-FROM node:16-alpine as build
+FROM node:lts-alpine as build
 COPY . /app
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


tv42:
SESSION_SECRET and DATABASE_URL complicate this horribly. Since `npm run build` imports the program, those values need to be set even at build time!

tv42:
Using a separate build image and `COPY --from=build /app /` means that the final image cannot reuse layers that didn't change; that flattens the whole ~1 GB node_modules pile into one layer, which must be transferred over the network on every change.

dcousens:
@tv42 you mentioned
> those values need to be set even at build time

You shouldn't need to configure `DATABASE_URL` for `build`, and `SESSION_SECRET` can be configured to be randomly generated if no `env` variable is present.

tv42:
It seems I am required to provide a `DatabaseConfig` with a `url`:
https://github.com/keystonejs/keystone/blob/c396f16a77798852208e93c2a8bb4a07c04c3af5/packages/core/src/types/config/index.ts#L24
https://github.com/keystonejs/keystone/blob/c396f16a77798852208e93c2a8bb4a07c04c3af5/packages/core/src/types/config/index.ts#L56

Providing a dummy value if the environment variable is set is not desirable to me; I want to check that the environment is configured correctly and give an error message if not.

For example:

```
const databaseUrl = (function () {
  const fromEnv = process.env.DATABASE_URL;
  if (fromEnv !== undefined) {
    return fromEnv;
  }
  if (isDev) {
    // Default convenience value during development.
    return "postgres://postgres:pgdev@localhost:5432/postgres?application_name=keystone&schema=cms";
  }
  throw new Error("missing environment variable DATABASE_URL");
})();
```

```
let sessionSecret = process.env.SESSION_SECRET;
if (!sessionSecret) {
  if (process.env.NODE_ENV === "production") {
    throw new Error(
      "The SESSION_SECRET environment variable must be set in production"
    );
  } else {
    sessionSecret = "dev-notReallySecret-session-key-dummy";
  }
}
```

The way the API has been structured prevents me from doing this.

gautamsi:
I tried with the basic example. some observations not sure if that is problem with keystone or due to prisma/nextjs

1. docker image size when pushed to registry is almost 900MB with node alpine
2. memory footprint with basic example with not lot of lists is ~200mb

I am not too much worried about the image size as in private registry, it says 264 mb total size. but the memory footprint should be optimized.

dcousens:
@gautamsi neat metrics!
I think the memory requirements could be tracked in a different issue, but I wonder if we could CI that graphed it over time.

PR #7006, opened 2021-11-29
🚀 General Availability CLI changes

gwyneplaine:
^ as it says on the tin. 
PR-ing this into the ga-release branch to make it easier to review the changes.


packages/keystone/README.md
```diff
@@ -1,3 +1,3 @@
-# @keystone-6/keystone
+# @keystone-next/keystone
```
mitchellhamilton, line 1: Was this intentional?

gwyneplaine: yep 👍

gwyneplaine: oh wait, nope! rebase problems

docs/pages/updates/index.tsx
```diff
@@ -715,7 +715,7 @@ export default function WhatsNew() {
             <li>
               Exchanged <InlineCode>deploy</InlineCode>, <InlineCode>reset</InlineCode> and{' '}
               <InlineCode>generate</InlineCode> commands for{' '}
-              <InlineCode>keystone-next prisma</InlineCode> commands.
+              <InlineCode>keystone prisma</InlineCode> commands.
```
mitchellhamilton, line 718: I would think we wouldn't want to update this(and the other bit in this file) given it's about a release and in that release, the command was `keystone-next`?

gwyneplaine: ah good catch, missed this exclude in the find-replace

emmatown (approved):

PR #7014, opened 2021-11-30
Improve console output

JedWatson:
A minor tweak to make important messaging stand out more, in particular the notice when keystone stops because of schema changes and you need to restart it.

### DIFF
--- a/.changeset/grumpy-humans-care.md
+++ b/.changeset/grumpy-humans-care.md
@@ -2,4 +2,4 @@
 "@keystone-6/keystone": patch
 ---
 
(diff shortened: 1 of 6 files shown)


.changeset/grumpy-humans-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-6/keystone": patch
+---
+
+Improve console output, especially when you need to restart the server because of schema changes
```
dcousens, line 5:
```suggestion
Improve console output for when you need to restart the server because of schema changes
```

PR #7017, opened 2021-11-30
Noviny/website fixes

Noviny:
Got some fixes in here to use MDX2, which will unbreak the live site.

There's at least 1 TS error I don't think is worth fixing in here.


gwyneplaine (approved):
lgtm

PR #7033, opened 2021-12-01
Reword homepage intro for easy of understanding

jesstelford:
I found the homepage intro very hard to understand when first reading it. I came across the following stumbling blocks:

- Very long sentences
- The first paragraph used the work `and` 4 times
- Too many ideas per sentence
- Generic words detracted from the powerful message in the sentences

I've taken a stab at rewriting the intro to have:

- Short sentences; makes the content much easier to digest
- Single idea per sentence; understanding the key benefits is much easier this way
- Removed a few filler + generic words; keeps it succinct and to the point. Every sentence encourages the user to read the next.

Screen Shot 2021-12-01 at 3 50 12 pm

### DIFF
--- a/docs/pages/index.tsx
+++ b/docs/pages/index.tsx
@@ -55,13 +55,13 @@ export default function IndexPage() {
             The <Highlight look="grad1">superpowered</Highlight> CMS for developers
           </IntroHeading>
           <IntroLead>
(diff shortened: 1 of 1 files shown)


dcousens:
We didn't land with a consensus that this was an improvement @jesstelford,  happy to re-open if you want to continue discussing that though :heart_decoration:

PR #7039, opened 2021-12-02
Improve performance of generating artifacts for projects in the repo

emmatown:
Currently we have some shell stuff that ends up being very slow(~2 minutes) because of the way the Keystone repo is setup, we end up basically re-compiling Keystone for every project. This script takes ~10 seconds instead which ofc isn't great but it's probably good enough for now. The big motivation here being that we'll be able to use the generated types in examples without making our postinstall take 2 minutes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,17 +94,6 @@ jobs:
     name: Package Unit Tests
     needs: should_run_tests
     runs-on: ubuntu-latest
-    services:
-      postgres:
-        image: postgres:12
-        env:
-          POSTGRES_USER: keystone5
-          POSTGRES_PASSWORD: k3yst0n3
-          POSTGRES_DB: test_db
-        ports:
-          - 5432:5432
-    env:
-      DATABASE_URL: 'postgres://keystone5:k3yst0n3@localhost:5432/test_db'
     strategy:
       fail-fast: false
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -353,8 +353,5 @@ jobs:
       - name: Remark
         run: yarn lint:markdown
 
-      - name: Example schemas
-        run: yarn lint:examples
-
       - name: Prisma Filters
         run: yarn lint:filters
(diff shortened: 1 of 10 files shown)


.github/workflows/tests.yml
```diff
@@ -353,8 +342,5 @@ jobs:
       - name: Remark
         run: yarn lint:markdown
 
-      - name: Example schemas
-        run: yarn lint:examples
```
mitchellhamilton, line 357: This is unnecessary now since it happens in our `prepare` script (which happens after install)

examples-staging/ecommerce/keystone.ts
```diff
@@ -46,18 +46,16 @@ export default withAuth(
...
-      : {
-          provider: 'sqlite',
-          url: databaseURL,
-          async onConnect(context) {
-            console.log('Connected to the database!');
-            if (process.argv.includes('--seed-data')) {
-              await insertSeedData(context);
-            }
-          },
-        },
```
mitchellhamilton, line 60: We don't support changing which database provider you're using dynamically like this, and doing this messed up running because on some jobs in CI, we have a `DATABASE_URL` env variable set.

packages/keystone/src/___internal-do-not-use-will-break-in-patch/require-source.ts
```diff
@@ -0,0 +1 @@
+export { requireSource } from '../lib/config/requireSource';
```
mitchellhamilton, line 1: I've intentionally not added a changeset for this, it'll happen to be released at some point but it's not public so I don't want to talk about and imply that it should be used. It's entirely because this is the easiest way to get that code into our script.

.github/workflows/tests.yml
```diff
@@ -94,17 +94,6 @@ jobs:
...
-      postgres:
-        image: postgres:12
-        env:
-          POSTGRES_USER: keystone5
-          POSTGRES_PASSWORD: k3yst0n3
-          POSTGRES_DB: test_db
-        ports:
-          - 5432:5432
-    env:
-      DATABASE_URL: 'postgres://keystone5:k3yst0n3@localhost:5432/test_db'
```
mitchellhamilton, line 107: This was only used for the ecommerce project tests which will now just always use SQLite so this isn't needed.

JedWatson (approved):

PR #7068, opened 2021-12-10
Update Embedded Mode guide

bladey:
Currently builds fail if you run `yarn build`, there is stricter type checking now, which has been tested and added to the guide, needs general type review.

Screen Shot 2021-12-10 at 2 06 23 pm

Screen Shot 2021-12-10 at 2 06 37 pm


docs/pages/docs/walkthroughs/embedded-mode-with-sqlite-nextjs.mdx
```diff
@@ -220,51 +224,47 @@ Now add a `/post` subdirectory in `/pages` and include the code below in `[slug]
...
   const posts = await query.Post.findMany({
     query: `slug`,
   });
 
   const paths = posts
-    .map(post => post.slug)
-    .filter((slug): slug is string => !!slug)
-    .map(slug => `/post/${slug}`);
+    .map((post: Post) => post.slug)
+    .filter((slug: string): slug is string => !!slug)
```
dcousens, line 258:
```suggestion
    .filter((slug: string): slug is string => !!slug)
```
This is a redundant type guard

docs/pages/docs/walkthroughs/embedded-mode-with-sqlite-nextjs.mdx
```diff
@@ -220,51 +224,47 @@ Now add a `/post` subdirectory in `/pages` and include the code below in `[slug]
...
+    .map((slug: string) => `/post/${slug}`);
```
dcousens, line 259: This shouldn't be necessary

dcousens (changes requested):
The only change that should be required is the need to change `posts` from `any` to `Post[]`,  which should be possible by providing the Keystone context type information.

The other changes are not necessary.

PR #7070, opened 2021-12-10
Rewrite of Image and Files (breaking)

rohan-deshpande:
> **Warning**: This new feature includes breaking changes that may affect you

This pull request replaces `image` and `files` configuration options in Keystone's configuration, and replaces them with a new `storage` configuration object.

- :heavy_plus_sign: Amazon S3 (and other compatible providers) are now supported when uploading images and files
- :heavy_plus_sign: New Guide coming - see #7563 will be merged soon but check it out if you are keen
- :rotating_light: Images and files are now - **DELETED BY DEFAULT** from the underlying storage provider when replaced or deleted from the database
    - > **Note**: A `preserve` flag has been added to the new storage configurations to default back to the previous behaviour
- :warning: If you were previously using refs in your application, you need to migrate your database

### How to upgrade
Previously, your images and files were managed in your root Keystone configuration, under the fields `.images` and `.files` respectively.
The underlying storage mechanism was denominated through an `upload` field on each respectively, and you had no easy way to modify this behaviour for different fields.

It probably looked something like this
```ts
export default config({
  image: { upload: 'local' },
  lists: {
    Image: { fields: { image: image() } },
    /* ... */
  },
  /* ... */
});
```

As of this pull request, there is now a unified `.storage` configuration object that can contain multiple different storage providers, which you can optionally share or differentiate between for different fields as desired.
A canonical example for storing some images on the local filesystem might look something like the following:

```ts
export default config({
  storage: {
    my_avatars: {
      kind: 'local',
      type: 'image',
      generateUrl: path => `http://localhost:3000/avatars${path}`,
      serverRoute: { path: '/avatars' },
      storagePath: 'public/avatars',
    },
  },
  lists: {
    User: {
		fields: {
			avatar: image({ storage: 'my_avatars' })
		}
	 },
    // ...
  },
  // ...
});
```

A canonical example for storing some images on an S3 compatible storage provider might look something like the following:
```ts
export default config({
  storage: {
    my_avatars: {
      kind: 's3',
      type: 'image',
      bucketName: process.env.S3_BUCKET_NAME,
      region: process.env.S3_REGION,
      accessKeyId: process.env.S3_ACCESS_KEY_ID,
      secretAccessKey: process.env.S3_SECRET_ACCESS_KEY,
    },
  },
  lists: {
    User: {
		fields: {
			avatar: image({ storage: 'my_avatars' })
		}
	 },
	// ...
  },
  // ...
});
```

### Goodbye refs
The previous implementation of refs were useful for moving images between different lists using the admin interface, allowing a user to refer to same image or file without copying that duplicating that image or file at the storage layer.
Unfortunately, this creates a precarious and imperformant situation when you want or need to delete these images or file; you then need to undertake a form of ref-counting across your different lists in a custom hook to ensure you won't 404 the the underlying image or file.

This complexity didn't really pay off for the benefits it provided, when compared with typical Keystone relationships; and in fact the inability to use GraphQL to introspect these references was inhibitive.
The end result is images or files are left to orphan and/or duplicate themselves until some external reason results in the manual database and storage layer interventions.

With refs removed, the new recommendation for sharing images between lists and fields is an approach you may already be familiar with: [Keystone Relationships](https://keystonejs.com/docs/guides/relationships).
A canonical examle for sharing an image between multiple lists might look something like the following:

```ts
export default config({
  storage: {
    my_images: {
      // ...
    },
  },
  lists: {
    Image: { fields: { image: image({ storage: 'my_images' }) } },
    User: { fields: { avatar: relationship({ ref: 'Image' }) } },
    Post: { fields: { photos: relationship({ ref: 'Image', many: true }) } },
    // ...
  },
  // ...
});
```

Using Keystone relationships allows you to mirror the previous functionality while supporting the needs and features required when scaling up.

#### What if I can't?
If you can't migrate your schema to use Keystone relationships, but you still want to update your Keystone instance to the new `storage` configuration, you can use `preserve: true` to ensure that files - that are replaced or deleted from the database - are not deleted from the underlying storage layer.
However, similar to the previous behaviour prior to this pull request, this will result in images and files becoming orphans within the relevant filesystems as they are removed from the database.

##### Local images no longer need a baseUrl
If you were previously using `baseUrl` for your `upload: 'local'`, the configuration approach has changed to `serverRoute.path`.

For example
```diff
{
-   baseUrl: '/images'
+   serverRoute: {
+       path: '/images'
+   }
}
```


packages/keystone/src/lib/context/createImagesContext.ts
```diff
@@ -63,40 +75,62 @@ export function createImagesContext(
...
+        case 'cloud': {
+          const metadata = await cloudAssets().images.upload(stream, id);
+          return { mode, id, ...metadata };
+          break;
+        }
+        case 's3': {
+          const metadata = await s3Assets(s3).images.upload(stream, id);
 
-      const chunks = [];
+          console.log('METADATA', metadata);
```
dcousens, line 115:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -121,7 +121,63 @@ jobs:
...
+  field_crud_tests:
+    name: Field CRUD Tests
+    needs: should_run_tests
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:12
+        env:
+          POSTGRES_USER: keystone5
+          POSTGRES_PASSWORD: k3yst0n3
```
dcousens, line 141:
```suggestion
          POSTGRES_USER: testuser
          POSTGRES_PASSWORD: testpassword
```

.github/workflows/tests.yml
```diff
@@ -149,6 +206,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: keystone
           # this doesn't mean anything when we're using a custom s3 endpoint but the sdk wants something so we just give it something
           S3_REGION: us-east-1
+          DATABASE_URL: ${{ matrix.adapter == 'sqlite' && 'file:./dev.db' || 'postgres://keystone5:k3yst0n3@localhost:5432/test_db' }}
```
dcousens, line 213:
```suggestion
          DATABASE_URL: ${{ matrix.adapter == 'sqlite' && 'file:./dev.db' || 'postgres://testuser:testpassword@localhost:5432/test_db' }}
```

packages/core/src/lib/assets/createImagesContext.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  if (width === undefined || height === undefined) {
+    throw new Error('Height and width could not be found for image');
+  }
+  return { width, height, filesize: buffer.length, extension };
+}
+
+export function createImagesContext(
+  config: KeystoneConfig,
+  s3Assets: () => Map<string, AssetsAPI>
```
mitchellhamilton, line 32: The whole function thing just existed because the cloud stuff had to do async setup that wouldn't necessarily have been done by the time this was called. Could you just pass it directly?

packages/core/src/lib/assets/createImagesContext.ts
```diff
@@ -0,0 +1,132 @@
...
+          return { storage, id, ...metadata };
+        }
+      }
+
+      throw new Error(
+        `attempted to get data from stream for storage ${storageConfig}, however could not find the config for it`
+      );
+    },
+    deleteAtSource: async (storageString, id, extension) => {
+      let storageConfig = config.storage?.[storageString];
```
mitchellhamilton, line 112: Could we throw here when the storage isn't there like the others do?

packages/core/src/lib/context/createContext.ts
```diff
@@ -22,18 +22,26 @@ export function makeCreateContext({
...
   gqlNamesByList: Record<string, GqlNames>;
   lists: Record<string, InitialisedList>;
-  cloudAssetsAPI: () => CloudAssetsAPI;
 }) {
-  const images = createImagesContext(config, cloudAssetsAPI);
-  const files = createFilesContext(config, cloudAssetsAPI);
+  const s3AssetsAPI =
+    config.storage && Object.values(config.storage).find(({ kind }) => kind === 's3')
+      ? s3Assets(config.storage)
+      : undefined;
```
mitchellhamilton, line 36: Could we not avoid doing things conditionally based on having 1 or more s3 assets vs 0?
```suggestion
  const s3AssetsAPI = s3Assets(config.storage);
```

packages/core/src/types/next-fields.ts
```diff
@@ -14,6 +14,9 @@ export type ListGraphQLTypes = { types: GraphQLTypesForList };
 export type FieldData = {
   lists: Record<string, ListGraphQLTypes>;
   provider: DatabaseProvider;
+  assets: {
+    getStorage: (storage: string) => StorageKind | undefined;
```
mitchellhamilton, line 18: This returns the storage config not the kind, right?

packages/core/src/types/config/index.ts
```diff
@@ -19,6 +19,35 @@ import type { BaseFields } from './fields';
...
+      removeFileOnDelete?: boolean;
+    }
+  | {
+      kind: 's3';
+      type: 'file' | 'image';
+      signed?: { expiry: number };
+      proxied?: {
+        baseUrl: string;
+        generate?: boolean; // if false, allow proxied requests but return urls directly to s3
+      };
```
mitchellhamilton, line 37: Is this stuff being implemented in this PR or later?

packages/core/src/lib/assets/createImagesContext.ts
```diff
@@ -0,0 +1,132 @@
...
+  return { width, height, filesize: buffer.length, extension };
+}
+
+export function createImagesContext(
+  config: KeystoneConfig,
+  s3Assets: () => Map<string, AssetsAPI>
+): ImagesContext | undefined {
+  if (!config.storage) {
+    return;
+  }
```
mitchellhamilton, line 36: I kinda feel like we should just always have the image and file stuff on the context rather than now if there's 1 or more configs now that there's N configs rather than 1 or 0 like before, thoughts?

packages/core/src/types/config/index.ts
Noviny: Plan is still this PR - got some base bits and will try and push tonight.

packages/core/src/types/next-fields.ts
Noviny: This `getStorage` function returns a storageKind, and is used for looking up the full info of storage.

Previously here we had a `getStorageKind` function that would get `s3` or `local` but since we occasionally needed more than that we just get the whole object.

mitchellhamilton: Oh, sorry, I expected `StorageKind` to be `'s3' | 'local'` based on the name. Makes sense

packages/core/src/lib/assets/createImagesContext.ts
Noviny: I still like having this for creating the local files, and for making this relatively long code block have its own file

mitchellhamilton: Oh, yeah, I don't mean moving anything, just changing `s3Assets: () => Map<string, AssetsAPI>` to `s3Assets: Map<string, AssetsAPI>`

packages/core/src/types/next-fields.ts
dcousens: @mitchellhamilton the naming of the types for this probably needs a bike-shed moment

examples-staging/assets-local/keystone.ts
```diff
@@ -7,10 +7,26 @@ export default config({
     url: process.env.DATABASE_URL || 'file:./keystone-example.db',
   },
   lists,
-  images: {
-    upload: 'local',
-  },
-  files: {
-    upload: 'local',
+  storage: {
+    images: {
```
dcousens, line 11:
```suggestion
    my_images: {
```

dcousens, line 11:
```suggestion
    my_images: {
```

examples-staging/assets-local/keystone.ts
```diff
@@ -7,10 +7,26 @@ export default config({
...
+      kind: 'local',
+      type: 'image',
+      generatedUrl: path => `http://localhost:3000/images${path}`,
+      addServerRoute: {
+        path: '/images',
+      },
+      storagePath: 'public/images',
+      preserve: false,
+    },
+    files: {
```
dcousens, line 21:
```suggestion
    my_files: {
```

examples-staging/assets-local/schema.ts
```diff
@@ -15,8 +15,8 @@ export const lists = {
       content: text(),
       publishDate: timestamp(),
       author: relationship({ ref: 'Author.posts', many: false }),
-      hero: image(),
-      attachment: file(),
+      hero: image({ storage: 'images' }),
+      attachment: file({ storage: 'files' }),
```
dcousens, line 19:
```suggestion
      hero: image({ storage: 'my_images' }),
      attachment: file({ storage: 'my_files }),
```

examples-staging/assets-local/keystone.ts
```diff
@@ -7,10 +7,26 @@ export default config({
...
-  images: {
-    upload: 'local',
-  },
-  files: {
-    upload: 'local',
+  storage: {
+    my_images: {
+      kind: 'local',
+      type: 'image',
+      generatedUrl: path => `http://localhost:3000/images${path}`,
```
dcousens, line 14:
```suggestion
      generateUrl: path => `http://localhost:3000/images${path}`,
```


This name isn't amazing

examples-staging/assets-s3/keystone.ts
```diff
@@ -0,0 +1,47 @@
...
+    my_images: {
+      kind: 's3',
+      type: 'image',
+      bucketName,
+      region,
+      accessKeyId,
+      secretAccessKey,
+      // proxied: { baseUrl: '/images-proxy' },
+      endpoint: 'http://127.0.0.1:9000/',
+      forcePathStyle: true,
```
dcousens, line 29:
```suggestion
      forcePathStyle: true, // if false, your endpoint string will default to using buckets as subdomains, for example `http://keystone-test.127.0.0.1:9000/'`
```

examples-staging/basic/keystone.ts
```diff
@@ -34,8 +34,28 @@ export default auth.withAuth(
...
+        kind: 'local',
+        type: 'image',
+        storagePath: 'public/images',
+        generatedUrl: path => `http://localhost:3000/images${path}`,
+        addServerRoute: {
+          path: '/images',
+        },
+        preserve: false,
+      },
+      files_basic: {
```
dcousens, line 46:
```suggestion
      my_images: {
        kind: 'local',
        type: 'image',
        storagePath: 'public/images',
        generatedUrl: path => `http://localhost:3000/images${path}`,
        addServerRoute: {
          path: '/images',
        },
        preserve: false,
      },
      my_files: {
```

examples-staging/basic/schema.ts
```diff
@@ -45,8 +45,8 @@ const User: Keystone.Lists.User = list({
     /** Email is used to log into the system. */
     email: text({ isIndexed: 'unique', validation: { isRequired: true } }),
     /** Avatar upload for the users profile, stored locally */
-    avatar: image(),
-    attachment: file(),
+    avatar: image({ storage: 'images_basic' }),
+    attachment: file({ storage: 'files_basic' }),
```
dcousens, line 49:
```suggestion
    avatar: image({ storage: 'my_images' }),
    attachment: file({ storage: 'my_files' }),
```

examples-staging/basic/keystone.ts
```diff
@@ -34,8 +34,24 @@ export default auth.withAuth(
...
+    storage: {
+      images_basic: {
+        kind: 'local',
+        storagePath: 'public/images',
+        generatedUrl: path => `http://localhost:3000/images${path}`,
+        addServerRoute: {
+          path: '/images',
+        },
+      },
+      files_basic: {
```
dcousens, line 46:
```suggestion
      my_files: {
```

examples-staging/basic/keystone.ts
```diff
@@ -34,8 +34,24 @@ export default auth.withAuth(
       // path: '/admin',
       // isAccessAllowed,
     },
-    images: { upload: 'local' },
-    files: { upload: 'local' },
+    storage: {
+      images_basic: {
```
dcousens, line 38:
```suggestion
      my_images: {
```

packages/core/package.json
```diff
@@ -90,6 +93,8 @@
     "execa": "^5.1.1",
     "express": "^4.17.1",
     "fast-deep-equal": "^3.1.3",
+    "fast-glob": "^3.2.7",
+    "file-type": "^16.5.3",
```
Noviny, line 97:
```suggestion
-    "file-type": "^16.5.3",
```

Looks like file-type isn't being used anywhere in the codebase

packages/core/src/fields/types/file/index.ts
```diff
@@ -1,128 +1,117 @@
...
-        mode: { kind: 'scalar', scalar: 'String', mode: 'optional' },
         filename: { kind: 'scalar', scalar: 'String', mode: 'optional' },
       },
     })({
       ...config,
+      hooks: storage.preserve
+        ? config.hooks
+        : {
+            ...config.hooks,
+            async beforeOperation(args) {
```
Noviny, line 77: @dcousens you talked about doing this as `afterOperations` in some code channel spitballing - was there a conclusion there?

packages/core/src/fields/types/file/index.ts
```diff
@@ -1,128 +1,117 @@
...
+    config: FileFieldConfig<ListTypeInfo>
   ): FieldTypeFunc<ListTypeInfo> =>
-  () => {
-    if ((config as any).isIndexed === 'unique') {
+  meta => {
+    const storage = meta.assets.getStorage(config.storage);
+
+    if (!storage) {
+      throw new Error(
+        `${meta.listKey}.${meta.fieldKey} has storage set to ${config.storage} but there is no storage config under that key`
```
dcousens, line 57:
```suggestion
        `${meta.listKey}.${meta.fieldKey} has storage set to ${config.storage}, but no storage configuration was found for  that key`
```

packages/core/src/fields/types/file/index.ts
```diff
@@ -1,128 +1,117 @@
...
-            filename === null ||
-            mode === null ||
-            (mode !== 'local' && mode !== 'cloud')
-          ) {
+        resolve({ value: { filesize, filename } }) {
+          if (filesize === null || filename === null) {
             return null;
           }
-          return { mode, filename, filesize };
+          return { filename, filesize, storage: config.storage };
```
Noviny, line 112: @mitchellhamilton you had a reason to return storage here yeah?

docs/pages/docs/apis/config.mdx
```diff
@@ -396,29 +396,54 @@ See the [schema extension guide](../guides/schema-extension) for more details on
...
+  storage: {
+    my_images: {
+      kind: 'local',
+      generatedUrl: path => `http://localhost:3000/images${path}`,
+      serverRoute: {
+        path: '/images',
+      },
+      storagePath: 'public/images',
+    },
+    my_files: {
```
Noviny, line 435: @borisno2 you good if we only provide examples of one storage option for this just to reduce code length or do you think having two is important?

borisno2: Hey @Noviny
Yeah Maybe we can cut this down a bit (ie one example of each `kind` ?) We can then expand on this in a more in-depth guide. I'll have a go at cutting this down this morning.

borisno2: Just pushed a new example @Noviny with both 'kind`s in one example - do you think makes sense?

.changeset/unlucky-ducks-raise.md
```diff
@@ -0,0 +1,122 @@
+---
+'@keystone-6/core': major
+---
+
+Large changes to the files and images fields.
+
```
dcousens, line 6:
```suggestion
```

.changeset/unlucky-ducks-raise.md
```diff
@@ -0,0 +1,122 @@
...
+```ts
+export default config({
+    image: { upload: 's3' },
+    s3: {
+        bucketName: S3_BUCKET_NAME,
+        region: S3_REGION,
+        accessKeyId: S3_ACCESS_KEY_ID,
+        secretAccessKey: S3_SECRET_ACCESS_KEY
+    },
```
dcousens, line 24:
```suggestion
```

packages/core/src/fields/types/file/index.ts
dcousens: Needed for the URL resolver

dcousens (commented):
A few changes

packages/core/src/lib/server/createExpressServer.ts
```diff
@@ -104,18 +108,46 @@ export const createExpressServer = async (
...
-    expressServer.use(
-      '/files',
-      express.static(config.files.local?.storagePath ?? DEFAULT_FILES_STORAGE_PATH)
-    );
-  }
+  if (config.storage) {
+    for (const val of Object.values(config.storage)) {
+      if (!val.serverRoute) continue;
+      if (val.kind === 'local') {
+        expressServer.use(val.serverRoute.path, express.static(val.storagePath));
```
dcousens, line 115: For files, we need to set `Content-Type: application/octet-stream`

packages/core/src/lib/server/createExpressServer.ts
```diff
@@ -104,18 +108,46 @@ export const createExpressServer = async (
...
+      } else if (val.kind === 's3') {
+        const endpoint = getS3AssetsEndpoint(val);
+        expressServer.use(`${val.serverRoute.path}/:id`, async (req, res) => {
+          const s3Url = new URL(endpoint);
+          s3Url.pathname += `${s3Url.pathname.endsWith('/') ? '' : '/'}${val.pathPrefix || ''}${
```
dcousens, line 120: Move this out

packages/core/src/types/config/index.ts
```diff
@@ -195,7 +262,7 @@ export type FilesConfig = {
   };
 };
 
-// config.images
+// config.storage
```
dcousens, line 260:
```suggestion
```

packages/core/src/types/config/index.ts
```diff
@@ -176,7 +243,7 @@ export type GraphQLConfig = {
 
 export type ExtendGraphqlSchema = (schema: GraphQLSchema) => GraphQLSchema;
 
-// config.files
+// config.storage
```
dcousens, line 241:
```suggestion
```

packages/core/src/types/config/index.ts
```diff
@@ -27,8 +90,14 @@ export type KeystoneConfig<TypeInfo extends BaseKeystoneTypeInfo = BaseKeystoneT
...
   graphql?: GraphQLConfig;
   extendGraphqlSchema?: ExtendGraphqlSchema;
-  files?: FilesConfig;
-  images?: ImagesConfig;
+  /** An object containing configuration about keystone's various external storages.
+   *
+   * Each entry should be of either `kind: 'local'` or `kind: 's3'`, and follow the configuration of each.
+   *
+   * When configuring a `file` or `image` field that uses the storage, use the key in the storage object
+   * as the `storage` config option for the field.
```
dcousens, line 93:
```suggestion
   * as the `storage` option for that field.
```

packages/core/src/lib/assets/local.ts
```diff
@@ -0,0 +1,58 @@
...
+    async url(id, extension) {
+      return storageConfig.generateUrl(`/${id}.${extension}`);
+    },
+    async upload(stream, id, extension) {
+      const buffer = await streamToBuffer(stream);
+
+      await fs.writeFile(path.join(storageConfig.storagePath, `${id}.${extension}`), buffer);
+    },
+    async delete(id, extension) {
+      fs.unlink(path.join(storageConfig.storagePath, `${id}.${extension}`));
```
mitchellhamilton, line 23:
```suggestion
      await fs.unlink(path.join(storageConfig.storagePath, `${id}.${extension}`));
```

packages/core/src/lib/assets/local.ts
```diff
@@ -0,0 +1,58 @@
...
+        await pipeStreams;
+        const { size: filesize } = await fs.stat(path.join(storageConfig.storagePath, filename));
+        return { filesize, filename };
+      } catch (e) {
+        await fs.remove(path.join(storageConfig.storagePath, filename));
+        throw e;
+      }
+    },
+    async delete(filename) {
+      fs.unlink(path.join(storageConfig.storagePath, filename));
```
mitchellhamilton, line 55:
```suggestion
      await fs.unlink(path.join(storageConfig.storagePath, filename));
```

.changeset/unlucky-ducks-raise.md
```diff
@@ -0,0 +1,111 @@
...
+    /* ... */
+})
+```
+
+This allows mirroring of the old functionality, while allowing us to add the below feature/breaking change.
+
+##### Images and Files will now be deleted when deleted
+
+Before this change, if you uploaded a file or image, Keystone would never remove it from where it was stored. The inability to tidy up unused
+files or images was unwelcome. With the removal of `ref`, we can no remove things from the source, and this will be done by default.
```
borisno2, line 79:
```suggestion
files or images was unwelcome. With the removal of `ref`, we can now remove things from the source, and this will be done by default.
```

emmatown (approved):

boyboy1995:
Hi, is this file config written in the right way to using s3 storage for file field?

Screenshot 2022-06-08 at 16 03 31

Andrey-Bazhanov:
looks like this is not released yet... documentation is updated but @keystonejs/core - not

PR #7076, opened 2021-12-13
Update virtual-fields.mdx

Noviny:
Was reading through the virtual example, and came across an error an a syntax weirdness so made some quick fixes.


dcousens (approved):

Noviny:
Replacing with less weird branch

PR #7091, opened 2021-12-17
add titles to pages in admin UI

Noviny:
Found that we didn't have titles on Keystone pages, so it just showed the URL path.

This would have been fine, EXCEPT the title didn't get updated on Next.js page transitions, meaning it was often misleading, and kind of annoying if you had multiple tabs open.

PR defaults it to being set to `Keystone` which will eliminate confusion.
- Item view pages use the item's label or id as 'Keystone - label/id'
- List view pages use the list's plural name: `Keystone - People' eg

Sign in and init both have custom titles

### DIFF
--- a/.changeset/moody-pianos-sell.md
+++ b/.changeset/moody-pianos-sell.md
@@ -1,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Add titles to initi and sign-in pages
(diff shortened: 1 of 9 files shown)


.changeset/moody-pianos-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-next/auth': patch
+---
+
+Add titles to initi and sign-in pages
```
dcousens, line 5:
```suggestion
Add titles to init and sign-in pages
```

dcousens (approved):

PR #7112, opened 2021-12-22
Update embedded tutorial docs

bladey:
The tutorial for the embedded Next.js example doesn't build due to types failing, this adds refined types from @dcousens and @bladey.


dcousens (approved):
LGTM, but this should be an example too

PR #7119, opened 2021-12-24
fixed typo in password and text field validations

moselhy:
* was showing "must be no longer than x.min" not x.max for longer inputs


dcousens (approved):
Thanks @moselhy!  Good catch

PR #7136, opened 2022-01-01
toggle popover open on triggerRenderer click

moselhy:
* as opposed to only opening, now it opens/closes


gwyneplaine:
thanks for this @moselhy  🎉

PR #7148, opened 2022-01-04
Update Community numbers

raveling:


dcousens:
@raveling how is this number determined?

raveling:
> @raveling how is this number determined?

@dcousens the `#announcements` channel in Keystone community Slack

All members are added to that channel by default. We're technically at 2600 as of today, I went with a round number for the website.

PR #7188, opened 2022-01-12
Add option to use BigInt in autoincrement idField

MurzNN:
Now for `autoincrement` idField only Integer field type is used, which sometimes is not enough.

In this PR I've implemented an option "useBigInt" in `idField` configuration, that switches to use BigInt as idField in database. Here is an example how to enable this feature:
```js
    db: {
      idField: {
        kind: 'autoincrement',
        useBigInt: true,
      }
    }
```

Related discussion: https://github.com/keystonejs/keystone/discussions/6969

### DIFF
--- a/.changeset/honest-lamps-bow.md
+++ b/.changeset/honest-lamps-bow.md
@@ -1,5 +1,6 @@
 ---
 '@keystone-6/core': minor
+'@keystone-6/auth': patch
 ---
 
 Added support for BigInt autoincrement id fields with `idField: { kind: 'autoincrement', type: 'BigInt' }`
--- a/.changeset/honest-lamps-bow.md
+++ b/.changeset/honest-lamps-bow.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Ability to use BigInt as autoincrement idField
--- a/.changeset/honest-lamps-bow.md
+++ b/.changeset/honest-lamps-bow.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': minor
 ---
 
-Ability to use BigInt as autoincrement idField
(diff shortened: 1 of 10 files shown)


dcousens:
Thinking about this,  I think we would want to add a `db.idField.type` as the discriminating factor instead of `useBigInt`.

For example
```
	db: {
		idField: {
			kind: 'autoincrement',
			type: 'Int' | 'BigInt' // the underlying Prisma type, the default is 'Int' when `kind: 'autoincrement'`
		}
	}
```


This will then be aligned with other advanced changes we intend to add, for example with the new [MySQL ](https://github.com/keystonejs/keystone/pull/7538) changes.

dcousens:
Thinking about this,  I think we would want to add a `db.idField.type` as the discriminating factor instead of `useBigInt`.

For example
```
	db: {
		idField: {
			kind: 'autoincrement',
			type: 'Int' | 'BigInt' // the underlying Prisma type, the default is 'Int' when `kind: 'autoincrement'`
		}
	}
```


This will then be aligned with other advanced changes we intend to add, for example with the new [MySQL ](https://github.com/keystonejs/keystone/pull/7538) changes.

MurzNN:
@dcousens yeah, adding type will simplify the whole construction, but will require more changes in core, but seems it almost implemented already, thanks!

.changeset/honest-lamps-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': minor
```
mitchellhamilton, line 2:
```suggestion
'@keystone-6/core': minor
'@keystone-6/auth': patch
```

mitchellhamilton (approved):

PR #7204, opened 2022-01-19
Add select and include to findOne and findMany

### LINKED ISSUE #7198: Missing some relationship fields in Database API find query results
1. Start the `examples/blog` project, fill with some data.
2. In `onConnect` function write two queries:
```
      console.log(await context.db.Post.findMany({ take: 1 }));
      console.log(await context.db.Author.findMany({ take: 1 }));
```
3. See in console that in `Author` list item the `posts` relationship field is missing.
But in `Post` the `author` relationship field is present.

Will be good to describe what rule does depend on this, maybe it is simple like "only single relation relationship fields are loaded"?

So please add this nuance to the documentation here https://keystonejs.com/docs/apis/db-items

MurzNN:
Now `findOne` and `findMany` functions in DB API misses ability to select specific fields only, and include relationship fields too.

It is my try to add this missing functionality.

Prisma allows query specific fields, here is the documentation https://www.prisma.io/docs/concepts/components/prisma-client/select-fields#select-specific-fields

So seems we simply can pass those variables from Keystone API to Prisma API.

Here is usage example of this feature directly in Prisma, based on `examples/blog`:
```js
    async onConnect(context) {
      const result = await context.prisma.author.findMany({
        take: 2,
        select: {
          name: true,
          posts: {
            select: {
              id: true,
            },
          },
        },
      });
      console.log(result[0]);
```

And here is desired behavior in Keystone DB API:
```js
      const result = await context.db.Author.findMany({
        take: 2,
        select: {
          name: true,
          posts: {
            select: {
              id: true,
            },
          },
        },
      });
      const result2 = await context.db.Author.findMany({
        take: 2,
        include: {
          posts: true,
        },
      });
```

Resolves https://github.com/keystonejs/keystone/issues/7198

### DIFF
--- a/.changeset/afraid-timers-divide.md
+++ b/.changeset/afraid-timers-divide.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Add select and include options to findOne and findMany DB API calls
--- a/packages/keystone/src/lib/core/queries/resolvers.ts
+++ b/packages/keystone/src/lib/core/queries/resolvers.ts
@@ -13,6 +13,7 @@ import { limitsExceededError, userInputError } from '../graphql-errors';
 import { InitialisedList } from '../types-for-lists';
 import { getDBFieldKeyForFieldOnMultiField, runWithPrisma } from '../utils';
 import { checkFilterOrderAccess } from '../filter-order-access';
+import { Prisma } from '@prisma/client';
 
 // we want to put the value we get back from the field's unique where resolver into an equals
 // rather than directly passing the value as the filter (even though Prisma supports that), we use equals
@@ -77,7 +78,7 @@ export async function accessControlledFilter(
 }
 
 export async function findOne(
-  args: { where: UniqueInputFilter },
+  args: { where: UniqueInputFilter; select: Prisma.UserSelect; include: Prisma.UserInclude },
   list: InitialisedList,
   context: KeystoneContext
 ) {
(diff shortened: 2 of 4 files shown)


MurzNN:
The problem now only is that Keystone's Prisma somehow misses the `Prisma` type, so this line:
```
import { Prisma } from '@prisma/client';
```
throws an error.

fkrauthan:
This would be a great feature for some more heavy processing in some custom mutations for example.

dcousens:
@mitchellhamilton if this was  _only_ on `find*` functions, wouldn't we be able to ignore hooks, and only the return type would need to be adjusted?

dcousens:
I think we can add this functionality with the caveat that no `read` hooks could exist for the underlying item, but maybe it can for it's fields?

Alternatively, we can change the type of `item` passed into `read` hooks to be `Partial<typeof item>`, to indicate that hooks could receive anything.

_Alternatively_, maybe this is an opt-in behaviour, with something like `db.selectable: true` which will change the types of `item` to `Partial<typeof item>`, that way it's non-breaking, but additionally not painful in the typical scenario? :thinking:

mariomnts:
Is there any plan to support this or even only `select` the fields that are being requested in the graphql query?

dcousens:
@mariomnts plans yes, but how we integrate this type of functionality with hooks is a problem that needs a resolution.

mariomnts:
> @mariomnts plans yes, but how we integrate this type of functionality with hooks is a problem that needs a resolution.

That makes sense.

In my particular use case I don't use hooks at all in any part of the app so something that comes to my mind that if hooks are not defined a possible optimization is just getting the requested fields.

dcousens:
@mariomnts indeed, this what I suggested previously

> I think we can add this functionality with the caveat that no read hooks could exist for the underlying item or fields...
(https://github.com/keystonejs/keystone/pull/7204#issuecomment-1736438736)

mariomnts:
> @mariomnts indeed, this what I suggested previously
> 
> > I think we can add this functionality with the caveat that no read hooks could exist for the underlying item or fields...
> > ([#7204 (comment)](https://github.com/keystonejs/keystone/pull/7204#issuecomment-1736438736))

@dcousens sorry I missed that comment, that sounds great then. @MurzNN do you plan to continue working on this PR?

PR #7231, opened 2022-01-27
Improving and expanding docs on how to use relationship filters in GraphQL

molomby:
This fills a reasonably large hole in our docs around how to use and filter on relationships in the GraphQL API.

Comes out of discussions [in Slack](https://keystonejs.slack.com/archives/C01STDMEW3S/p1640679267492200).

### DIFF
--- a/docs/pages/docs/guides/filters.mdx
+++ b/docs/pages/docs/guides/filters.mdx
@@ -43,7 +43,7 @@ Keystone provides a wide range of different filters. If we want to find all thos
 
 ```graphql
 {
-  tasks(where: { label: { NOT: { equals: "Hello" } } }) {
+  tasks(where: { label: { not: { equals: "Hello" } } }) {
     id
     label
   }
--- a/docs/pages/docs/guides/filters.mdx
+++ b/docs/pages/docs/guides/filters.mdx
@@ -217,7 +217,7 @@ This indicate that no conditions should be applied to tasks so all tasks are mat
 
 #### None
 
-The `none` filter returns true when at there are zero related items that match the conditions supplied.
+The `none` filter returns true when there are zero related items that match the conditions supplied.
 
 For example, if we had an urgent task we were looking to assign, we could query for people who didn't have any task that were both incomplete and high-priority:
 
@@ -230,7 +230,7 @@ For example, if we had an urgent task we were looking to assign, we could query 
 }
 ```
 
-Or, if we wanted people with no assigned tasks at all (not even any completed ones), we could pass `none` an empty object:
+Or, if we wanted people with no assigned tasks at all (not even completed tasks), we could pass `none` as an empty object:
 
 ```graphql
 {
@@ -245,7 +245,7 @@ Or, if we wanted people with no assigned tasks at all (not even any completed on
 
 The `every` filter returns true when all related items match the conditions supplied.
 
-For example, if we wanted to find people who have completed all their assigned tasks, we could write:
+For example, if we wanted to find people who have completed their assigned tasks, we could write:
 
 ```graphql
 {
@@ -257,7 +257,7 @@ For example, if we wanted to find people who have completed all their assigned t
 ```
 
 Note the results will also include people with no tasks assigned at all.
-If you wanted people who have completed all their assigned tasks and actually have at least one task assigned, you'd could add a `some` filter:
+If you only wanted people who have both completed their assigned tasks _and_ actually have at least one task assigned, you could add a `some` filter:
 
 ```graphql
 {
--- a/docs/pages/docs/guides/filters.mdx
+++ b/docs/pages/docs/guides/filters.mdx
@@ -43,7 +43,7 @@ Keystone provides a wide range of different filters. If we want to find all thos
 
 ```graphql
 {
-  tasks(where: { label: { not: { equals: "Hello" } } }) {
+  tasks(where: { label: { NOT: { equals: "Hello" } } }) {
     id
     label
   }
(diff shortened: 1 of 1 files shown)


docs/pages/docs/guides/filters.mdx
```diff
@@ -43,7 +43,7 @@ Keystone provides a wide range of different filters. If we want to find all thos
 
 ```graphql
 {
-  tasks(where: { label: { not: { equals: "Hello" } } }) {
+  tasks(where: { label: { NOT: { equals: "Hello" } } }) {
```
mitchellhamilton, line 46: This change isn't correct, the previous casing is correct in this case
```suggestion
  tasks(where: { label: { not: { equals: "Hello" } } }) {
```

molomby: Oh you're right. Wow, that's pretty confusing, no? :/

molomby: Good catch though, thanks

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,167 @@ For example, to find any tasks that have no assigned user, we can run the follow
 }
 ```
 
-### Many
+Conversely, if we wanted tasks where any user was assigned, we can invert the above using the `NOT` operator:
 
-If you have `many: true` configured on the relationship field, then you can find items based on whether `some`, `none`, or `every` of the related items match a `where` filter using the fields from the related list.
+```graphql
+{
+  tasks(where: { NOT: { assignedTo: null } }) {
```
dcousens, line 179:
```suggestion
  tasks(where: { not: { assignedTo: null } }) {
```

dcousens (approved):

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,167 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+  }
+}
+```
+
+### Filtering Related Items
+
+In the relationship examples above we've been filtering items based on their related items but we haven't been loading any information about the related items themselves.
+What makes GraphQL so powerful that it gives developers the ability to write queries that retrieve data from across a graph of interconnected objects.
+In Keystone, relationships form the linkages between objects this graph.
+Queries that traverse these links apply filters both the top-level "list" fields (ie. `tasks` and `people` in these examples) and when selecting data about related items themselves.
```
dcousens, line 300:
```suggestion
In the relationship examples above we have been filtering items based on their related items, but we haven't been loading any information about the related items themselves.
GraphQL gives us the ability to write queries that retrieve data across a graph of interconnected items - in Keystone, relationships form the links between these items.
Queries that traverse these links apply filters both the top-level "list" fields (ie. `tasks` and `people` in these examples) and when selecting data about related items themselves.
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,167 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+```graphql
+{
+  people(where: { tasks: { every: { isComplete: { equals: true } } } }) {
+    id
+    name
+  }
+}
+```
+
+Note the results will also include people with no tasks assigned at all.
```
dcousens, line 259:
```suggestion
Note the results will also include `people` with no tasks assigned at all.
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,167 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+If you wanted people who have completed all their assigned tasks and actually have at least one task assigned, you'd could add a `some` filter:
```
dcousens, line 260:
```suggestion
If you wanted `people` who have completed all their assigned tasks and actually have at least one task assigned, you'd could add a `some` filter:
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,166 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+    id
+    name
+  }
+}
+```
 
-For example, to find all the people which have `some` posts with the label `"Hello"`, we can run the following query:
+#### None
+
+The `none` filter returns true when at there are zero related items that match the conditions supplied.
```
dcousens, line 220:
```suggestion
The `none` filter returns true when there are zero related items that match the conditions supplied.
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,166 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
 {
-  people(where: { tasks: { some: { label: { equals: "Hello" } } } }) {
+  people(where: { tasks: { none: { priority: { equals: high }, isComplete: { equals: false } } } }) {
     id
     name
   }
 }
 ```
 
+Or, if we wanted people with no assigned tasks at all (not even any completed ones), we could pass `none` an empty object:
```
dcousens, line 233:
```suggestion
Or, if we wanted people with no assigned tasks at all (not even completed tasks), we could pass `none` an empty object:
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,166 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+    name
+  }
+}
+```
+
+#### Every
+
+The `every` filter returns true when all related items match the conditions supplied.
+
+For example, if we wanted to find people who have completed all their assigned tasks, we could write:
```
dcousens, line 248:
```suggestion
For example, if we wanted to find people who have completed all of their assigned tasks, we could write:
```

docs/pages/docs/guides/filters.mdx
```diff
@@ -171,21 +172,166 @@ For example, to find any tasks that have no assigned user, we can run the follow
...
+{
+  people(where: { tasks: { every: { isComplete: { equals: true } } } }) {
+    id
+    name
+  }
+}
+```
+
+Note the results will also include people with no tasks assigned at all.
+If you wanted people who have completed all their assigned tasks and actually have at least one task assigned, you'd could add a `some` filter:
```
dcousens, line 260:
```suggestion
If you only wanted people who have both completed their assigned tasks _and_ actually have at least one task assigned, you could add a `some` filter:
```

PR #7232, opened 2022-01-27
Adding an intro to the GraphQL API docs

molomby:
Adding an intro to the GraphQL API docs with basic info and context, like where to find the endpoint and what to do with it.

Not 100% sure this is the right place for this content (possibly should be part of a more general getting started-type guide?) but I feel that giving beginner devs some basic context here is helpful. Not everyone has used GraphQL before and our docs should probably do a better job of orienting people within the ecosystem.

### DIFF
--- a/docs/pages/docs/apis/graphql.mdx
+++ b/docs/pages/docs/apis/graphql.mdx
@@ -11,16 +11,13 @@ Keystone generates a CRUD (create, read, update, delete) GraphQL API based on th
 
 ## Using The API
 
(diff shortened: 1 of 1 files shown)


Noviny:
I like the things and would be happy to approve. One comment is I'd love to split out 'here is how it works (not mentioning options) and then do a second section on customising the defaults, as they feel like they appeal to different people, and reading about the default case feels weird in that moment

molomby:
Yeah this content is a bad fit for the API docs. So we should establish a "Working with GraphQL" guide page (or similar) and put this content there? We do get a lot of general questions about GraphQL in community.

How does this fit with the other doc holes, eg..

> A guide on a very basic calling to the graphql API from a frontend to show how this works

Same doc?

Noviny (approved):

PR #7234, opened 2022-01-28
Add documentation for using the Keystone GraphQL API from an example frontend

Noviny:
This has two things, one of which will be removed:

- The document that has a guide on setting up a frontend with keystone
- the KS example that this builds

- [x] move the KS example into its own repo so it can be referenced
- [ ] Another editing pass to make sure things make sense


examples/task-manager/seed-data/data.ts
```diff
@@ -44,28 +44,24 @@ export const tasks = [
   {
     label: 'Query content over GraphQL🔎',
     isComplete: false,
-    finishBy: '2021-03-21T05:41:37.000Z',
```
dcousens, line 47: Why removing these?

Noviny: Mostly because they weren't being used - when they are used, they should be future dates, so this existing string doesn't make sense.

Issue #7230, opened 2022-01-26
Missing license SPDX for @keystone-6/document-renderer & @keystone-6/fields-document


alex-boutin-ingeno:
@keystone-6/auth and @keystone-6/core 
mention the MIT license in `package.json` but it is not mentioned in @keystone-6/document-renderer and @keystone-6/fields-document. I believe it should be included for consistency.

bladey:
Thanks @alex-boutin-ingeno, I'll run this by the team.

dcousens:
Conceptually,  we'd want the license at the root to be distributed automatically,  but short of over-complicating the publish process,  I think we'll simply add the `LICENSE` to each sub package.

raveling:
https://github.com/keystonejs/keystone/pull/7244

dcousens:
Resolved in #7244

PR #7244, opened 2022-01-30
Added license info to Keystone packages

raveling:
This adds the SPDX identifier and the license text to each of the monorepo's packages


Noviny:
https://github.com/keystonejs/keystone/blob/main/LICENSE

@dcousens do we want to add this in to our packages, or do you reckon the license field is enough?

bladey:
https://github.com/keystonejs/keystone/issues/7230

dcousens (approved):

PR #7249, opened 2022-01-31
Improved `context.query` types

emmatown:
This types the return types of `context.query` based on the GraphQL return types rather than being `Record<string, any>`. All the fields are optional because the types do not know what fields were selected in the `query` field. The generated output types are in an `OutputTypes` namespace, we might want to move the input types into an `InputTypes` namespace but I have not done that here. There's also an example change because it had to change because of the type changes but I ended up changing how it works because previously, it was doing Bad things.


dcousens (approved):
LGTM

PR #7252, opened 2022-01-31
rename keystone folder to core

Noviny:
When we renamed the 'core' package from `keystone` to `core`, we didn't rename the folder. Let's do that


gwyneplaine (approved):

mitchellhamilton:
https://github.com/keystonejs/keystone/blob/48f3cb23649166ee3fd78caf1f32fc2657d376ad/prisma-utils/src/index.ts#L109 needs to change

mitchellhamilton:
Could you search for `packages/keystone` in the repo, looks like there are a couple other things?

dcousens:
Do we consider this a breaking change or at least something to mention in a CHANGESET?

Noviny:
This change has no impact on the published packages, or things on NPM, and is just a change for the repository. Like changing our linting rules isn't a change, I don't think this is.

NB: This renames the folder, but the package name change went out pre-GA

PR #7264, opened 2022-02-07
Move relationship config for relationships in component blocks

emmatown:
The rationale behind this to remove the current awkward config with relationships in component blocks, especially so that we don't inherit it for components outside the document editor. See the diff in the docs changes for a before -> after of configuring it.

I also ended up adding docs for child fields and early validation for checking that relationships in the document field(both inline relationships and props in component blocks) refer to lists that actually exist

### DIFF
--- a/docs/pages/docs/guides/document-fields.mdx
+++ b/docs/pages/docs/guides/document-fields.mdx
@@ -172,7 +172,6 @@ export default config({
         content: document({
           relationships: {
             mention: {
-              kind: 'inline',
               listKey: 'Author',
               label: 'Mention',
               selection: 'id name',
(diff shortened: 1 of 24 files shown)


docs/pages/docs/guides/document-fields.mdx
```diff
@@ -568,42 +617,19 @@ fields.object({
...
-You can reference the key of the relationship in the relationship and in the form, it will render a relationship select like the relationship field on lists.
+To use relationship fields on component blocks, you need to add a relationship field and provide a list key, label and options selection.
+In the form, it will render a relationship select like the relationship field on lists.
+Similarly to inline relationships, they will be hydrated with this selection if `hydrateRelationships: true` is provided when fetching.
 
 ```tsx
 import { fields } from '@keystone-6/fields-document/component-blocks';
 
-fields.relationship({ label: 'Authors', relationship: 'featuredAuthors' });
+fields.relationship({
```
dcousens, line 627: Could you put this in context of what the surrounding code might look like,  `...` is fine

dcousens (approved):

PR #7265, opened 2022-02-07
Use defined type PersonData

schoening:
Current example tries to use the value `Person = list({ ... })` as a type (and fails) and the `type PersonData` is never used.

### DIFF
--- a/docs/pages/docs/guides/auth-and-access-control.mdx
+++ b/docs/pages/docs/guides/auth-and-access-control.mdx
@@ -503,11 +503,11 @@ const isAdmin = ({ session }: { session: Session }) =>
   session?.data.isAdmin;
 
 // Validate the current user is updating themselves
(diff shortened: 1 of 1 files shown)


schoening:
However the `item` type here is `BaseItem`. It should probably be upgraded to `BaseItem<T>` in the future for more specific type checking?

```
item: {
    update: ({ session, context, listKey, operation, inputData, item }) => boolean;
}
```

schoening:
The actual working item example for me becomes:

```
// Validate if current user is an Admin, or updating themselves
const isAdminOrPersion = ({ session, item }: { session: Session, item: BaseItem }) =>
  isAdmin({ session }) || isPerson({ session, item });
```

dcousens (approved):

dcousens:
Thanks @schoening

Issue #7274, opened 2022-02-07
startup - async config


pazams:
### Discussed in https://github.com/keystonejs/keystone/discussions/7273

<div type='discussions-op-text'>

<sup>Originally posted by **pazams** February  7, 2022</sup>
`keystone.ts` interface is a default export of `KeystoneConfig`.
However, let's consider that some of the values in `KeystoneConfig` (e.g. db.url) needs to be fetched from a remote secret storage service. How should this type of async call work considering the interface for `keystone.ts` is a synchronous export of `KeystoneConfig`?</div>

dcousens:
See #7273 instead, this is not a bug

PR #7275, opened 2022-02-07
Add support for asynchronous configuration

pazams:
This is a **Non-breaking** feature.
This allows keystone.ts to now return either `KeystoneConfig` or `Promise<KeystoneConfig>` 
Ref https://github.com/keystonejs/keystone/issues/7274

### DIFF
--- a/packages/core/src/lib/config/normalizeConfig.ts
+++ b/packages/core/src/lib/config/normalizeConfig.ts
@@ -1,0 +1,10 @@
+import { KeystoneConfig } from '../../types';
+
+/*
+  This function normalizes an input of KeystoneConfig or Promise<KeystoneConfig>
+  and just returns Promise<KeystoneConfig>
+*/
+
+export async function normalizeConfig(config: KeystoneConfig | Promise<KeystoneConfig>) : Promise<KeystoneConfig> {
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/config/normalizeConfig.ts
```diff
@@ -0,0 +1,10 @@
+import { KeystoneConfig } from '../../types';
+
+/*
+  This function normalizes an input of KeystoneConfig or Promise<KeystoneConfig>
+  and just returns Promise<KeystoneConfig>
+*/
+
+export async function normalizeConfig(config: KeystoneConfig | Promise<KeystoneConfig>) : Promise<KeystoneConfig> {
+	return config;
```
pazams, line 9: Alternatively, we can rewrite this with `Promise.resolve` instead of an `async function` to achieve the same thing.

dcousens:
I don't think we want to support an asynchronous configuration from the `keystone.ts`,  but with new pathways like `getContext` this should free users to do any asynchronous actions as needed.

dcousens:
Followed up in https://github.com/keystonejs/keystone/discussions/7273#discussioncomment-4190847

PR #7280, opened 2022-02-10
Error Access Denied large number of mutation with relationship connect

gautamsi:
this is the error thrown (works with 200, does not work for 3000 or 5000)

```sh
Error: An error occured while resolving relationship fields.
  - Post.author: Access denied: You cannot perform the 'connect' operation on the item '{"id":"ckzgl6lkz0002l4dnz95e7h7c"}'. It may not exist.
    at relationshipError (D:\dr\gh\keystonejs\keystone\packages\core\src\lib\core\graphql-errors.ts:57:10)
    at getResolvedData (D:\dr\gh\keystonejs\keystone\packages\core\src\lib\core\mutations\create-update.ts:289:11)
    at resolveInputForCreateOrUpdate (D:\dr\gh\keystonejs\keystone\packages\core\src\lib\core\mutations\create-update.ts:355:27)
    at createSingle (D:\dr\gh\keystonejs\keystone\packages\core\src\lib\core\mutations\create-update.ts:49:36)
    at D:\dr\gh\keystonejs\keystone\packages\core\src\lib\core\mutations\create-update.ts:112:38 {
  path: [ 'createPosts', 2 ],
  extensions: { debug: [ [Object] ], code: 'KS_RELATIONSHIP_ERROR' }
}
```


examples/blog/seed-data/index.ts
```diff
@@ -53,6 +53,23 @@ export async function insertSeedData(context: KeystoneContext) {
...
 
+  console.log('generate large number of data');
+  const existing = await context.query.Author.findMany({
+    take: 1,
+    query: 'id',
+  });
+
+  const data = [];
+  for (let index = 0; index < 5000; index++) {
+    // data.push({ title: `title: ${index}`, author: { connect: { email: 'arthur.cd@email.com' } } });
```
gautamsi, line 64: whether connecting based on email or id it always fail for large queries

mitchellhamilton:
The root cause of this is Prisma timing out when doing this many operations, this is pretty much expected given the number of operations (and the number of operations that can happen successfully will vary based on the machine you're running on, I got it to fail/succeed at different numbers than you). When doing this many operations, you should chunk them so that they're not all trying to happen at once.

Issue #2929, opened 2020-05-08
socket.io example


pahaz:
# Feature request

At the moment the Keystone doesn't have examples for an HTTP server customization like `socket.io`. There is only `configureExpress` hook but it's useless for the HTTP customization like socket.io.

What should I do if I need to customize the HTTP server?

jesstelford:
We allow you to fully customise your server, as long as you're able to consume the express middleware Keystone provides: https://keystonejs.com/guides/custom-server

pahaz:
Yes, you can, but it not the same as just define `configureExpress`

jesstelford:
I'm not sure I understand. `configureExpress` just passes you the express app for you to do what you need. At that point you can do pretty much anything like attaching new routes or middleware.

If you need more customisation, I linked to the docs which show a bunch of examples of how to build a fully custom server.

bonniss:
@pahaz it's been a while, have you got any solutions?

PR #7292, opened 2022-02-21
Add support for accessing the http server instance through config

lachieh:
Adds support for the underlying node http server to be extended using config. This would enable the use of [Socket.io](https://socket.io/docs/v4/server-initialization/#with-express) which has been asked for [here](https://github.com/keystonejs/keystone/issues/2929), as well as on the Keystone Slack channel. All of the current solutions require running a second node http server along side the keystone one and this isn't really a great solution.

I expect there is more to be discussed here but I wanted to get a proof-of-concept PR going to get the conversation started.

### DIFF
--- a/.changeset/hot-geese-push.md
+++ b/.changeset/hot-geese-push.md
@@ -1,0 +1,6 @@
+---
+'@keystone-6/core': patch
+'@keystone-6/api-tests-legacy': patch
+---
+
+added support for extending the underlying node http server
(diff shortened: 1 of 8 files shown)


dcousens:
@lachieh I think this looks great!  I'll check back whether there is anything technically holding us back from opening this up.

lachieh:
Thanks, @dcousens!

I wasn't sure if this would break any APIs so just assumed patch, but it's probably a minor release.

If you've got some pointers on where the test should go for this, that'd be great. Looks like 8(?) is failing. What is the significance of `8`? Is there be a better place that I should write the tests for this?

dcousens (approved):
Nice work @lachieh,  I concur we should update the `CONTRIBUTING.md` to be unambiguous about how to test changes locally :+1:

dcousens:
Sorry I hadn't replied to these comments earlier.

> I wasn't sure if this would break any APIs so just assumed patch, but it's probably a minor release.

It is a minor,  but we can update that when preparing the release notes.  Your work is super appreciated! 

> Looks like 8(?) is failing. What is the significance of 8? Is there be a better place that I should write the tests for this?

The tests are split across multiple GitHub action jobs for parallelization,  you can determine the breakdown from the following

https://github.com/keystonejs/keystone/blob/81e663d3157b3c49a191aaa1bc1c179dd8cc34d4/.github/workflows/tests.yml#L88-L89

https://github.com/keystonejs/keystone/blob/81e663d3157b3c49a191aaa1bc1c179dd8cc34d4/tests/jest/jest-sequencer.js#L6-L18

In this instance,  your test _was_ incorrectly setting the header (as you resolved in https://github.com/keystonejs/keystone/pull/7292/commits/c5393e9711482b03b55b4e9ec863b18634a98e1a).

Again,  great work here @lachieh!

PR #7302, opened 2022-02-23
Revert "Improved `context.query` types"

emmatown:
Reverts keystonejs/keystone#7249

The problem is that a `context.query.*.find*` result cannot be statically translated to produce accurate Typescript types, and as such the result currently is that we return a `Record<string, any>`.

For example, casting to a `Post` is the approach typically taken by users, but that can be wrong.
Even if you could assume `Partial<Post>` (or similar for `null`s), the inherent capability of GQL aliases means the naming may not be 1:1 and thus types cannot be automatically inferred.

### DIFF
--- a/.changeset/wicked-ties-warn.md
+++ b/.changeset/wicked-ties-warn.md
@@ -1,5 +1,0 @@
----
-'@keystone-6/core': minor
----
-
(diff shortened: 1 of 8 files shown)


dcousens:
We are actively planning an alternative solution to this problem,  the resultant solution as merged ended up having problems in that it easily misrepresented the actual types when aliasing and other GraphQL subtleties were involved.

dcousens (approved):

PR #7304, opened 2022-02-24
Add the ability to send additional Apollo links to the admin ui

acoreyj:
This was useful for the ability to add headers, in my case authorization, to every request to the server


borisno2:
Hi @acoreyj thanks for the PR!

Something like this would also be handy with subscriptions (#7722) where we needed to instead create a 2nd Apollo Client.

Not sure how using `from()` impacts subscriptions though. For subscriptions you use `split()` and set mutations and queries down one link and subscriptions down another. I would be more inclined to have `apolloLink` and let the end developer do the `from()` or `split()` and just swap out the default `link` with what is provided in `adminConfig.apolloLink`. Would replacing the `link` remove the need for the `adminConfig.providerLogic` in your use case? 

It would be great to see an example of using `adminConfig.providerLogic`
Thanks!

PR #7314, opened 2022-03-02
Fix context for ApolloServer

bichikim:
🤣


dcousens:
Hi @bichikim,

Thanks for the pull request, I think this highlights a problem we need to adjust, but maybe not exactly as this pull request is doing.

- We shouldn't be replacing the Apollo `context` with our own, without at least ensuring our users understand that; maybe by restricting the types and or documentation.
- Or, we should be carrying/merging/injecting the Apollo `context` into our own `context`

PR #7324, opened 2022-03-03
Adds http.Server options as configuration server.options 

dcousens:
This pull request adds the `http.Server` options as an configuration option `server.options` for your Keystone configuration.

This will enable developers to configuration things like the `host` interface, or to bind the server to an existing `fd` descriptor (like if you are using systemd's `LISTEN_FDS`).

### DIFF
--- a/packages/core/src/scripts/run/start.ts
+++ b/packages/core/src/scripts/run/start.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import type { ListenOptions } from 'net';
 import * as fs from 'fs-extra';
 import { createSystem } from '../../lib/createSystem';
 import { initConfig } from '../../lib/config/initConfig';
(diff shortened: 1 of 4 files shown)


packages/core/src/scripts/run/start.ts
```diff
@@ -44,9 +44,13 @@ export const start = async (cwd: string) => {
     console.log(`✅ Admin UI ready`);
   }
 
-  const port = config.server?.port || process.env.PORT || 3000;
-  httpServer.listen(port, (err?: any) => {
+  // use env.PORT if supplied
+  const port = parseInt(process.env.PORT || '') || config.server?.port;
+  const options = config.server?.options;
+  if (options) options.port = port;
```
dcousens, line 50: - [ ] only .port or .options should be supplied

mitchellhamilton (approved):
I want to note that this makes `process.env.PORT` preferred over `config.server.port` which I agree makes more sense though you could argue that it's breaking but as long as you're aware of that and think that's fine then I'm ok with it.

dcousens:
I think the previous behaviour is indicative of a bug, and I'm happy to fix that as a `patch`,  however I will identify append that to the changeset.

Issue #6262, opened 2021-08-05
Anonymize Google Analytics Tag


raveling:
Check out Google Analytics status re: GDPR & CCPA. Low risk posture typically involves:

- Ensuring the tracking tag is set to anonymise user data
- Setting GA admin to discard data after a certain period of time (as opposed to forever)

bladey:
Moved to planning doc.

PR #7335, opened 2022-03-07
Remove analytics services from website

bladey:
Co-Authored-By: Daniel Cousens <dcousens@users.noreply.github.com>
Co-Authored-By: Ronald Aveling <6447754+raveling@users.noreply.github.com>


dcousens:
As a follow up to https://github.com/keystonejs/keystone/issues/6262 and https://github.com/keystonejs/keystone/pull/6631,  we'll re-evaluate this soon,  but for now,  let's drop it

PR #7337, opened 2022-03-07
Fix file and image routes

Olli1080:
The options baseUrl configurations of both files and images have been ignored.
This is because the baseUrl is registered as a hardcoded path inside the construction of the express server.
The paths generated from the graphql query on the other hand are correct.

This PR exposes the previously hidden default basePaths and fixes the routes for images and files.

### DIFF
--- a/.changeset/stale-poems-lay.md
+++ b/.changeset/stale-poems-lay.md
@@ -1,0 +1,6 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix routes for files and images in case baseUrl exists.
+Replace hardcoded paths with default or config paths.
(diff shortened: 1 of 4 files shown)


Olli1080:
Thank you for your response.

I see now where my confusion came from. In the documentation it uses a relative path ('/images') as an example.
I have some questions for comprehension. At the current state all urls would have to contain '/images' to be valid? E.g. http://localhost:3000/images or https://localhost:3000/images.
Wouldn't it make more sense if the path would be omitted and added to the provided baseUrl?

E.g. (1)
```
baseUrl = http://localhost:3000  
fullPath = baseUrl + '/images'
```
or (2)
```
baseUrl = http://localhost:3000  
basePath = '/images'  
fullPath = baseUrl + basePath
```
Also would you be open to a pr adding options (1) or (2) as a new feature?

NasserBvB (approved):

dcousens:
Replaced by #7070

dcousens:
Thanks @Olli1080,  but these fixes and many other changes were completed in #7070 instead.  Thanks for your work though,  and sorry we didn't update you sooner

PR #7350, opened 2022-03-13
Support Prisma SHADOW_DATABASE_URL

chelkyl:
Rationale: When using shadow databases, you must use the `shadowDatabaseUrl` API field per [the Prisma documentation](https://www.prisma.io/docs/concepts/components/prisma-migrate/shadow-database#cloud-hosted-shadow-databases-must-be-created-manually).

Implementation: We supported passing `DatabaseConfig.shadowDatabaseUrl` in the Keystone config object.

Relevant Links:
* https://github.com/prisma/prisma-engines/pull/1672

Follow-up From:
* https://github.com/keystonejs/keystone/pull/6982


jlarmstrongiv:
@dcousens thanks for running the pipeline!

@bladey we’ve double checked the GraphQL schema updates https://github.com/keystonejs/keystone/pull/6982#issuecomment-1055945012

@molomby we implemented your suggestions from https://github.com/keystonejs/keystone/pull/6982#issuecomment-1057670458

Thank you all for your help!  Let me know if there’s anything else needed to move this PR along.

jlarmstrongiv:
@dcousens any updates? 🤞

dcousens:
Hi @jlarmstrongiv ,  we haven't forgotten this,  the only remaining block is to verify the behaviour we want around how this changes the `schema.prisma`.  We're wanting to see if we can always provide the shadow database URL,  and hoping that it being blank reverts to the default behaviour.

jlarmstrongiv:
Thanks for the update @dcousens much appreciated 👍 

Yes, we implemented it [similar to](https://github.com/keystonejs/keystone/blob/6b45119a1d5e6ed98d61977868a87a6be0984a2a/packages/core/src/lib/core/prisma-schema.ts#L206) [prisma flags](https://github.com/keystonejs/keystone/blob/6b45119a1d5e6ed98d61977868a87a6be0984a2a/packages/core/src/lib/core/prisma-schema.ts#L212)

borisno2:
Hi @jlarmstrongiv Thanks for your reply, we are wanting to avoid the `prisma.schema` file changing depending on environment variables.

If you were to alway print `url = env("DATABASE_URL")  shadowDatabaseUrl = env("SHADOW_DATABASE_URL")` in the prisma schema but `process.env.SHADOW_DATABASE_URL` was blank does it just revert back to default behaviour?

Let me know if you have any questions.

jlarmstrongiv:
@borisno2 Thank you for following up!  Oh I see, theoretically yes.  

Back in February, the https://github.com/prisma/prisma-engines/pull/1672 was recently merged and KeystoneJS was depending on an older version of prisma, so that was not possible at the time.

Now, with KeystoneJS depending on prisma@3.14.0 (released a month ago), including `shadowDatabaseUrl = env("SHADOW_DATABASE_URL")` in every schema should work.

jlarmstrongiv:
@borisno2 we updated the PR with your suggestions 🚀 would you run the the workflow?

packages/core/src/lib/migrations.ts
```diff
@@ -17,11 +17,13 @@ import { confirmPrompt, textPrompt } from './prompts';
...
 // not in control of their Prisma version.
 // https://www.prisma.io/docs/reference/api-reference/environment-variables-reference#prisma_hide_update_message
-function runMigrateWithDbUrl<T>(dbUrl: string, cb: () => T): T {
+function runMigrateWithDbUrl<T>(dbUrl: string, shadowDbUrl: string | undefined, cb: () => T): T {
   let prevDBURLFromEnv = process.env.DATABASE_URL;
+  let prevShadowDBURLFromEnv = process.env.SHADOW_DATABASE_URL;
   let prevHiddenUpdateMessage = process.env.PRISMA_HIDE_UPDATE_MESSAGE;
   try {
     process.env.DATABASE_URL = dbUrl;
+    process.env.SHADOW_DATABASE_URL = shadowDbUrl;
```
borisno2, line 26: Might need to only set `process.env.SHADOW_DATABASE_URL` if `shadowDbUrl` is set.
ie
```
if (shadowDbUrl) {
    process.env.SHADOW_DATABASE_URL = shadowDbUrl;
}
```
otherwise it looks like ` process.env.SHADOW_DATABASE_URL` gets set to `undefined` which prisma doesn't seem to like.

jlarmstrongiv:
Interesting!  We now set it to an empty string instead, which should work https://github.com/prisma/prisma-engines/pull/1672/files#diff-c1428df746767db7244ee782e01a8c5cd941e55218ee25ace07da1eece5907e7R169

Would you re-run the workflow @borisno2 ?

jlarmstrongiv:
@borisno2 fixed some hard-coded schema tests, let’s try again 🚀

emmatown (approved):

Issue #6112, opened 2021-07-15
`ui.labelField` option does not render on relationship fields


bladey:
# Bug report

Setting `ui.labelField` on a relationship field doesn't render the chosen data.

You can see below I'm attempting to use the `email` field from the user as the value (first arrow) however it doesn't pick up on this, it continues to default to the list default, which is the name of the `user`.

If I set it on a list level (second arrow, commented out) it comes through fine.

<img width="595" alt="export const lists = createSchena({" src="https://user-images.githubusercontent.com/737821/125738152-6e1ae11b-7a95-4b26-bc74-abbec481140d.png">

Field level:

<img width="509" alt="image" src="https://user-images.githubusercontent.com/737821/125738517-b0dc51cb-ada0-420f-8a2e-03f395c25057.png">

List level:

<img width="491" alt="image" src="https://user-images.githubusercontent.com/737821/125738633-b31d98da-ca3f-4b9b-bb8e-8c5093048280.png">

Are we missing an option on this line [here](https://github.com/keystonejs/keystone/blob/3f03b8c1fa7005b37371e1cc401c3a03334a4f7a/packages/fields/src/types/relationship/index.ts#L133)?

It is currently this:

<img width="663" alt="image" src="https://user-images.githubusercontent.com/737821/125738726-bb6d02a7-7354-48ab-a101-935b4dbe06b2.png">

Changing this to:

<img width="763" alt="image" src="https://user-images.githubusercontent.com/737821/125738423-61d140b8-5001-48b3-af69-350ad2d98be0.png">

Does not have any effect, but it seems to be a step in the right direction, as we probably need to gather meta data somewhere for data to be available (?).

bladey:
Slack reference - https://keystonejs.slack.com/archives/C01STDMEW3S/p1626240048270500

jigz:
+1, I came upon this too

supermx1:
+1

cpt-n3mo:
+1

willemmulder:
I have the same issue. Apparently this should be fixed when #7352 is merged! 👍

georgekrax:
+1 | Same here. I added a pull request, but still the `labelField` is not changed on the `<RelationshipSelect />` component dropdown options.

dcousens:
Fixed for the `relationship` field in https://github.com/keystonejs/keystone/pull/8049, #7352 did resolve the problem too, but the other changes there aren't ready for merge.

PR #7352, opened 2022-03-13
Order relationship select

skeary:
Updates the admin UI for the relation type fields so that the values in the select are ordered (ascending) by the label field. This makes it easier to find an option when dealing with larger select lists.

The behavior could be made configurable within `schema.ts` so that ordering could be done on another field or descending rather than ascending order. I wasn't sure of the value of this so I've not made it configurable but explore this if felt useful.


dcousens:
Hi @skeary ,

This is nice to have functionality!
Unfortunately,  we don't necessarily want to assume what might be a natural order for this field generally.
We might want to expand this functionality to include configuration in the schema to what the default ordering is,  and use that instead.

If you want to try and do that,  awesome!  Otherwise we'll try and expand on this pull request when we have time :+1:

skeary:
Hi @dcousens,
Thanks for the feedback. My thinking was that this was a better default - while it may not solve all use cases it wouldn't be worse for anyone and in general would be what most people wanted? It's not a problem though to add in extra configuration, and it's definitely more flexible, so am happy to have a go at this!

I think the following might be a reasonable configuration approach but just wanted to checked what you thought before I go ahead:
```
    ui?: {
        sort?: 'none' | 'labelField' | { field: string } | undefined;
        sortOrder?: 'asc' | 'desc' | undefined;
    } | undefined;
```

With examples of use in schema.ts:

```
      author: relationship({ ref: 'Author.posts', many: false }), // order on label field, ascending
      author: relationship({ ref: 'Author.posts', many: false, ui: { sort: 'labelField', sortOrder: 'asc' } }), // same
      author: relationship({ ref: 'Author.posts', many: false, ui: { sort: 'none' } }), // no ordering
      author: relationship({ ref: 'Author.posts', many: false, ui: { sort: { field: 'email' }  } }), // order on email, ascending
      author: relationship({ ref: 'Author.posts', many: false, ui: { sort: { field: 'email' }, sortOrder: 'desc'  } }), // order on email, descending
```

What do you think?

skeary:
Note this changeset now (I think) includes a working fix for #6112

skeary:
Note this changeset now (I think) includes a working fix for https://github.com/keystonejs/keystone/issues/6112. This issue was that the ui.labelField option under the relationship field config wasn't having any affect. I noticed this problem while testing my changes and have now hopefully fixed the issue. Unless I'm missing something, the support for this feature wasn't really there rather than a logic bug or similar. So I've now implemented support for this.

skeary:
Hi @dcousens,
I've implemented the approached I outlined in a earlier comment where the configuration now supports:

```
 ui?: {
        orderBy: false | Array<{ labelField:  'asc' | 'desc' } | { field: string, order: 'asc' | 'desc' }> | undefined,
    } | undefined;
```
To me, this seems a reasonably simple way to configure the ordering and should hopefully cover the significant number of use cases. If someone doesn't want ordering they can turn off the ordering which will revert to the current behaviour. By default the ordering is still based on label field (ascending) and I think that is what is useful for most people?

I've updated the documentation to describe the new configuration options. See https://keystone-next-docs-ahungkcd8-keystonejs.vercel.app/docs/apis/fields#relationship for a rendered version.

As part of doing this changeset, I also think I've addressed #6112. 

It would be great if you could review my overall approach and implementation and let me know what you think. I'm happy to tweak as needed.

Cheers,
Simon

willemmulder:
Nice work @skeary ; let's hope this can be merged soon!

stuible:
Would love the ability to sort relationship select and cards.  Excited to see this feature merged!

dcousens:
Replaced by https://github.com/keystonejs/keystone/pull/9648

PR #7394, opened 2022-03-24
Fix off by one date selection issue 

### LINKED ISSUE #6115: Date Picker always changes to the previous day
# Bug report

## Describe the bug

When selecting a date, the DatePicker always changes the previous day

## To Reproduce

- Setup a list with a `timestamp` 
```
  orderDate: timestamp(),
```
- create a new record
- select a date
- date will change to the previous day

## Expected behaviour

I expect the date a selected to continue to be the date I selected

## Screenshots

WY8H3nzUSC

## System information

- OS: macos
- Browser: chrome & brave

## Additional context

The date stored in the database also seems to be the new date "choosen" by the DatePicker...

IOW:

- If I'm in MST and I select  `7/15/2021 07:00`
- The UI updates to `7/14/2021 07:00`
- I save
- The database displays `2021-07-14 13:00:00`

gwyneplaine:
Resolves #6115 

The root of the problem as highlighted in #6115 is inconsistency in how browsers handle `new Date(value)` where `value` is some string representation of a Date. 

This PR opts for explicitly passing in year month and day to Date for more deterministic behaviour.

### DIFF
--- a/design-system/packages/fields/src/DatePicker/index.tsx
+++ b/design-system/packages/fields/src/DatePicker/index.tsx
@@ -31,6 +31,11 @@ export function useEventCallback<Func extends (...args: any) => any>(callback: F
   return cb as any;
 }
 
+function deserializeDate(value: string): Date {
+  const [year, month, day] = value.split('-').map(el => parseInt(el, 10));
+  return new Date(year, month - 1, day);
+}
+
 export const DatePicker = ({
   value,
   onUpdate,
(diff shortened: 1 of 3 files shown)


design-system/packages/fields/src/DatePicker/index.tsx
```diff
@@ -31,6 +31,11 @@ export function useEventCallback<Func extends (...args: any) => any>(callback: F
   return cb as any;
 }
```
gwyneplaine, line 33: couldn't think of a better name for this fn. suggestions welcome

emmatown (approved):

design-system/packages/fields/src/DatePicker/index.tsx
```diff
@@ -31,6 +31,11 @@ export function useEventCallback<Func extends (...args: any) => any>(callback: F
...
 
+function deserializeDate(value: string): Date {
+  const [year, month, day] = value.split('-').map(el => parseInt(el, 10));
+  return new Date(year, month - 1, day);
```
dcousens, line 36: Alternatively,  I think you could have done `return new Date(value.replace(/-/g, '/'))`

dcousens, line 36: Alternatively,  I think you could have

```suggestion
  return new Date(value.replace(/-/g, '/'))
```

dcousens: Actually, we should use `import { parse} from 'date-fns'`

Issue #7364, opened 2022-03-15
Reset to defaults resets pagination 


gwyneplaine:
We store filter information for the user in local-storage such that when users revisit a page with filters applied, they don't need to spend time reconfiguring it manually. 

We also have a `reset to defaults` button to enable users to manually clear the filters from localStorage should they go stale. Presently this is also clearing stored pagination data. We want to omit this from the filters cleared by reset to defaults. 

https://user-images.githubusercontent.com/12119389/161702716-04893d86-1ceb-44cc-90cb-8a227f5e3953.mov

This is likely happening because we are naively calling `router.push` in the `resetToDefaults` function without also passing in the current page in as a query param. 

The buggy code can be found in `packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx` on `line 87`

PR #7456, opened 2022-04-11
Fixed Reset to Defaults Button

Achisingh:
Co-authored-by: Charles <gwyneplaine@users.noreply.github.com>

#7364 - Fixed Reset to Default pagination issue


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -254,7 +254,7 @@ const ListPage = ({ listKey }: ListPageProps) => {
               <FilterAdd listKey={listKey} filterableFields={filterableFields} />
             ) : null}
             {filters.filters.length ? <FilterList filters={filters.filters} list={list} /> : null}
-            {Boolean(Object.keys(query).length) && (
+            {Boolean(filters.filters.length || sort || query.fields) && (
```
mitchellhamilton, line 257: `sort` will be truthy when there is an initial sort so we'll want to only show the button when there's a custom sort
```suggestion
            {Boolean(filters.filters.length || query.sortBy || query.fields) && (
```

emmatown (approved):

PR #7463, opened 2022-04-13
Add Codesandbox configs to allow launching online sandboxes for example projects

MurzNN:
As continuation to https://github.com/keystonejs/keystone/discussions/7222 I've added sandbox configs to all example projects, which allow launching Keystone examples on https://codesandbox.io/ service.

With this feature Keystone users can play with Keystone without instantly in web browser, without any manual setup on local machine. I think it will be very convenient especially for new Keystone users, that will can touch Keystone Admin UI and frontend directly in their web browsers with few clicks, and even fork them to add own changes.

Current example of URL: https://codesandbox.io/s/github/MurzNN/keystone/tree/codesandbox-configs/examples/blog

Also you can now point to that playground urls from Keystone documentation.

### DIFF
--- a/.changeset/shy-cycles-check.md
+++ b/.changeset/shy-cycles-check.md
@@ -1,0 +1,25 @@
+---
+'@keystone-6/example-assets-cloud': patch
+'@keystone-6/example-assets-local': patch
+'@keystone-6/example-auth': patch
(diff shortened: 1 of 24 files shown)


dcousens (approved):
Awesome! Thanks @MurzNN!

Issue #7395, opened 2022-03-24
Document field toolbar Z-index bug


gwyneplaine:
## Description 
Presently any dropdowns or dialogs in the toolbar of the `document-editor` will sink under the sticky primary actions on the item view. (see screenshot below)
![image](https://user-images.githubusercontent.com/413395/161677986-a2125350-c641-40d2-993d-697bd9fab4d0.png)
(image credit @Vopa37 from #7343)

This is not intended behavior, dropdowns and dialogs in general should consistently sit at a higher z-index than main content. We should ideally audit and resolve this for all dropdowns and dialogs in keystone, and align on z-index values across the codebase, but for now fixing this for all dialogs in the document-editor toolbar is sufficient. 

## Expected Behaviour 
All document-editor toolbar dialogs and dropdowns should **sit above** the Item Page toolbar. 

## Helpful context
Relevant code is in:
* Document editor Toolbar: `packages/fields-document/src/DocumentEditor/Toolbar.tsx` 
* Inline Dialogs used by the document editor: `packages/fields-document/src/DocumentEditor/primitives/inline-dialog.tsx`
* Admin UI Item Page toolbar: `packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ItemPage/index.tsx`  `line 525`

To reproduce this error and sandbox the issue try running the `@keystone-6/example-document-field` example. 
`yarn workspace @keystone-6/example-document-field dev` 

Follow up bug, transferred from #7343

jordie23:
@Achi06 @dcousens 

FYI the fix for this issue has the consequence of the toolbar appearing on top of the sticky bar when scrolling:


https://user-images.githubusercontent.com/712360/169222730-8d2398e0-36d3-4e27-8504-08688c49b1a1.mov

It happens only on the edit screen. The side create drawer works fine.

In fact it appears the editor has some sort of functionality to move the drop down menus up or down depending on position on the screen. 

This works on the create drawer:

https://user-images.githubusercontent.com/712360/169223081-9b467b00-294b-445c-a05d-b94bd0a56175.mov


But only partially works on the edit page (likely some offset value or something?):

https://user-images.githubusercontent.com/712360/169223071-c33daa6d-239f-4434-98aa-eeebb7d530db.mov

dcousens:
Related https://github.com/keystonejs/keystone/issues/6157

dcousens:
Should be fixed by https://github.com/keystonejs/keystone/pull/7464

PR #7464, opened 2022-04-14
#7395 document z index bug

Achisingh:
Fixed z-index issue for the toolbar of the editor #7395

### DIFF
--- a/packages/fields-document/src/DocumentEditor/Toolbar.tsx
+++ b/packages/fields-document/src/DocumentEditor/Toolbar.tsx
@@ -188,7 +188,7 @@ const ToolbarContainer = ({ children }: { children: ReactNode }) => {
         paddingTop: spacing.small,
         position: 'sticky',
         top: 0,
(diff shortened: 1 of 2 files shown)


packages/fields-document/src/DocumentEditor/Toolbar.tsx
```diff
@@ -188,7 +188,7 @@ const ToolbarContainer = ({ children }: { children: ReactNode }) => {
         paddingTop: spacing.small,
         position: 'sticky',
         top: 0,
-        zIndex: 2,
+        zIndex: 30,
```
dcousens, line 191: Any particular reason to use `30`?

Achi06: The zIndex for the sticky bar is 20, so I thought that 30 would be better for the toolbar.

dcousens (approved):
LGTM

PR #7474, opened 2022-04-19
Update image UI

gwyneplaine:
(replaced by https://github.com/keystonejs/keystone/pull/7520)


gwyneplaine:
@mitchellhamilton  updated to use `object-fit:contain` over `object-fit: cover` I've decided not to add an anchor to the image though, as it degrades the experience of screen reader and VO users

>For a keyboard user, it is tedious to navigate through redundant links. For users of assistive technologies, it can be confusing to encounter successive identical links. When the text alternative for the icon is a duplicate of the link text, it is repetitive as screen readers read the description twice.
- [WCAG 2.1](https://www.w3.org/WAI/WCAG21/Techniques/html/H2)

We can probably implement a "lightbox" modal later on though, should we want to improve the user experience in the way that you described. I've attached a POC

https://user-images.githubusercontent.com/12119389/164371185-2aba3b55-cb91-4f84-8090-a43c56057041.mov

packages/core/src/fields/types/image/views/Field.tsx
```diff
@@ -92,126 +90,167 @@ function ImgView({
...
-            inputRef.current?.click();
-          }}
-          tone="positive"
-        >
-          Upload Image
-        </Button>
-        {value.kind === 'remove' && value.previous && (
+        )
+      ) : (
+        <Fragment>
```
mitchellhamilton, line 221: Why the removal of a stack here?
screenshot of Keystone Admin UI showing Upload Image and Undo Removal buttons with no spacing between them

gwyneplaine: Moving code is hard 😂, thanks for the catch 🎉

emmatown (approved):

gwyneplaine:
@raveling design changes enacted, see screencapture below for success states (and screenshot for error state) 

https://user-images.githubusercontent.com/12119389/165230670-8f756b83-9986-46d0-ae4f-578607e04a2c.mov

Screen Shot 2022-04-26 at 3 46 32 pm

raveling:
@gwyneplaine great stuff! 🙌 

A few smol things:

1. Can we please change the error copy to `Sorry, that file type isn’t accepted. Please try .gif, .jpg, .png, or .webp.` (Open to improving this further of course. Preference being to soften it a little, and move the solutions to the end of string).
2. The error message needs `margin-top: 8px` to match vertical rhythm on other field inputs. (think there's a lobotomised own selector in the code base that isn't being applied to the error element.
3. Can we set something like `line-height: 1.15` for the save overlay text at [line 135](https://github.com/keystonejs/keystone/pull/7474/files#diff-ecd7df09b9cc2caf9785d9c0ca4f89174688373e82f927fe242e2c451dec0e89R135). I'm not firm on it being exactly `1.15`, so if there's a line height token you can reuse from the DS that's great. My preference is that it not be equivalent to 1 seeing as we know the string wraps. And the current line height has the descenders and ascenders nearly touching:

image

Thank you! 🙇

gwyneplaine:
@raveling sweet, thanks for the feedback 👍 screenshot below for parity 
Screen Shot 2022-04-27 at 8 53 25 am

gwyneplaine:
Screen Shot 2022-04-28 at 10 51 07 am

gwyneplaine:
replaced by #7520

PR #7490, opened 2022-04-27
Prompt users in the Admin UI to confirm they want to navigate away when there are unsaved changes

emmatown:
We could show one of our own modals in the local client-side navigation case but we can always add that later.

### DIFF
--- a/packages/core/src/admin-ui/utils/usePreventNavigation.tsx
+++ b/packages/core/src/admin-ui/utils/usePreventNavigation.tsx
@@ -1,28 +1,31 @@
 import { useEffect } from 'react';
 import { useRouter } from '../router';
 
-export function usePreventNavigation(shouldPreventNavigation: boolean) {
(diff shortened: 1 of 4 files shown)


packages/core/src/admin-ui/utils/usePreventNavigation.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const router = useRouter();
+
+  useEffect(() => {
+    if (shouldPreventNavigation) {
+      const clientSideRouteChangeHandler = () => {
+        if (!window.confirm('There are unsaved changes, are you sure you want to exit?')) {
+          // throwing from here seems to be the only way to prevent the navigation
+          // we're throwing just a string here rather than an error because throwing an error
+          // causes Next to show an error overlay in dev but it doesn't show the overlay when we throw a string
+          throw 'Navigation cancelled by user';
```
gwyneplaine, line 14: 😂

gwyneplaine (approved):
lgtm

PR #7504, opened 2022-04-29
Fix read only segmented control

Achisingh:
Fix read only display for segmented control. 
Previously the segmented control had focus-styles that were visible even in readOnly mode.
This is because under the hood the implementation is a list of radio inputs.
Radio inputs do not have a notion of `readOnly` unlike other input fields. 

We've implemented what we believe to be decent `readOnly` ux for this component. Briefly outlined below are the changes we've made: 
* `isReadOnly` prop added to the `SegmentedControl` component. When this is true, all unselected radio inputs are marked as disabled, and the focus styles for the selected input reflects its read only state. We keep the selected radio input enabled so that screen reader users browsing the form via tabbing know what field this is and what the selected value is. 
* A hidden input with a value of `none` is added to the `SegmentedControl`, this input is checked when no other input values are checked. This none input is not a field value, and is purely there for a11y purposes, such that screen reader users browsing the form via tabbing know what field this is and that no value is presently selected (even in read only mode) 

https://user-images.githubusercontent.com/28669082/167352435-d32865c1-7e82-4722-bb68-dec8de50ee5a.mov

### DIFF
--- a/.changeset/gold-glasses-melt.md
+++ b/.changeset/gold-glasses-melt.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Fixed read only view for segmented control display mode in the select field
+Fixed read only view for segmented control display mode in the select field. When a `segmented-control` select field is in `read` mode, the values no longer appear editable. 
--- a/.changeset/gold-glasses-melt.md
+++ b/.changeset/gold-glasses-melt.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fixed read only view for segmented control display mode in the select field
--- a/.changeset/real-pigs-tease.md
+++ b/.changeset/real-pigs-tease.md
@@ -1,0 +1,5 @@
+---
+'@keystone-ui/segmented-control': minor
+---
+
+Added isDisabled prop
--- a/.changeset/real-pigs-tease.md
+++ b/.changeset/real-pigs-tease.md
@@ -2,4 +2,4 @@
 '@keystone-ui/segmented-control': minor
 ---
 
-Added isDisabled prop
+Added `isReadOnly` prop and custom read only logic for the Segmented Control component.
--- a/design-system/packages/segmented-control/src/SegmentedControl.tsx
+++ b/design-system/packages/segmented-control/src/SegmentedControl.tsx
@@ -31,7 +31,6 @@ import {
   useId,
   useTheme,
   VisuallyHidden,
-  css,
 } from '@keystone-ui/core';
 
 import { SizeKey, WidthKey, useControlTokens } from './hooks/segmentedControl';
@@ -50,6 +49,8 @@ type SegmentedControlProps = {
   onChange: ManagedChangeHandler<Index>;
   /** Provide labels for each segment. */
   segments: string[];
+  /** To Disable */
+  isDisabled?: boolean;
   /** The the selected index of the segmented control. */
   selectedIndex: Index | undefined;
   /** The size of the controls. */
@@ -63,6 +64,7 @@ export const SegmentedControl = ({
   fill = false,
   onChange,
   segments,
+  isDisabled = false,
   size = 'medium',
   width = 'large',
   selectedIndex,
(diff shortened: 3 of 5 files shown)


gwyneplaine:
Tricky one, maybe we can catchup tmr morning w Matt Stow about this one.
Mostly conscious of us wanting to be able to focus on a read only  control if a value is selected for screen reader benefits.
So i'm not entirely convinced that disabled is the right call here 

Styling looks good to me though

gwyneplaine:
Couple of changes: 
* We want a isReadOnly prop for the segmented control 
* This should disable all items that are not selected 
* Selected items that are readOnly should have a grayed out focus ring on focus 
* We want a visually hidden `none selected` item that is selected if the segmented control does not have a selected value. (so the screen reader can read it out) 
* If no values are selected we want the entire segmented control to have a grayed out focus outline.

.changeset/real-pigs-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-ui/segmented-control': minor
+---
+
+Added isDisabled prop
```
gwyneplaine, line 5: We added a isReadOnly prop
```suggestion
Added isReadOnly prop
```

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -138,6 +144,16 @@ export const SegmentedControl = ({
...
           );
         })}
+        <VisuallyHidden as="label">
+          None Selected
+          <VisuallyHidden
+            as="input"
+            type="radio"
+            checked={noneSelected}
+            disabled={getDisabled(isDisabled, isReadOnly, noneSelected)}
+            value="None Selected"
```
gwyneplaine, line 154:
```suggestion
            value="none"
```

packages/core/src/fields/types/select/views/index.tsx
```diff
@@ -60,6 +60,8 @@ export const Field = ({
                   ? field.options.findIndex(x => x.value === value.value!.value)
                   : undefined
               }
+              isReadOnly={onChange === undefined}
+              // isDisabled={onChange === undefined}
```
gwyneplaine, line 64:
```suggestion
```

Achi06:
Update made to the display of segmented control in **read-only** mode is displayed below:

https://user-images.githubusercontent.com/28669082/167352435-d32865c1-7e82-4722-bb68-dec8de50ee5a.mov

.changeset/gold-glasses-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fixed read only view for segmented control display mode in the select field
```
gwyneplaine, line 5:
```suggestion
Fixed read only view for segmented control display mode in the select field. When a `segmented-control` select field is in `read` mode, the values no longer appear editable. 
```

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -100,19 +105,18 @@ export const SegmentedControl = ({
       });
     }
   }, [animate, selectedIndex]);
+  const noneSelected = selectedIndex === undefined;
+
+  const getDisabled = (disabled?: boolean, readOnly?: boolean, selected?: boolean) => {
```
mitchellhamilton, line 110: These parameters are unused, no? And there's a duplicate of this in the Item component?

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -58,11 +61,19 @@ type SegmentedControlProps = {
   width?: WidthKey;
 } & BoxProps;
 
+const getDisabled = (disabled?: boolean, readOnly?: boolean, selected?: boolean) => {
```
mitchellhamilton, line 64:
```suggestion
const getDisabled = (disabled: boolean, readOnly: boolean, selected: boolean) => {
```

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -126,6 +127,8 @@ export const SegmentedControl = ({
               fill={fill}
               isAnimated={animate}
               isSelected={isSelected}
+              disabled={isDisabled}
+              readOnly={isReadOnly}
```
dcousens, line 131: No strong opinion, but the mix of `isAdjective` and `adjective` is inconsistent

dcousens (approved):
LGTM

PR #7505, opened 2022-04-29
Fix read only json fix

Achisingh:
Fix read only display for json field

### DIFF
--- a/.changeset/young-hotels-smoke.md
+++ b/.changeset/young-hotels-smoke.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix the read only view for json field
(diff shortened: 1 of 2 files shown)


gwyneplaine:
This is on the right track @Achi06 but lets talk tomorrow with @raveling  and @mattbx  about how to visually differentiate the read-only ui a bit more, at the moment it looks a bit **too** similar to an editable field, especially since its now focusable (which imo is the correct behaviour) 

https://user-images.githubusercontent.com/12119389/166193747-425be8c3-f528-48f6-b16d-f54d3e962998.mov

gwyneplaine (approved):

gwyneplaine (approved):

PR #7533, opened 2022-05-14
Remove unused name from Auth session documentation

Greenheart:
Quick fix to make the code sample in the guide work as expected.

The auth guide now consistently uses `id` instead of mentioning `id` but actually using `name` like it did in a previous commit https://github.com/keystonejs/keystone/blob/d0713111f1cdb76d553eb0d65a597620e0d3e056/docs/pages/docs/guides/auth-and-access-control.mdx#add-a-sessiondata-query

This is such a small change so I thought no changeset was needed, but let me know if that's required, and I'll update the PR.

### DIFF
--- a/docs/pages/docs/guides/auth-and-access-control.mdx
+++ b/docs/pages/docs/guides/auth-and-access-control.mdx
@@ -136,7 +136,7 @@ You configure it like this:
 ```ts
 const { withAuth } = createAuth({
   // ...
-  sessionData: 'id isAdmin',
+  sessionData: 'isAdmin',
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/pages/docs/guides/auth-and-access-control.mdx
```diff
@@ -136,7 +136,7 @@ You configure it like this:
 ```ts
 const { withAuth } = createAuth({
   // ...
-  sessionData: 'name isAdmin',
+  sessionData: 'id isAdmin',
```
dcousens, line 139: When using `createAuth`, `session.itemId` is actually the `id` of the authenticated list.
This is redundant.

See https://github.com/keystonejs/keystone/blob/43a73e5bd6fddd0404dad87ab3ff406aef6143e9/packages/auth/src/index.ts#L242-L245

This documentation needs work,  we need to outline these interfaces and their guarantees properly.

dcousens:
```suggestion
  sessionData: 'isAdmin',
```

dcousens (approved):

Issue #7325, opened 2022-03-03
The `description` schema field is not used in the Admin UI


JMLX42:
## How to reproduce

1. Define [the `description` field of a schema](https://keystonejs.com/docs/apis/schema#description).
2. Open the Admin UI on that schema.

## Expected behavior

The expected behavior is [the documented one](https://keystonejs.com/docs/apis/schema#description):

> The description option defines a string which will be used as a description in the Admin UI

## Actual behavior

The schema description is not in the Admin UI.

```
"@keystone-6/auth": "^1.0.1",
"@keystone-6/core": "^1.0.1",
"@keystone-6/fields-document": "^1.0.1",
```

pkuwyn:
I asked the same in slack;
the keystone team said:"This feature isn’t fully operational yet ,We’ll look into this next sprint."

dcousens:
Relevant discussion for fields (not yet implemented) can be found at https://github.com/keystonejs/keystone/discussions/5875

dcousens:
Fixed in #7537

PR #7537, opened 2022-05-16
Display list descriptions in the Admin UI

Achisingh:
Fixed the display of lists `description` in Admin UI #7325 

Screenshot 2022-05-16 at 3 47 12 pm
Co-authored-by: Mitchell Hamilton <mitchellhamilton@users.noreply.github.com>

### DIFF
--- a/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
+++ b/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
@@ -248,7 +248,9 @@ const ListPage = ({ listKey }: ListPageProps) => {
         'Error...'
       ) : data && metaQuery.data ? (
         <Fragment>
-          {list.description !== null && <p>{list.description}</p>}
+          {list.description !== null && (
+            <p css={{ marginTop: '24px', maxWidth: '704px' }}>{list.description}</p>
+          )}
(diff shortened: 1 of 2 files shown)


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -248,6 +248,7 @@ const ListPage = ({ listKey }: ListPageProps) => {
         'Error...'
       ) : data && metaQuery.data ? (
         <Fragment>
+          {list.description !== null && <p>{list.description}</p>}
```
raveling, line 251: 24px top margin, and 704px max width please. Otherwise great! 🤚 

image

raveling (changes requested):
Small style changes please. Otherwise great!

packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
Achi06: Applied the above stylus changes in https://github.com/keystonejs/keystone/pull/7537/commits/479624a28e4fc4f7550a64edf0920f73105b6527 commit

dcousens (approved):
LGTM

PR #7538, opened 2022-05-16
Add support for MySQL

borisno2:
## Adding support for MySQL
Allows the setting of `db.provider` to `mysql` in `keystone.ts` config. 

URL for MySQL as per [Prisma connection details](https://www.prisma.io/docs/concepts/database-connectors/mysql)

Things to keep in mind when using MySQL - doc page added to cover this as well:

- When using `autoincrement` MySQL requires the field to also be indexed - added check to ensure this is set when using MySQL.
- Added tests for `mysql` to match current `postgresql` tests
- MySQL is not case sensitive - currently, the `postgres` tests are case sensitive and `sqlite` skips the `string` tests which is also what this current PR does for `mysql` - this is not a blocker for this PR, however, an issue is logged - #7574 - to complete this work. Also adding documentation
- Prisma string uses varchar(191) in MySQL instead of text in Postgres - added `nativeType` to `text` field so this can be easily overridden.


mitchellhamilton:
>  Prisma error different tests/api-tests/fields/unique.test.ts

For this, we should expect whatever it returns based on the database we're using. (Potentially in the future, we might want to have a consistent thing rather than what Prisma gives us but that is a totally separate thing to supporting MySQL)

>  Too many keys specified; max 64 keys allowed in tests/api-tests/access-control/schema.test.ts

I'm guessing that's because those tests create a huge number of lists. The fix here is probably going be to have a bunch of keystone instances with 1 list(or however many is convenient for how the tests are written) vs just a single keystone instance with a bunch of lists

dcousens:
A few thoughts on this,  to be wary of

For POSTGRESQL, `NULL` is a distinct value from the empty string `''`,  but in **MySQL the empty string is equivalent to NULL.**
This might explain some of the tests failures in our string comparisons/filtering; in addition to case sensitivity differences.

For the failing filtering tests, let's push this pull request forward and skip them for now,  with a follow GitHub issue to resolve that.
In any event, we want our API documentation for the `text` field to be updated to reflect these differences, preferably by linking to relevant Prisma documentation generically, but maybe specifically linking to MySQL documentation if necessary.

The same is for the case sensitivity, as MySQL is insensitive by default, we don't want to be too smart here, we can link to these differences in our API documentation,  again, preferably linking to relevant Prisma documentation where possible.

Lastly,  the `varchar(191)` difference,  interesting,  that might be painful.  Same approach for our API documentation.
That said,  adding support for a `{field}.db.nativeType` where you can provide a literal string that refers to the underlying database type,  **would help others who want a `text` field that might not be limited to 191 characters.**
For auto-complete and Typescript support, we could use a template literal type for these.

docs/components/docs/Navigation.tsx
```diff
@@ -161,6 +161,9 @@ export function DocsNavigation() {
         <PrimaryNavItem href="/docs/guides">Guides</PrimaryNavItem>
         <NavItem href="/docs/guides/cli">Command Line</NavItem>
         <NavItem href="/docs/guides/relationships">Relationships</NavItem>
+        <NavItem href="/docs/guides/choosing-a-database">
+          Choosing a Database<Badge look="success">New</Badge>
```
raveling, line 165:
```suggestion
          Choosing a Database <Badge look="success">New</Badge>
```

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,52 @@
+import { Markdown } from '../../../components/Markdown';
+import { Well } from '../../../components/primitives/Well';
+import { RelatedContent } from '../../../components/RelatedContent';
+
+# Choosing the right Database
+
+Keystone supports Postgres, MySQL and SQLite as database engines for your backend. There might be different reasons why you would choose one over the other.
```
raveling, line 7: Should we link to something for `Postgres`, `MySQL`, `SQLite`?

Perhaps other parts of our documentation that show how to configure each? 🤷

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+This page seeks to clarify some of the differences in the databases supported by Keystone.
+
+!> **Note:** SQLite is not recommended in production unless for readonly senarios like the [embedded Keystone](../walkthroughs/embedded-mode-with-sqlite-nextjs) example
```
raveling, line 11: Is the use of `readonly` consistent with rest of docs?

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,52 @@
...
+see [Primsa Schema Reference](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#model-field-scalar-types).
+
+The Keystone `text` field supports the `db.nativeType` option, allowing you to override this - [Fields API](../apis/fields). Over time this option will be added to other field types.
+
+## Prisma Resources
+
+- [Database Features](https://www.prisma.io/docs/reference/database-reference/database-features)
+- [Supported Databases](https://www.prisma.io/docs/reference/database-reference/supported-databases)
+
+!> **Note**: Keystone currently supports Postgres, MySQL, SQLite
```
raveling, line 33: Not sure we need this due to the presence in the top of page callout.

If you'd like to keep it, suggest including `Prisma` somewhere to situate it in the Prisma H2 in a more obvious way.

docs/pages/docs/guides/choosing-a-database.mdx
raveling, line 7:
```suggestion
Keystone supports Postgres, MySQL and SQLite database engines. This guide highlights the differences between these engines to help you choose the right one for your project.
```

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,52 @@
+import { Markdown } from '../../../components/Markdown';
+import { Well } from '../../../components/primitives/Well';
+import { RelatedContent } from '../../../components/RelatedContent';
+
+# Choosing the right Database
+
+Keystone supports Postgres, MySQL and SQLite as database engines for your backend. There might be different reasons why you would choose one over the other.
+
+This page seeks to clarify some of the differences in the databases supported by Keystone.
```
raveling, line 9:
```suggestion
```

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,52 @@
...
+  </Well>
+  <Well
+    heading="DB Config API"
+    href="../apis/config#db"
+    >
+    Documentation for the DB Config API 
+  </Well>
+</RelatedContent>
+
+export default ({ children }) => <Markdown description="Learn which database is right for your project">{children}</Markdown>;
```
raveling, line 52:
```suggestion
export default ({ children }) => <Markdown description="How to choose the right database for your Keystone project">{children}</Markdown>;
```

raveling (changes requested):
looking great!
a few small suggestions for the guide

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,51 @@
+import { Markdown } from '../../../components/Markdown';
+import { Well } from '../../../components/primitives/Well';
+import { RelatedContent } from '../../../components/RelatedContent';
+
+# Choosing the right Database
+
+Keystone supports Postgres, MySQL and SQLite database engines. This guide highlights the differences between these engines to help you choose the right one for your project.
+
+
+!> **Note:** SQLite is not recommended in production unless for readonly senarios like the [embedded Keystone](../walkthroughs/embedded-mode-with-sqlite-nextjs) example
```
borisno2, line 10:
```suggestion
!> **Note:** SQLite is not recommended in production except for scenarios like the [embedded Keystone](../walkthroughs/embedded-mode-with-sqlite-nextjs) example
```

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+# Choosing the right Database
+
+Keystone supports Postgres, MySQL and SQLite database engines. This guide highlights the differences between these engines to help you choose the right one for your project.
+
+!> **Note:** SQLite is not recommended in production except for scenarios like the [embedded Keystone](../walkthroughs/embedded-mode-with-sqlite-nextjs) example
+
+## Case Sensitivity
+
+By default, Postgres is case sensitive, while MySQL and SQLite are not case sensitive. This means when filtering and sorting the results will differ depending on the database type and collation used.
```
mitchellhamilton, line 13: SQLite is only case insensitive for `contains`, `startsWith` and `endsWith`

docs/pages/docs/guides/choosing-a-database.mdx
```diff
@@ -0,0 +1,51 @@
...
+When using `StringFilter` in a GraphQL query the `mode` defaults to that of the database - case sensitive for Postgres, and collation dependent for MySQL. You can make a query case insensitive by using `mode: insensitive`, there is no way to make a `StringQuery` case sensitive if your database is not case sensitive.
```
mitchellhamilton, line 14: `mode` only exists on Postgres. Also, `StringFilter` & `StringQuery` -> string filter imo

docs/pages/docs/guides/choosing-a-database.mdx
borisno2: Thanks @raveling I have just rewritten this section

docs/pages/docs/guides/choosing-a-database.mdx
borisno2: Thanks @mitchellhamilton, just updated

docs/pages/docs/guides/choosing-a-database.mdx
borisno2: Thanks @mitchellhamilton, I just rewrote this section a bit.

docs/pages/docs/guides/choosing-a-database.mdx
borisno2: Added some links...

mitchellhamilton (approved):

raveling (approved):

PR #7543, opened 2022-05-19
Fix the viewport shifting when opening the date picker in a modal

### LINKED ISSUE #6534: Date field shifts browser viewport when clicked
Affects date time field with date picker ONLY when in the create new item modal when clicked.

Not browser specific, but depending on the browser used, the viewport is moved in different widths.

**See video below:**

https://user-images.githubusercontent.com/737821/133006739-dbcc7dd2-0c7e-4a8a-900a-51e90dfd3f0b.mov

**Chrome:**

image

**Firefox:**

image

emmatown:
This probably isn't the _ideal_ fix but it works fine and figuring out positioning things is annoying.

Closes #6534

### DIFF
--- a/.changeset/lazy-knives-hammer.md
+++ b/.changeset/lazy-knives-hammer.md
@@ -1,0 +1,5 @@
+---
+'@keystone-ui/popover': patch
+---
+
+Fixed bug where focustrap wasn't deactivating on unmount.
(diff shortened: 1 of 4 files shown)


gwyneplaine (approved):

.changeset/lazy-knives-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-ui/popover': patch
+---
+
+Fixed bug where focustrap wasn't deactivating on unmount.
```
mitchellhamilton, line 5:
```suggestion
Fixed bug where focus trap wasn't deactivating on unmount.
```

PR #7547, opened 2022-05-20
Fix document editor preventing tabbing out of the editor

mitchellhamilton:
This is slightly different logic than we discussed, it's "allow the event normally to continue normally(and therefore focus the next element) if the tab in a list didn't do anything" rather than "don't allow the event to continue if in a list, otherwise allow it to continue" but I think that's probably Fine and Good?


gwyneplaine (approved):
yeah this seems fine to me

PR #7549, opened 2022-05-20
Associate document editor label with editable element

mitchellhamilton:
`<label>` is not for `contenteditable` elements, I've used `aria-labelledby`.


gwyneplaine (approved):
nice

Issue #6157, opened 2021-07-21
Wrong z-index for dropdown in a document field


Chaptykov:
# Bug report

## Describe the bug

Dropdown list appears under a form footer. 

## To Reproduce

1. Add document field
2. Hover „More formatting“

## Expected behaviour

Dropdown list appears above UI. 

## Screenshots

![image](https://user-images.githubusercontent.com/1920602/126560885-9fe6aad7-8e3f-439f-b53c-20c12b1f12cd.png)

## System information

- OS: OS X
- Browser: Chrome

bladey:
Thanks @Chaptykov!

dcousens:
(An earlier) Duplicate of https://github.com/keystonejs/keystone/issues/7395

PR #7556, opened 2022-05-24
Fix z-index issues for document editor text-area and toolbar

Achisingh:
The document field toolbar that was over-lapping the Action Bar at the bottom (i.e. Save Changes, Delete Button) caused by z-index is fixed by using `<Portal>` instead. This also resolves z-index issues within the text-area (e.g. the popover for links in the document editor is now visible in the create modal). #6157 #7464


gwyneplaine (approved):
nice one! lgtm

PR #7564, opened 2022-05-25
Update Keystone display images on website

mattbx:
Site images updated based on

https://github.com/keystonejs/keystone-private/issues/83


raveling (changes requested):
looking great!

the "why keystone" image needs to be a transparent PNG in order to render properly when the website is set to dark mode.

image

raveling (approved):
LGTM!

PR #7591, opened 2022-06-02
Support body-parser options when configuring GraphQL

dcousens:
Following the discussion in https://github.com/keystonejs/keystone/discussions/7582

[`body-parser`](https://www.npmjs.com/package/body-parser) has a number of configuration options that may be useful for developers to configure themselves. It looks like [Apollo 4 will be changing](https://github.com/apollographql/apollo-server/commit/81c7f7b2692d57bb45e3755dc44e7e3bfe1a03c4) how express interacts with user code in future so we'll need to do this anyway; might as well do it now.

### DIFF
--- a/packages/core/src/lib/server/createExpressServer.ts
+++ b/packages/core/src/lib/server/createExpressServer.ts
@@ -48,6 +48,7 @@ const addApolloServer = async ({
     app: server,
     path: config.graphql?.path || '/api/graphql',
     cors: false,
+    bodyParserConfig: config.graphql?.bodyParser === undefined ? true : config.graphql?.bodyParser,
   });
   return apolloServer;
 };
--- a/packages/core/src/lib/server/createExpressServer.ts
+++ b/packages/core/src/lib/server/createExpressServer.ts
@@ -48,7 +48,7 @@ const addApolloServer = async ({
     app: server,
     path: config.graphql?.path || '/api/graphql',
     cors: false,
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/server/createExpressServer.ts
```diff
@@ -48,6 +48,7 @@ const addApolloServer = async ({
     app: server,
     path: config.graphql?.path || '/api/graphql',
     cors: false,
+    bodyParserConfig: config.graphql?.bodyParser === undefined ? true : config.graphql?.bodyParser,
```
mitchellhamilton, line 51: Is `true` equivalent to the default? and if so, why doesn't just `config.graphql?.bodyParser` work?

dcousens: `true` is equivalent to the `express` defaults, and anything `!== false` is explicitly provided.

Reference https://github.com/apollographql/apollo-server/blob/e5144f160f4d49f9651473ff092c7ae11d433d82/packages/apollo-server-express/src/ApolloServer.ts#L128-L132

That said, reading https://github.com/expressjs/body-parser/blob/1f6f58e1f8dc222f2b6cfc7eb3a3bf5145ff2b56/lib/types/json.js#L51 - it doesn't matter anyway.

mitchellhamilton (commented):
Other than adding a changeset, this looks good to me

mitchellhamilton (approved):

PR #7593, opened 2022-06-03
Fix website search, breadcrumbs popping in, key warning and hydration mismatch

emmatown:
Was just trying to solve the search and got sidetracked 🙃


borisno2 (approved):
Thanks @mitchellhamilton

PR #7595, opened 2022-06-03
Ensure Prisma engines exist before attempting to format a Prisma schema

emmatown:
This should resolve the issues discussed in https://github.com/keystonejs/keystone/discussions/7461.

This will also make Keystone work when using `--ignore-scripts` with a package manager.

Note we always format the schema before attempting to do migrations and generate a Prisma client so this will ensure the engines will exist whenever they're necessary.

### DIFF
--- a/.changeset/dirty-apples-build.md
+++ b/.changeset/dirty-apples-build.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-The Prisma binaries are now downloaded just before they're needed if the Prisma's install script to download them fails. Note this will never happen in production, they will always be downloaded before.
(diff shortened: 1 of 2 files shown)


.changeset/dirty-apples-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+The Prisma binaries are now downloaded just before they're needed if the Prisma's install script to download them fails. Note this will never happen in production, they will always be downloaded before.
```
mitchellhamilton, line 5:
```suggestion
The Prisma binaries are now downloaded just before they're needed if Prisma's install script to download them fails. Note this will never happen in production, they will always be downloaded before.
```

borisno2 (approved):

PR #7616, opened 2022-06-10
Fix spacing on relationships fields using the cards display mode

mitchellhamilton:
See the sandbox to see what it looks like now


raveling (approved):
LGTM 🙌

PR #7621, opened 2022-06-13
CI cleanup

emmatown:
This is basically two things:
- Defining how to install node and cache & install deps in one place
- Only starting services for things that actually need them 

Note that GitHub will be like "Some checks haven’t completed yet" since various jobs now have different names, that's expected, I'll bypass the checks when merging and then update the branch protections with the new names.


borisno2 (approved):
Looks good, thanks @mitchellhamilton

PR #7639, opened 2022-06-16
Hide clear for segmented-control

### LINKED ISSUE #7413: select shouldn't show Clear button if `validation.isRequired` is true
Define a field like

```
      status: select({
        options: [
          { label: "Published", value: "published" },
          { label: "Draft", value: "draft" },
        ],
        defaultValue: "draft",
        validation: {
          isRequired: true,
        },
        ui: {
          displayMode: "segmented-control",
        },
      }),
```

Since the field is required, and has a default value, I expected the UI to show a widget that lets the user pick one of the options, but *be forced to pick one*.

Right now, user can click "Clear" and "Save changes" and get an error message "Status is required". But why let them get into that dead end, in the first place?

image

u-ishii:
Fix #7413 

Hide the clear button if is-required is true for segmented-control.

I confirmed that this fix works as expected in sandbox, but if you find a problem, please comment.

### DIFF
--- a/.changeset/forty-stingrays-destroy.md
+++ b/.changeset/forty-stingrays-destroy.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Hide the clear button if is-required is true for segmented-control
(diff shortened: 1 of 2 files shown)


dcousens (approved):

dcousens:
Thanks for this @u-ishii!

PR #7657, opened 2022-06-23
Add ui.{label,plural,singular,path}

### LINKED ISSUE #7385: Error: Unable to use Firmware as a List name - it has an ambiguous plural (Firmware).
As the title described I got:
`
Error: Unable to use Firmware as a List name - it has an ambiguous plural (Firmware). Please choose another name for your list.`

And here is my list:

```
Firmware: list({
    fields: {
      name: text({ validation: { isRequired: true } }),
    }
  }),
```

I am not sure what's wrong with the name. `Firmware`'s plural is still the same I think, but how to solve it? And I got this link in Stackoverflow https://stackoverflow.com/questions/49771957/keystonejs-plural-option-not-working-in-adminui

but seems it's V5 version solution.

Thanks!

emmatown:
This pull request removes the error for ambiguous plurals on lists with ambiguous names.
For example [`Shrimp` (singular) and `Shrimp` (plural)](https://github.com/keystonejs/keystone/discussions/7550) and [`Firmware` (singular) and `Firmware` (plural)](#7385 ).

Only the GraphQL plural needs to be unambiguous.

A warning that configuring any of the `ui` options does not affect the GraphQL plural; and vice-versa.
An example on how this might look:
``` typescript
export const lists = {
  Shrimp: list({
    fields: {
      name: text(),
    },
    graphql: {
      plural: 'Shrimps'
    },
    // ui: { // un-comment to change the default plural from `Shrimp` in the admin interface
    //   plural: 'Shrimps' 
    // },
  })
}
```

- Closes https://github.com/keystonejs/keystone/discussions/6231
- Closes #7385 
- Closes #7550


packages/core/src/lib/core/utils.ts
```diff
@@ -121,48 +121,35 @@ export async function promiseAllRejectWithAllErrors<T extends unknown[]>(
...
+  return { pluralGraphQLName, adminUILabels };
 }
 
-const keyToLabel = (str: string) => {
-  let label = humanize(str);
-
-  // Retain the leading underscore for auxiliary lists
-  if (str[0] === '_') {
-    label = `_${label}`;
-  }
```
mitchellhamilton, line 162: This was about a special kind of list that was magically created for certain things in Keystone 5 which does not exist in Keystone 6 so it's no longer relevant

borisno2 (approved):

PR #7658, opened 2022-06-23
Add calendarDay field

Achisingh:
This adds a new field type which store a date without a time or timezone. 
On PostgreSQL and MySQL this is represented with the native date type.
On SQLite this is represented as a string.
- https://github.com/keystonejs/keystone/discussions/7473

~The naming of `plainDate` is based on [the PlainDate type in the Temporal proposal ](https://tc39.es/proposal-temporal/docs/plaindate.html).~

### DIFF
--- a/.changeset/strange-panthers-eat.md
+++ b/.changeset/strange-panthers-eat.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Added `calendarDay` field to store a date without a time or timezone attached
--- a/docs/pages/docs/apis/fields.mdx
+++ b/docs/pages/docs/apis/fields.mdx
@@ -657,6 +657,57 @@ export default config({
 });
 ```
 
+### calendarDay
+
+A `calendarDay` field represents a date value only in ISO8601 format.
+
+On PostgreSQL and MySQL this is represented with the native date type.
+On SQLite this is represented as a string.
+
+Options:
+
+- `defaultValue` (default: `undefined`): Can be a string value with a date string in ISO8601 format.
+  This value will be used for the field when creating items if no explicit value is set.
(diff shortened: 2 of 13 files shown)


packages/core/src/fields/types/plainDate/index.ts
```diff
@@ -0,0 +1,214 @@
...
+        type: config.graphql?.read?.isNonNull
+          ? graphql.nonNull(graphql.PlainDate)
+          : graphql.PlainDate,
+        resolve({ value }) {
+          if (value instanceof Date) {
+            return formatISO(
+              new Date(value.getUTCFullYear(), value.getUTCMonth(), value.getUTCDate()),
+              { representation: 'date' }
+            );
+          }
```
dcousens, line 146: This is OK up until year 10000,  at which point, all of this code will break anyway

```suggestion
            return value.toISOString().slice(0, 10)
          }
```

dcousens:
I think we should place this in in a `temporal` namespace if we are going to follow the naming conventions from [that proposal](https://github.com/tc39/proposal-temporal).

Although, at this time we don't actually use any code related to that proposal, I think the concepts becoming the canonical Javascript vernacular is quite likely.  If anything changes, we can easily change that too.

For now, that would mean having
- `plainDate` as `temporal.plainDate`
- `timestamp` as `temporal.instant`

I'm happy to open this in another pull request though and gather feedback on that there.

dcousens:
We didn't reach a consensus on https://github.com/keystonejs/keystone/pull/7658#issuecomment-1170717471, and thus for now we are going to move forward with this as `calendarDay`.
That name may change in the future, but we consider that naming to be more intuitive than `plainDate` when absent from a like-TC39 namespace.

dcousens (approved):

Issue #7659, opened 2022-06-23
schema.prisma is broken when you declare a map + index


TonnyORG:
I’m getting a weird prisma error here, I'm not sure how to fix it:

```
    deletedAt: timestamp({
      db: {
        isNullable: true,
        map: 'deleted_at',
      },
      isIndexed: true,
      validation: { isRequired: false },
    }),
```

turned into:
```
  deletedAt DateTime?
  @@index([deletedAt]) @map("deleted_at")
```

and throw this:
```
error: Error validating: This line is not a valid field or attribute definition.
  -->  schema.prisma:47
   | 
46 |   deletedAt DateTime?
47 |   @@index([deletedAt]) @map("deleted_at")
48 |   updatedAt DateTime? @updatedAt
```

It should build the  the deletedAt field in a single line but it's doing line-break. If I manually "edit" the `schema.prisma` file this way it does work.
```
  deletedAt DateTime? @@index([deletedAt]) @map("deleted_at")
```

TonnyORG:
PR sent: #7664

PR #7664, opened 2022-06-25
Resolve issue with line-break for the @@map instruction for prisma #7659

TonnyORG:


mitchellhamilton:
Thanks for the issue and this PR! The issue will be fixed by #7666, the issue ended up being different to what's in this PR

TonnyORG:
@mitchellhamilton sounds good, tyvm for your help here!

PR #7683, opened 2022-07-01
Add multiselect field

Achisingh:
This adds a `multiselect` field that mirrors the `select` field options. 

-  It is always non-nullable
- It is stored as a json array in databases

In this PR it does not have any filters and it is not orderable. It may change in the future PRs.

### DIFF
--- a/.changeset/light-balloons-repair.md
+++ b/.changeset/light-balloons-repair.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Adds `multiselect` field
--- a/.changeset/light-balloons-repair.md
+++ b/.changeset/light-balloons-repair.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': minor
 ---
 
-Adds `multiselect` field
+Adds a new `multiselect` field type
--- a/docs/pages/docs/apis/fields.mdx
+++ b/docs/pages/docs/apis/fields.mdx
@@ -544,6 +544,58 @@ export default config({
 });
 ```
 
+### multiselect
+
+A `multiselect` field represents the selection of a set of values from the defined `options`.
+Values can be either strings, integers, or enum values, as determined by the `type` option.
+This will determine their GraphQL data type.
+Unlike the `select` field, the `type` will not change the database type, `multiselect` fields are always stored as a json array in the database.
+
+Options:
+
+- `type` (default: `'string'`): Sets the type of the values of this field.
+  Must be one of `['string', 'enum', 'integer']`.
+- `options`: An array of `{ label, value }`.
+  `label` is a string to be displayed in the Admin UI.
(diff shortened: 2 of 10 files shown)


docs/pages/docs/apis/fields.mdx
```diff
@@ -486,6 +486,58 @@ export default config({
...
+  when access is denied, `null` will be returned which will cause an error since the field is non-nullable and the error
+  will propagate up until a nullable field is found which means the entire item will be unreadable and when doing an `items` query, all the items will be unreadable.
+- `graphql.create.isNonNull` (default: `false`): If you have no create access control and you want to explicitly show that this is field is non-nullable in the create input
+  you can set this to true and the create field will be non-nullable and have a default value at the GraphQL level.
+  This is only allowed when you have no create access control because otherwise, the item will always fail access control
+  if a user doesn't have access to create the particular field regardless of whether or not they specify the field in the create.
+
+```typescript
+import { config, list } from '@keystone-6/core';
+import { select } from '@keystone-6/core/fields';
```
mitchellhamilton, line 517:
```suggestion
import { multiselect } from '@keystone-6/core/fields';
```

gautamsi:
what about adding this as `,` separated string in database? this `,` should be configurable `delimiter`

caveat is that search for `new` will also fetch `news` but that can have workaround

packages/core/src/fields/types/multiselect/index.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export const multiselect =
+  <ListTypeInfo extends BaseListTypeInfo>({
+    ui,
+    defaultValue = [],
+    ...config
+  }: MultiselectFieldConfig<ListTypeInfo>): FieldTypeFunc<ListTypeInfo> =>
+  meta => {
+    if ((config as any).isIndexed === 'unique') {
+      throw Error("isIndexed: 'unique' is not a supported option for field type password");
```
borisno2, line 61: Should this be 'field type multiselect'?

mitchellhamilton:
> what about adding this as , separated string in database? this , should be configurable delimiter

What would be the advantage of doing that?

packages/core/src/fields/types/multiselect/index.ts
mitchellhamilton:
```suggestion
      throw Error("isIndexed: 'unique' is not a supported option for field type multiselect");
```

gautamsi:
>What would be the advantage of doing that? 

* this avoids json fields and you can add Filters in this.

If there are filters (maybe I missed) for multiselect value combinations I am good with this.

.changeset/light-balloons-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Adds `multiselect` field
```
dcousens, line 5:
```suggestion
Adds a new `multiselect` field type
```

dcousens (approved):

rafenden:
Any updates on the filters?

PR #7701, opened 2022-07-11
Update Slate

emmatown:
Nothing very interesting here, mostly just removing some code around soft breaks because Slate added a `insertSoftBreak` method on the `Editor` type which `slate-react` will call on <kbd>shift</kbd>+<kbd>enter</kbd> (though `Editor.insertSoftBreak` defaults to inserting a hard break so we still have to override it)

### DIFF
--- a/.changeset/smooth-mugs-divide.md
+++ b/.changeset/smooth-mugs-divide.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/fields-document': patch
+---
+
+Updates Slate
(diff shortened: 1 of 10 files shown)


.changeset/smooth-mugs-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/fields-document': patch
+---
+
+Updates Slate
```
dcousens, line 5:
```suggestion
Updates slate and slate-react to ^0.81.1
```

dcousens (approved):

PR #7575, opened 2022-05-31
Singletons

Noviny:
This pull request introduces [Singletons](https://en.wikipedia.org/wiki/Singleton_pattern) as a new feature for configuring global, editable database entries when it's not necessary to have more than one representation of the schema (aka, not a normal list).

Singletons can be useful for things like: 

- System Settings
- Homepage Layout
- Site Navigation

Before this feature, developers could write a typical schema for a list, and then using Keystone hooks limit the number of items within that list to approximate this functionality.
Unfortunately the GraphQL schema didn't reflect these constraints,  and neither did the AdminUI.

This pull requests brings that functionality native to Keystone's schema configuration as a first class primitive.

### Tasks
- [ ] Add enforcement of the databases (where able) to ensure that the singleton list only ever has one item
- [x] Alter the graphql layer so that queries on singletons make more sense (remove querying for multiple items, remove the ‘where’ query on fetching a single item, as there will only ever be one)
- [x] Update the Admin UI - mostly subtractive changes to give a good experience for singletons where navigating items UI is removed, removing clicks

PR #7733, opened 2022-07-21
Rename `.keystone/types` to `.keystone` and add `graphql` export to `.keystone`

emmatown:
This adds a `graphql` export to `.keystone` (renamed from `.keystone/types` because with the `graphql` export, it is no longer just types) that is the same as the `graphql` export from `@keystone-6/core` but uses the specific generated context type for the project. This is important for #7575 since not using the specific context will be more painful but it also generally improves the experience of using graphql-ts for schema extensions and virtual fields.


flexdinesh (approved):

.changeset/bright-pots-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Renames `.keystone/types` to `.keystone` and adds `graphql` export to `.keystone` that is the same as the `graphql` export from `@keystone-6/core` but uses the specific generated context type for the project
```
dcousens, line 5:
```suggestion
Moves `.keystone/types` to `.keystone` and adds `graphql` export to `.keystone` that is the same as the `graphql` export from `@keystone-6/core` but uses the specific generated context type for the project
```

dcousens:
I don't think we want to do this at this time.

That may change in the future, but for now I will close this pull request.

PR #7744, opened 2022-07-28
Lockfile update

emmatown:
This is part "use fixed versions of things so that our lockfile is fresh"  and part "fix the things on updated versions"


packages/core/src/admin-ui/components/__tests__/Pagination.test.tsx
```diff
@@ -1,122 +0,0 @@
...
- */
-
-import React from 'react';
-import userEvent from '@testing-library/user-event';
-import { render, screen } from '@testing-library/react';
-import { Pagination } from '../Pagination';
-const useRouter = jest.spyOn(require('next/router'), 'useRouter');
-describe('<Pagination/>', () => {
-  it('should render a nav element with role navigation', () => {
-    useRouter.mockImplementation(() => ({
```
mitchellhamilton, line 12: I think that the cost of fixing these tests (now and over time) would be more costly than the benefit of the tests so I just deleted them

packages/core/package.json
```diff
@@ -111,7 +110,7 @@
     "normalize-path": "^3.0.0",
     "object-hash": "^3.0.0",
     "p-limit": "^2.3.0",
-    "pirates": "^4.0.1",
+    "pirates": "4.0.4",
```
mitchellhamilton, line 113: This is a "it's our fault, we were relying on unintentional behavior that broke now" thing but I don't want to deal with it in this PR

.changeset/chatty-needles-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Pins `@types/graphql-upload` to `0.8.7`
```
dcousens, line 5:
```suggestion
Pins `@types/graphql-upload` to `8.0.7`
```

dcousens (approved):

PR #7750, opened 2022-07-31
Add an example of seeding the first user

zizzfizzix:
Add documentation to seed the first user during development for people frequently resetting their database, so they don't have to set up a user every time.

### DIFF
--- a/docs/pages/docs/config/auth.md
+++ b/docs/pages/docs/config/auth.md
@@ -194,6 +194,31 @@ const { withAuth } = createAuth({
 });
 ```
 
+#### Bootstrapping a default user for local development at seed stage
+
+If you frequently reset your database during development and get frustrated having to create a user each time you want to log in, you can bootstrap a user in your seeding script.
+
+```typescript
+// seed-admin-user.ts
+import { KeystoneContext } from '@keystone-6/core/types';
+
+export default async (context: KeystoneContext) => {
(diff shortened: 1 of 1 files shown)


docs/pages/docs/apis/auth.mdx
```diff
@@ -195,6 +195,31 @@ const { withAuth } = createAuth({
...
+
+If you frequently reset your database during development and get frustrated having to create a user each time you want to log in, you can bootstrap a user in your seeding script.
+
+```typescript
+// seed-admin-user.ts
+import { KeystoneContext } from '@keystone-6/core/types';
+
+export default async (context: KeystoneContext) => {
+  // For local development convenience bootstrap the first admin user
+  if (process.env.NODE_ENV !== 'production') {
```
dcousens, line 208: Rather than environment, we tend to recommend `process.argv`, as seen in https://github.com/keystonejs/keystone/blob/44d94482770bdfbaf1be2e224912df854f7ee934/examples/blog/keystone.ts

```suggestion
  if (process.argv.includes('--seed-data')) {
```

dcousens:
Thinking about this @zizzfizzix,  I think we should do everything we can to show better data seeding practices in the examples rather than in the documentation.
If users are trying the project for the first time, we don't want to introduce any confusion about whether they _need_ to seed data or not.

PR #7766, opened 2022-08-03
Support pasting a url onto text to create a link in the document editor

emmatown:
It's a nice thing that lots of other editors do, so let's do it as well


JedWatson (approved):
Nice tests 👌

PR #7771, opened 2022-08-04
Updated redis tests to fail when redis is not running

Noviny:
Previously they would just hang indefinitely

### DIFF
--- a/tests/api-tests/auth-stored-session.test.ts
+++ b/tests/api-tests/auth-stored-session.test.ts
@@ -7,6 +7,7 @@ import { createClient } from '@redis/client';
 import { createAuth } from '@keystone-6/auth';
 import type { KeystoneContext } from '@keystone-6/core/types';
 import { setupTestRunner, TestArgs } from '@keystone-6/core/testing';
+import fetch from 'node-fetch';
 import { apiTestConfig, expectAccessDenied, seed } from './utils';
 
 const initialData = {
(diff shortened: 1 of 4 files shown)


Noviny:
Currently unsure if we should update `packages/session-store-redis/src/index.ts` line 12, which is

```
    async connect() {
      client.on('error', err => console.log('Redis Client Error', err)); // this one

      await client.connect();
    },
```

and sort of just swallows errors, and seems to lead to reattempting rather than failing out.

tests/api-tests/auth-stored-session.test.ts
```diff
@@ -99,6 +100,19 @@ async function login(
 }
 
 describe('Auth testing', () => {
+  beforeAll(async () => {
+    try {
+      await fetch('http:127.0.0.1:6379', {
```
dcousens, line 105:
```suggestion
      await fetch('http://127.0.0.1:6379', {
```

tests/api-tests/auth-stored-session.test.ts
```diff
@@ -99,6 +100,19 @@ async function login(
...
 describe('Auth testing', () => {
+  beforeAll(async () => {
+    try {
+      await fetch('http://127.0.0.1:6379', {
+        method: 'post',
+        body: 'ping',
+      });
+    } catch (e) {
+      if ((e as Error).message.includes('ECONNREFUSED')) {
+        throw new Error('redis is not running on this computer so this test suite cannot be run');
```
dcousens, line 111:
```suggestion
        throw new Error('redis not found: ECONNREFUSED');
```

dcousens: if debugging a redis instance, this is usually an easier search term

dcousens (approved):

dcousens (approved):

PR #7796, opened 2022-08-15
Add informations in CONTRIBUTING.md on how to install a dev version of this repo

anthonyivol:
No changeset has been done because this PR doesn't concern any packages but the CONTRIBUTING.md file only.


CONTRIBUTING.md
```diff
@@ -27,6 +27,81 @@ KeystoneJS adheres to the [Contributor Covenant Code of Conduct](/CODE-OF-CONDUC
 
 KeystoneJS follows the [Thinkmill Monorepo Style Guide](https://github.com/Thinkmill/monorepo). For more information on the reasoning behind using certain tooling, please refer to it.
 
+### Installation
```
dcousens, line 30: We should be verbose that these steps are for linking a local copy (or fork) of the `keystone-6/*` monorepo for local development in an external project.

If you are contributing to `keystone-6/*` generally, you don't need to follow these steps beyond `git clone https://github.com/keystonejs/keystone.git` and `yarn` in the monorepo root.

anthonyivol: Well I missed `packages/core/bin/cli.js dev` to run keystone from the monrepo as `keystone dev` return `keystone command not found`. Maybe adding a script line "dev" in package.json would help. You also need to create a config file at the monorepo root to run it. Or I missed something.

Actually I first try to run a fork as an Embed keystone in next with `withKeystone`, this is why I messed up (and this is why I did first a commit to prevent using env.PORT as it is used by next).

So possible changes :
- For clarity : add a script `"dev": "packages/core/bin/cli.js dev"` command in package.json at monorepo root. Add information about the config file and `yarn dev` after `yarn install` in CONTRIBUTING.md
- complete the CONTRIBUTING.md example as you suggest for those who want to work with an embed keystone in next or an external project.
- create an issue about env.PORT priority in `packages/core/src/scripts/run/dev.ts` which is used by next and not configurable without .env

dcousens:
@anthonyivol if you're going to update the branch, please rebase not merge, reviewing the diff becomes difficult otherwise :sos:

dcousens:
@anthonyivol why did you open a different pull request?

anthonyivol:
It has been automatically closed after `git reset --hard upstream/main` & `git push origin main --force`

dcousens:
Shame, next time you can use `git rebase upstream/main --interactive` :+1:

PR #7805, opened 2022-08-17
Change view module resolution

mitchellhamilton:
This pull request changes field type `.views` resolution from being path based to a module path.
This means instead of using `require.resolve(...)` to resolve your custom views, you will now use a path as though it was resolving from where `keystone dev` is run.

This additionally means you can use typical modules or packages without needing relative or absolute pathing workarounds.

For example, if you wanted to use the `@keystone-6` text field views for your custom field type, you can now do that through canonical module resolution:

```typescript
  fieldType({ ... })({
    input: { ... },
    output: graphql.field({ ... }),
    views: '@keystone-6/core/fields/types/text/views',
  })
```

PR #7809, opened 2022-08-18
Use esbuild to build server-side code

emmatown:
Requires #7805

While reviewing https://github.com/keystonejs/keystone/pull/7744 - the `pirates` changes demonstrated that we should simplify the build process for end developer configurations.

Keystone passes compiling to `next/babel` at the moment, but that is executed in a way that isn't actually helping us drop the responsibility, nor the maintenance burden.  We are operating in a constrained environment, with no configuration options for our users. This could be simpler.

Some options
- `tsc`, slower 🐌 , no bundling :face_in_clouds: 
- `esbuild`, fast 💨, easy bundling 📦, limited configuration ❤ 

This pull request switches `@keystone-6/core` to use `esbuild` for the compilation process.

This change should ideally not break most users, the reality of modules in the JS ecosystem is that every tools does things ever so subtly different so there will likely be small breaks, the most notable change though is that `__dirname`, `__filename`, and `require.resolve` will not be treated specially, this essentially means in most cases, they should just be avoided. If you need to read a file/etc., you should do it relative to `process.cwd()`.


jlarmstrongiv:
Following up from Slack, are the new functions `loadBuiltConfig`, `getBuiltConfigPath`, and `getAdminPath` exported from any part of the @keysone-6/core package?  It’s very important to have those functions exposed.

dcousens:
Hi @jlarmstrongiv,  I don't think they are and we typically want to be conservative in what we `export`.  Did you have a particular use case in mind?

jlarmstrongiv:
@dcousens Ahh, I had missed that `initConfig` was already exported.  I can redefine what’s not included 👍 

I’m building my own “start” function to return the underlying express server for experimental aws lambda support and bundling.  If I can get that to work, then I should be able to make a pr for an experimental `generateExpressApp` and a bundling example.  Supporting serverless (even just experimentally or a community example) would be very nice and I think it’s doable.

mitchellhamilton:
Before merging this, we should make sure source maps work (we might want to skip source maps for the prod build but they should definitely work in dev)

mitchellhamilton:
Source maps turned into a rabbit hole, doing it Correctly would be very subtle so let's go with no source maps for now rather than wrong source maps. The bundle that esbuild generates is quite readable so I don't think it's a huge concern.

dcousens (approved):

PR #7841, opened 2022-08-24
Add a search input to the list view

Noviny:
This pull request adds a search input field to the list view.

The search input text will be used in a `contains` query across each of the `ui.searchFields`,  and then joined with the typical list view filters.

This feature may not be suitable for databases with large numbers of rows, as the `contains` query is not indexed; never-the-less the feature should be useful if the `.ui.searchFields` are curated and useful in context.

If `ui.searchFields` is undefined or the empty `[]`, the field defaults to searching by `.id`.

**This searching behaviour is unchanged from the behaviour of how the relationship field already operates.**


Noviny:
@dcousens proposed alternate implementation:

The search field is typed, and on enter, it clears the field, and adds the filters to the existing filters. This makes it more explicit what this does (apply filters) and encourages people to explore filters some more, but makes undo a bit weirder, stops live updating

mitchellhamilton (approved):
Might want to update the PR description but the rest of this all looks good to me

dcousens (approved):

PR #7844, opened 2022-08-24
Refactor Admin Meta GraphQL Schema

emmatown:
This refactoring is to make #7575 easier.

The only publicly visible change here is adding some things on to the types named `*RootVal`.
While these are technically publicly accessible, they should not have been used by developers as they are for the admin only.

We additionally remove some unused fields, `enableSessionItem` and `enableSignout`.


examples/custom-admin-ui-logo/schema.graphql
```diff
@@ -254,7 +254,6 @@ type KeystoneMeta {
 
 type KeystoneAdminMeta {
   enableSignout: Boolean!
```
dcousens, line 256: I don't think `enableSignout` is used either

flexdinesh (approved):

dcousens (approved):

PR #7846, opened 2022-08-25
Define Singleton Admin Meta GraphQL API 

emmatown:
This defines the GraphQL types needed for singletons in the admin meta and updates the usages to continue working for lists. The implementation assumes singletons are never actually used. The Admin UI and the back-end portions of singletons can be properly implemented independently after this.

Based on #7845

### DIFF
--- a/.changeset/fair-terms-beg.md
+++ b/.changeset/fair-terms-beg.md
@@ -1,0 +1,6 @@
+---
+'@keystone-6/core': patch
+'@keystone-6/fields-document': patch
+---
+
+Updates admin meta GraphQL schema to support singletons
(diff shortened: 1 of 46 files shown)


.changeset/fair-terms-beg.md
```diff
@@ -0,0 +1,6 @@
+---
+'@keystone-6/core': patch
+'@keystone-6/fields-document': patch
+---
+
+Updates admin meta GraphQL schema to support singletons
```
flexdinesh, line 6: @mitchellhamilton will this be a separate patch to support singleton or will this go out together with singleton release as a major?

flexdinesh (approved):
Probably the docs change for `lists` => `models` will be up in some other PR yeah?

.changeset/fair-terms-beg.md
mitchellhamilton: This will go out in the same release

PR #7848, opened 2022-08-25
Make operations permissions required

Noviny:
When looking at operations permissions, we have observed that this can often be misconfigured.
We additionally don't want to rely on `true` or `false` defaults that may be confusing or unexpected to users.

As a result, we are changing `access.operations` on lists to be explicitly required.

The basic change is the following:

```diff
- list({
-   access?: {
-     operation?: {
-       query?: ({ session, context, listKey, operation }) => boolean,
-       create?: ({ session, context, listKey, operation }) => boolean,
-       update?: ({ session, context, listKey, operation }) => boolean,
-       delete?: ({ session, context, listKey, operation }) => boolean,
-     }
-     ...
-   },
-   ...
- })
+ list({
+   access: {
+     operation: {
+       query: ({ session, context, listKey, operation }) => boolean,
+       create: ({ session, context, listKey, operation }) => boolean,
+       update: ({ session, context, listKey, operation }) => boolean,
+       delete: ({ session, context, listKey, operation }) => boolean,
+     }
+     ...
+   },
+   ...
+ })
```

To make this a bit easier however, we are adding two new options.

```ts
list({
  access: ({ session, context, listKey, operation }) => boolean
  ...
})
```

allows you to set all access controls at once, and

```ts
list({
  access: {
    operations: ({ session, context, listKey, operation }) => boolean,
    ...
  },
  ...
});
```

allows you to set all `operations` access controls for a list at once.

### New exports

We are providing three new helper exports to help out with this change so you can have code that reads well:

```ts
import { allowAll, denyAll, allOperations } from "@keystone-6/core/access"
```

These functions should allow quickly applying access controls to lists. For example, to get the existing default access controls, you could do:

```ts
list({
  access: allowAll,
  ...
});
```

or

```ts
list({
  access: allowAll,
  ...
});
```

```ts
list({
  access: { operations: allowAll },
  ...
});
```

If you were setting access for create operations, you can quickly update to:

```ts
list({
  access: { operations: { ...allOperations(allowAll), create: myCustomCreateAccessCheck } },
  ...
});
```


Noviny:
Commit history should be a good guide here in understanding actual changes vs updating all our various configs to comply with the change.

TODO:

- [ ] update docs to reflect the new reality
- [ ] changeset

packages/core/src/access/index.ts
```diff
@@ -0,0 +1,20 @@
...
+export function allowAll() {
+  return true;
+}
+export function denyAll() {
+  return false;
+}
+
+export function allOperations<ListTypeInfo extends BaseListTypeInfo>(
+  func: (args: BaseAccessArgs<ListTypeInfo>) => boolean
+) {
```
Noviny, line 13: @mitchellhamilton or @dcousens This type functions but getting here felt like flailing not knowledge, and would love some scrutiny and feedback of this.

packages/core/src/types/config/access-control.ts
```diff
@@ -63,37 +63,27 @@ export type DeleteListItemAccessControl<ListTypeInfo extends BaseListTypeInfo> =
   args: DeleteItemAccessArgs<ListTypeInfo>
 ) => MaybePromise<boolean>;
 
+type Operations = 'create' | 'query' | 'update' | 'delete';
```
mitchellhamilton, line 66:
```suggestion
type Operation = 'create' | 'query' | 'update' | 'delete';
```

packages/core/src/access/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function allowAll() {
+  return true;
+}
+export function denyAll() {
+  return false;
+}
+
+export function allOperations<ListTypeInfo extends BaseListTypeInfo>(
+  func: (args: BaseAccessArgs<ListTypeInfo>) => boolean
```
mitchellhamilton, line 12:
```suggestion
  func: (args: BaseAccessArgs<ListTypeInfo> & { operation: Operation }) => MaybePromise<boolean>
```

mitchellhamilton: (might want to name and export the intersected type from the access control config types)

docs/pages/docs/apis/access-control.md
```diff
@@ -90,6 +100,43 @@ export default config({
...
+import { config, list } from '@keystone-6/core';
+import { denyAll, allOperations } from '@keystone-6/core/access';
+
+export default config({
+  lists: {
+    ListKey: list({
+      access: {
+        operation: {
+          ...allOperations(denyAll)
+          query: ({ session, context, listKey, operation }) => true,
```
dcousens, line 132:
```suggestion
          // hint: unconditionally returning `true` is equivalent to using allowAll for this operation
          query: ({ session, context, listKey, operation }) => true,
```

packages/core/src/lib/core/access-control.ts
```diff
@@ -101,10 +101,24 @@ export type ResolvedFieldAccessControl = {
...
+    return {
+      operation: {
+        create: access,
+        query: access,
+        update: access,
+        delete: access,
+      },
+
+      filter: { query: access, update: access, delete: access },
+      item: { create: access, update: access, delete: access },
```
dcousens, line 118:
```suggestion
      filter: {
        create: () => true,
        update: () => true,
        delete: () => true
      },
      item: {
        create: () => true,
        update: () => true,
        delete: () => true
      },
```
Actually, I don't think we are defaulting these functions in this way, they are still optional as they are additive only

dcousens (approved):

PR #7863, opened 2022-08-30
Singletons

mitchellhamilton:
Adds new flag `isSingleton` to list, which indicates that there will only ever be one item in this list.

here's a config for a singleton:

```ts
import { list, singleton } from '@keystone-6/core';
import { allowAll } from '@keystone-6/core/access';
import { relationship, text } from '@keystone-6/core/fields';

 const lists = {
  Settings: singleton({
    access: allowAll,
    fields: {
      websiteName: text(),
      copyrightText: text(),
      highlightedPosts: relationship({ ref: 'Post', many: true }),
    },
  }),
  Post: list({
    access: allowAll,
    fields: {
      title: text({ validation: { isRequired: true } }),
      content: text(),
    },
  }),
  ...
};

```

Setting this flag does a few things:

## In Graphql

Graphql queries for singletons are all the same as those for lists, however with two major differences:

- Graphql will error if you ever make a query that doesn't have the `where: { id: '1' }`, you will get an error - you can only read, create, update, or delete an item with an ID of 1 in singleton.
- Upside, we are defaulting the `where` query for singletons, meaning you can simply not provide it, so you don't need to think about the above

So for example, the query for the above singleton would be:

```graphql
query {
  settings {
    copyrightText
  }
}
```

NB: This is a query for a single `settings`

## In the admin UI

- Singletons do not display a count or an add button.
- Clicking on a singleton's dashboard card or dashboard nav link will take you to the single singleton item
- If the singleton doesn't exist when you visit its page, you get a create button, and a message `{ThisSingleton} doesn't exist or you don't have access to it.`

## In Keystone

- You cannot have a relationship *to* a singleton. This is done because, since there is only ever one, there's very little reason to ever have a relationship to a singleton, and allowing it added some edge-cases that were easier just to simply not to support

## DX Design Notes

Singletons use bright new technology to have a plural of `Many{SingletonName}` when there's an ambiguous tech, so you never need to define a `pluralName` for your singletons.

For maximum compatibility, we've left the `many` queries on the graphql for singletons, as removing them causes large breaks in Keystone that, after exploring them, were really bad. They should almost certainly never be used, and really just pretend they're not there.


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/CreateItemPage/index.tsx
```diff
@@ -30,6 +30,9 @@ function CreatePageForm(props: { list: ListMeta }) {
           event.preventDefault();
           const item = await createItem.create();
           if (item) {
+            // if a user came from the item page, got prompted to create the item
+            // we need to essentially refetch the item so the user doesn't see
+            // "that item doesn't exist" because the cache was used
```
mitchellhamilton, line 35:
```suggestion
```

(this was a comment from me that I fixed in a different way)

tests/api-tests/queries/singletons.test.ts
```diff
@@ -0,0 +1,210 @@
...
+  test(
+    'delete many',
+    runner(async ({ context }) => {
+      await initialiseData({ context });
+
+      const { data, errors } = await context.graphql.raw({
+        query: `mutation { deleteSingulars(where: { id: 1 }) { id, field } }`,
+      });
+
+      expect(errors).toEqual(undefined);
```
mitchellhamilton, line 191:
```suggestion
      const data = await context.graphql.run({
        query: `mutation { deleteSingulars(where: { id: 1 }) { id, field } }`,
      });

```

I don't really care about this too much and don't mind if you want to change it or not and I know we have a bunch of tests that don't work like this but ideally, we should use `context.graphql.run` here.

examples/testing/schema.ts
```diff
@@ -44,4 +44,10 @@ export const lists: Lists = {
       tasks: relationship({ ref: 'Task.assignedTo', many: true }),
     },
   }),
+  Settings: singleton({
+    access: allowAll,
+    fields: {
+      siteName: text(),
+    },
+  }),
```
mitchellhamilton, line 52: `examples/testing` is meant to be about showing how to write tests for a Keystone project, not be an example for us to test Keystone itself, could you move this to `tests/sandbox` (and possibly add a singleton example if you'd like?)

docs/pages/docs/apis/schema.md
```diff
@@ -230,6 +231,10 @@ export default config({
 The `description` option defines a string that will be used as a description in the Admin UI and GraphQL API docs.
 This option can be individually overridden by the `graphql.description` or `ui.description` options.
 
+## isSingleton
+
+The `isSingleton` flag sets whether there should only ever be one of an item. Setting it to true ensures only one item can be created or read on the list, makes some updates to the admin UI, and defaults graphql queries so you can leave off your `where` option for singletons, as we default it.
```
dcousens, line 236:
```suggestion
The `isSingleton` flag defines if a list should limit the number of list items to `0 | 1`, and adjusts the user interface configuration to default to the item view.
GraphQL queries default the `*UniqueWhereFilter` to only refer to an item with `id: 1` (the default for a singleton item). 
```

dcousens (approved):

docs/pages/docs/apis/auth.md
```diff
@@ -459,7 +459,7 @@ The authenticated item will be returned as `item`.
 {% related-content %}
 {% well
 heading="Example Project: Authentication"
-href="https://github.com/keystonejs/keystone/tree/main/examples/with-auth" %}
+href="<https://github.com/keystonejs/keystone/tree/main/examples/with-auth>" %}
```
mitchellhamilton, line 462:
```suggestion
href="https://github.com/keystonejs/keystone/tree/main/examples/with-auth" %}
```

(we might want to disable prettier on markdoc files)

PR #7876, opened 2022-08-31
Rename 'Schema API' to 'Lists API'

dcousens:
Do we want to change `/docs/apis/schema` to `/docs/apis/lists` too?
And redirect?


mitchellhamilton (approved):
I'd say yeah to do renaming and a redirect but don't mind if it's in this PR or not

PR #7877, opened 2022-09-01
Add Context types for graphql-ts export

dcousens:
This pull request adds support for the capability to add `: Context` as a type refinement when doing GraphQL extensions.
Fundamentally this is still no different to a cast of `as Context`.

For the future, we will look at things like https://github.com/keystonejs/keystone/pull/7733, but for now this will be an improvement for many using these types.

```typescript
export const extendGraphqlSchema = graphql.extend(base => {
  ... = graphql.object<{ authorId: string }>()({
    name: 'Statistics',
    fields: {
      authorPosts: graphql.field({
        type: graphql.Int,
        resolve({ authorId }, args, context: Context) { // <-- adding `: Context` now contextualizes the types
          return context.db.Post.count({
            where: { 
              author: {
                id: {
                  equals: authorId
                }
              }
          });
        },
      }),
    },
  })
})
```


packages/core/src/types/schema/graphql-ts-schema.ts
```diff
@@ -219,28 +217,44 @@ export const CalendarDay = graphqlTsSchema.graphql.scalar<string>(
...
-  Fields extends Record<string, graphqlTsSchema.InterfaceField<any, OutputType, Context>>
+  Fields extends Record<string, graphqlTsSchema.InterfaceField<any, OutputType, Context>>,
+  Context extends KeystoneContext<any>
 > = graphqlTsSchema.InterfaceType<Source, Fields, Context>;
 export type InterfaceField<
   Args extends Record<string, graphqlTsSchema.Arg<any>>,
-  TType extends OutputType
+  TType extends OutputType,
+  Context extends KeystoneContext<any>
 > = graphqlTsSchema.InterfaceField<Args, TType, Context>;
```
mitchellhamilton, line 260: Why make these generic over a context?

mitchellhamilton (approved):

PR #7878, opened 2022-09-01
Fixes `context.graphql` types and adds `ts-gql` example

borisno2:
Currently the return types for both `context.graphql.run` and `context.graphql.raw` do not allow for graphql type generation tools like `ts-gql`. This PR fixes the types for these so the correct types are returned, an `ts-gql` example is also added.


packages/core/src/types/context.ts
```diff
@@ -140,13 +140,33 @@ export type KeystoneDbAPI<KeystoneListsTypeInfo extends Record<string, BaseListT
...
+  ) => Promise<ExecutionResult<TData, TVariables>['data']>;
+  raw: <TData = any, TVariables = Record<string, any>>(
+    args: GraphQLExecutionArguments<TData, TVariables>
+  ) => Promise<ExecutionResult<TData, TVariables>>;
 };
 
-type GraphQLExecutionArguments = {
-  query: string | DocumentNode;
-  variables?: Record<string, any>;
+export interface TypedDocumentNode<
```
mitchellhamilton, line 151: We should import this from `@graphql-typed-document-node/core`, not duplicate it. (the only dependency it has is a peer dep on `graphql` so I'm not worried about dependency bloat from it)

borisno2: Cool, I was wondering about that.. I'll make that change now...

dcousens (approved):

examples/graphql-ts-gql/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@keystone-6/example-graphql-ts-gql",
+  "version": "0.0.3",
+  "private": true,
+  "license": "MIT",
+  "scripts": {
+    "postinstall": "ts-gql build",
```
mitchellhamilton, line 7: We need to run `cd examples/graphql-ts-gql && yarn ts-gql build` from the root `prepare` script because yarn doesn't consistently run postinstall/etc. scripts from within packages

mitchellhamilton (changes requested):

packages/core/src/types/context.ts
```diff
@@ -140,13 +141,17 @@ export type KeystoneDbAPI<KeystoneListsTypeInfo extends Record<string, BaseListT
...
+    args: GraphQLExecutionArguments<TData, TVariables>
+  ) => Promise<ExecutionResult<TData, TVariables>>;
 };
 
-type GraphQLExecutionArguments = {
-  query: string | DocumentNode;
-  variables?: Record<string, any>;
+type GraphQLExecutionArguments<TData, TVariables> = {
+  query: string | DocumentNode | TypedDocumentNode<TData, TVariables>;
+  variables?: TVariables | Record<string, any>;
```
mitchellhamilton, line 154:
```suggestion
  variables?: TVariables;
```

packages/core/src/types/context.ts
```diff
@@ -140,13 +141,17 @@ export type KeystoneDbAPI<KeystoneListsTypeInfo extends Record<string, BaseListT
...
 export type KeystoneGraphQLAPI = {
   schema: GraphQLSchema;
-  run: (args: GraphQLExecutionArguments) => Promise<Record<string, any>>;
-  raw: (args: GraphQLExecutionArguments) => Promise<ExecutionResult<Record<string, any>>>;
+  run: <TData = any, TVariables = Record<string, any>>(
+    args: GraphQLExecutionArguments<TData, TVariables>
+  ) => Promise<ExecutionResult<TData, TVariables>['data']>;
+  raw: <TData = any, TVariables = Record<string, any>>(
+    args: GraphQLExecutionArguments<TData, TVariables>
+  ) => Promise<ExecutionResult<TData, TVariables>>;
```
mitchellhamilton, line 149:
```suggestion
  run: <TData, TVariables = Record<string, any>>(
    args: GraphQLExecutionArguments<TData, TVariables>
  ) => Promise<ExecutionResult<TData, TVariables>['data']>;
  raw: <TData, TVariables = Record<string, any>>(
    args: GraphQLExecutionArguments<TData, TVariables>
  ) => Promise<ExecutionResult<TData, TVariables>>;
```

We should require people to cast the result if they're not using a typed document node

examples/graphql-ts-gql/package.json
```diff
@@ -0,0 +1,36 @@
...
```
mitchellhamilton, line 7:
```suggestion
```
(since it's at the root)

mitchellhamilton (approved):
The example seems a bit strange given you wouldn't want to use this for those use cases but ehhh, it's fine

PR #7882, opened 2022-09-05
Improve tests/api-tests performance

emmatown:
The result of this is basically that running `yarn jest api-tests` now does this:

```
Test Suites: 80 passed, 80 total
Tests:       3140 passed, 3140 total
Snapshots:   15 passed, 15 total
Time:        41.229 s
Ran all test suites matching /api-tests/i.
```

- Previously, it was basically impractical to run all of `tests/api-tests` locally because they wouldn't be run in parallel since all the tests used the same database, this fixes that by using one database per Jest worker.
- The tests now default to using SQLite and `DATABASE_URL` isn't used for SQLite so you can just do `yarn jest api-tests`
- This skips generating an actual Prisma Client to disk and creates a Prisma Client in memory instead. This is going to be prone to breakage given it's using very internal Prisma things, I think that's a worthwhile trade-off for the gain in performance in our tests.
- The tests override `console.warn` to avoid printing expected warnings
- The tests slightly use Prisma Migrate in a slightly more efficient way (with more specific optimizations for SQLite since that's what you'll generally use for testing locally when making changes that shouldn't differ per-database provider)

(I intentionally didn't create a changeset for the small change in packages/core because it doesn't change anything public)


dcousens (approved):
LGTM, I don't like the fact we are reaching into `@prisma/internals`, I'd prefer faster tests in the interim though - and we rely on `@prisma/internals` in a similar way elsewhere, so this is nothing new

tests/api-tests/utils.ts
```diff
@@ -2,21 +2,94 @@ import { initConfig, createSystem } from '@keystone-6/core/system';
...
+      'There are already 10 instances of Prisma Client actively running.'
+    ) ||
+      // we should really enforce a safe default for this though
+      args[0] ===
+        'Persisted queries are enabled and are using an unbounded cache. Your server is vulnerable to denial of service attacks via memory exhaustion. Set `cache: "bounded"` or `persistedQueries: false` in your ApolloServer constructor, or see https://go.apollo.dev/s/cache-backends for other alternatives.')
+  ) {
+    return;
+  }
+  prevConsoleWarn.apply(this, args);
+};
```
dcousens, line 21: @mitchellhamilton is there an alternative to this?

PR #7888, opened 2022-09-06
Add `cache: "bounded"` and `persistedQueries: false` to default Apollo Server config

emmatown:
This PR changes the default of `cache` in Apollo Server to `"bounded"` to ensure that an attacker cannot fill Apollo's automatic persisted queries cache to the point of memory exhaustion. This will be default in a future version of Apollo Server: https://www.apollographql.com/docs/apollo-server/performance/cache-backends#ensuring-a-bounded-cache.

This _also_ sets `persistedQueries: false`. While setting `cache: "bounded"` ensures that an attacker cannot exhaust a server's memory, an attacker could still send a large number of different queries constantly to fill up the cache rendering automatic persisted queries pointless and actually degrade performance because trusted actors would have to frequently send a second request since the server would have replaced the queries from trusted actors in the cache with malicious queries and the client would have to send the actual query text in a second request. A safer and more predictable approach would be to persist queries at compile time from the actual applications the server is meant to serve such as [what Relay suggests](https://relay.dev/docs/guides/persisted-queries/#compile-time-push).

If you're aware of the above caveats of persisted queries, you can re-enable persisted queries with `config.graphql.apolloConfig`.

### DIFF
--- a/.changeset/many-bulldogs-scream.md
+++ b/.changeset/many-bulldogs-scream.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Adds `cache: "bounded"` and `persistedQueries: false` to default Apollo Server config.
--- a/.changeset/many-bulldogs-scream.md
+++ b/.changeset/many-bulldogs-scream.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': major
 ---
 
-Adds `cache: "bounded"` and `persistedQueries: false` to default Apollo Server config.
+Changes default Apollo Server configuration to use `cache: "bounded"` and `persistedQueries: false`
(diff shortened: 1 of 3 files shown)


flexdinesh (approved):

.changeset/many-bulldogs-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Adds `cache: "bounded"` and `persistedQueries: false` to default Apollo Server config.
```
dcousens, line 5:
```suggestion
Changes default Apollo Server configuration to use `cache: "bounded"` and `persistedQueries: false`
```

dcousens (approved):

PR #7894, opened 2022-09-07
Tidy and format

dcousens:
This is plainly the result of running `yarn format` locally.
This should decrease the pain when running `yarn format` on unrelated branches like in https://github.com/keystonejs/keystone/pull/7863

PR #7895, opened 2022-09-07
Drop remark formatting

dcousens:
Followup to https://github.com/keystonejs/keystone/pull/7894 causing unnecessary busy work


mitchellhamilton (approved):

mitchellhamilton (changes requested):
Actually, i have thoughts, one sec

mitchellhamilton (approved):

PR #7897, opened 2022-09-08
Add filters to pair field example

dcousens:
This pull request adds an example of a custom GraphQL filter to the pair field type example.

### DIFF
--- a/examples/custom-field/3-pair-field/index.ts
+++ b/examples/custom-field/3-pair-field/index.ts
@@ -62,7 +62,10 @@ export function pair<ListTypeInfo extends BaseListTypeInfo>(
       input: {
         where: {
           arg: graphql.arg({ type: PairFilter }),
-          resolve: resolveWhere,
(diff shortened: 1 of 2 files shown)


mitchellhamilton (approved):
obligatory "I don't think this filter behavior is _actually_ what you would want" but it's a fine illustration

examples/custom-field/3-pair-field/index.ts
```diff
@@ -39,18 +50,20 @@ export function pair<ListTypeInfo extends BaseListTypeInfo>({
...
           scalar: 'String',
-          index: isIndexed === true ? 'index' : isIndexed || undefined,
         },
       },
     })({
       ...config,
       input: {
+        where: {
+          arg: graphql.arg({ type: PairFilter }),
+          resolve: resolveWhere,
```
dcousens, line 65:
```suggestion
          // eslint-disable-next-line @typescript-eslint/no-unused-vars
          resolve(value, context) {
            return resolveWhere(value);
          },
```

PR #7914, opened 2022-09-12
Unify access-control errors, and fix access-control types

dcousens:
This pull request unifies most access-control errors to the following two messages:

```typescript
You cannot ${operation} that ${listKey} - it may not exist
```

```typescript
You cannot ${operation} that ${listKey} - you cannot ${operation} the fields ["${fieldKey}", ...]
```

This removes the burden of stringifying the `inputData` or the user provided filters in different scenarios.
That _shouldn't be a problem_, but, we can eliminate that risk with this.

The only differentiation of routine error types now is between list and fields.

This pull request additionally fixes a few unreleased type problems that were merged in https://github.com/keystonejs/keystone/pull/7848.


packages/core/src/lib/core/mutations/access-control.ts
```diff
@@ -64,240 +63,224 @@ export async function checkUniqueItemExists(
...
+        returned: typeof listResult,
       },
     ]);
   }
 
-  if (!result) {
-    throw accessDeniedError(
-      `You cannot perform the '${operation}' operation on the item '${JSON.stringify(
-        uniqueWhere
-      )}'. It may not exist.`
```
dcousens, line 108: I think this was the wrong error, we know the `item` exists - if we're trying to make that fact opaque, we can call `missingItem` to throw and make that explicit

packages/core/src/types/config/access-control.ts
```diff
@@ -89,22 +84,22 @@ type ListAccessControlObject<ListTypeInfo extends BaseListTypeInfo> = {
...
   // - boolean true/false. If false, treated as a filter that never matches.
-  filter?: {
-    query?: ListFilterAccessControl<'query', ListTypeInfo>;
-    update?: ListFilterAccessControl<'update', ListTypeInfo>;
-    delete?: ListFilterAccessControl<'delete', ListTypeInfo>;
-    // create: not supported: FIXME: Add explicit check that people don't try this.
-    // FIXME: Write tests for parseAccessControl.
-  };
+  filter?:
+    | ListFilterAccessControl<FilterOperation, ListTypeInfo>
```
dcousens, line 88: This may not be desirable, but, it is what the code currently supports

https://github.com/keystonejs/keystone/blob/c3a3789c6e4a1761a24cd02904b7549a62113363/packages/core/src/lib/core/access-control.ts#L149

packages/core/src/types/config/access-control.ts
```diff
@@ -89,22 +84,22 @@ type ListAccessControlObject<ListTypeInfo extends BaseListTypeInfo> = {
...
   // These rules are applied to each item being operated on individually. They return `true` or `false`,
   // and if false, an access denied error will be returned for the individual operation.
-  item?: {
-    // query: not supported
-    create?: CreateListItemAccessControl<ListTypeInfo>;
-    update?: UpdateListItemAccessControl<ListTypeInfo>;
-    delete?: DeleteListItemAccessControl<ListTypeInfo>;
-  };
+  item?:
+    | {
```
dcousens, line 98: We technically support a `boolean` here

https://github.com/keystonejs/keystone/blob/c3a3789c6e4a1761a24cd02904b7549a62113363/packages/core/src/lib/core/access-control.ts#L160

I don't think we want that...

packages/core/src/types/config/access-control.ts
```diff
@@ -133,25 +128,44 @@ export type ListAccessControl<ListTypeInfo extends BaseListTypeInfo> =
...
+    fieldKey: string;
+    /**
+    * The item being updated
+    */
+    item: ListTypeInfo['item'];
+    /**
+    * The input passed in from the GraphQL API
+    */
+    inputData: ListTypeInfo['inputs']['update'];
+  };
```
dcousens, line 162: I opted for explicit types here, they are only related to the list access control functions by convention

packages/core/src/types/config/access-control.ts
```diff
@@ -89,18 +81,17 @@ type ListAccessControlObject<ListTypeInfo extends BaseListTypeInfo> = {
...
+        query?: ListFilterAccessControl<'query', ListTypeInfo>;
+        update?: ListFilterAccessControl<'update', ListTypeInfo>;
+        delete?: ListFilterAccessControl<'delete', ListTypeInfo>;
+      };
 
   // These rules are applied to each item being operated on individually. They return `true` or `false`,
   // and if false, an access denied error will be returned for the individual operation.
   item?: {
-    // query: not supported
     create?: CreateListItemAccessControl<ListTypeInfo>;
```
dcousens, line 95: Technically this supports a top-level function or `boolean` too
https://github.com/keystonejs/keystone/blob/c3a3789c6e4a1761a24cd02904b7549a62113363/packages/core/src/lib/core/access-control.ts#L160

packages/core/src/lib/core/access-control.ts
```diff
@@ -82,7 +82,7 @@ export async function getAccessFilters(
 export function parseFieldAccessControl(
   access: FieldAccessControl<BaseListTypeInfo> | undefined
 ): ResolvedFieldAccessControl {
-  if (typeof access === 'boolean' || typeof access === 'function') {
+  if (typeof access === 'function') {
```
dcousens, line 85: This is accurate to the types, but, is this what we want?

.changeset/errors-now-equal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Errors are now mostly context free - and non-differentiable as to which access-control was rejected
```
dcousens, line 5: .. except field level access-control

dcousens, line 5:
```suggestion
Errors are now mostly context free - only list and field access-control is differentiable by the error message now
```

packages/core/src/types/config/access-control.ts
mitchellhamilton: We shouldn't support this

dcousens: https://github.com/keystonejs/keystone/pull/7914/commits/51022994db76e68a4d90979cb48ae1eb065e73ff

.changeset/errors-now-equal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Errors are now mostly context free - only list and field access-control is differentiable by the error message now
```
mitchellhamilton, line 5: This description doesn't explain the change well imo. Like what errors? What do you mean by context?

mitchellhamilton (approved):
Besides making the changeset more clear, this looks good to me

mitchellhamilton (approved):

Issue #7638, opened 2022-06-16
is one of filter not working properly for integer field


pawan123e:
![filter](https://user-images.githubusercontent.com/26507643/174059709-a6e90cb6-1445-446a-8031-a65f78fe9e6b.png)

Here in the input box, we can add only 1 item.

mitchellhamilton:
You can have multiple items by separating numbers with a comma. We're planning on improving the UX around this in the future so it's clear how to use it.

pawan123e:
Okay, great!

gautamsi:
you can not use `,` in this field as this is clearing non numeric input on entry in field. a fix is still needed

PR #7930, opened 2022-09-16
Fix filter `input` types for integer, bigInt, decimal and float fields in the UI

nya1:
Closes https://github.com/keystonejs/keystone/issues/7638

In `integer` and `bigInt` fields I have changed the `TextInput` type to `text` (instead of `number`), this allows to enter commas, the filtering of the value (only digits, commas and minus sign) is already performed by a [regex on the `onChange` handler](https://github.com/keystonejs/keystone/blob/df88fe6c49c6e15acea45e00192a3c8d16a3d461/packages/core/src/fields/types/integer/views/index.tsx#L218-L220).
With this change the filtering works as expected.

There is a comment above the `TextInput` regarding the type change but I'm not sure this is an old comment (prior to the onChange regex) or if we want to have a more advanced input.

https://github.com/keystonejs/keystone/blob/df88fe6c49c6e15acea45e00192a3c8d16a3d461/packages/core/src/fields/types/integer/views/index.tsx#L213-L216

### DIFF
--- a/packages/core/src/fields/types/bigInt/views/index.tsx
+++ b/packages/core/src/fields/types/bigInt/views/index.tsx
@@ -227,19 +227,18 @@ export const controller = (
     validate: value =>
       validate(value, validation, config.label, hasAutoIncrementDefault) === undefined,
     filter: {
-      Filter(props) {
+      Filter({ autoFocus, type, onChange, value }) {
         return (
           <TextInput
-            // this should not be type=number since it shoud allow commas so the one of/not one of
-            // filters work but really the whole filtering UI needs to be fixed and just removing type=number
-            // while doing nothing else would probably make it worse since anything would be allowed in the input
-            // so when a user applies the filter, the query would return an error
-            type="text"
             onChange={event => {
-              props.onChange(event.target.value.replace(/[^\d,\s-]/g, ''));
+              if (type in ['in', 'not_in']) {
+                onChange(event.target.value.replace(/[^\d,\s-]/g, ''));
+                return;
+              }
+              onChange(event.target.value.replace(/[^\d\s-]/g, ''));
             }}
-            value={props.value}
-            autoFocus={props.autoFocus}
+            value={value}
+            autoFocus={autoFocus}
           />
         );
       },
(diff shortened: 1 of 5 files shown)


packages/core/src/fields/types/bigInt/views/index.tsx
```diff
@@ -248,8 +247,8 @@ export const controller = (
         const valueWithoutWhitespace = value.replace(/\s/g, '');
         const parsed =
           type === 'in' || type === 'not_in'
-            ? valueWithoutWhitespace.split(',').map(x => BigInt(x))
-            : BigInt(valueWithoutWhitespace);
+            ? valueWithoutWhitespace.split(',')
+            : valueWithoutWhitespace;
```
dcousens, line 251: This fixes a JSON serialization error

mitchellhamilton (approved):

dcousens:
Thanks @nya1 and @gautamsi - it is unfortunate that these fields are not locale friendly.
I have updated the pull request to try and increase the consistency of the filters across each of the numeric fields, but honestly these need work and attention. :yellow_heart: 

We're currently experimenting with Adobe Spectrum in Keystone's updated design system, and that, or something like it, should help us unify and smooth the user experience across many of these field views.

gautamsi:
> We're currently experimenting with Adobe Spectrum in Keystone's updated design system, and that, or something like it, should help us unify and smooth the user experience across many of these field views.

does this mean end of `@keystone-ui` or just use their design system with `@keystone-ui` for implementing the design? you mentioned locale, is this going to help improve locale in admin-ui and content? Also new design system should make it possible to use new Admin-ui over the headless api instead of tightly coupled in terms of generation and consumption of admin meta.

jossmac:
@gautamsi We're interested in [react-aria](https://react-spectrum.adobe.com/react-aria/index.html) specifically for accessibility. Since they also provide I18N solutions we'll be considering that too.

The intention is to build a component library that's informed by and in service of, but decoupled from, the admin UI. Ideally consumers will be able to pick up the same _pieces_ to build custom interfaces that fit seamlessly within Keystone's overall aesthetic and behaviour.

PR #7941, opened 2022-09-20
Prefer isSingleton, not a singleton function

Noviny:


dcousens:
I don't know if this will be the name we continue using by the next major, but I think this is the best compromise for now

dcousens (approved):

PR #7951, opened 2022-09-26
Add support for writing conditional documentation on the docs site

emmatown:
This will allow us to remove the `website_live` branch while still allowing us to document new features that will only be available in the next release. The conditionals are removed when doing a release. I was planning on removing them by transforming Markdoc's AST but Markdoc's formatter is experimental and seems to break on some of our content so this uses a regex. I was also planning to add conditionals around documentation on unreleased features but it seems as though there isn't any.

### New
```markdown
{% if $nextRelease %}
You can filter a post using `post (...)`
{% /if %}
```

### Updates
```markdown
{% if $nextRelease %}
You can filter a post using `post (...)`
{% else /%}
You cannot filter posts.
{% /if %}
```

### Removals
```markdown
{% if $nextRelease %}
{% else /%}
You cannot filter posts.
{% /if %}
```

### DIFF
--- a/docs/markdoc/config.ts
+++ b/docs/markdoc/config.ts
@@ -1,9 +1,24 @@
-import { Config, nodes, Tag, ValidationError, Node } from '@markdoc/markdoc';
+import {
+  Config,
+  nodes,
+  Tag,
+  ValidationError,
+  Node,
+  functions,
+  ConfigFunction,
+} from '@markdoc/markdoc';
 import slugify from '@sindresorhus/slugify';
 
 export type Pages = Map<string, { ids: Set<string> }>;
 
-export const markdocConfig: Config = {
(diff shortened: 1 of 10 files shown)


docs/pages/docs/config/config.md
```diff
@@ -5,7 +5,7 @@ description: "API reference for configuring your Keystone system: Lists, Databas
 
 The `keystone` [CLI](../guides/cli) expects to find a module named `keystone.ts` with a default export of a Keystone system configuration returned from the function `config()`.
 
-```typescript filename=keystone.ts
```
mitchellhamilton, line 8: (This `filename=keystone.ts` had no effect)

dcousens (approved):

PR #7952, opened 2022-09-26
Fallback to default AWS credential chain

ryzr:
By setting the S3 Client credentials to undefined, the AWS SDK will fallback to a default credential chain, attempting to load credentials from the instance profile, .ini file, environment variables etc:
https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html

This allows users to follow AWS best practices by using temporary credentials:
https://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html

To achieve this, I've made the s3Storage.accessKeyId and s3Storage.secretAccessKey configuration items optional.

### DIFF
--- a/.changeset/silver-shirts-remain.md
+++ b/.changeset/silver-shirts-remain.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Allow falling back to default AWS credential provider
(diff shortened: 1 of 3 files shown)


dcousens:
The changes LGTM, but I don't know if we want to commit to supporting default credential providers.

We are contemplating moving away from the AWS SDK to minimize the dependency requirements.

dcousens:
@ryzr how did this change help you follow AWS best practices?

dcousens (approved):

PR #7954, opened 2022-09-26
Add new getContext function and script example

dcousens:
This pull request adds a new `getContext` function for working with Keystone directly using your pre-generated Prisma client.

This is incredibly useful for oneshot scripts that don't need Keystone's web middleware.


examples/script/script.ts
```diff
@@ -0,0 +1,24 @@
...
+import config from './keystone';
+import * as PrismaModule from '.prisma/client';
+
+async function main() {
+  const { connect, context, disconnect } = getContext(config, PrismaModule);
+
+  console.log('(script.ts)', 'connect');
+
+  // if you don't call connect, db.onConnect will not be run
+  // but Prisma will automatically connect when Prisma is used
```
borisno2, line 11:
```suggestion
  // connect() here also runs db.onConnect from your keystone config, 
  // to avoid running db.onConnect you can not run connect here, 
  // in which case Prisma will automatically connect when it is used
```

borisno2 (approved):

PR #7957, opened 2022-09-26
Remove experimental.generateNodeAPI

dcousens:
This pull request removes the `experimental.generateNodeAPI` flag, in favour of using `getContext` in relevant applications instead.


docs/pages/docs/walkthroughs/embedded-mode-with-sqlite-nextjs.md
```diff
@@ -226,7 +224,7 @@ export default function Home({ posts }: InferGetStaticPropsType<typeof getStatic
 // Here we use the Lists API to load all the posts we want to display
 // The return of this function is provided to the `Home` component
 export async function getStaticProps() {
-  const posts = (await query.Post.findMany({ query: 'id title slug' })) as Post[];
```
dcousens, line 229: - [ ] Add the missing `context` here after we verify the example

packages/core/src/admin-ui/system/buildAdminUI.ts
```diff
@@ -1,8 +1,10 @@
+const { env } = process;
```
dcousens, line 1:
```suggestion
// destructured to prevent inlining to `"production" = "production"`
const { env } = process;
```

.changeset/less-next-things.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': minor
```
dcousens, line 2:
```suggestion
'@keystone-6/core': major
```

Although the flag is `experimental`, so I am marking the change as `major` as this will be releasing with other major changes, and it will easily be seen by anyone reviewing the CHANGELOG.

packages/core/src/artifacts.ts
```diff
@@ -146,22 +146,6 @@ export async function generateCommittedArtifacts(
   return artifacts;
 }
 
-const nodeAPIJS = (
-  cwd: string,
-  config: KeystoneConfig
-) => `import keystoneConfig from '../../keystone';
-import { PrismaClient } from '.prisma/client';
-import { createQueryAPI } from '@keystone-6/core/___internal-do-not-use-will-break-in-patch/node-api';
```
borisno2, line 154: Are we going to remove `packages/core/src/___internal-do-not-use-will-break-in-patch/node-api.ts` in this PR?

borisno2 (approved):

PR #7968, opened 2022-09-29
Replace `@keystone-6/core/testing` exports with `resetDatabase`

emmatown:
A combination of `getContext` from `@keystone-6/core/context` and this `resetDatabase` function makes the previous testing helpers unnecessary and avoids the overhead of generating a Prisma Client dynamically which isn't needed when there is a single Keystone config so the Prisma Client can be imported


dcousens (approved):
LGTM

PR #7971, opened 2022-09-29
Remove `connect`, `disconnect` for session strategies

emmatown:
This pull request removes `connect` and `disconnect` from the SessionStrategy interface, and replaces the `session-store-redis` package with an example.

We shouldn't have created a package for what can be effectively reproduced with:
```typescript
const sessionStrategy = storedSessions({
  store: ({ maxAge }) => ({
    async get(key) {
      let result = await redis.get(key)
      if (typeof result === 'string') {
        return JSON.parse(result)
      }
    },
    async set(key, value) {
      await redis.setEx(key, maxAge, JSON.stringify(value))
    },
    async delete(key) {
      await redis.del(key)
    },
  }),
  // ...
})

export default withAuth(
  config({
    db: {
      // ...
      async onConnect() {
        await redis.connect()
      }
    },
    // ...
  })
)
```

And an `redis.connect()` in the `db.onConnect` hook.

Unfortunately `db.onConnect` probably should have been called `onStartup[but before HTTP]`.

At the moment, `db.onConnect` happens to be sequentially when we connect the Prisma client, but strictly has no relationship to if and when the Prisma client is connected or disconnected.
This can lead to DX confusion when you use a `context` directly from `getContext`.


dcousens (approved):
LGTM

PR #7982, opened 2022-10-04
Fixes JSON field type view assuming the empty string is equivalent to `null`

dcousens:
When a JSON value is the empty JSON string `""` (not an empty input), the JSON field type view would de-serialize the value to an empty input.

It should have been de-serialized to `'""'` (an empty JSON string).

This pull request fixes that.

### DIFF
--- a/packages/core/src/fields/types/json/views/index.tsx
+++ b/packages/core/src/fields/types/json/views/index.tsx
@@ -94,6 +94,7 @@ export const controller = (config: Config): FieldController<string, string> => {
     },
     deserialize: data => {
       const value = data[config.path];
+      // null is equivalent to Prisma.DbNull, and we show that as an empty input
       if (value === null) return '';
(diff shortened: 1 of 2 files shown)


packages/core/src/fields/types/json/views/index.tsx
```diff
@@ -94,18 +94,16 @@ export const controller = (config: Config): FieldController<string, string> => {
     },
     deserialize: data => {
       const value = data[config.path];
-      if (!value) return '';
+      if (value === null) return '';
```
dcousens, line 97: This is the fix, the rest is tidy up

dcousens, line 97:
```suggestion
      // null is equivalent to Prisma.DbNull, and we show that as an empty input
      if (value === null) return '';
```

mitchellhamilton (approved):

PR #7995, opened 2022-10-12
No automatic many-prefixing when isSingleton is set 

dcousens:
This isn't intuitive, and we have a robust solution for resolving ambiguities already: `graphql.plural`

#### Related
- https://github.com/keystonejs/keystone/pull/7863

### DIFF
--- a/docs/pages/docs/config/lists.md
+++ b/docs/pages/docs/config/lists.md
@@ -277,15 +277,15 @@ Using GraphQL, to query a list named `settings`, with `isSingleton` set, you can
 ```graphql
 query {
   # singular (null or an item)
-  settings {
-    seoTitle
-    seoDescription
+  seoConfiguration {
+    title
+    description
   }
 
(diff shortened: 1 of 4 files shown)


docs/pages/docs/config/lists.md
```diff
@@ -277,15 +277,15 @@ Using GraphQL, to query a list named `settings`, with `isSingleton` set, you can
...
+    description
   }
 
   # plural (0 or 1 items)
-  settingsMany {
-    seoTitle
-    seoDescription
+  seoConfigurations {
+    title
+    description
```
mitchellhamilton, line 288: Not a blocking thing but showing the many query feels unnecessary. It's not like you would actually want to use it. A note like "hey, the many things still exist for reasons" would certainly be helpful but actually showing it feels like we're saying that there's a reason people would want it when there isn't

mitchellhamilton (approved):

PR #7997, opened 2022-10-12
Add documentation for getContext

dcousens:


docs/pages/docs/context/get-context.md
```diff
@@ -0,0 +1,56 @@
+---
+title: "Get Context"
+description: "getContext is a function exported by @keystone-6/core/context to support operations without starting a HTTP server"
+---
+
+Keystone's command line interface is helpful, but sometimes you don't want to startup the HTTP server, or kickoff any build processes.
+Maybe you need to insert data into the database, maybe you only want Keystone for your internal schema management behind a custom protocol or tiny REST API.
+*Or maybe you want to write unit tests.*
+
+If you have previously built Keystone using `keystone build` or used `keystone dev`, then you can use `getContext`.
```
flexdinesh, line 10: Should we clarify what _built_ means here? Something along the lines of _if you have already generated the schema and prisma client (and whatever build command does) using `keystone build` or `keystone dev`_?

docs/pages/docs/context/get-context.md
```diff
@@ -0,0 +1,56 @@
...
+Using the `getContext` function does not need the adhere to the typical Keystone entry point - it only and requires that your Prisma client has been built and can be provided as a parameter with your Keystone configuration.
+
+```typescript
+import { getContext } from '@keystone-6/core/context';
+import config from './keystone';
+import * as PrismaModule from '.prisma/client';
+
+const context = getContext(config, PrismaModule);
+
+// ... whatever you need
```
flexdinesh, line 21: Maybe an example one-liner with the available APIs in context?

```
// You can now access all context APIs
// Eg. context.db.User
```

flexdinesh (approved):

docs/pages/docs/context/get-context.md
dcousens: I don't think so, not unless we want to introduce documentation the explains each of the artifacts of Keystone - the hint is that you need to run `keystone build` or `keystone dev`, I think we can leave at that.

I might change the wording to remove `built` completely

PR #7999, opened 2022-10-12
Add snapshot release workflow

emmatown:
This will allow publishing everything to `0.0.0-some-tag-DATETIME` with a workflow dispatch so that we can easily do pre-releases or etc. (I'm avoiding [Changesets' prerelease thing](https://github.com/changesets/changesets/blob/main/docs/prereleases.md) because it's more complicated). You can read more about snapshot releases at https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md. The most important thing to note here is that the version updates/changelogs are never committed.


.github/workflows/publish.yml
```diff
@@ -9,23 +9,20 @@ jobs:
...
-          version: yarn no-run-version-packages
-          publish: yarn publish-changed
+          version: exit 1
+          publish: yarn changeset publish
           createGithubReleases: false
         env:
-          # note that we're not using the GH token provided by Actions here because Actions has a rule that Actions
-          # will not run as the result of another Action so CI wouldn't run on the release PRs then
-          # we can get around it by using a personal access token from a GH account
-          GITHUB_TOKEN: ${{ secrets.KEYSTONE_RELEASE_BOT_GITHUB_TOKEN }}
```
mitchellhamilton, line 30: the action errors if GITHUB_TOKEN isn't set.

package.json
```diff
@@ -13,7 +13,7 @@
     "format:file": "prettier --write",
     "format": "prettier --write \"**/*.{js,json,ts,tsx,md,mdx}\"",
     "fresh": "yarn clean && yarn",
```
mitchellhamilton, line 15:
```suggestion
```

dcousens (approved):

PR #8005, opened 2022-10-17
Fix bigInt deserialisation

dcousens:
If a `bigInt` is not parsed early, it will throw when trying to format the value in the view.
This is checked as an error condition, but the branch throwing in lieu of this is throwing when trying to `JSON.stringify` a `BigInt`.

This pull request resolves the lack of parsing in the deserialisation (bug 1) and in the event an error condition does occur, this pull request fixes the error message by not using `JSON.stringify`,  instead showing only the `typeof` value types that were involved.

### DIFF
--- a/packages/core/src/fields/types/bigInt/views/index.tsx
+++ b/packages/core/src/fields/types/bigInt/views/index.tsx
@@ -22,7 +22,7 @@ type Validation = {
 
 type Value =
   | { kind: 'create'; value: string | bigint | null }
-  | { kind: 'update'; value: string | bigint | null; initial: bigint | null };
+  | { kind: 'update'; value: string | bigint | null; initial: unknown | null };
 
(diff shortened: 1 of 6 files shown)


packages/core/src/fields/types/bigInt/views/index.tsx
```diff
@@ -214,7 +214,11 @@ export const controller = (
           ? BigInt(config.fieldMeta.defaultValue)
           : null,
     },
-    deserialize: data => ({ kind: 'update', value: data[config.path], initial: data[config.path] }),
+    deserialize: data => ({
+      kind: 'update',
+      value: BigInt(data[config.path]),
```
mitchellhamilton, line 219: needs to handle `data[config.path]` being null

mitchellhamilton (approved):

PR #8007, opened 2022-10-17
Fix multiselect field type not propagating common config options

marekryb:
Currently multiselect field type completely ignores some common config options - most notably 'access' and 'graphql'.

This pull request fixes that by 'spreading' user config to final type.

### DIFF
--- a/.changeset/two-ghosts-jog.md
+++ b/.changeset/two-ghosts-jog.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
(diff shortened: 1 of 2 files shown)


dcousens (approved):

dcousens:
Thanks @marekryb for reporting and fixing this.

I have credited you in the related [security advisory](https://github.com/keystonejs/keystone/security/advisories/GHSA-6mhr-52mv-6v6f), and I will be looking into preventing this from happening again.

PR #8017, opened 2022-10-19
Fix FieldCreateItemAccessArgs inputData typings

acburdine:
I'm not 100% sure if this change was intentional, but I'm updating my project to `@keystone-6/core@3.0.0` and I noticed some types breaking in a field-level create access control function. I checked the commit that introduced this typing change, and didn't see any functionality difference that would necessitate a change in the typings, so I'm assuming the typing change was accidental.

### DIFF
--- a/.changeset/plenty-fireants-deliver.md
+++ b/.changeset/plenty-fireants-deliver.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


dcousens (approved):

dcousens (approved):

acburdine:
thanks for getting the release out quickly @dcousens 😄

PR #8020, opened 2022-10-19
[POC] Allow custom app

gautamsi:
Allow using custom `_app` template for admin ui

@flexdinesh this how we can do natively in keystone instead of my file watcher workaround.

I have made changes to `examples/basic` template to use custom `theme`. notice button color changes when running the modified example.

image


gautamsi:
let me know if this approach is acceptable then I will cleanup the PR and write documentation

flexdinesh (changes requested):
Hey @gautamsi! Thanks for sending this PR. Keen to get this in. 🙌  

FYI, there is work planned to make Admin UI customisation early next year which will make this PR redundant. In the meanwhile this is a nice escape hatch for users to be able to customise their `_app.tsx` component. This approach is a bit brittle since unintended changes in this custom `app.tsx` file or internal templating changes could break the entire Admin UI but I think that's okay. We will add a note to the example saying _try this at your own risk_.

Here are a few things that we need to do before this PR is merge ready:
- Please create a separate example for customising the app component instead of using the `examples/basic` example. Recommended title is `customising-admin-ui-app`.
- Add a comment in the example `app.tsx` file please - _This approach is a bit brittle and try it at your own risk. If any of the Keystone components break, your entire Admin UI might stop working._
- Instead of copying all the utils, you could import `identityType` from `'@keystone-ui/core`. Eg. `import { identityType } from '@keystone-ui/core';`
- In the next example, please remove `colors_old.ts` and just replace the colors in `colors.ts` file you've created.
- Could we maybe just check for existence of `app.tsx` instead of passing around a boolean `appFileExists`?

deibeljc:
@flexdinesh I am currently really looking for this capability. Are there still plans/work being done to replicate this behavior in some manner? Or is finishing out this PR the only way to get this functionality in the short term?

PR #8021, opened 2022-10-20
Fix selection for component blocks without child fields

emmatown:
Now you can actually put your text cursor in a component block (and therefore insert content below it or etc.) that doesn't have child fields and the bar on the side will change colour to indicate your selection is inside there.

Try navigating your caret to the carousel block on the docs site:
on this branch: https://keystone-next-docs-git-fix-selection-for-compo-824f49-thinkmill.vercel.app/docs/guides/document-field-demo
before this branch: https://keystone-next-docs-pm1f8fo8c-thinkmill.vercel.app/docs/guides/document-field-demo

### DIFF
--- a/.changeset/rich-tools-share.md
+++ b/.changeset/rich-tools-share.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/fields-document': patch
+---
+
+Fixes selection for component blocks without child fields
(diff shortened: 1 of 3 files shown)


flexdinesh (approved):

mitchellhamilton:
(found a problem)

mitchellhamilton:
The problem was that you would see all of the child fields after the form (yes, that could be desirable behavior but it would need to be presented well which this was not), I've fixed that by reverting the change for edit mode. So selection won't work right when in the edit mode of a component block, I think that's good enough since people can click done and then it'll work right again.

Issue #8024, opened 2022-10-20
TypeError: Cannot destructure property 'operation' of 'access' as it is undefined.


riccardi1997:
### How to reproduce the problem
Try to reproduce this example by clicking [here](https://github.com/keystonejs/keystone/tree/main/examples/embedded-nextjs) and run `npm run dev` after cloned the example.

### What I'm expecting
I'm expecting the server start but I receive an error after `keystone postintstall` command. 

### The error
![error](https://user-images.githubusercontent.com/29570607/196919205-e38a7b84-9d6e-48cb-9d29-f2891a563594.png)

### Contextual informations
There is also a syntax error [here](https://github.com/keystonejs/keystone/blob/main/examples/embedded-nextjs/with.js) at line 15 on `const { connect, context }` with error code ts(2339). Now getContext is available but it doesn't work properly.

### Greetings
I hope to be helpful. Thanks

mmachatschek:
I ran into the same issue. I suppose [this change](https://github.com/keystonejs/keystone/pull/7968/files#diff-4e89868a26b75da76e42152ddc2af3928f661aecafb350b195578de76574696e) breaks this.

dcousens:
Thanks @riccardi1997, that is a different issue and I have opened a pull request to fix that Javascript example.

PR #8026, opened 2022-10-20
Fix embedded-nextjs Javascript example

dcousens:
Thanks to @riccardi1997 for pointing this out in https://github.com/keystonejs/keystone/issues/8024


flexdinesh (approved):

mmachatschek:
@dcousens just for clarification, how does one go about calling `connect` and `disconnect` if you only get `context` from `import { getContext } from '@keystone-6/core/context';`?

is keystone calling it in the background?

PR #8032, opened 2022-10-21
Fix where query for embedded-nextjs example

bketelsen:
Fix query in `getStaticProps()` for next's embedded example.


dcousens (approved):

dcousens:
Thanks @bketelsen :pray: , not having Typescript in that example is a bit of a blind spot

PR #8035, opened 2022-10-23
Fixed integer filter one-of and no one-of

gautamsi:
This has been annoying for long. 
I tried to fix it by minimal changes while preserving other case and concerns when other single values filters are used.

I also find that using regex already prevents the non numeric values.

this helps filtering on multiple integer values especially in case of auto-increment.

### DIFF
--- a/.changeset/ten-pears-worry.md
+++ b/.changeset/ten-pears-worry.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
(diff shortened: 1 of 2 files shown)


flexdinesh:
Relevant PR #7930. Fix seems to be the same.

gautamsi:
sorry, did not notice. 

do you plan to merge soon?

flexdinesh:
Yeah that's the plan. Once it's reviewed we'll merge soon. I'll close this PR as duplicate now.

PR #8038, opened 2022-10-23
Fix changes to some config not being updated in live reloads

### LINKED ISSUE #8028: Infinite reload when session secret is reset
When session secret is reset between hot reloads during development, Admin UI goes into an infinite reload loop. 

It's easy to reproduce with the latest `create-keystone-app@8.0.0` where [we generate a new session secret with every reload](https://github.com/keystonejs/create-keystone-app/issues/372).

## Steps to reproduce

1. `npx create-keystone-app@latest session-bug`
2. Create your first user and login to the Admin UI
3. Add a random `console.log()` anywhere in the code
4. Refresh Admin UI and it will infinitely reload

emmatown:
Fixes #8028

### DIFF
--- a/packages/core/src/lib/server/createAdminUIMiddleware.ts
+++ b/packages/core/src/lib/server/createAdminUIMiddleware.ts
@@ -10,28 +10,33 @@ const adminErrorHTMLFilepath = path.join(
   'admin-error.html'
 );
 
-export const createAdminUIMiddleware = async (
-  config: KeystoneConfig,
-  createContext: CreateContext,
-  dev: boolean,
-  projectAdminPath: string
-) => {
+type NextApp = {
+  prepare(): Promise<void>;
+  getRequestHandler(): express.Application;
+  render(req: express.Request, res: express.Response, url: string): void;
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/server/createAdminUIMiddleware.ts
```diff
@@ -10,28 +10,33 @@ const adminErrorHTMLFilepath = path.join(
...
+  const handle = nextApp.getRequestHandler();
+
+  const { ui, session } = config;
   const publicPages = ui?.publicPages ?? [];
   return async (req: express.Request, res: express.Response) => {
     const { pathname } = url.parse(req.url);
-    if (
-      pathname?.startsWith('/_next') ||
-      pathname === (graphql?.path || '/api/graphql') ||
-      pathname === '/api/__keystone_api_build'
```
mitchellhamilton, line 33: These deletions don't fix anything, they just became unnecessary after #7910 and #7809 respectively

flexdinesh (approved):

PR #8046, opened 2022-10-25
Adds CLI flags to `dev` `build` and `start` to change what is built and run 

borisno2:
Currently, Keystone provides the `useMigrations` config option to either enable or disable the generation of Prisma Migrations. When `useMigrations` is `false`, the default, running `keystone dev` performs a schema push to the DB and forces a reset if there is any drift. When `useMigrations` is `true` running `keystone dev` will generate the prisma migrations and also reset the DB if there is any drift.

The aim of this PR to give the developer more options when it comes to migrations. The current behaviour remains however the following is added:
- Updates `keystone prisma` CLI to use the built configuration when the `--frozen` flag is used so that uncompiled `ts` files are not required after a build step - helpful when building Docker images without needing `keystone.ts` and all its dependencies.
- Adds a `--no-db-push` flag for `keystone dev` that skips the schema push to the DB when `useMigrations` is set to `false` or `undefined`, used if you don't want to force keystone to use the database you have, for example, an existing DB you don't want to have migrations for.
- Changes the behaviour of `build` so that it generates the schema files and adds a `--frozen` flag to `build` to force it to just validate instead (current behaviour)
- Adds `--no-server` to `dev` to run dev without express, useful when not using express - ie Keystone in Next
- Adds `--no-ui` (`dev`, `build` and `start`) and `--no-prisma` (`dev` and `build`) to build and/or run keystone either without the AdminUI or without Prisma
- Marks `useMigrations`  as depreciated - this config option will be removed in the next major release - use `keystone prisma migrate dev` to generate database migrations. To mirror existing behaviour of `useMigrations` use `keystone prisma migrate dev && keystone dev --no-db-push`


.changeset/plenty-hairs-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': minor
```
borisno2, line 2: We are changing the default behaviour of `build` so maybe should be `major`?

packages/core/src/scripts/cli.ts
```diff
@@ -80,7 +81,7 @@ export async function cli(cwd: string, argv: string[]) {
   // WARNING: postinstall is an alias for `build --no-ui`
   if (command === 'postinstall') {
     flags.ui = false;
-    flags.frozen = !('fix' in flags);
+    flags.frozen = !flags.fix;
```
dcousens, line 81: This is unfortunate, but, `fix` needs to take precedence until it's removed

packages/core/src/scripts/prisma.ts
```diff
@@ -1,12 +1,18 @@
...
 
 export async function prisma(cwd: string, args: string[]) {
-  const config = await loadConfigOnce(cwd);
-
+  let config: KeystoneConfig;
+  if (process.env.NODE_ENV === 'production') {
+    config = loadBuiltConfig(cwd);
+  } else {
+    config = await loadConfigOnce(cwd);
+  }
```
dcousens, line 15: Use `--frozen` for this, so we are consistent.
Unexpected behaviour from `NODE_ENV` is not fun

packages/core/src/scripts/tests/artifacts.test.ts
```diff
@@ -37,15 +37,17 @@ describe('postinstall', () => {
...
     const files = await getFiles(tmp, schemasMatch);
     expect(files).toEqual(await getFiles(`${__dirname}/fixtures/basic-project`, schemasMatch));
     expect(recording()).toMatchInlineSnapshot(`
       "Your Prisma and GraphQL schemas are not up to date
-      Prompt: Would you like to update your Prisma and GraphQL schemas? true
-      ✨ GraphQL and Prisma schemas are up to date"
+      Prompt: Replace the Prisma and GraphQL schemas? true
+      ✨ GraphQL and Prisma schemas are up to date
+      ✨ Building Keystone
+      ✨ Skipping Admin UI code generation"
```
dcousens, line 50: Let's remove "Building Keystone" and "Skipping..." as messages to be displayed.  Their verbosity is unhelpful.

gautamsi:
what is the use case for keystone without `prisma`

borisno2:
> what is the use case for keystone without `prisma`

Good question @gautamsi - It will allow the building of just the Admin UI for example, to deploy the UI separately to the API

dcousens (approved):

gautamsi:
@borisno2 Am I able to build the admin ui separately yet?

dcousens:
Moved the removal of `--reset-db` to https://github.com/keystonejs/keystone/pull/8302, so we can merge this pull request and move forward.

gautamsi:
What is the reasoning behind deprecation of `useMigration`? I see it will be removed but I do not understand why.

PR #8073, opened 2022-11-07
Remove `createContext`

borisno2:
Also passes a generic `context` into `session` functions instead of the previous `createContext`

Providing `res` to `context.withRequest` is optional, however, if not provided `startSession` and `endSession` will not be available on the newly created context as these both currently require `res` to start a new session and end an existing one.

This PR also changes the types returned from the `start` and `end` in `SessionStrategy` to be more generic, the types are now `Promise<any>` instead of `Promise<string>` and `Promise<void>` for start and end respectively. 

## Request Context
This PR also removes `createRequestContext` which was passed into `config.extendExpressApp` and `config.extendHttpServer` and instead passes through a blank context which a new request context can be created from using `context.withRequest(req,res)`.

Previously creating a request context in `extendExpressApp` looked like this
```ts
extendExpressApp: (app, createContext) => {
  app.use('/rest', async (req, res, next) => {
    (req as any).context = await createContext(req, res);
    next();
  });
```

After this PR it will look like this:
```ts
extendExpressApp: (app, context) => {
   app.use('/rest', async (req, res, next) => {
     (req as any).context = await context.withRequest(req, res);
     next();
  });
```


dcousens (approved):

jmas:
Cool. Waiting when it will be released. Thanks!

Issue #7852, opened 2022-08-26
`server.options.host` is sporadically used


genu:
## TLDR
The `server.options.hosts` property should allow a user to provide a custom hostname for the server. If one is provided, like so:

```ts
server: {
      options: {
        host: 'my-special-app.local',
      },
    },
```

The host is properly set for the dev server at boot up, but the host is not properly handled across the rest of the stack.

## Some observations:
- Many of the tests assume the host is always localhost
- The admin-ui seems to have localhost hardcoded because setting the `host` property breaks the `admin-ui`
- Many the logs assume host is always `localhost` (see below)

https://github.com/keystonejs/keystone/blob/c2275621d7e96576acd59de7636095c432b3570e/packages/core/src/scripts/run/dev.ts#L297-L305

dcousens:
Thanks for reporting this @genu, please don't hesitate to open a pull request if you have the time :blue_heart:

vikasvmads:
Hi, Can I work on this issue. :)

josemf:
I don't think this is well documented either.

Some environments (some containers for example) really need need the http and graphql servers to be bound to a network interface other than localhost. 

Please add server option `server.options.host` to the documentation page: https://keystonejs.com/docs/config/config#server

Thank you!

dcousens:
@josemf pull requests are welcome! No doubt this is important to document, but we haven't had the opportunity to do this yet.

dcousens:
Added in https://github.com/keystonejs/keystone/pull/8212

PR #8078, opened 2022-11-08
Log configured host

u-ishii:
Log the configured host instead of localhost when starting Keystone.

I checked that the log has changed by sandbox.

This PR is related to #7852.

I searched [all codes that uses `localhost`](https://github.com/keystonejs/keystone/search?q=localhost), and it looks like there is no other code that needs fixing.
They were mostly example READMEs and test codes.
Also, #7852 reported that the Admin UI was broken, but now it seems to work.
Please let me know if I have overlooked anything.

### DIFF
--- a/.changeset/loud-moose-hunt.md
+++ b/.changeset/loud-moose-hunt.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Log the configured host instead of localhost when starting Keystone.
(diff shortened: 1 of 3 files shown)


u-ishii:
I noticed that the behavior is incorrect when not configured.

I mark this PR as a draft until I finish fixing it.

u-ishii:
Made it work correctly even if host is not configured.

dcousens (approved):
Thanks @u-ishii! Nice pull request

PR #8088, opened 2022-11-11
Add field groups

emmatown:
This adds the ability to group fields together in the Admin UI with a label and description

```tsx
list({
  ...etc,
  fields: {
    ...group({
      label: "Some group",
      description: "Some group description"
      fields: {
        something: text(),
        ...etc,
      }
    }),
    ...etc,
  }
})
```

Keystone's Admin UI on an item with a group of fields expanded
Keystone's Admin UI on an item with a group of fields collapsed


packages/core/src/admin-ui/system/adminMetaSchema.ts
```diff
@@ -187,9 +203,12 @@ const KeystoneAdminUIListMeta = graphql.object<ListMetaRootVal>()({
     fields: graphql.field({
       type: graphql.nonNull(graphql.list(graphql.nonNull(KeystoneAdminUIFieldMeta))),
     }),
+    groups: graphql.field({
+      type: graphql.nonNull(graphql.list(graphql.nonNull(KeystoneAdminUIFieldGroupMeta))),
+    }),
     initialSort: graphql.field({ type: KeystoneAdminUISort }),
     ...contextFunctionField('isHidden', graphql.Boolean),
-    isSingleton: graphql.field({ type: graphql.Boolean }),
+    isSingleton: graphql.field({ type: graphql.nonNull(graphql.Boolean) }),
```
emmatown, line 211: This is an unrelated I noticed when re-generating the types for the admin meta query, I don't think it was intentionally nullable. Types elsewhere assumed it was non-nullable so I've made it non-nullable.

dcousens (approved):

PR #8089, opened 2022-11-14
fix ui.labelField use multiselect type throw label must be scalar types bug.

dongwenxiao:
When I custom admin UI, I will use `authenticatedItem.label` from `NavigationProps`. 
It default use `name` field,But my logic need use array field, like `authenticatedItem.label` is `["student", "teacher"]`, so I set `ui.labelField` is `role`, like below 👇🏻

```ts
// my example code
const User = list({
  fields: {
    role: multiselect({ options: [ROLE.student, ROLE.teacher] }),
  },
  ui: {
    labelField: 'role',   // <== this use multiselect type, when you build code, it will throw 
  },
});
```
But, when I build code, throw theres errors 👇🏻

`Error: Label fields must be scalar GraphQL types but the labelField "role" on the list "User" is not a scalar type
    at getLazyMetadataQuery (/[PROJECT_PATH]/node_modules/@keystone-6/core/scripts/cli/dist/keystone-6-core-scripts-cli.cjs.dev.js:197:17)`

Then, I found the [scalar types](https://keystonejs.com/docs/fields/overview#scalar-types) include `multiselect`, so I think this is a bug and I want fix it.
image

### DIFF
--- a/packages/core/src/admin-ui/templates/app.ts
+++ b/packages/core/src/admin-ui/templates/app.ts
@@ -3,6 +3,7 @@ import {
   executeSync,
   GraphQLNonNull,
   GraphQLScalarType,
+  GraphQLList,
   GraphQLSchema,
   GraphQLUnionType,
   parse,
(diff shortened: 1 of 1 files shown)


dcousens:
Hi @dongwenxiao,

This isn't actually a bug, where `NonNull<T>` is trivially unwrapped to `T`, we don't necessarily agree that `List<T>` should be unwrapped to `T`.

The implicit Javascript behaviour of `['a', 'b', 'c'].toString() === 'a,b,c'` isn't necessarily something we want to rely on for this.

I think this would be better left to a virtual field.

dongwenxiao:
Thank you for your reply. @dcousens 

I see your recommend virtual field function can be used. 👍🏻

And, if it do not fix, then anyone use `multiselect` type field on `ui.labelField` stil it will throw error when code build.
I hope keystoneJS friendly support it, let us write less.

PR #8092, opened 2022-11-16
Add types for examples/blog

dcousens:

PR #8096, opened 2022-11-17
Tidy up examples, and add types

dcousens:
- Follow up to https://github.com/keystonejs/keystone/pull/8092 (branch name was annoying locally)
- Needs https://github.com/keystonejs/keystone/pull/8298

The examples are, honestly, disorganized and incomplete.
We actively need examples that show-case how to structure your project, and examples that demonstrate and explain the complexity of different features.  Some examples might show how a particular use-case might play out, that is OK, but they should be identifiably different from our feature examples.

Before this pull request, our access control and authentication examples did not showcase things like `isFilterable` or `isOrderable`, nor did they showcase `filter` or `item` access control.

**As of this pull request, many of our examples still need plenty of work, but they are an improvement on what we had.**


examples/access-control/schema.ts
```diff
@@ -0,0 +1,54 @@
...
+        access: {
+          read: isAdmin,
+        },
+        isFilterable: false,
+        isOrderable: false,
+      }),
+
+      isAdmin: checkbox({
+        access: {
+          read: isAdmin,
```
borisno2, line 44:
```suggestion
          read: isAdmin,
          update: isAdmin
```

examples/auth/schema.ts
```diff
@@ -1,67 +1,139 @@
...
+  // otherwise, no
+  return false;
+}
+
 export const lists: Lists = {
   User: list({
     access: {
       operation: {
-        ...allOperations(allowAll),
-        // Only allow admins to delete users
```
dcousens, line 11: Probably the example should have prevented other operations too...

examples/auth/schema.ts
```diff
@@ -1,67 +1,147 @@
...
+        //   the access.filter.* and access.item.* access controls
+        update: hasSession,
+
+        // only allow admins to delete users
+        delete: isAdmin
+      },
+      filter: {
+        update: isAdminOrSameUserFilter
       },
+      item: {
```
borisno2, line 71: Is this doing the same thing as `filter` - maybe a better example of `item` access control is something like you can't update an email and a password at the same time??

borisno2 (approved):

PR #8013, opened 2022-10-18
Add ability to create a new context with a `req` using `context.withRequest`

borisno2:
Adds the ability to create a new context from a `req` and `res` using `context.withRequest` - this returns a new `context` with a `session` based on the provided `req`.

PR #8097, opened 2022-11-17
Allow customisation of the session object through new `getSession` config

borisno2:
This PR continues the context and session work from #8013 

The goal of this PR to simplify how keystone handles different session implementations. With that in mind This PR does the following:

### Moves `statelessSessions` and `storedSessions` functions into the `auth` package.
These Session Strategies are fairly simple cookie-based session management which do the following
 - `start` - starts a session by setting a cookie on the server `res` and in the case of `storedSessions` add the session data to the store.
 - `get` - decodes the data from the cookie in `req` returns data for `context.session` and
 - `end` - sends a blank cookie back in the server `res`
These functions are based around keystone handling the start and end of a session, however, this is not always the case - for example, SSO or other token-based auth.

By moving this more prescriptive session setup into auth the aim is to allow more complex auth and session management strategies to be implemented

This is a BREAKING change to config and means the following:
- Change location of import for sessionStrategy
```ts
import { statelessSessions } from '@keystone-6/core/session';
```
becomes:
```ts
import { statelessSessions } from '@keystone-6/auth/session';
```
- Move `session` config from main config into `createAuth` `sessionStrategy` config and move `sessionData` into `sessionStrategy` config - this means the following:
```ts
const { withAuth } = createAuth({
 /* ... Existing Auth Config */
sessionData: 'name isAdmin',
});

export default withAuth(
  config({
 /* ... Existing Keystone Config */
    session: statelessSessions({
         maxAge: sessionMaxAge,
         secret: sessionSecret,
       }),
   })
 );
});
```
Becomes:
```ts
const { withAuth } = createAuth({
 /* ... Existing Auth Config */
    sessionStrategy: statelessSessions({
         maxAge: sessionMaxAge,
         secret: sessionSecret,
         data: 'name isAdmin',
       }),
});

export default withAuth(
  config({
 /* ... Existing Keystone Config */
   })
 );
});
```

### Adds new `getSession` object to `config` - removing current `session` and `sessionStrategy`
With the current `session` and `sessionStrategy` implementation moved to auth, Keystone still needs a way to populate the `session` object on the `context`. In order to facilitate this a new `getSession` option is added to `config` this passes the current `context` - which if `withRequest` has been used will contain the current `req` and `res` - and accepts a function that returns a valid session, or `undefined` if the session is not valid. The return of `getSession` is added to the `context` as `session`.

This should allow for easier configuration of more complex session strategies.


emmatown:
Superseded by #9579

Issue #8085, opened 2022-11-09
initFirstItem not working with ui.isAccessAllowed (config)


perlohmann:
Create a regular test project (with empty db) that uses authentication and enable initFirstItem
```
initFirstItem: {
    fields: ["email", "password"],
    itemData: {
      isAdmin: true,
    },
    skipKeystoneWelcome: true,
  },
```
add the property isAccessAllowed to the ui config e.g. 
```
ui: {
      isAccessAllowed: (context) => {
        return (
          process.env.NODE_ENV === "development"
        );
      },
    },
```
You can use any project you have which uses auth and have those properties enabled, as long as it is an empty db.

if running on local host e.g. npm run dev then I would expect the /init form to show the first time you visit localhost, but it simply just goes directly to the dashboard where you cant really do anything since you are not logged in.

flexdinesh:
@perlohmann this is the expected behaviour. `initFirstItem` is for auth. When you bypass auth completely using `isAccessAllowed`, you can only browse public pages and cannot do anything in pages that are behind access control (Eg. dashboard.)

Perhaps we could find another way to achieve what you want. Could you explain a bit more on what you want to allow public access during development?

perlohmann:
@flexdinesh so we are using keystone as a backend api and admin cms for our frontend so we need to prevent our regular user from accessing the keystone cms. 
Besides the development flag in the example we also have a "|| isAdmin" condition which I removed to make the case simpler.

UI.isAccessAllowed does not bypass authentication. if I try to access our keystone UI on our production environment then I am prompted to login and it will determine if I have access based on "who" I log in as.

When developing then we would like to have access to the cms from all user types we create (because we filter available data based on the "owner" of the data and therefore we "cant" see the data from other accounts).
This works flawlessly so it is not that much of a consern. 

My consern and where this problem arises is if you create a new environment e.g. new developer joins, new test environment etc. and you don't have any users in the db. 
If this is the case you cant create a new first user because isAccessAllowed overwrites the initFirstItem and it will actually open up the admin interface anyways, but you cant view any data, because you are not authenticated.

I do not understand why UI.isAccessAllowed should overwrite the behaviour of the initFirstItem when it does not overwrite e.g. authentication, that has to be a bug in my mind.

flexdinesh:
@perlohmann I stand corrected. `ui.isAccessAllowed` does not bypass auth. We have identified the root cause and we're tracking the fix here.

- #8115

Issue #7970, opened 2022-09-29
Auth package's `pageMiddleware` doesn't respect `ui.publicPages`


flexdinesh:
Config `ui.publicPages` is bugged at the moment because of this line — https://github.com/keystonejs/keystone/blob/e322ce1223ef00dfb77f22b2f669e05d91b09814/packages/auth/src/index.ts#L116

All pages (including publicPages) are still checked for the existence of session object and only `/signin` and `/init` are allowed to be browsed without a session. 

## Temporary workaround

```js
const publicPageRoutes = ["/public-page", "/signup"];
...
ui: {
  isAccessAllowed: ({ req, session }) => {
    // don't check access for public pages
    if (publicPageRoutes.includes((req as any)?.originalUrl)) {
      return true;
    }

    // allow all logged in users
    return !!session?.data;
  }
}
```

**Codesandbox:** https://codesandbox.io/s/access-control-and-public-pages-uk06ez

But this workaround doesn't seem to work when you query params in the url path. Eg. `/public-page?hello=world`

reza27:
When is this going to be fixed?
Reset password requires a token query param. Is there a workaround for this in the meanwhile?

borisno2:
This was resolved in https://github.com/keystonejs/keystone/releases/tag/2022-12-06

Issue #7662, opened 2022-06-24
Infinite loop on setting config.ui.isAccessAllowed = false


chiu0602:
Actually I have set 
```isAccessAllowed: context.session?.data?.role === 'admin'```
, but for easy demonstration, I used `false` instead.

```
export default withAuth(
  config({
    ui: {
      isAccessAllowed: () => false,
      ...
    },
    ...
  },
  ...
```

Expected:
User cannot login.

Actual:
After inputting correct credentials and press "Sign in", infinite loop (successful login but jump back to login page).

mitchellhamilton:
We can’t seem to reproduce this issue based on the code you’ve shown. Could you provide a GitHub repo that reproduces this problem?

dcousens:
@chiu0602 can you still reproduce this?

flexdinesh:
Tracking here #8028

PR #8115, opened 2022-11-22
Fixes `initFirstItem` bypass when `ui.isAccessAllowed` is defined

dcousens:
This pull request rewrites the `@keystone-6/auth` middleware and `isAccessAllowed` composition.

Fixes
- https://github.com/keystonejs/keystone/issues/7970
- https://github.com/keystonejs/keystone/issues/8085

Related
- https://github.com/keystonejs/keystone/issues/7662

### DIFF
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -213,7 +213,7 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
   async function hasInitFirstItemConditions(context: KeystoneContext) {
     if (!initFirstItem) return false;
 
-    const count = await context.sudo().query[listKey].count({});
+    const count = await context.sudo().db[listKey].count({});
     return count === 0;
   }
 
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -109,9 +109,6 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
    * Must be added to the ui.publicPages config
    */
   const authPublicPages = ['/signin'];
-  if (initFirstItem) {
-    authPublicPages.push('/init');
-  }
 
   /**
    * extendGraphqlSchema
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -173,7 +173,7 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
    * withItemData
    *
    * Automatically injects a session.data value with the authenticated item
-  */
+   */
   const withItemData = (
     _sessionStrategy: SessionStrategy<Record<string, any>>
   ): SessionStrategy<{ listKey: string; itemId: string; data: any }> => {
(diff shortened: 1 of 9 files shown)


packages/auth/src/index.ts
```diff
@@ -255,44 +210,93 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
...
         isAccessAllowed: async (context: KeystoneContext) => {
-          // Allow access to the adminMeta data from the /init path to correctly render that page
-          // even if the user isn't logged in (which should always be the case if they're seeing /init)
-          const headers = context.req?.headers;
-          const host = headers ? headers['x-forwarded-host'] || headers['host'] : null;
-          const url = headers?.referer ? new URL(headers.referer) : undefined;
-          const accessingInitPage =
-            url?.pathname === '/init' &&
-            url?.host === host &&
-            (await context.sudo().query[listKey].count({})) === 0;
```
dcousens, line 289: I replaced this with
```
if (pathname === '/__nextjs_original-stack-frame') return
```

And for the meta endpoint the replacement for `isAccessAllowed` is simply to return `true` for this when `hasInitFirstItemConditions`.

packages/core/src/admin-ui/context.tsx
```diff
@@ -113,13 +113,6 @@ export const useKeystone = (): {
     throw new Error('useKeystone must be called inside a KeystoneProvider component');
   }
   if (value.adminMeta.state === 'error') {
-    // If we get an "Access denied" error, it probably means the user doesn't have access to the
-    // adminMeta but has navigated (probably client-side) to a page that requires it. We reload
-    // the page so the server-side access control can run which should bounce them to the right
-    // place (or display the no-access page)
-    if (value.adminMeta.error.message === 'Access denied') {
-      window.location.reload();
-    }
```
dcousens, line 122: We might want to think about how to do this differently, the infinite loop that can happen with this isn't helpful

packages/auth/src/index.ts
```diff
@@ -255,44 +210,93 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
...
-          return (
-            accessingInitPage ||
-            (keystoneConfig.ui?.isAccessAllowed
-              ? keystoneConfig.ui.isAccessAllowed(context)
-              : context.session !== undefined)
-          );
+          if (await hasInitFirstItemConditions(context)) return true;
```
dcousens, line 298: With effectively being a bypass, I wanted the function to be isolated

dcousens:
This will probably need to go out as a patch, independent of the upcoming work

packages/auth/src/index.ts
dcousens: Now replaced with https://github.com/keystonejs/keystone/blob/f6fb299cb987a27ce9c0885620dc2b68a6186ef8/packages/core/src/lib/server/createAdminUIMiddleware.ts#L46

packages/core/src/lib/server/createAdminUIMiddleware.ts
```diff
@@ -24,37 +24,47 @@ export async function getNextApp(dev: boolean, projectAdminPath: string): Promis
...
   context: KeystoneContext,
   nextApp: NextApp
 ) {
   const handle = nextApp.getRequestHandler();
 
-  const { ui, session } = config;
-  const publicPages = ui?.publicPages ?? [];
+  const {
+    ui: { isAccessAllowed = defaultIsAccessAllowed, pageMiddleware, publicPages = [] } = {},
+  } = config;
```
dcousens, line 41: Defaults > branching everywhere

flexdinesh (approved):

PR #8122, opened 2022-11-24
Fix `access.filter.*` passing `listKey: undefined`

### LINKED ISSUE #8121: list.access.filter.query "listKey" is undefined... but did get a "list" parameter
https://keystonejs.com/docs/config/access-control#filter

```
   access: {
        filter: {
          query: ({ session, context, listKey, operation }) => {
            return { isReadable: { equals: true } };
          },
          update: ({ session, context, listKey, operation }) => {
            return { isUpdatable: { equals: true } };
          },
          delete: ({ session, context, listKey, operation }) => {
            return { isDeletable: { equals: true } };
          },
        }
```
        
filter function got listKey as a param. however, listKey is undefined but got a list param here

image
for exmaple, here in the function,since the ts defination do not contain list param, so the red underline shows, but actually the list got the listKey and listKey got undefined, the actual log is this image below

image
Obviously there is a typo in the source code, when passing params to the filter function.

dcousens:
Fixes https://github.com/keystonejs/keystone/issues/8121 where `access.filter.*` operations were being passed `list: list.listKey` instead of `listKey: list.listKey`.

The typo fell through our types because we had an `as any` with the type being difficult to resolve.
Prior to c740ba630c we had `@ts-ignore` for this check:
https://github.com/keystonejs/keystone/blob/6c60565c940a9e27b8bb9b3e3e61260f22ee1e1a/packages/core/src/lib/core/access-control.ts#L63-L67

And then with #7914 we changed it to `as any` and introduced the regression (as shown in the pull request diff).

With this pull request, we resolve the types verbosely.  Not ideal, but, it's better than the alternative.

### DIFF
--- a/packages/core/src/lib/core/access-control.ts
+++ b/packages/core/src/lib/core/access-control.ts
@@ -82,7 +82,6 @@ export async function getAccessFilters(
         listKey: list.listKey,
         context,
       });
-
     } else if (operation === 'update') {
       filters = await list.access.filter.update({
         operation,
@@ -90,7 +89,6 @@ export async function getAccessFilters(
         listKey: list.listKey,
         context,
       });
-
     } else if (operation === 'delete') {
       filters = await list.access.filter.delete({
         operation,
--- a/packages/core/src/lib/core/access-control.ts
+++ b/packages/core/src/lib/core/access-control.ts
@@ -101,7 +101,7 @@ export async function getAccessFilters(
     }
 
     if (typeof filters === 'boolean') return filters;
-    if (filters === undefined) return false; // shouldn't happen, but, Typescript
+    if (!filters) return false; // shouldn't happen, but, Typescript
 
     const schema = context.sudo().graphql.schema;
     const whereInput = assertInputObjectType(schema.getType(getGqlNames(list).whereInputName));
--- a/packages/core/src/lib/core/access-control.ts
+++ b/packages/core/src/lib/core/access-control.ts
@@ -71,17 +71,37 @@ export async function getOperationAccess(
 export async function getAccessFilters(
   list: InitialisedList,
   context: KeystoneContext,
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/core/access-control.ts
```diff
@@ -71,17 +71,37 @@ export async function getOperationAccess(
...
+  operation: keyof typeof list.access.filter
 ): Promise<boolean | InputFilter> {
-  const access = list.access.filter[operation];
   try {
-    const filters = await access({
-      operation,
-      session: context.session,
-      list: list.listKey,
-      context,
-    } as any); // TODO: FIXME
```
dcousens, line 83: :fire:

packages/core/src/lib/core/access-control.ts
```diff
@@ -71,17 +71,35 @@ export async function getOperationAccess(
...
+        listKey: list.listKey,
+        context,
+      });
+    } else if (operation === 'delete') {
+      filters = await list.access.filter.delete({
+        operation,
+        session: context.session,
+        listKey: list.listKey,
+        context,
+      });
```
dcousens, line 98: Verbose, but, at least it type checks

packages/core/src/lib/core/access-control.ts
```diff
@@ -71,17 +71,35 @@ export async function getOperationAccess(
 export async function getAccessFilters(
   list: InitialisedList,
   context: KeystoneContext,
-  operation: 'update' | 'query' | 'delete'
+  operation: keyof typeof list.access.filter
```
dcousens, line 74: Unnecessary, but, nice

flexdinesh (approved):

dcousens:
Released https://github.com/keystonejs/keystone/releases/tag/2022-11-25

PR #7594, opened 2022-06-03
Remove Create Item Modal from Dashboard and List views

emmatown:
Closes #7583

See the CodeSandbox link in the comment below to try it (it will take a bit to load)

PR #8123, opened 2022-11-24
Prevent inline related item form from submitting the parent form

januzis:
Inline related item creation is currently semi-broken, because the related item form in a drawer is submitting the main form as well. This results either in the main form showing validation errors just after the related item is created, or, if all the fields in the main form were filled correctly, in the main form being submitted automatically without any value in the relation field.

I think the https://github.com/keystonejs/keystone/pull/7594 introduced the regression, when a separate page started being used to create new items, instead of using the drawer component.

**Steps to reproduce the problem (using `tests/test-projects/basic`):**
1. Go to the new task page (http://localhost:3000/tasks/create);
2. Fill in the "Label" field to make the main form valid (it's the only required field);
3. Click "Create related Person" button to open the person form in a drawer;
4. Fill in any name and click "Create Person" to submit the form;
5. You should see that the Task form has been submitted as well, and the value of the "Assigned To" field isn't set.

I've created a test following these steps, which fails in the current `main` branch, and passes with my fix. The fix itself is very simple - stop form submit event propagation in the drawer component.

### DIFF
--- a/.changeset/early-days-thank.md
+++ b/.changeset/early-days-thank.md
@@ -1,0 +1,5 @@
+---
+"@keystone-ui/modals": patch
+---
+
+Fixes inline related item form submitting the parent form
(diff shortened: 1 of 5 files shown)


dcousens (approved):
Works for me

emmatown (approved):

PR #8132, opened 2022-11-26
Changes default `listen` to remove `host`

### LINKED ISSUE #8131: Using `::` as default for host does not work in all intances
It looks like keystone if nothing is provided specifies `::` as `host` when creating the httpServer. However depending on the system and OS this does not bind to ipv4 AND ipv6. Areas that I've noticed is having issues like this:

* Docker on Windows (seen by myself)
* codesandbox.io samples (reported by someone in slack)

Other tools like Remix-Serve use the `listen` method without a host unless a host is specified ([code](https://github.com/remix-run/remix/blob/main/packages/remix-serve/cli.ts)).

**Current workaround:**

```
        server: {
            options: {
                 host: "",
            },
        }
```

The only downside the console log message when starting is now saying `http://:3000` but it listens on all interfaces now.

fkrauthan:
This fixes #8131 by doing the following changes:

1. Removed the default `host` provided inside `start` and `dev` command
2. Added support to set `host` with the env variable `HOST`
3. Adjusted log message to resolve `undefined`, `''`. `'::'`, and `'0.0.0.0'` to the message `(http://localhost:3000/)`

### DIFF
--- a/packages/core/src/scripts/run/start.ts
+++ b/packages/core/src/scripts/run/start.ts
@@ -42,7 +42,6 @@ export const start = async (cwd: string) => {
   }
 
   const httpOptions: ListenOptions = {
-    host: '::', // the default for nodejs httpServer
     port: 3000,
   };
 
@@ -59,12 +58,17 @@ export const start = async (cwd: string) => {
     httpOptions.port = parseInt(process.env.PORT || '');
   }
 
+  // preference env.HOST if supplied
+  if ('HOST' in process.env) {
+    httpOptions.host = process.env.HOST || '';
+  }
+
   httpServer.listen(httpOptions, (err?: any) => {
     if (err) throw err;
 
-    const easyHost = httpOptions.host === '::' ? 'localhost' : httpOptions.host;
+    const easyHost = [undefined, '', '::', '0.0.0.0'].indexOf(httpOptions.host) !== -1 ? 'localhost' : httpOptions.host;
     console.log(
-      `⭐️ Server listening on ${httpOptions.host}:${httpOptions.port} (http://${easyHost}:${httpOptions.port}/)`
+      `⭐️ Server listening on ${httpOptions.host || ''}:${httpOptions.port} (http://${easyHost}:${httpOptions.port}/)`
     );
   });
 };
--- a/packages/core/src/scripts/run/start.ts
+++ b/packages/core/src/scripts/run/start.ts
@@ -66,7 +66,7 @@ export const start = async (cwd: string) => {
   httpServer.listen(httpOptions, (err?: any) => {
     if (err) throw err;
 
-    const easyHost = [undefined, '', '::', '0.0.0.0'].indexOf(httpOptions.host) !== -1 ? 'localhost' : httpOptions.host;
+    const easyHost = [undefined, '', '::', '0.0.0.0'].includes(httpOptions.host) ? 'localhost' : httpOptions.host;
     console.log(
       `⭐️ Server listening on ${httpOptions.host || ''}:${httpOptions.port} (http://${easyHost}:${httpOptions.port}/)`
     );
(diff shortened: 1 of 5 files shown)


fkrauthan:
This might require a changeset? Not 100% sure

packages/core/src/scripts/run/start.ts
```diff
@@ -59,12 +58,17 @@ export const start = async (cwd: string) => {
...
+  // preference env.HOST if supplied
+  if ('HOST' in process.env) {
+    httpOptions.host = process.env.HOST || '';
+  }
+
   httpServer.listen(httpOptions, (err?: any) => {
     if (err) throw err;
 
-    const easyHost = httpOptions.host === '::' ? 'localhost' : httpOptions.host;
+    const easyHost = [undefined, '', '::', '0.0.0.0'].indexOf(httpOptions.host) !== -1 ? 'localhost' : httpOptions.host;
```
dcousens, line 69:
```suggestion
    const easyHost = [undefined, '', '::', '0.0.0.0'].includes(httpOptions.host) ? 'localhost' : httpOptions.host;
```

dcousens:
If we nominate that https://github.com/keystonejs/keystone/issues/8131 is a regression, then we need a `patch` changeset for that.

> Added support to set host with the env variable HOST

Additionally, this new feature needs a `minor` changeset :+1:

fkrauthan:
> If we nominate that #8131 is a regression, then we need a `patch` changeset for that. (I think it is a bug, personally)
> 
> > Added support to set host with the env variable HOST
> 
> Additionally, this new feature needs a `minor` changeset 👍

Sounds good. Which package does the changeset need to be applied to? core?

dcousens:
Thanks for the pull request @fkrauthan,  what a rollercoaster this "logging" change https://github.com/keystonejs/keystone/pull/8078 has been.
We logged one behaviour (listening on localhost), but actually did the other (listening on `::`).

Unfortunate that `::` isn't actually working on every platform, but I wonder, what _is_ the behavior now? The exact behaviour isn't necessarily strictly defined in the [node documentation](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback).

I think `host: undefined` is probably the best option, shame about the variability.

dcousens:
> Sounds good. Which package does the changeset need to be applied to? core?

Yup
```
@keystone-6/core
```

fkrauthan:
> Thanks for the pull request @fkrauthan, what a rollercoaster this "logging" change #8078 has been. We logged one behaviour (listening on localhost), but actually did the other (listening on `::`).
> 
> Unfortunate that `::` isn't actually working on every platform, but I wonder, what _is_ the behavior now? The exact behaviour isn't necessarily strictly defined in the [node documentation](https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback).
> 
> I think `host: undefined` is probably the best option, shame about the variability.

I believe by default (if nothing is specified, undefined or empty string) it listens to all available interfaces. So if someone wants it to be explicit they would configure host explicitly. Given that a lot of people running there software in docker etc you usually want to just listen to all interfaces

dcousens:
> Added support to set host with the env variable HOST

I'm ambivalent about this, users can easily add to their own Keystone configuration, and personally I'd encourage that for `PORT` too. (the only difference there, is `PORT` already exists so it isn't worth removing for a breaking change).

fkrauthan:
> > Added support to set host with the env variable HOST
> 
> I'm ambivalent about this, users can easily add to their own Keystone configuration, and personally, I'd encourage that for `PORT` too. (the only difference there, is `PORT` already exists so it isn't worth removing for a breaking change).

Well, I was kind of following to train of thought here:

1. We already have `PORT` so it only make sense to have `HOST` as well.
2. This seem to be pretty standard across software to have `PORT` and `HOST` configurable through env variables

I personally don't need it (I just need default host `::` removed) so if you feel strongly about this can be removed.

dcousens (approved):

PR #8167, opened 2022-12-10
Added @latest scope to npm create-keystone-app commands

### LINKED ISSUE #8154: npm init/create reuses cached version without @latest scope
### Reproduction

Bit tricky, but if you have a previously created a keystone app using the `npm init keystone-app` there is a chance it will try to reuse a cached version. It does notify you by printing: 

```bash
⚠️  You're running an old version of create-keystone-app, please update to 8.1.0
✔ What directory should create-keystone-app generate your app into? · ks-auth-unresponsive
```

I hadn't managed to find a way to update it cleanly.

### Proposed solution

Use `@latest` scope in docs to guide people to always pull the latest create-keystone-app. This seems to be "industry standard" way of telling people to use their create app. [NextJS](https://nextjs.org/docs/getting-started#automatic-setup), [SvelteKit](https://kit.svelte.dev/docs/creating-a-project)

Happy to create a PR!

adamjkb:
Fixes #8154

### DIFF
--- a/docs/pages/docs/getting-started.md
+++ b/docs/pages/docs/getting-started.md
@@ -36,7 +36,7 @@ Use [`npm init`](https://docs.npmjs.com/cli/v7/commands/npm-init):
 
 ```sh
 cd your/path/
-npm init keystone-app
+npm init keystone-app@latest
 ```
 
 {% hint kind="warn" %}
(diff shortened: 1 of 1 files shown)


flexdinesh (approved):
Thanks @adamjkb! 🙌

dcousens (commented):
Why is this necessary?

dcousens (approved):

PR #8173, opened 2022-12-13
New blog post: Use Keystone in Next.js applications

flexdinesh:
- Updated old embedded keystone blog post and added a note that it's outdated with a link to the new blog post
- New blog post on _how to use Keystone in Next.js applications_
  - Visual representation to understand the mental model
  - Link to the example boilerplate
  - Link to demo


docs/pages/blog/nextjs-keystone.md
```diff
@@ -0,0 +1,137 @@
+---
+title: "Use Keystone in Next.js applications"
+description: "Learn how to use Keystone in Next.js apps and deploy to Vercel."
+publishDate: "2022-12-14"
+authorName: "Dinesh Pandiyan"
+authorHandle: "https://twitter.com/flexdinesh"
+metaImageUrl: ""
+---
+
+Keystone is an excellent headless CMS that provides a GraphQL API to access your data. But did you also know that you could use **Keystone as a data layer within Next.js applications**? Yes, you don't need to host Keystone on a separate server. Keystone's entire data layer is so well abstracted that you could just import the `context` API in a Next.js application and CRUD your data. In other words, Keystone can be imported and used in serverless environments like Next.js routes, `getStaticProps`, `getServerSideProps` and `next@13` server components.
```
emmatown, line 10: >`next@13` server components

Have you checked that? I wouldn't be surprised if it didn't work

flexdinesh: Uhm no. I've seen examples with prisma so I thought the context API would work as well. To be safer, I will remove this from the post until we test if it will work.

docs/pages/blog/nextjs-keystone.md
```diff
@@ -0,0 +1,137 @@
...
+- You deploy a separate frontend and a separate keystone server.
+- Your frontend will send network requests to Keystone server's GraphQL API to CRUD data. Eg. During server rendering, in `getStaticProps`, `getServerSideProps` or in `next@13` server components, you will send http requests to Keystone server's GraphQL API to query or mutate your data. 
+
+With the release of [core v3](https://github.com/keystonejs/keystone/releases/tag/2022-10-19), **Keystone can now be used within other servers**. Keystone's data layer is well abstracted and exported as node APIs that you could import and use in node based servers like Next.js and deploy to [Vercel](https://vercel.com). 
+
+![Keystone in Next.js](/assets/blog/images/nextjs-keystone/keystone-in-nextjs.png)
+
+This is how Keystone works within a Next.js application:
+
+- You import Keystone's `getContext` in your Next.js application. It is a wrapper around Keystone's data layer.
```
borisno2, line 31: It might be worth calling out that `getContext` returns a `commonContext` and you can create a context specific to the request using `withRequest` - which includes a `session` if enabled and available - or an elevated context using `sudo`

borisno2 (commented):
Nice one @flexdinesh

docs/pages/blog/nextjs-keystone.md
flexdinesh: In the spirit of keeping the content simple for readers who don't have much context about keystone, it will be good to keep the complexity to minimum @borisno2. That's why we didn't get into the details of how auth/session is handled and left it for the reader to explore it themself by looking at the example. I will update the example with more information about `commonContext` vs `context.withRequest`. That would be much better for anyone going through our examples to understand the deets.

borisno2: Sounds good @flexdinesh

borisno2 (approved):

docs/pages/docs/examples.tsx
```diff
@@ -248,6 +248,16 @@ export default function Docs() {
...
         </Well>
+
+        <Well
+          grad="grad2"
+          heading="Next.js + Keystone"
+          href="https://github.com/keystonejs/keystone/blob/main/examples/nextjs-keystone"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          Shows you how to use Keystone as a data engine withing Next.js applications.
```
emmatown, line 259:
```suggestion
          Shows you how to use Keystone as a data engine within Next.js applications.
```

flexdinesh: thanks for catching this. I've pushed this change along with another change in a separate commit.

emmatown (approved):

PR #8177, opened 2022-12-15
Fix creating new relationships in card view on create item page

### LINKED ISSUE #8149: Inline related item not working when using cards. Refresh the page
I have noticed this error after upgrading `keystone-6/core` from 1.1.0 to 3.1.2

I created a repository https://github.com/davidherdu/keystonejs-example with the example from https://github.com/keystonejs/keystone/tree/main/examples/task-manager

The only change is that I am using `displayMode: 'cards'` to create a Person when creating a Task, everything else being the same.

### How to reproduce the problem

1. Clone the repository https://github.com/davidherdu/keystonejs-example
2. Start the app in development mode and open the admin UI in the browser
3. Try to create a task
4. While creating a task, try to create a Person, after clicking the Create Person button, the page refreshes and the data entered for the Task and the Person are not saved

### What I'm expecting

I would expect the same behavior as before, after clicking Create Person, the person would be created, the page would not refresh, and the data entered in the task form would be there.

### Screenshot
I add a record where you can see after clicking Create Person that neither the Person nor the Task are created and the page is refreshed.

https://user-images.githubusercontent.com/5971103/205259102-0a48903a-26ee-4c77-96b2-10a779e4725f.mov

### Contextual informations

I have tried that in chrome, firefox and safari and it all happens

borisno2:
Fixes #8149

### DIFF
--- a/design-system/packages/segmented-control/src/SegmentedControl.tsx
+++ b/design-system/packages/segmented-control/src/SegmentedControl.tsx
@@ -146,7 +146,7 @@ export const SegmentedControl = ({
           <VisuallyHidden
             as="input"
             type="radio"
-            onChange={() => ''}
+            onChange={() => {}}
             checked={nothingIsSelected}
             disabled={actuallyDisabled}
             value="none"
--- a/design-system/packages/segmented-control/src/SegmentedControl.tsx
+++ b/design-system/packages/segmented-control/src/SegmentedControl.tsx
@@ -146,6 +146,7 @@ export const SegmentedControl = ({
           <VisuallyHidden
             as="input"
             type="radio"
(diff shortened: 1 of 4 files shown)


flexdinesh (approved):

design-system/packages/segmented-control/src/SegmentedControl.tsx
```diff
@@ -146,6 +146,7 @@ export const SegmentedControl = ({
           <VisuallyHidden
             as="input"
             type="radio"
+            onChange={() => ''}
```
borisno2, line 149:
```suggestion
            onChange={() => {}}
```

PR #8179, opened 2022-12-18
Add file upload in nextjs-keystone example

### LINKED ISSUE #8176: File upload does not work in 'keystone-in-nextjs' example
Thank you for providing new example (https://github.com/keystonejs/keystone/pull/8148) and other changes leading into it. I tried embedded-nextjs example in the past, but abandoned it after realising that uploading files does not work and there is no easy way to replace generated 'withKeystone' with something else. This new way looks much brighter, but file upload functionality also does not work out-of-the-box.

## Steps to Reproduce
1. Add some image/file field to the schema. Eg. https://github.com/marekryb/keystone/commit/2aca0fedb5e085f67c9d3c078b8ac4c13a1e0a5e
2. Run the keystone server in standalone mode (port 3000)
> yarn keystone:dev

3. Run mutation that uploads a file
NOTE: Standard playground does not support file uploads, so I am using curl. You can also use Altair GraphQL client or do it programatically.
> curl localhost:3000/api/graphql \\
  -F operations='{ "query": "mutation Test($id: ID!, $file: Upload!) {updateUser(where:{id:$id},data:{avatar:{upload:$file}}) { id } }", "variables": { "id": "{{USER_ID}}", "file": null } }' \\
  -F map='{ "0": ["variables.file"] }' \\
  -F 0=@{{FILE_NAME}}

Need to replace USER_ID and FILE_NAME in above. In my case final query is eg.
> curl localhost:3000/api/graphql \\
  -F operations='{ "query": "mutation Test($id: ID!, $file: Upload!) {updateUser(where:{id:$id},data:{avatar:{upload:$file}}) { id } }", "variables": { "id": "clbafdgom00135y3upyadgilw", "file": null } }' \\
  -F map='{ "0": ["variables.file"] }' \\
  -F 0=@test.jpg

Result is as expected
> {"data":{"updateUser":{"id":"clbafdgom00135y3upyadgilw"}}}

4. Run nextjs with embedded keystone (port 4000)
> yarn next:dev

5. Run same mutation as in point 3, but for port 4000
>curl localhost:4000/api/graphql \\
  -F operations='{ "query": "mutation Test($id: ID!, $file: Upload!) {updateUser(where:{id:$id},data:{avatar:{upload:$file}}) { id } }", "variables": { "id": "clbafdgom00135y3upyadgilw", "file": null } }' \\
  -F map='{ "0": ["variables.file"] }' \\
  -F 0=@test.jpg
6. Observe error
> {"errors":[{"message":"Variable \"$file\" got invalid value {}; Upload value invalid.","locations":[{"line":1,"column":25}]}]}

## Expected Behavior
File upload should also work in embedded mode.

## Possible Solution
You can fix this problem by manually pre-processing requests with graphql-upload, before passing them to Yoga. See: https://github.com/marekryb/keystone/commit/bae9841622d237a2c9b40b248c34a66fd0a49447

- Is there some way to make yoga work with keystone upload type by default?
- Would you like my changelist as PR to keystone-in-nextjs example?

marekryb:
Fixes https://github.com/keystonejs/keystone/issues/8176


marekryb:
@flexdinesh There is one unsolved issue. In keystone.ts / generateUrl there is hardcoded port 4000. This will work fine for 'yarn next:dev', but will not work for 'yarn keystone:dev' (port 3000). Normally I would pass this as some env variable, but not sure how to do this in this example without complicating things. Some guidance please.

examples/nextjs-keystone/src/keystone/schema.ts
```diff
@@ -19,7 +19,7 @@ const permissions = {
 export const lists: Lists = {
   User: list({
     // readonly for demo purpose
-    access: permissions.readOnly,
+    access: allowAll,
```
flexdinesh, line 22: We probably have to change the permission to `readOnly`. This example is deployable to vercel in one click using the button in the docs by anyone. `allowAll` gives CRUD access to all fields and it's probably not safe for an example deployed on vercel with public write access.

examples/nextjs-keystone/keystone.ts
```diff
@@ -17,5 +17,16 @@ export default withAuth(
     },
     lists,
     session,
+    storage: {
+      my_images: {
+        kind: 'local',
+        type: 'image',
+        generateUrl: path => `http://localhost:4000/images${path}`,
```
flexdinesh, line 24: We might have to set the URL using environment variables. Apps deployed on vercel will have their URL defined in the env variable `process.env.NEXT_PUBLIC_VERCEL_URL`.

flexdinesh (changes requested):

flexdinesh:
@marekryb re: generateUrl there is hardcoded port 4000

That's a tricky one. I think we could use environment variables to identify which server we're running (keystone or next). Perhaps we could replace the current `next:dev` script with `"next:dev": "PORT=4000  next dev"` and use the `PORT` env variable to conditionally setup the localhost URL.

But this probably adds an unwarranted layer of complexity to a simple example. Maybe we should just create a separate example to showcase file uploads with nextjs-keystone example and leave this example be as simple as it could be.

examples/nextjs-keystone/src/keystone/schema.ts
marekryb: Done

marekryb:
@flexdinesh I added passing public url through env variable (through cross-env as otherwise someone will come and complain that it doesn't work under Win), but vercel url is prioritized.

See how you like the changes. Alternatives are:
1) split this to separate example,
2) have two graphql apis - one plain and one with upload (upload one can be commented out),
3) leave 'processRequest' block as commented alternative and don't change schema at all.

mmachatschek:
@borisno2 this fixes the issue with file uploads and graphql-yoga for me.

If someone imports keystone in a different project (workspace setup), you need to make sure to export the `processRequest` from the keystone project and not adding `graphql-upload` to the dependencies of the other project (because of `instanceof` checks in graphql-upload).

PR #8182, opened 2022-12-18
Fixes GraphQL errors when using `@keystone-6/auth`

### LINKED ISSUE #8163: The authenticatedItem query always fails for postgresql users with autoincrement idFields
## Reproduction Steps

To reproduce, configure a starter Keystone.js project using `postgresql` and `autoincrement` id fields:

```javascript
export default withAuth(
  config({
    db: {
      provider: 'postgresql',
      url: process.env.DATABASE_URL || '',
      useMigrations: true,
      idField: { kind: 'autoincrement' },
```

Create a test user, then authenticate with it using `authenticateUserWithPassword`.

Then, try to call the following query:
```graphql
query authenticatedItem {
  authenticatedItem {
    ... on User {
      id
      username
      email
    }
  }
}
```
## Expected Result

```json
{
  "data": {
    "authenticatedItem": {
      "id": "1337",
      "username": "Example",
      "email": "example@gmail.com"
    }
  }
}
```
## Actual Result

```json
{
    "data": {
        "authenticatedItem": null
    }
}
```

## Probable Cause

The body of the resolver for `authenticatedItem` only executes if `itemId` is a string -- but in the case of autoincrement ids, it is a number.

Code location: https://github.com/keystonejs/keystone/blob/94e12b40872cd5814fe6d980b68070906f0a9307/packages/auth/src/gql/getBaseAuthSchema.ts#L57

Convenience screenshot:

image

Console log output of the value of `session` in the code block above, demonstrating that `itemId` is indeed not a string:

image

If I manually hack `node_modules\@keystone-6\auth\dist\keystone-6-auth.cjs.dev.js` and remove the `typeof` string check, the authenticatedItem query works as expected.

## Possible Fix

Allow for typeof `string` or `number` instead of just `string`.

## Closing Thoughts

Thank you for Keystone.js! I'm mid-migrating from an early version to v6 and ran into this item.

If the Possible Fix above looks reasonable to seasoned contributors who may have more context, I'd be willing to take a stab at it myself -- but it would take me some time to learn how to properly contribute to this repo, even for just a one-line change.

### LINKED ISSUE #8178: Bad generated schema resulting in "Cannot return null for non-nullable field"
# Steps to reproduce
Clone and set up the [nextjs-keystone example](https://github.com/keystonejs/keystone/tree/b81fb4522a733302c0cdaf7c8b7b2fd11c3cef10/examples/nextjs-keystone)

Create a jest test suite with a utility to log in via Graph QL:

```
async function logIn(email: string, password: string): Promise<Context> {
  const { data, errors }: ExecutionResult<{
    authenticateUserWithPassword: {
      __typename: string;
      sessionToken: string | null | undefined;
      message?: string | null | undefined;
    };
  }> = await context.graphql.raw({
    query: gql`
      mutation logIn($email: String!, $password: String!) {
        authenticateUserWithPassword(email: $email, password: $password) {
          __typename
          ...on UserAuthenticationWithPasswordSuccess {
            sessionToken
          }
          ...on UserAuthenticationWithPasswordFailure {
            message
          }
        }
      }
    `,
    variables: {
      email,
      password
    }
  });

  if (errors != null) {
    throw new Error(`Unable to log in: ${errors.join(', ')}`);
  } else if (
    (data?.authenticateUserWithPassword == null) ||
    (data.authenticateUserWithPassword.__typename === 'UserAuthenticationWithPasswordFailure')
  ) {
    throw new Error(`Unable to log in: ${data?.authenticateUserWithPassword.message ?? 'Unknown Error'}`);
  }

  return context.withSession(data.authenticateUserWithPassword.sessionToken);
}
```

Call this method in a test, e.g.

```
test('list users > authenticated > returns all users with email', async () => {
    const sudoContext = context.sudo();
    const users = generate.array.ofLengthBetween(generate.user, 1, 15); // generates users with { name, email, password }
    await seedUsers(sudoContext, users); // analogous to the demo seed in the example

    // error thrown on next line
    const loggedInContext: Context = await logIn(users[0].email, users[0].password);
    const { data } = await userQuery(loggedInContext); // analogous to client-side user query

    expect(data).not.toBeNull();
    if (data == null) return;
    expect(data.users.length).toBe(users.length);
    for (const user of users) {
      expect(data.users.some(u => (u.name === user.name) && (u.email === user.email))).toBe(true);
    }
  });
```

# Result
Keystone generates a `schema.graphql` with this definition
```
type UserAuthenticationWithPasswordSuccess {
  sessionToken: String!
  item: User!
}
```

This error is thrown
```
Unable to log in: Cannot return null for non-nullable field UserAuthenticationWithPasswordSuccess.sessionToken.

    GraphQL request:5:7
    4 |     ... on UserAuthenticationWithPasswordSuccess {
    5 |       sessionToken
      |       ^
    6 |     }

      151 |
      152 |   if (errors != null) {
    > 153 |     throw new Error(`Unable to log in: ${errors.join(', ')}`);
          |           ^
      154 |   } else if (
      155 |     (data?.authenticateUserWithPassword == null) ||
      156 |     (data.authenticateUserWithPassword.__typename === 'UserAuthenticationWithPasswordFailure')
```

# Environment
Node: 16.14.2
jest: 29.3.1
yarn: 3.3.0
@keystone-6/core: 4.0.1
@keystone-6/auth: 6.0.0
OS: Ubuntu 20.04.5 LTS (GNU/Linux 5.10.16.3-microsoft-standard-WSL2 x86_64)

borisno2:
Fixes #8163 - Allows for auth list id to be `autoincrement`
Fixes #8178 - Returns an authentication failure when `sessionStrategy.start` does not return a valid string.

### DIFF
--- a/.changeset/early-suns-knock.md
+++ b/.changeset/early-suns-knock.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/auth': patch
+---
+
+Fixes GraphQL error when using `autoincrement` for auth list id
--- a/.changeset/good-monkeys-complain.md
+++ b/.changeset/good-monkeys-complain.md
@@ -1,0 +1,6 @@
+---
+'@keystone-6/auth': patch
+---
+
+Fixes GraphQL error when `sessionStrategy.start` returns null or undefined  
+
--- a/.changeset/good-monkeys-complain.md
+++ b/.changeset/good-monkeys-complain.md
@@ -2,5 +2,4 @@
 '@keystone-6/auth': patch
 ---
 
-Fixes GraphQL error when `sessionStrategy.start` returns null or undefined  
-
+Fixes GraphQL error when `sessionStrategy.start` returns null or undefined
--- a/packages/auth/src/gql/getBaseAuthSchema.ts
+++ b/packages/auth/src/gql/getBaseAuthSchema.ts
@@ -54,7 +54,10 @@ export function getBaseAuthSchema<I extends string, S extends string>({
           resolveType: (root, context) => context.session?.listKey,
         }),
         resolve(root, args, { session, db }) {
-          if (typeof session?.itemId === 'string' && typeof session.listKey === 'string') {
+          if (
+            (typeof session?.itemId === 'string' || typeof session.itemId === 'number') &&
+            typeof session.listKey === 'string'
+          ) {
             return db[session.listKey].findOne({ where: { id: session.itemId } });
           }
           return null;
--- a/packages/auth/src/gql/getBaseAuthSchema.ts
+++ b/packages/auth/src/gql/getBaseAuthSchema.ts
@@ -91,13 +91,17 @@ export function getBaseAuthSchema<I extends string, S extends string>({
           }
 
           // Update system state
-          const sessionToken = (await context.sessionStrategy.start({
+          const sessionToken = await context.sessionStrategy.start({
             data: {
               listKey,
               itemId: result.item.id,
             },
             context,
-          })) as string;
+          });
+          // return Failure if sessionStrategy.start() returns null
+          if (typeof sessionToken !== 'string' || sessionToken.length === 0) {
+            return { code: 'FAILURE', message: 'Failed to start session.' };
+          }
           return { sessionToken, item: result.item };
         },
       }),


austinmilt (approved):
💯

flexdinesh (approved):

PR #8221, opened 2023-01-04
Upgrade Apollo Server to version 4

borisno2:
Upgrades Apollo Server to `@apollo/server` using `4.3.0` 
Removes the experimental 'generateNextGraphqlAPI` as this is now replaced with the keystone in next example this also removes the requirement for extra next/micro packages.


packages/core/src/lib/core/queries/resolvers.ts
```diff
@@ -147,8 +148,8 @@ export async function findMany(
     })
   );
 
-  if (info.cacheControl && list.cacheHint) {
-    info.cacheControl.setCacheHint(
+  if (list.cacheHint) {
+    cacheControlFromInfo(info).setCacheHint(
```
emmatown, line 152: Could you use `maybeCacheControlFromInfo` instead of `cacheControlFromInfo` everywhere in Keystone internals since the schema will often be used outside of Apollo Server

packages/core/src/lib/server/createApolloServer.ts
```diff
@@ -1,50 +1,18 @@
...
-  const serverConfig = _createApolloServerConfig({ graphQLSchema, graphqlConfig });
-  return new ApolloServerMicro({ ...serverConfig, context: userContext });
-};
+import { GraphQLFormattedError, GraphQLSchema } from 'graphql';
+import { ApolloServer, ApolloServerOptions } from '@apollo/server';
+import { ApolloServerPluginLandingPageDisabled } from '@apollo/server/plugin/disabled';
+import { ApolloServerPluginLandingPageLocalDefault } from '@apollo/server/plugin/landingPage/default';
+import type { KeystoneContext, GraphQLConfig } from '../../types';
 
 export const createApolloServerExpress = ({
```
dcousens, line 7: Naming is off now

tests/api-tests/fields/types/document.test.ts
```diff
@@ -287,7 +287,7 @@ describe('Document field type', () => {
       });
       // FIXME: The path doesn't match the null field here!
       expect(body.data).toEqual({ author: { badBio: null } });
-      expectInternalServerError(body.errors, true, [
+      expectInternalServerError(body.errors, false, [
```
dcousens, line 290: What happened here?

> The error is now a standard GraphQL error type, not internally typed

dcousens (approved):
LGTM, maybe we can inline the apollo server creation

dcousens:
As mentioned in https://github.com/keystonejs/keystone/pull/8258, if you use `graphql-upload` on your frontend, and you use COR's, you will need to resolve a CSRF vulnerability that Apollo 4 (this pull request) now prevents by default :yellow_heart:

PR #8247, opened 2023-01-15
Set `isFilterable` to `false` in examples where read access is false

borisno2:
Both the basic and custom-sesson-validation examples have fields with field-level access control that don't allow reads, this PR also sets those fields to `isFilterable: false`

### DIFF
--- a/examples/basic/schema.graphql
+++ b/examples/basic/schema.graphql
@@ -56,7 +56,6 @@ input UserWhereInput {
   name: StringFilter
   email: StringFilter
   password: PasswordFilter
-  isAdmin: BooleanFilter
   roles: StringFilter
   phoneNumbers: PhoneNumberManyRelationFilter
   posts: PostManyRelationFilter
@@ -103,11 +102,6 @@ input NestedStringFilter {
 
 input PasswordFilter {
   isSet: Boolean!
-}
-
-input BooleanFilter {
-  equals: Boolean
-  not: BooleanFilter
 }
 
 input PhoneNumberManyRelationFilter {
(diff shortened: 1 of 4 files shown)


dcousens:
Thoughts about
- https://github.com/keystonejs/keystone/blob/04025f747cf4a65a6b80e11594f880e393396358/examples/nextjs-keystone/src/keystone/schema.ts#L30 
- https://github.com/keystonejs/keystone/blob/04025f747cf4a65a6b80e11594f880e393396358/docs/pages/docs/guides/auth-and-access-control.md#create-a-list-for-users

borisno2:
> Any thoughts about
> 
> * https://github.com/keystonejs/keystone/blob/04025f747cf4a65a6b80e11594f880e393396358/examples/nextjs-keystone/src/keystone/schema.ts#L30
> * https://github.com/keystonejs/keystone/blob/04025f747cf4a65a6b80e11594f880e393396358/docs/pages/docs/guides/auth-and-access-control.md#create-a-list-for-users

The first one requires filtering for auth to work - this is where `isFilterable: false` would be too blunt - this would generally fall to list filtering however in this example it is `allowAll` on read, so not the most security-centered example. If I was writing something like this where some parts of `user` needed to be public, I would separate those out into a separate list, and if required, use virtual fields to not replicate data.

As for the guide maybe this needs some rework to align with some "best practices", ie there is no access control at all on the `User` list. I would also generally recommend separate lists for User Authentication/Authorization and Application concerns with the Auth list completely locked down at the list level.

dcousens (approved):

Issue #7636, opened 2022-06-16
Problem with setting up a project without TypeScript


Achisingh:
### Discussed in https://github.com/keystonejs/keystone/discussions/7535

<div type='discussions-op-text'>

<sup>Originally posted by **zeepsopje** May 14, 2022</sup>
I've been trying to setup Keystone 6 for a project, but it's unclear to me how to setup it up _without_ typescript. Although the docs point out the steps to do so, I'm getting the following message after exactly following the steps mentioned in [this](https://keystonejs.com/docs/walkthroughs/lesson-1) guide from the docs:
```
It looks like you're trying to use TypeScript but do not have the required package(s) installed.

Please install typescript by running:

        npm install --save-dev typescript

If you are not trying to use TypeScript, please remove the tsconfig.json file from your package root (and any TypeScript files in your pages directory).

error Command failed with exit code 1.
```
This happens after running `yarn keystone dev`. I can confirm there is no usage of TS inside of my project.

Any one who's managed to get this up and running without typescript? The `create-keystone-app` utility hasn't given me the choice to not use TS so I was hoping a manual approach would do the trick.

**Context:**
Node: 16.15.0
Yarn: 1.22.18
OS: Arch Linux</div>

fkrauthan:
Yeah I think I had to add typescript to my dev dependencies. For some reason the `keystone dev` command requires it. But you can remove the typescript config file etc.

mattkrins:
FYI workaround for this currently is to just have typescript in the dependency list but not actually use it.
(will generate IDE intellisense errors).

dcousens:
Cannot reproduce, `keystone.js` works as intended now.

dcousens:
Reproduced this problem, `keystone build` and `keystone dev`, using a plain `keystone.js` entry, still results in `yarn` adding `typescript` to your `devDependencies`.

```
$ keystone build
✨ Building Keystone
✨ Generating Admin UI code
✨ Building Admin UI
info  - Skipping validation of types
info  - Skipping linting
It looks like you're trying to use TypeScript but do not have the required package(s) installed.
Installing dependencies

If you are not trying to use TypeScript, please remove the tsconfig.json file from your package root (and any TypeScript files in your pages directory).

Installing devDependencies (yarn):
- typescript
```

PR #8255, opened 2023-01-17
Add missing access control for walkthrough documentation

dcousens:
This is required in recent versions of `keystone-6`, this amends the guide to reflect that.

Found while trying to reproduce https://github.com/keystonejs/keystone/issues/7636

### DIFF
--- a/docs/pages/docs/walkthroughs/lesson-1.md
+++ b/docs/pages/docs/walkthroughs/lesson-1.md
@@ -87,7 +87,7 @@ export default config({
   db: {
     provider: 'sqlite',
     url: 'file:./keystone.db',
-  },
+  }
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/pages/docs/walkthroughs/lesson-1.md
```diff
@@ -87,7 +87,7 @@ export default config({
   db: {
     provider: 'sqlite',
     url: 'file:./keystone.db',
-  },
+  }
```
dcousens, line 90:
```suggestion
  },
  lists: {} // ...
```

If the developer uses `keystone dev` without this, the error is:

```
TypeError: Cannot convert undefined or null to objec
```

Adding this changes the error into
```
Error:
You don't have any models defined in your schema.prisma, so nothing will be generated.
```

Which, in my opinion, is quite helpful when you are running keystone without Typescript

borisno2 (approved):

PR #8260, opened 2023-01-19
Add digital ocean spaces link for S3 providers (and tidy up)

Meetcpatel:
This PR includes small information to use S3 comaptible storage services from Digital Ocean spaces

### DIFF
--- a/docs/pages/docs/guides/images-and-files.md
+++ b/docs/pages/docs/guides/images-and-files.md
@@ -21,7 +21,7 @@ The `storage` object defines how and where the assets are stored and accessed by
 
 First, we are going to use [dotenv](https://www.npmjs.com/package/dotenv) to retrieve the S3 bucket and URL details from a `.env` file or set environment variables.
 
-In your Keystone Config file, pull in your environment variables and maps them to some easy to use names
+Before your configuration file, you can map your environment variables to some easy to use names:
 
 ```typescript{24-54}
 import { config } from '@keystone-6/core';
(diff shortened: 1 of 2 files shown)


jhohlfeld:
Hi @Meetcpatel,

may I ask how you managed to set your permissions from 'Private' to 'Public' on DigitalOcean? I was trying to set the ACL of newly uploaded assets automatically and found no way to do it out of the box, so I created #8264.

Is there any other way I am not aware of?

cheers
Jakob

docs/pages/docs/guides/images-and-files.md
```diff
@@ -3,11 +3,11 @@ title: "Images and Files"
...
 
 ## How asset storage works in Keystone
 
-Keystone manages file and image assets through a `storage` object you define in Keystone’s [configuration file](/docs/config/config). Any number of stores can be set up within the `storage` object, and you can mix and match between `local` and `s3` (by [Amazon](https://aws.amazon.com/s3/)) depending on your use case.
+Keystone manages file and image assets through a `storage` object you define in Keystone’s [configuration file](/docs/config/config). Any number of stores can be set up within the `storage` object, and you can mix and match between `local` and `s3` (by [Amazon](https://aws.amazon.com/s3/)) and `S3-Compatible Cloud Object Storage` (by [DigitalOcean Spaces](https://www.digitalocean.com/products/spaces)) depending on your use case.
```
dcousens, line 10:
```suggestion
Keystone manages file and image assets through a `storage` object you define in Keystone’s [configuration file](/docs/config/config). Any number of stores can be set up within the `storage` object, and you can mix and match between `local` and `s3` compatible providers (akin to [Amazon S3](https://aws.amazon.com/s3/, or [DigitalOcean Spaces](https://www.digitalocean.com/products/spaces)).
```

We have no preference for a provider, only that it is compatible.  Maybe this is an improvement?

docs/pages/docs/guides/images-and-files.md
```diff
@@ -66,6 +94,8 @@ storage: {
     accessKeyId,
     // The S3 Secret pulled from the S3_SECRET_ACCESS_KEY environment variable
     secretAccessKey,
+    // the S3 endpoint pulled from S3_ENDPOINT environment variable
```
dcousens, line 97:
```suggestion
    // the S3 endpoint pulled from S3_ENDPOINT environment variable (if defined)
```

docs/pages/docs/guides/images-and-files.md
dcousens: I'm weary of adding many S3 providers to this section, but don't want to bias either.  Maybe we can use generic language after _one_ section that lists a number of _different_ providers.

Meetcpatel:
> Hi @Meetcpatel,
> 
> may I ask how you managed to set your permissions from 'Private' to 'Public' on DigitalOcean? I was trying to set the ACL of newly uploaded assets automatically and found no way to do it out of the box, so I created #8264.
> 
> Is there any other way I am not aware of?
> 
> cheers Jakob

Sure, 
Yes, it stores as private asset on DigitalOcean, but I think for our use case we don't  need to make it public  because the asset link that we have after keystone uploads the asset is fine to access the same asset.

But, if you have other use case and it requires to make all asset public, I think that your idea would be helpful to make it public.

asset link from Digital Ocean: 
`https://bucketname.region.digitaloceanspaces.com/1f7fcab9-39f3-47c1-bbdc-0b5938bbc495.jpg` (we can't access, it gives access denied)

asset link keystone received after upload: 
`https://bucketname.region.digitaloceanspaces.com/1f7fcab9-39f3-47c1-bbdc-0b5938bbc495.jpg?X-Amz-Algorithm....` (we can access)

docs/pages/docs/guides/images-and-files.md
Meetcpatel: Yes, Thank You!

dcousens (approved):

PR #8264, opened 2023-01-19
Add acl option for s3 storage

jhohlfeld:
This PR adds ACL settings to the s3 storage config.

## Why

This change was necessary because on DigitalOcean I wasn't able to see my uploaded images publicly. Instead, I had to set them to public manually. 

DO does not let you set default permissions:
https://docs.digitalocean.com/products/spaces/how-to/set-file-permissions/

> By default, file permissions are set to Private. Some third-party tools provide support for changing the default permissions, but there is currently no way to do so in the control panel.

## Usage

```typescript
config({
  storage: {
    remote: {
      kind: 's3',
      type: 'image',
      bucketName,
      region,
      accessKeyId,
      secretAccessKey,
      endpoint,

      /** Optionally add one of the ACL options from AWS-SDK.
       * 
       * Example: 'private', 'public-read' or 'public-read-write'.
       * 
       * See https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl
       * for more details.
       */
      acl: 'public-read'
    },    
  }
});
```

### DIFF
--- a/docs/pages/docs/config/config.md
+++ b/docs/pages/docs/config/config.md
@@ -483,6 +483,7 @@ S3 options:
 - `endpoint`: The endpoint to use - if provided, this endpoint will be used instead of the default amazon s3 endpoint
 - `forcePathStyle`: Force the old pathstyle of using the bucket name after the host
 - `signed.expiry`: Use S3 URL signing to keep S3 assets private. `expiry` is in seconds
+{% if $nextRelease %}
 - `acl`: Set the permissions for the uploaded asset. If not set, the permissions of the asset will depend on your S3 provider's default settings.
 These values are supported:
   - `'private'` No public assess.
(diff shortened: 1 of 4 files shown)


dcousens:
Nice pull request @jhohlfeld!
I'm trying to determine if any part of this should be captured as part of the field type or individually for uploads; but maybe that's a different use-case that we can leave out and move forward with this as-is.

borisno2 (changes requested):
Looks good, thanks @jhohlfeld 
Just a couple of website fixes needed so that the docs don't update until this change is ~deployed~ released.
Thanks,

docs/pages/docs/config/config.md
```diff
@@ -483,6 +483,19 @@ S3 options:
 - `endpoint`: The endpoint to use - if provided, this endpoint will be used instead of the default amazon s3 endpoint
 - `forcePathStyle`: Force the old pathstyle of using the bucket name after the host
 - `signed.expiry`: Use S3 URL signing to keep S3 assets private. `expiry` is in seconds
+{% if $nextRelease %}
+- `acl`: Set the permissions for the uploaded asset. If not set, the permissions of the asset will depend on your S3 provider's default settings.
+These values are supported:
+  - `'private'` No public assess.
```
dcousens, line 489:
```suggestion
  - `'private'` No public access.
```

borisno2 (approved):

PR #8274, opened 2023-01-24
Fix data proxy unsupported before exit event

mmachatschek:
This is based on the branch of PR #7964 as I tested all prisma versions from 4.3.1 (current keystone supported version) to 4.9.0. Earlier versions had problems with parsing the `prisma://` database urls

- https://github.com/prisma/prisma/blob/5e4322b0441cb16cea0bcfa3877550e4a1fa8ec7/packages/engine-core/src/data-proxy/DataProxyEngine.ts#L91


packages/core/src/lib/createSystem.ts
```diff
@@ -82,13 +82,17 @@ export function createSystem(config: KeystoneConfig) {
...
-      prismaClient.$on('beforeExit', async () => {
-        // Prisma is failing to properly clean up its child processes
-        // https://github.com/keystonejs/keystone/issues/5477
-        // We explicitly send a SIGINT signal to the prisma child process on exit
-        // to ensure that the process is cleaned up appropriately.
-        prismaClient._engine.child?.kill('SIGINT');
-      });
+
+      // beforeExit event is not supported by Prisma Data Proxy
+      if (!prismaClient._dataProxy) {
```
mmachatschek, line 87: I'm accessing the internal variable `prismaClient._dataProxy` here because the only alternative to check for dataProxy usage would be:

```js
prismaClient.getEngine().constructor.name === 'DataProxyEngine'
```

packages/core/src/lib/createSystem.ts
```diff
@@ -82,13 +82,17 @@ export function createSystem(config: KeystoneConfig) {
...
+        prismaClient.$on('beforeExit', async () => {
+          // Prisma is failing to properly clean up its child processes
+          // https://github.com/keystonejs/keystone/issues/5477
+          // We explicitly send a SIGINT signal to the prisma child process on exit
+          // to ensure that the process is cleaned up appropriately.
+          prismaClient._engine.child?.kill('SIGINT');
+        });
+      }
```
dcousens, line 95: Do we actually need this still?

borisno2: I am unable to replicate the original issue (#5477) if I remove this whole `prismaClient.$on` block

dcousens: Let's drop it

borisno2:
Moved changes over to #8347 - Closing the PR... thanks @mmachatschek

PR #7004, opened 2021-11-26
Add capability to always send field in update* mutations

rburgst:
- this serves as the basis for optimistic locking
- you can now configure a `version` field as readonly and have it
  included in all update calls
- on the server side you can then validate the version with the
  `item.version` and then throw a validation error in case it does not
  match
- that way a concurrent modification on the server can be identified

fixes #6851

PR #8283, opened 2023-01-31
Add graphql.isNonNull.update

dcousens:
(replaces https://github.com/keystonejs/keystone/pull/7004)

TODO
- [x] Review `defaultValue` changes for field types


examples/versioning/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This example includes sample data. To add it to your database:
+
+1. Ensure you’ve initialised your project with `yarn dev` at least once.
+2. Run `yarn seed-data`. This will populate your database with sample content.
+3. Run `yarn dev` again to startup Admin UI with sample data in place.
+
+## Try it out in CodeSandbox 🧪
+
+You can play with this example online in a web browser using the free [codesandbox.io](https://codesandbox.io/) service. To launch this example, open the URL <https://githubbox.com/keystonejs/keystone/tree/main/examples/blog>. You can also fork this sandbox to make your own changes.
```
rburgst, line 32: code sandbox link appears to be wrong

packages/core/src/types/config/fields.ts
```diff
@@ -32,17 +32,25 @@ export type CommonFieldConfig<ListTypeInfo extends BaseListTypeInfo> = {
...
-    //   'read': Does this field exist on the Item type? Will also disable filtering/ordering/admimMeta
-    // Input Types
-    //   'create': Does this field exist in the create Input type?
-    //   'update': Does this field exist in the update Input type?
-    //
-    // If `true` then the field will be completely removed from all types.
-    //
-    // Default: undefined
-    omit?: true | readonly ('read' | 'create' | 'update')[];
+    isNonNull?: {
```
rburgst, line 35: IMHO `isNonNull` does not fully explain (at least for me) the purpose and the fact that this field has to be included in every update. But then again, https://www.quora.com/Would-someone-be-kind-enough-to-explain-this-joke-in-more-detail-%E2%80%9CThere-are-two-hard-problems-in-computer-science-naming-things-cache-invalidation-and-off-by-1-errors%E2%80%9D

dcousens: I understand, this does alter your proposed `alwaysSend` feature to something that is an emergent behaviour required b to interoperate with your GraphQL  schema without errors.

I'm personally OK with that, as we can resolve the rest with documentation changes?

rburgst: I am not complaining at all, I know that your implementation is more encompassing than just my `alwaysSend` feature. I just mean that to me as a keystone user the name `graphql.isNonNull` is not as intuitive as I would have hoped. But as stated above, naming is really hard. I am sure that documentation could remedy that but it would be even better to have the code be more self-explanatory. But again, I am not sure I could come up with a better alternative name. Intuitively I would have opted for `graphql.required` but I am sure there are perspectives where this would also be not ideal.

dcousens: I actually opted for `graphql.isRequired` at first, and I'm not against it. 
`isNonNull` was the precedent, but seeing as I am breaking that precedent by re-ordering it from `*.isNonNull` to `isNonNull: { * }` anyway, I'm open to suggestions.

rburgst: I like `graphql.isRequired`

dcousens: With the pull request as-is, we have `isNonNull`, `isNullable`, and `isRequired`, in different contexts.
I'll continue pushing the pull request to :green_circle:, and see if I can organise a bikeshed.

packages/core/src/fields/types/text/index.ts
```diff
@@ -161,10 +158,8 @@ export const text =
         },
         create: {
           arg: graphql.arg({
-            type: config.graphql?.create?.isNonNull
-              ? graphql.nonNull(graphql.String)
-              : graphql.String,
-            defaultValue: config.graphql?.create?.isNonNull ? defaultValue : undefined,
+            type: graphql.String,
+            defaultValue: typeof defaultValue === 'number' ? defaultValue : undefined,
```
borisno2, line 162:
```suggestion
            defaultValue: typeof defaultValue === 'string' ? defaultValue : undefined,
```
`string` maybe??

packages/core/src/types/config/fields.ts
dcousens: @rburgst we aren't happy with the situation, but, in context each version makes sense.

- `isNonNull` is used within the negative context of "adding a !" to the GraphQL schema, which **stops** something from being `null`
- `isNullable` is used within the affirmative context of "adding a ?" to the Prisma schema, which **allows** something to be `null`
- `isRequired`, defaults `isNullable` to `false`, throws an error for `null` via GraphQL, and prevents empty `text` strings

If anything, `isRequired` is probably the outlier in it's weird behaviour, in that, really, it's only a shorthand for `isNullable` and enforcing useful behaviour for `text`.

`isRequired` didn't really make sense for `isNonNull.read` either, because that is **not** a GraphQL input type, and it's weird to say that the server "isRequired" to make a field not null, compared to the `create` and `update` which are GraphQL input types.

We might look at unifying these concepts in the future, but for now, `graphql.isNonNull` as an idea that is solid on it's own.

borisno2 (approved):

PR #8300, opened 2023-02-13
Fix undefined types when omitted from GraphQL

dcousens:
As is, if you use `graphql.omit`, some printed Typescript types will not be defined, but they are still printed.
This pull request fixes that by not printing the types if they are not defined.


dcousens:
This pull request has additionally revealed a soft spot, that we don't have tests for our printed Typescript types.
We should fix that.

borisno2 (approved):

PR #8301, opened 2023-02-13
Reduce code complexity

dcousens:
This pull request refactors `getDbAPIFactory` to not use complicated type resolution and casting, but instead rely on the return type.

### DIFF
--- a/scripts/changelog.js
+++ b/scripts/changelog.js
@@ -85,7 +85,7 @@ async function fetchData(tag) {
   const changes = {};
   for (const commit of revs) {
     let { user, pull } = await getInfo({ repo: 'keystonejs/keystone', commit });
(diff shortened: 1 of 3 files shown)


scripts/changelog.js
```diff
@@ -85,7 +85,7 @@ async function fetchData(tag) {
   const changes = {};
   for (const commit of revs) {
     let { user, pull } = await getInfo({ repo: 'keystonejs/keystone', commit });
-    pull = pull || gitCommitDescription(commit).match(/#([0-9]+)/)[1];
+    pull = pull || gitCommitDescription(commit).match(/#([0-9]+)/)?.[1]
```
dcousens, line 88: Unrelated, but this is a minor fix for the upcoming release

borisno2 (approved):

PR #8302, opened 2023-02-13
Remove `keystone dev --reset-db`

dcousens:
This pull request removes `--reset-db`.

This command was dangerous at the best of times and frustrating almost every other time you accidentally used it from your shell history.

Resetting your database in this way is not a typical way to work with a local development database.
We recommend using `keystone prisma db push --force-reset` for the limited number of occasions that you need this.

This pull request is part of the work from https://github.com/keystonejs/keystone/pull/8046 which cements the idea that `keystone dev` is about a workflow, and not about kicking off specific actions.

(review and merge https://github.com/keystonejs/keystone/pull/8046 before this pull request is rebased)

### DIFF
--- a/packages/core/src/scripts/cli.ts
+++ b/packages/core/src/scripts/cli.ts
@@ -94,7 +94,7 @@ export async function cli(cwd: string, argv: string[]) {
   }
 
   if (command === 'prisma') {
-    return prisma(cwd, argv.slice(1), defaultFlags(flags, { frozen: false }).frozen);
+    return prisma(cwd, argv.slice(1), !!flags.frozen);
   }
 
   if (command === 'telemetry') return telemetry(cwd, argv[1]);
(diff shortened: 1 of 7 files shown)


packages/core/src/scripts/cli.ts
```diff
@@ -1,55 +1,114 @@
...
-    return prisma(cwd, argv.slice(1));
-  } else if (command === 'postinstall') {
-    return postinstall(cwd, flags.fix);
-  } else if (command === 'dev') {
-    return dev(cwd, flags.resetDb);
-  } else if (command === 'telemetry') {
-    return telemetry(cwd, argv[1]);
-  } else {
-    return commands[command](cwd);
+    return prisma(cwd, argv.slice(1), defaultFlags(flags, { frozen: false }).frozen);
```
borisno2, line 97: The default flags function doesn't work fore when you want to pass through a Prisma flag - for example `--force-reset`

borisno2 (approved):

PR #8304, opened 2023-02-14
Update `file-type` to `18.2.0`

dcousens:
This pull request mitigates [CVE-2022-36313](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-36313) by upgrading and using `file-type` directly.

The `image-type` package wraps `file-type` and then limits the returned result to a subset of filename extensions.
In our usage, we then limited the return of `image-type` to a smaller subset of filename extensions.
Knowing this, we have omitted the `image-type` package and used `file-type` directly.

Unfortunately we don't have static ESM imports at this time, as a result we are kind of hacking around this for now.
Rather than not mitigate the security vulnerability, I have opted for this as a temporary solution to move forward.

The tests require `NODE_OPTIONS=--experimental-vm-modules`, but on GitHub CI, `jest` is failing with a segmentation fault for that configuration. To pass CI, this pull request disables the test for now, so we can move forward with resolving the security vulnerability. You can run the tests locally using `NODE_OPTIONS=--experimental-vm-modules` with success.


borisno2 (approved):

tsconfig.json
```diff
@@ -13,7 +13,7 @@
     "noUnusedLocals": true,
     "types": ["@types/jest"],
     "typeRoots": [],
-    "skipLibCheck": false,
```
dcousens, line 16: Unfortunate, but hopefully only temporary

PR #8307, opened 2023-02-15
Add `db.prismaPath` configuration

borisno2:
This pull request removes our assumptions about the `@prisma/client`'s output location.
Previously Keystone was assuming that `@prisma/client` could always be found in the project's relative `node_modules/.prisma`.

This isn't a safe assumption, if for example you are hoisting to a monorepo `node_modules`.
In actuality, we were making this assumption so our own monorepo can work with multiple Keystone configurations, as seen in our `examples/` and `tests/`.

By adding a new `db.prismaPath` configuration option to configure this, we can default back to the normal behaviour of letting Prisma resolve `@prisma/client` as it would, and then otherwise let users define where a suitable location might be.


dcousens (approved):

dcousens:
Renamed to `db.prismaClientPath` in https://github.com/keystonejs/keystone/pull/8323

jlarmstrongiv:
@borisno2 

I’m glad this PR was merged :tada: which helps a lot with monorepo environments.  Being able to move the `.prisma` folder is great, but what about the `.keystone` folder with types, etc.?

When I use keystone, I was pretty much required to use npm because that particular package manager didn’t hoist packages to the root or global folder.

jlarmstrongiv:
To clarify, there are two `.keystone` folders.  One is in the project root.  The other is in the `node_modules` folder.  I’m referring to the `node_modules` folder one.

jlarmstrongiv:
The relevant references I found are:
- the actual artifact generation https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/packages/core/src/artifacts.ts#L85
- the artifact generation tests
  - https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/packages/core/src/scripts/tests/artifacts.test.ts#L87
  - https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/packages/core/src/scripts/tests/__snapshots__/artifacts.test.ts.snap#L4
- code comments https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/packages/core/src/lib/core/utils.ts#L70
- extend-graphql-schema-nexus example https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/examples/extend-graphql-schema-nexus/nexus/index.ts#L17
- experimental generateNextGraphqlAPI https://github.com/keystonejs/keystone/blob/72529e8c1d0d4bd9c72e05d5b65f15ff62cf390c/docs/pages/docs/config/config.md?plain=1#L556

jlarmstrongiv:
Being able to move the `node_modules/.keystone` folder is the last piece (I know of) for broad monorepo support

dcousens:
@jlarmstrongiv you might have noticed, that function is _primed_ for that feature.

jlarmstrongiv:
@dcousens great!  Two quick followup questions:
- Would that be a separate config option and what would it be called?
- Does this need a new issue or is a followup PR fine?
- Will we need to use a similar `...fixKeystonePath` everywhere or how would you prefer that to look?

Issue #34, opened 2018-05-06
More pagination options


JedWatson:
We currently support two pagination options on List queries:

* `first: Int` (limits the number of items returned)
* `skip: Int` (skips that many items from the start, based on the sort criteria)

In addition, if we're hoping to get close to API-compatible with [graph.cool](https://www.graph.cool/docs/reference/graphql-api/query-api-nia9nushae/#pagination) / [Prisma](https://www.prisma.io/docs/reference/prisma-api/queries-ahwee4zaey#pagination) we should add support for additional pagination options in the List GraphQL APIs:

* `last: Int` (reverse of `first`)
* `after: ID` (alternative to `skip`, starts after the provided ID based on sort criteria)
* `before: ID` (reverse of `after`)

I'm not sure if these are easy/possible to implement with MongoDB so it may be a Postgres-adapter specific thing.

PR #8313, opened 2023-02-15
Add support for cursor-based pagination

marekryb:
Currently Keystone supports only offset-based pagination through `take` and `skip` arguments in GraphQL queries.
However Prisma, which is used under the hood, [supports both offset and cursor pagination](https://www.prisma.io/docs/concepts/components/prisma-client/pagination). 
Above article summarizes pros and cons of each solution very well. My motivation comes from the fact that I am using infinite scrolling exclusively on the frontend.

Related issue: https://github.com/keystonejs/keystone/issues/34

This PR add cursor-based pagination that follow Prisma's syntax by:
- adding `cursor` argument to GraphQL queries that returns array (also through relation) 
- adding `cursor` to context.query / context.db API
- passing `cursor` value to Prisma

`cursor` argument is an object with same type as eg. `connect` in relation - all unique fields are allowed. Existing `take` and `skip` do not change and `cursor` usage is optional. Therefore this is (as fat as I can tell) non-breaking change.

Currently this PR includes some very basic unit tests. I can add more, but keep in mind that no special logic is introduced here, so extensive testing would actually test Prisma and not Keystone.

Please have a look and let me know whether you are interested in going forward with this PR. What did I miss? What other changes are needed? I would say documentation, but surprisingly I could not find anything in the docs on `take` and `skip` either ^_^


.changeset/sour-singers-glow.md
```diff
@@ -0,0 +1,43 @@
...
+'@keystone-6/example-roles': minor
+'@keystone-6/example-virtual-field': minor
+'@keystone-6/example-assets-local': minor
+'@keystone-6/example-custom-field': minor
+'@keystone-6/example-usecase-blog': minor
+'@keystone-6/example-assets-s3': minor
+'@keystone-6/singleton': minor
+'@keystone-6/example-testing': minor
+'@keystone-6/example-limits': minor
+'@keystone-6/example-script': minor
```
dcousens, line 36: Can you please remove all the examples from this

dcousens:
A few tests to satisfy, but the on the face of it, this pull request looks great @marekryb! 

> surprisingly I could not find anything in the docs on take and skip either ^_^

I'm happy to not block on this pull request for that, but we should do add documentation for `cursor`, `take` and `skip` in another PR

.changeset/sour-singers-glow.md
marekryb: Done

dcousens:
@marekryb when https://github.com/keystonejs/keystone/pull/8320 is merged, could you please rebase?

marekryb:
> @marekryb when #8320 is merged, could you please rebase?

Done. Also added few more tests.

marekryb:
@dcousens Seems everything is passing now. Looks good from my point of view.

borisno2 (approved):

tests/api-tests/queries/cursor-pagination.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            take: 6,\
+            skip: 1,\
+            cursor: { order: 5 }\
+            orderBy: { order: asc }\
+          ) { id order }\
+        }\
+      }`,
+    });
+    expect(errors).toEqual(undefined);
+    let currentOrder = 6;
```
dcousens, line 114: We should alter this to use an `autoincrement` identifier rather than something named `order`, but we won't block merging on that

PR #8318, opened 2023-02-16
Fix withRequest behavior without session strategy

### LINKED ISSUE #8317: Context withRequest edge case bug
Ran into this rather strange edge case behavior in my app, thankfully I was able to reproduce it in a clean slate app here: https://github.com/acburdine/keystone-withrequest-repro

TL;DR calling `await context.withRequest(req, res)` and then `.sudo()` on the returned context object results in a context object with the _original_ request and response properties (i.e. the ones set prior to running withRequest.)

### Repro Steps
This is what I did specifically, though I'm sure the behavior can be triggered in other ways.
First, remove _any_ session/auth configuration from Keystone (the code path containing this bug isn't triggered when a session strategy is configured).

Add the following code to your keystone config:

```js
server: {
  extendExpressApp(app, commonCtx) {
    app.get("/api/testing", async (req, res, next) => {
      try {
        const ctx = await commonCtx.withRequest(req, res);
        const sudoCtx = ctx.sudo();

        console.log(Boolean(ctx.req));
        console.log(Boolean(sudoCtx.req));
        console.log(ctx.req === sudoCtx.req);

        res.json({ success: true });
      } catch (error) {
        next(error);
      }
    });
  }
}
```
Expected: three `true` lines should be output when the `/api/testing` endpoint is invoked.

Actual Output: one `true` output, then two `false` outputs, indicating that sudoCtx doesn't have a request object set, and that the request object on the base `ctx` object doesn't match that on the `sudoCtx` object.

I should also note - I have a fix for this that I'll be submitting in a PR shortly.

acburdine:
closes #8317
- previously, if no session strategy was configured in the Keystone config, the withRequest function would return early without calling `createContext`. If `sudo()` was called on the returned context from withRequest, the context object returned from sudo would have the req/res objects present on the initial context object (i.e. the one prior to withRequest being called)
- this change makes withRequest consistent with the other context 'change' functions (sudo/exitSudo/withSession)

### DIFF
--- a/packages/core/src/lib/context/createContext.ts
+++ b/packages/core/src/lib/context/createContext.ts
@@ -75,14 +75,19 @@ export function makeCreateContext({
       return result.data as any;
     };
 
-    async function withRequest(req: IncomingMessage, res?: ServerResponse) {
-      contextToReturn.req = req;
-      contextToReturn.res = res;
-      if (!config.session) {
-        return contextToReturn;
+    async function withRequest(newReq: IncomingMessage, newRes?: ServerResponse) {
+      const newContext = createContext({
+        session,
+        sudo,
+        req: newReq,
+        res: newRes,
+      });
+
+      if (config.session) {
+        newContext.session = await config.session.get({ context: newContext });
       }
-      contextToReturn.session = await config.session.get({ context: contextToReturn });
-      return createContext({ session: contextToReturn.session, sudo, req, res });
+
+      return newContext;
     }
     const dbAPI: KeystoneContext['db'] = {};
     const itemAPI: KeystoneContext['query'] = {};
--- a/.changeset/thin-turtles-beg.md
+++ b/.changeset/thin-turtles-beg.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix withRequest behavior when no session strategy is configured


packages/core/src/lib/context/createContext.ts
```diff
@@ -75,14 +75,19 @@ export function makeCreateContext({
...
+    async function withRequest(newReq: IncomingMessage, newRes?: ServerResponse) {
+      const newContext = createContext({
+        session,
+        sudo,
+        req: newReq,
+        res: newRes,
+      });
+
+      if (config.session) {
+        newContext.session = await config.session.get({ context: newContext });
```
acburdine, line 87: a note here: I considered doing this instead:
```ts
        const newSession = await config.session.get({ context: newContext });
        return createContext({
          session: newSession,
          sudo,
          req: newReq,
          res: newRes,
        });
```

however, that would end up calling createContext twice. Also, the `session` property in the createContext initial arguments isn't used by any of the "modification" functions (sudo, exitSudo, withSession), so it wouldn't end up having the same issue that the req/res objects did initially.

acburdine:
also, I tried finding a good place to add tests for this edge case but didn't see any obvious spots to put it 😬 

If there's a "correct" place to write tests for this let me know and I'll gladly add them 😄

dcousens (approved):
LGTM, thanks @acburdine! Nice find

dcousens:
Failing tests is unrelated, this looks good to go :+1:

PR #8118, opened 2022-11-23
Add Telemetry to `keystone dev`

borisno2:
As the Keystone project grows, Thinkmill Labs would like to learn more about how Keystone is being used by developers and projects that may not be actively participating on our GitHub, Twitter and slack.
The questions we want to currently answer using telemetry are the following:

- How many active developers are using Keystone, compared to how many are active in the community
- How complex are projects that use Keystone, and
- What field types and `@keystone-6` or community packages are being used by developers, and
- What operating systems and node versions are being used by the community

Knowing this information will help Thinkmill Labs prioritise future features or maintenance efforts for ongoing development.

## What's in this pull request
- [x] This pull request includes the changes required to trial this telemetry with a new minor version of Keystone.
- [ ] This pull request additionally includes a privacy policy to be published as part of our documentation to formalize our commitment to how the telemetry data will be used, shared and retained. **(work in progress)**

For notifying users who may not be watching this pull request or repository, on the first run of `keystone dev`, a notification is logged in the console informing the developer that keystone will now collect anonymous data, including a link to instructions on how to opt-out.
No telemetry is sent before this, which should prevent any user from missing the opportunity to opt-out before telemetry is sent.

### How to get involved?
If you have any concerns, comments or feedback on that policy, *please* comment on this pull request!
Our privacy policy is intended to be a living document and this is ultimately an experiment that should **grow with the community.**

We don't want to personally identify developers, nor do we want to collect any personal information (no thanks!).
We are interested in answering the aforementioned questions above, and it's our thought that telemetry will be the easiest way to do that in a way that scales with the project.

### Sharing with the community
Part of this experiment is that we want to share the aggregated form of the telemetry reports with the community where reasonable.  We don't as of yet have a timeline or community plan for this, but we're open to feedback from the community about what that could look like.

As part of that, in respect to the project telemetry report, we are open to additional package namespaces being collected for the the `versions` data to allow us to provide community package maintainers members to learn from the same insights that the Keystone team will have about `@keystone-6` packages.

From an implementation perspective,  if you are writing custom field types that you want to have reported in this data, you can add the new `__ksTelemetryFieldTypeName` to your `fieldType` configuration.  Preferably use a format similar to `@keystone-6/text`, for example `@yourorganisation/yourfieldtypename`.

## Opting Out
This pull request includes two ways that you can explicitly opt out of telemetry.
With there being two ways to opt-out, any `false` has precedence and will opt out. 

[...]

PR #8326, opened 2023-02-20
Fix malformed telemetry configuration

dcousens:
This pull request fixes a users telemetry configuration which may have been malformed (but not actually broken).
We noticed recently that some of our project keys were malformed within the file, with keys of `projects${cwd}` when they should have been `projects.${cwd}`.

An unfortunate but inevitable error when using stringly typed keys.
To prevent any regression or weird behaviours, this pull request stops using stringly typed keys when updating the users configuration.  As part of these changes, we add a version number to the configuration for posterity.

In respect to data integrity, only an `informedAt` was kept within these malformed project keys; and seeing as it is a direct copy of `device.informedAt`, we can omit it.  This pull requests thereby removes that data completely from the local configuration.  

We additionally tidy up the `keystone telemetry status` command output

---

#### This pull request **does not** change what data is reported.
No update to our privacy policy is required.

For more about telemetry, see the primary pull request in https://github.com/keystonejs/keystone/pull/8118.


packages/core/src/lib/telemetry.ts
```diff
@@ -17,199 +17,318 @@ const packageNames: PackageName[] = [
...
         // skip id fields
         if (fieldPath.endsWith('id')) continue;
         fields.unknown++;
         continue;
       }
-      if (!fields[fieldType]) {
-        fields[fieldType] = 0;
-      }
+
+      fields[fieldType] ||= 0;
```
dcousens, line 96:
```suggestion
      fields[fieldType] ??= 0;
```

borisno2 (approved):

packages/core/tests/telemetry.test.ts
```diff
@@ -78,214 +102,217 @@ const lists: Record<string, InitialisedList> = {
...
-    expect(mockFetch).toHaveBeenCalledTimes(0);
-    expect(mockTelemetryConfig?.device).toBeUndefined();
-    expect(mockTelemetryConfig?.projects).toBeUndefined();
   });
 
-  test('Telemetry Does not send when env KEYSTONE_TELEMETRY_DISABLED is set to 1', async () => {
-    defaultFetchMock();
-    // @ts-ignore
-    process.env.KEYSTONE_TELEMETRY_DISABLED = '1';
+  test(`Telemetry is not sent telemetry if disabled`, async () => {
```
borisno2, line 212:
```suggestion
  test(`Telemetry is not sent if telemetry is disabled`, async () => {
```

PR #8344, opened 2023-02-28
Remove `process.cwd` change for Prisma in tests

borisno2:
In the Admin UI and Smoke Test utils we where changing `process.cwd` to the `projectDir` this should no longer be required


tests/admin-ui-tests/utils.ts
```diff
@@ -99,7 +88,7 @@ export const adminUITests = (
     let cleanupKeystoneProcess = () => {};
 
     afterAll(async () => {
-      await cleanupKeystoneProcess();
+      cleanupKeystoneProcess();
```
dcousens, line 91: why not `await`

borisno2: oh, it is a `let` that is changing to a `Promise<void>` below. I'll undo this

dcousens (approved):

PR #8350, opened 2023-03-01
Support prisma directUrl

mmachatschek:
If we can add any tests for this, I would appreciate any guidance to which test the keystone team wants me to add a test case

### DIFF
--- a/.changeset/gentle-seahorses-knock.md
+++ b/.changeset/gentle-seahorses-knock.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Adds support for the new directUrl feature available since prisma 4.10.0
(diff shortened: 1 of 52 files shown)


mmachatschek:
this provides the same challenges as https://github.com/keystonejs/keystone/pull/7350

IMO directUrl should behave the same as shadowDatabaseUrl and should not throw an error if the env var could not be found.

mmachatschek:
I came to the conclusion that it is way easier for users to use the `extendPrismaSchema` config released in v5.0. Making assumptions about the `directUrl` settings is a bit harder than it is with the shadowDatabaseUrl` it seems

dcousens:
@mmachatschek interesting conclusion! :yellow_heart:
What problems did you run into? (out of interest)

mmachatschek:
@dcousens I hoped that it would be easier to implement the directUrl flag on the prisma schema, but the prisma migrate commands are really restrictive when you don't provide the `DIRECT_URL` env var and exits (fails) when it is missing.

The shadowDatabaseUrl feature kind of feels like the same workflow to me and therefore I expect the DIRECT_URL feature to support empty/missing env vars when executing. So it would be easier on keystone side to just add the flag to the schema. (if you think the same, then you could add a comment on the linked prisma issue and maybe answer the question of janpio 😅)

So initially I hoped to activate this for all users (just like the shadowDBURL feature) and just add it to the schema. but this fails if users don't (want) to provide a DIRECT_URL env var

If it is OK for you (@dcousens) to require the users to add the `DIRECT_URL` env var if they want to use this feature, than I can reopen this with the requirement, of only adding the DIRECT_URL if its activated in the config.

PR #8354, opened 2023-03-01
Fix behavior of `keystone start --with-migrations`

### LINKED ISSUE #8353: `keystone start --with-migrations` exits after applying migrations
Reproduction Repo here: https://github.com/acburdine/keystone-migration-bug

1. Add migrations to a keystone project with `keystone prisma migrate dev`
2. Build Keystone with `keystone build`.
3. Attempt to start Keystone using `keystone start --with-migrations`
4. See the following output:

```
✨ Starting Keystone
✨ Connecting to the database
✨ Applying database migrations
```

then it exits with a code of 0.

Based on the code [here](https://github.com/keystonejs/keystone/blob/main/packages/core/src/scripts/run/start.ts), I'd expect the `keystone start --with-migrations` command to continue to start up the Keystone server after applying database migrations.

The previous workflow of `keystone prisma migrate deploy && keystone start` still works, so it's something specific to the `--with-migrations` functionality.

acburdine:
This pull request updates the `deployMigrations` function to take `schemaPath` (similar to devMigrations function), which was exiting when it should have migrated and then stayed open.

Closes #8353

### DIFF
--- a/.changeset/tidy-keys-own.md
+++ b/.changeset/tidy-keys-own.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix behavior of `keystone start --with-migrations`
--- a/.changeset/tidy-keys-own.md
+++ b/.changeset/tidy-keys-own.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Fix behavior of `keystone start --with-migrations`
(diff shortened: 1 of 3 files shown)


borisno2 (approved):

dcousens:
Thanks @acburdine! Good catch, we'll see if we can add a test for this separately

PR #8355, opened 2023-03-01
Update schema-type-printer to output "never" for omitted GraphQL types

### LINKED ISSUE #8349: .keystone/types broken when list-level graphql.omit option provided
Full repo with reproduction here: https://github.com/acburdine/keystone-type-mismatch

1. Generate a new Keystone app
2. Modify one of the list schemas by adding the following:

```ts
list({ 
  fields: { /* ...list fields here */ },
  graphql: {
    omit: ["update"],
  }
})
```

3. Run `keystone postinstall --fix` to regenerate the built types
4. Notice type errors anywhere that list's type is referenced.

This definitely worked prior to `@keystone-6/core^5.0.0`, we noticed the type errors in our apps when upgrading to the latest version.

My guess is - the changes in [this commit](https://github.com/keystonejs/keystone/commit/7c723d9618e57ca4323fce808e5741172c522b5a) and [this commit](https://github.com/keystonejs/keystone/commit/7ebc68ebef2215e57f2317e9db9a50c68cf421a8) caused the removal of the previously-invalid "input" types from each list's generated "TypeInfo" type, thus making it such that it no longer "extends" from [BaseListTypeInfo](https://github.com/keystonejs/keystone/blob/5051e5a724f13008ce1b606858b565830f349b98/packages/core/src/types/type-info.ts#L6). This causes an individual list TypeInfo to no longer satisfy the generic type constraints required by many of the different Keystone schema building functions.

I'm not entirely sure what the best fix would be (otherwise I'd submit a PR with said fix 😉) - I _think_ making the different input fields [here](https://github.com/keystonejs/keystone/blob/5051e5a724f13008ce1b606858b565830f349b98/packages/core/src/types/type-info.ts#L12-L16) optional would do the trick? that may cause other issues, however.

As a semi-related sidenote, is there a reason the list-level graphql omit config wasn't changed to match the field-level syntax? As it currently stands, if you wanted to remove "create" from a specific field, and "update" from the whole list, you'd end up with a configuration like this:

```ts
list({
  fields: {
    somefield: text({
      graphql: {
        omit: { create: true }
      }
    }),
  },
  graphql: {
    omit: ["update"]
  }
});
```
other than the aforementioned typing issues, this config _works_ just fine, but it might be confusing to some to have one syntax in one spot and a different one in another 😅

acburdine:
closes #8349

Originally, this PR marked the `create`/`update` fields of the BaseListTypeInfo type as "optional", since that seemed like the easiest approach to fix the typing issues in #8349. However, that approach broke a number of the built in keystone tests.

After evaluating a couple of other options, having the schema type printer output "never" seems like the best approach to fix the issue - it still satisfies the "extends BaseListTypeInfo" constraint, but if anyone attempts to actually _use_ a value of that type, they'll get a typescript compile error.

### DIFF
--- a/.changeset/lemon-cougars-peel.md
+++ b/.changeset/lemon-cougars-peel.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Update schema type printer to output `never` for omitted types.
+Fixes schema type printer to output `never` for omitted types.
--- a/.changeset/lemon-cougars-peel.md
+++ b/.changeset/lemon-cougars-peel.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Update schema type printer to output `never` for omitted types.
--- a/packages/core/src/lib/schema-type-printer.tsx
+++ b/packages/core/src/lib/schema-type-printer.tsx
@@ -196,21 +196,13 @@ function printListTypeInfo<L extends InitialisedList>(
     `    inputs: {`,
     `      where: ${whereInputName};`,
     `      uniqueWhere: ${whereUniqueInputName};`,
-    ...(list.graphql.isEnabled.create ? [
-    `      create: ${createInputName};`,
-    ] : []),
-    ...(list.graphql.isEnabled.update ? [
-    `      update: ${updateInputName};`,
-    ] : []),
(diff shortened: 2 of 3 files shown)


dcousens (approved):

dcousens:
Thanks @acburdine, this looks great :green_heart: - thanks for the speedy turn-around on these contributions! :fire:

PR #8369, opened 2023-03-05
Add newline compatibility to description text in admin ui

DiesIrae:
Added a simple `white-space: pre-wrap` to field description to allow usage of `\n` in description string

### DIFF
--- a/.changeset/late-birds-clean.md
+++ b/.changeset/late-birds-clean.md
@@ -1,0 +1,5 @@
+---
+'@keystone-ui/fields': minor
+---
+
+Add newline compatibility to description text in admin ui
--- a/design-system/packages/fields/src/FieldDescription.tsx
+++ b/design-system/packages/fields/src/FieldDescription.tsx
@@ -19,6 +19,7 @@ export const FieldDescription = (props: FieldDescriptionProps) => {
         color: palette.neutral700,
         marginBottom: spacing.small,
         minWidth: 120,
+        whiteSpace: 'pre-wrap'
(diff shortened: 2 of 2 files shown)


dcousens:
@DiesIrae nothing wrong with the code, out of interest what was your use case for this?

DiesIrae:
> @DiesIrae nothing wrong with the code, out of interest what was your use case for this?

We wanted to write multi-line descriptions, it seemed natural to be able to (contrary to labels that need to be short). In For example in Google forms you can write rich descriptions (multi-line, bold...).

dcousens (approved):

dcousens:
I don't know to what extend we'd support rich descriptions in some future, but, I think this is OK as is.
Customisation is still a story yet to be told when it comes to components in the AdminUI.

Thanks for the easy addition @DiesIrae :yellow_heart:

dcousens:
> I discovered read-only versions of text fields didn't have multi-line either. What do you think about it?

Interesting, I don't think I would expect this to work, because what if a read-only is transitioned back to a read-write input field, what would the input field show?

DiesIrae:
Well, inputs fields already have this "problem" as we can inject multi-line data via graphql. The way I see it, multi-line in read-only text will:

- Improve read-only for text fields with teaxtarea
- Improve virtual text fields
- do no harm to text fields with teaxtarea

dcousens:
@DiesIrae fair, I'm not against your proposal then. Pull request accepted!

PR #8370, opened 2023-03-05
Generate Prisma Data Proxy client based on `db.url` provided

borisno2:
Generated the Prisma Client for data proxy based on the URL provided. If a `prisma://` URL is provided to `db.url` the Client is generated for the Prisma Data Proxy.

https://www.prisma.io/docs/data-platform/data-proxy/deploy

### DIFF
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -191,8 +191,7 @@ export async function generateNodeModulesArtifacts(
   ]);
 }
 
-async function generatePrismaClient(prismaSchemaPath: string, dataProxyFlag: boolean) {
-  const dataProxy = dataProxyFlag || process.env.PRISMA_GENERATE_DATAPROXY === 'true';
+async function generatePrismaClient(prismaSchemaPath: string, dataProxy: boolean) {
   const generator = await getGenerator({
     schemaPath: prismaSchemaPath,
     dataProxy,
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -181,17 +181,21 @@ export async function generateNodeModulesArtifacts(
   graphQLSchema: GraphQLSchema
 ) {
   const paths = getSystemPaths(cwd, config);
-
+  const dataProxy = config.db.url.startsWith('prisma:');
+  if (dataProxy === true) {
+    console.log('✨ Generating Prisma Client (data proxy)');
+  }
   await Promise.all([
-    generatePrismaClient(paths.schema.prisma),
+    generatePrismaClient(paths.schema.prisma, dataProxy),
     generateNodeModulesArtifactsWithoutPrismaClient(cwd, config, graphQLSchema),
   ]);
 }
 
-async function generatePrismaClient(prismaSchemaPath: string) {
+async function generatePrismaClient(prismaSchemaPath: string, dataProxyFlag: boolean) {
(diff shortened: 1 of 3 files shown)


dcousens:
Maybe we can automate this to automatically setting this to `true` when `"db.url": "prisma://somepathontheinternet/..."` is provided.

The problem that might happen then is that when you `keystone build`, you might not have your `db.url` set (from `env`), and then a subsequent `keystone start` will not work, it will blow up on the first attempted Prisma connection.

This is similar to the experience with this pull request... maybe we can improve _that_ by detecting what kind of Prisma client has been generated as part of `keystone start`?

Alternatively, we can add a flag, something like `db.useDataProxy`.
Or, we could encourage users to to provide `"db.url": "prisma://", as a way to flag to `build` they are using a data proxy, without needing to build with their production `env`.

packages/core/src/artifacts.ts
```diff
@@ -181,17 +181,21 @@ export async function generateNodeModulesArtifacts(
...
   await Promise.all([
-    generatePrismaClient(paths.schema.prisma),
+    generatePrismaClient(paths.schema.prisma, dataProxy),
     generateNodeModulesArtifactsWithoutPrismaClient(cwd, config, graphQLSchema),
   ]);
 }
 
-async function generatePrismaClient(prismaSchemaPath: string) {
+async function generatePrismaClient(prismaSchemaPath: string, dataProxyFlag: boolean) {
+  const dataProxy = dataProxyFlag || process.env.PRISMA_GENERATE_DATAPROXY === 'true';
```
borisno2, line 195: This env variable aligns with the prisma docs to generate for the data-proxy https://www.prisma.io/docs/data-platform/data-proxy/deploy

dcousens (approved):

packages/core/src/artifacts.ts
dcousens: I don't think we need that variable, it will only result in confusion about how it interacts with `"db.url": "prisma://"`

mmachatschek:
@borisno2 this doesn't work for docker builds where there are no environment vars available yet.

I would prefer a CLI flag when doing the postinstall/build so I can set if the dataproxy should be generated or not (similar to `--data-proxy` flag on the `prisma generate` cli command.

Furthermore it should listen to the `PRISMA_GENERATE_DATAPROXY` which tells the `prisma generate` command to generate the with data-proxy support 🤔

packages/core/src/artifacts.ts
mmachatschek: @dcousens  this env var and cli flag is actually required for build envs where we have no access to the production environment variables.

E.g. docker builds in CI that prepare the code for production

mmachatschek: adding a `DATABASE_URL` which has `prisma://` in it just for the resulting prisma generate to build with proxy support feels kind of weird to me

mmachatschek: hmmm...so maybe I'm the only one to whom this kind of feels weird 😅 the postinstall/build/start commands where always a pain to me when doing CI stuff. It is really hard (without reading the code) what each of the commands does.

Most of the time, I found my self diving into the cli implementations of keystone, because we found weird behaviour when executing `keystone ...` commands (but this probably is something for a discussion thread)

dcousens: Please don't hold back with the feedback, it's really valuable! :yellow_heart: 
Adding `--data-proxy` is effectively free, but, the error messages seemed confusing if you accidentally mismatch.

PR #8380, opened 2023-03-08
fix with-auth link

kennedybaird:
I think this wasn't updated


dcousens:
Thanks @kennedybaird !

dcousens (approved):

PR #8381, opened 2023-03-08
Add new `types.path` property

dcousens:
This pull request adds a new `.types.path` configuration option that will enable developers to specify where their Keystone types are built.

Previously you imported your types with `.keystone/types`, which are automatically built in `node_modules/.keystone/types.d.ts`.  This is suitable for many usecases, but if you're using a monorepo, you might want to import these types in your other workspaces. Using `node_modules/.keystone/...` prevented that.

You can now specify a `.types.path` that works for you.
If you were relying on `.keystone/types`, that will break only _if_ you use this feature.
Thereby, this is not a breaking change.

We thought about opting to try and keep `.keystone/types` resolving, even with a custom `types.path`, but that actually ended up in a number of weird edge-cases that aren't easy to debug.


examples/custom-output-paths/keystone.ts
```diff
@@ -0,0 +1,19 @@
...
+  db: {
+    provider: 'sqlite',
+    url: process.env.DATABASE_URL || 'file:./keystone-example.db',
+
+    // WARNING: this is only needed for our monorepo examples, dont do this
+    ...fixPrismaPath,
+  },
+  lists,
+
+  // TODO: add explanation about why you might want to do this
```
jlarmstrongiv, line 15: // specifying a custom types path may be needed in monorepos or with alternative package managers

examples/extend-graphql-schema-nexus/schema.ts
```diff
@@ -94,7 +94,7 @@ export function extendGraphqlSchema(baseSchema: GraphQLSchema) {
     },
 
     // Typescript output settings, probably something you might commit in dev
-    shouldGenerateArtifacts: process.env.NODE_ENV !== 'production',
+    shouldGenerateArtifacts: false ?? process.env.NODE_ENV !== 'production',
```
jlarmstrongiv, line 97: I think `false ?? process.env.NODE_ENV !== 'production',` will always evaluate to false.  The nullish coalescing operator only detects `null` or `undefined`.

packages/core/src/types/config/index.ts
```diff
@@ -94,12 +94,21 @@ export type StorageConfig = (
...
   ui?: AdminUIConfig<TypeInfo>;
   server?: ServerConfig<TypeInfo>;
   session?: SessionStrategy<any>;
-  graphql?: GraphQLConfig;
+  types?: {
+    // TODO: rename to contextTypesPath?
+    // TODO: rename to schemaTypesPath?
+    // TODO: rename to typeInfoPath?
+    // TODO: rename to internalTypesPath?
+    listTypesPath?: string;
```
jlarmstrongiv, line 108: Hard to say, I’ve seen code examples, such as:
- `import { Lists } from '.keystone/types';`
- `import type { Context } from '.keystone/types';`
- `import { TypeInfo } from '.keystone/types';`

I’m not sure what the best name would be.  If everything is in that file, then maybe a more generalized name would be good:
- `internalTypesPath`
- `generatedTypesPath`
- `keystoneTypesPath`

jlarmstrongiv (changes requested):
Looking good 😄 can’t wait to give keystone with monorepos another try.  Between this PR and https://github.com/keystonejs/keystone/pull/8307, it should be more doable now

examples/extend-graphql-schema-nexus/schema.ts
dcousens: sorry, this is a work-around for a bug in that particular example that needs to be fixed

examples/custom-output-paths/my-types.d.ts
```diff
@@ -0,0 +1,188 @@
+/* eslint-disable */
```
dcousens, line 1: Hopefully this prevents someone's day being ruined by `{linter}`

borisno2 (approved):

packages/core/src/types/config/index.ts
dcousens: @jlarmstrongiv we went with `path`, because we don't have any other examples

packages/core/src/artifacts.ts
```diff
@@ -79,15 +79,25 @@ export function getBuiltKeystoneConfigurationPath(cwd: string) {
 }
 
 export function getSystemPaths(cwd: string, config: KeystoneConfig) {
+  const builtTypesPath = config.types?.path
+    ? path.join(cwd, config.types.path)
+    : 'node_modules/.keystone/types.d.ts';
+
+  const relativePrismaPath = config.db.prismaClientPath
+    ? `./${path.relative(path.dirname(builtTypesPath), config.db.prismaClientPath)}`
```
borisno2, line 87: @dcousens is Windows going to be happy with this?

dcousens: I don't know

dcousens: It's the same problem though, we probably need to `JSON.stringify` it at the least

packages/core/src/artifacts.ts
```diff
@@ -79,15 +79,25 @@ export function getBuiltKeystoneConfigurationPath(cwd: string) {
 }
 
 export function getSystemPaths(cwd: string, config: KeystoneConfig) {
+  const builtTypesPath = config.types?.path
+    ? path.join(cwd, config.types.path)
+    : 'node_modules/.keystone/types.d.ts';
```
borisno2, line 84:
```suggestion
    : path.join(cwd, 'node_modules/.keystone/types.d.ts');
```

packages/core/src/artifacts.ts
borisno2, line 87: I think `config.db.prismaClientPath` might also need a `path.join(cwd, config.db.prismaClientPath)`

borisno2: Although the `JSON.stringify` in the `schema-type-printer` should excape the string

dcousens: Actually yes, this string replaced what was passed previously

examples/custom-output-paths/schema.ts
```diff
@@ -0,0 +1,33 @@
+import { list } from '@keystone-6/core';
+import { allowAll } from '@keystone-6/core/access';
+import { text, timestamp } from '@keystone-6/core/fields';
+import type { Lists } from '.keystone/types';
```
borisno2, line 4:
```suggestion
import type { Lists } from './my-types';
```

PR #8387, opened 2023-03-09
fix(docs): replace non-existing migrate command

kporten:
The command `keystone migrate generate` does not seem to exist. The described behavior can be achieved with `keystone migrate dev` according to the Prisma documentation: https://www.prisma.io/docs/reference/api-reference/command-reference#migrate-dev


dcousens (approved):
Thanks @kporten!

PR #8397, opened 2023-03-20
Remove `ExtendGraphqlSchema` type alias

dcousens:
This pull request removes the `ExtendGraphqlSchema` type alias, which is a breaking change.
This type alias was unnecessarily exported, and is easily replaced by the more informative to the reader, explicit type:

```typescript
import type { GraphQLSchema } from 'graphql'

// ...

(schema: GraphQLSchema) => GraphQLSchema
```

This pull request should only be merged when we have other breaking changes.


ganobrega:
@dcousens Leveraging that you are selling ExtendGraphQLSchema, I would like to request a support for asynchronous!! Can it be possible?

dcousens:
@ganobrega I can't see why not, what's your usecase?

borisno2 (approved):

PR #8399, opened 2023-03-20
Enable corepack for CI

dcousens:
This pull request updates our GitHub actions to use the latest `actions/setup-node`,  and updates our CI to use corepack to ensure it picks up the same `yarn` version as we have in our `package.json`.

The newest `setup-node` has inbuilt cache support, so we can ignore that boilerplate too.

### DIFF
--- a/.github/actions/ci-setup/action.yml
+++ b/.github/actions/ci-setup/action.yml
@@ -2,28 +2,17 @@ name: 'CI setup'
 runs:
   using: 'composite'
   steps:
-    - name: Setup Node.js LTS
-      uses: actions/setup-node@v3
+    - run: corepack enable
+      shell: bash
+
+    - uses: actions/setup-node@main
       with:
         # preferably lts/*, but we hit API limits when querying that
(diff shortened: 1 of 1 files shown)


.github/actions/ci-setup/action.yml
```diff
@@ -2,28 +2,17 @@ name: 'CI setup'
...
-        key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
-        restore-keys: |
-          ${{ runner.os }}-yarn-
-
-    - name: Install Dependencies
-      run: yarn --frozen-lockfile
+    - run: yarn --frozen-lockfile
       shell: bash
+      env:
+        PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1'
```
dcousens, line 18: We are manually running `yarn playwright install chromium` for our tests, so we can speed this up with https://playwright.dev/docs/browsers#skip-browser-downloads

borisno2 (approved):

PR #8403, opened 2023-03-21
Fix Document editor with `"use client"` for `getContext` in NextJS App Directory

borisno2:
Currently, if you add the document field to a project and use `getContext` in the NextJs app directory, the next build will fail due to how the document editor validates the schema on the server side - the result being `useState` in a server-side component. This PR marks two files with `"use client"` so Next knows these are client-side files.

Thanks to @gautamsi for working this one out - see comment on #8371

### DIFF
--- a/.changeset/short-impalas-hide.md
+++ b/.changeset/short-impalas-hide.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/fields-document': patch
+---
+
+Adds `"use client"` to Document field files that allow it to be used in the Next App Dorectory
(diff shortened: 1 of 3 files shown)


dcousens (approved):
LGTM, though I wonder how scalable this is for poor nextjs@13 if you need upstream libraries to add directives

borisno2:
This is failing due to preconstruct and rollup bundling files together where Module Level directives are ignored. That makes this approach not achievable. 

The root cause of this is the crossover of React/Next code in the `getContext` - in the document field case, this is used for validation of the JSON object to stop runtime errors in the frontend (Keystone effectively renders the Document Editor on the backend to parse and validate the JSON). Ideally, the root cause of this needs to be resolved - remove the need for and decouple the Admin UI (Studio) from the API framework (`getContext`) this will also help with reducing the weight of keystone when using `getContext`

Closing this PR, while we continue to explore this use case in #8371

borisno2:
Giving this another go in light of the latest preconstruct release #8423

gautamsi:
great to see this back, please release soon :)

PR #8404, opened 2023-03-22
Remove typescript dev dependency from packages

borisno2:


dcousens:
Unfortunately we need `tsc` to resolve in the packages for local development, this isn't actually going to work

PR #8411, opened 2023-03-25
Fix resolveData hook example

DustinWoods:
Should be fieldKey instead of fieldName.

(Sorry, haven't read contributing guidelines, just saw this mistake in docs)


dcousens (approved):
Thanks @DustinWoods

PR #8438, opened 2023-03-29
Tidyup examples and internal `InitialisedList` type

dcousens:
This pull request removes a number of `CHANGELOG`s from the examples that are unused.
These files are routinely copied between examples and are not useful for consumers, or us.
I have opted to remove them. 

This pull request additionally changes the internal `InitialisedList` type to follow suite with a more consistent structure centered around the property keys we are used to.  This should help future work currently ongoing in this area.

This pull request has no changeset, as nothing has changed in respect to our [publicly documented interfaces](https://keystonejs.com/docs).
Users of `context.experimental.lists` will be impacted, but that outcome is part of using `experimental` and will be trivially caught by Typescript.


packages/core/src/lib/context/api.ts
```diff
@@ -0,0 +1,102 @@
+import { GraphQLSchema } from 'graphql';
+import type { InitialisedList } from '../core/types-for-lists';
+import { KeystoneContext } from '../../types';
+import { executeGraphQLFieldToRootVal } from './executeGraphQLFieldToRootVal';
+import { executeGraphQLFieldWithSelection } from './executeGraphQLFieldWithSelection';
+
+export function getQueryFactory(list: InitialisedList, schema: GraphQLSchema) {
+  function f(operation: 'query' | 'mutation', field: string) {
```
borisno2, line 8: Could the `field` `fieldName` naming be standardised? - `Field` in my head always goes to a Keystone Field instance, where this is referring to the GrpahQL field which is the query/mutation name

borisno2 (approved):
Looks Good

PR #8439, opened 2023-03-29
Always use POSIX path separator when printing types

borisno2:
#8381 added the ability to set an output path for Keystone types to be generated.
The generated types printed a relative path to the Prisma client when a `prismaClientPath` is provided.

On Windows, this would use `\\` as the path separator, compared to `/` on POSIX operating systems.
Windows is interoperable with either.  Using `\\` resulted in unreadable path strings for Windows users.

This pull request changes the behaviour to always use a POSIX path separator (`/`), ensuring if types are committed to source control they are readable on both platforms, without any change in behaviour.

### DIFF
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -78,6 +78,10 @@ export function getBuiltKeystoneConfigurationPath(cwd: string) {
   return path.join(cwd, '.keystone/config.js');
 }
 
+function posixify(s: string) {
+  return s.split(path.sep).join('/');
+}
+
 export function getSystemPaths(cwd: string, config: KeystoneConfig) {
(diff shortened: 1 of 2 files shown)


packages/core/src/artifacts.ts
```diff
@@ -78,6 +78,10 @@ export function getBuiltKeystoneConfigurationPath(cwd: string) {
   return path.join(cwd, '.keystone/config.js');
 }
 
+function posixify(s: string) {
+  return s.split(path.sep).join('/');
```
dcousens, line 82: This might be OK for Windows, as `/` is not allowed in Windows filenames, but for Unix the `split` would be wrong, but probably still OK as it's immediately rejoined with `/`...

dcousens (approved):

PR #8443, opened 2023-04-01
Extend `examples/extend-graphql-schema-graphql-ts` to show chained queries/mutations

wchorski:
expanding on the documented `examples/extend-graphql-schema-graphql-ts`. This is a simple mutation extension that helps explain how to construct a chained `query` and mutation


dcousens:
Hi @wchorski,
I don't think we want to add an example of doing two queries in this particular example. (showing how to do a GraphQL schema extension)

Did you have any particular concerns that you wanted to help others with?

wchorski:
apologies, I wanted to contribute in some way. I'm working on Wes Bos's 'Advance React Course' while also porting it to the newer keystone 6 and want to help others make that conversion in keystone syntax

dcousens:
No need to apologize @wchorski, being proactive is a good thing!

Do you think changing the usage of `resolve` to an `async` function would have helped you without needing two queries explicitly?

PR #8445, opened 2023-04-02
package document-editor

mingfang:


Thinkscape:
Actually this would be quite useful for using the editor outside of Admin, or in case one has a custom admin that needs WYSIWYG `document` editing.

PR #8470, opened 2023-04-11
Change preview deployment trigger to pull_request_target

dcousens:
See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

This should hopefully allow us to (with approval) run the preview deployments for Vercel on (potentially) untrusted pull requests, using the base branch workflows instead of the pull request workflows (which may be modified).

### DIFF
--- a/.github/workflows/website_preview.yml
+++ b/.github/workflows/website_preview.yml
@@ -1,7 +1,7 @@
 name: Website (Preview)
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


borisno2 (approved):

dcousens:
Actually, this still needed "Required reviewers" on the environment to prevent secret leaks, I think we'll revert

PR #8471, opened 2023-04-11
Add Astro SSR example using `getContext`

borisno2:


examples/astro-keystone/astro.config.mjs
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'astro/config';
+import vercel from '@astrojs/vercel/serverless';
+
+// https://astro.build/config
+export default defineConfig({
+  output: 'server',
+  adapter: vercel(),
```
dcousens, line 7: Maybe we shouldn't have this `adapter` for the example?

borisno2: Agreed, or use the node adapter - was mainly seeing if it made a difference to the Prisma error I am getting. I think an adapter is only required to "deploy" - https://docs.astro.build/en/guides/server-side-rendering/#adding-an-adapter

dcousens (approved):

PR #8480, opened 2023-04-12
Add a new `graphql.typeName` for the `structure` field to support deduplication of GraphQL types

borisno2:
Adds `graphql.typeName` to structured JSON field to allow structured JSON to have global types 
Note if the types don't match it will cause GraphQL runtime errors, this will also not work with relationships

### DIFF
--- a/.changeset/early-feet-heal.md
+++ b/.changeset/early-feet-heal.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/fields-document': minor
+---
+
+Allow the global Graphql Types for Structured JSON
--- a/.changeset/early-feet-heal.md
+++ b/.changeset/early-feet-heal.md
@@ -2,4 +2,4 @@
 '@keystone-6/fields-document': minor
 ---
 
-Allow the global Graphql Types for Structured JSON
+Add a new `graphql.typeName` for the `structure` field to support deduplication of GraphQL types
--- a/packages/fields-document/src/structure.ts
+++ b/packages/fields-document/src/structure.ts
@@ -52,46 +52,47 @@ export const structure =
 
     const defaultValue = getInitialPropsValue(schema);
     const unreferencedConcreteInterfaceImplementations: graphql.ObjectType<any>[] = [];
-    if (!globalStructureTypes[name]) {
-      globalStructureTypes[name] = {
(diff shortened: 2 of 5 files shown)


dcousens:
Can you add an example to the test sandbox?

molomby:
This relates to a more general problem I've hit around GraphQL type re-usage across fields and custom mutations. Currently there's no way (that I know of; please correct me if wrong) that let's you share/reuse a GraphQL types between multiple virtual fields, nor is it possible to share a GraphQL type between a virtual fields and GraphQL extensions.

In my current project we have managed to reuse types across multiple GraphQL extensions (by closuring over `base` then passing around an object of types) but it kinda sucks. 

Ends up looking something like this:

```ts
export const extendGraphqlSchema: ExtendGraphqlSchema = graphql.extend((base) => {
  const loaders = buildLoaders(base);
  const dynamicGqlTypes = buildDynamicGqlTypes(base, loaders);

  return {
    query: {
      dodads: dodadsQuery({ dynamicGqlTypes }),
      thingamaJigs: thingaMaJigsQuery({ dynamicGqlTypes }),
      search: searchQuery({ base, dynamicGqlTypes }),
    },
    mutation: {
      createDodad: createDodadMutation({ dynamicGqlTypes }),
      updateDodad: updateDodadMutation({ dynamicGqlTypes }),
      createThingamaJig: createThingamaJigMutation({ dynamicGqlTypes }),
      updateThingamaJig: updateThingamaJigMutation({ dynamicGqlTypes }),
    },
  };
});
```

If you could lazily create and cache them like this...

```ts
let DodadGraphQLType;

function getDodadGraphQLType(base) {
  if (DodadGraphQLType) return DodadGraphQLType;
 DodadGraphQLType = graphql.object<DepositTransactionLeg>()({
    name: 'Dodad',
    fields: {
      // ...
    },
  });
  return DodadGraphQLType;
}
```

Then you could just pull the types you needed into each mutation separately but this doesn't work because Keystone creates _two schemas_ (the standard one and another for `sudo` operations) – you need two distinct GraphQL type objects to be created so they can closure the two different `base` objects.

**Shouldn't Keystone have a single place we can define GraphQL types for a schema?** Then we can refer to those types (by name, which is always unique) when they're needed in custom queries/mutations, virtual fields, structure fields, etc.

PR #8488, opened 2023-04-16
chore(deps): include all `@keystone-6/core` files inside npm package

pahaz:
Currently, we don't have an `src` folder, `CHANGELOG.md`, and so on ...

Related to https://github.com/keystonejs/keystone/discussions/8487

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -188,23 +188,6 @@
   "bin": {
     "keystone": "bin/cli.js"
   },
-  "files": [
-    "session",
-    "bin",
-    "___internal-do-not-use-will-break-in-patch",
-    "next",
-    "schema",
-    "scripts",
-    "dist",
-    "static",
-    "admin-ui",
-    "testing",
-    "fields",
-    "types",
-    "system",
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -188,23 +188,6 @@
...
-    "dist",
-    "static",
-    "admin-ui",
-    "testing",
-    "fields",
-    "types",
-    "system",
-    "context",
-    "access"
-  ],
```
dcousens, line 207: I am OK with this, we don't stipulate exactly what files are in other packages.
For defining our public exports, we use `"exports"`.

@borisno2 any objections?

dcousens (approved):

packages/core/package.json
borisno2: LGTM @dcousens - the only possible concern would be bundle size, however, given the `src` directory is only 1MB  I think there are better ways to reduce that.

dcousens: @borisno2 I am only seeing 168K for the `src` when compressed

PR #8491, opened 2023-04-17
Add declarationMap for source mapping

dcousens:
Follow up to https://github.com/keystonejs/keystone/pull/8488, one of the benefits of having the source available, is we can actually reference that instead of the compiled bundles.

This pull request adds `"declarationMap": true`, and then sorts the fields in the `tsconfig.json` lexicographically.


emmatown (approved):

dcousens:
Failing test is unrelated

PR #8521, opened 2023-04-27
Remove env.SKIP_PROMPTS for CLI

dcousens:
Duplicate of https://github.com/keystonejs/keystone/pull/8517 (reverted temporarily in https://github.com/keystonejs/keystone/pull/8520)

---

Previously added in https://github.com/keystonejs/keystone/pull/5758, this is being removed.
If you don't want an interactive CLI, don't run the command in an interactive CLI.

For example, using the same format as the [node documentation](https://nodejs.org/api/tty.html#tty):

```bash
keystone/examples/auth $ pn postinstall

Your Prisma schema is not up to date
Replace the Prisma schema? … (Y/n)
...

keystone/examples/auth $ pn postinstall | cat

Your Prisma schema is not up to date
Use keystone dev to update the Prisma schema
... ELIFECYCLE  Command failed with exit code 1.
```

### DIFF
--- a/.changeset/no-skip-prompts.md
+++ b/.changeset/no-skip-prompts.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Removes CLI environment variable `SKIP_PROMPTS`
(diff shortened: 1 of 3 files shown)


dcousens:
Removed in https://github.com/keystonejs/keystone/pull/8545 anyway

PR #8545, opened 2023-05-03
Add `ui.basePath` which sets the `basePath` in the AdminUI next config 

borisno2:
### Problem
We want to deploy Keystone to somewhere not at the root route `/`.
While you can use `graph.path` to modify the GraphQL API path, the AdminUI needs to have the base path changed too.
Advanced users can do this by replacing their `next.config.js` as part of the `ui.getAdditionalFiles` step, but, that isn't ideal.
We additionally have `withPreconstruct` as part of our project runtime dependencies, which isn't great.

### Proposal

```typescript
  ui: {
    basePath: '/foo'
  }
```

This is written to a now dynamic `next.config.js` (replacing the static variant we use now).
If advanced users are modifying the next configuration for their own purposes, well, they can continue doing that anyway.

This will break anyone using `__internal...`, that's OK.


tests/sandbox/utils.ts
```diff
@@ -28,6 +34,30 @@ export const localStorageConfig: Record<string, StorageConfig> = {
...
+  getAdditionalFiles: [
+    () => [
+      {
+        mode: 'copy' as const,
+        inputPath: Path.resolve('./test-next-config-fix.js'),
+        outputPath: 'next.config.js',
+      },
+    ],
+  ],
+};
```
dcousens, line 60:
```suggestion
```

dcousens (approved):
Awesome! Nice find with [`transpilePackages`](https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages)

PR #8546, opened 2023-05-04
Remove ansi-strip dependency for tests

dcousens:
The new version is having ESM issues since https://github.com/keystonejs/keystone/pull/8531.

We don't actually need this typically, we only used it for tests to remove some ANSI character strings.
For nearly all the calls we can replace with an ASCII only regex and for the ANSI parts, we can use `stringContaining` or something else.

Tbh, I think this is better than when the snapshot strings had emojis anyway.


packages/core/src/scripts/tests/prisma.test.ts
```diff
@@ -15,11 +14,11 @@ test('keystone prisma exits with the same code as the prisma child process exits
     all: true,
     cwd: tmp,
   });
-  expect(stripAnsi(result.all!)).toMatchInlineSnapshot(`
+  expect(result.all!.replace(/[^ -~]\+/, '?')).toMatchInlineSnapshot(`
     "
     ! Unknown command "bad-thing"
 
-    ◭  Prisma is a modern DB toolkit to query, migrate and model your database (https://prisma.io)
+    ?  Prisma is a modern DB toolkit to query, migrate and model your database (https://prisma.io)
```
dcousens, line 21: `?` is easier than an emoji

packages/core/src/scripts/tests/migrations.test.ts
```diff
@@ -429,11 +429,11 @@ describe('useMigrations: true', () => {
     const tmp = await testdir({
       ...symlinkKeystoneDeps,
       'app.db': await fs.readFile(`${prevCwd}/app.db`),
-      'keystone.js': await fs.readFile(`${__dirname}/fixtures/one-field-with-migrations.ts`),
+      'keystone.js': await fs.readFile(`${__dirname}/fixtures/one-field-with-migrations-2.ts`),
     });
     const recording = recordConsole({
       'Do you want to continue? All data will be lost.': true,
-      'Name of migration': 'init',
+      'Name of migration': 'init2',
```
dcousens, line 436:
```
{
  "oldMigrationName": "20230504081137_init",
  "migrationName": "20230504081138_init2"
}
```

This was previously failing when the timestamps were the same, and hence the migration names were the same

packages/core/src/lib/migrations.ts
```diff
@@ -235,11 +235,10 @@ We need to reset the ${credentials.type} database "${credentials.database}" at $
...
     if (appliedMigrationNames.length) {
-      console.info(
-        `✨ The following migration(s) have been applied:\n\n${printFilesFromMigrationIds(
-          appliedMigrationNames
-        )}`
-      );
+      console.info(`✨ The following migration(s) have been applied:`);
+      for (const id of appliedMigrationNames) {
+        console.info(`  - ${chalk.cyan.bold(id)}`);
+      }
```
dcousens, line 241: Not necessary, but simpler

packages/core/src/scripts/tests/migrations.test.ts
```diff
@@ -405,35 +405,35 @@ describe('useMigrations: true', () => {
...
     const { migrationName: oldMigrationName, prevCwd } = await setupInitialProjectWithMigrations();
     const tmp = await testdir({
       ...symlinkKeystoneDeps,
       'app.db': await fs.readFile(`${prevCwd}/app.db`),
-      'keystone.js': await fs.readFile(`${__dirname}/fixtures/one-field-with-migrations.ts`),
+      'keystone.js': await fs.readFile(`${__dirname}/fixtures/one-field-with-migrations-2.ts`),
     });
     const recording = recordConsole({
       'Do you want to continue? All data will be lost.': true,
-      'Name of migration': 'init',
```
dcousens, line 436:
```
{
  "oldMigrationName": "20230504081137_init",
  "migrationName": "20230504081138_init2"
}
```

This was previously failing when the timestamps were the same, and hence the migration names were the same

borisno2 (approved):

PR #8552, opened 2023-05-05
Add custom session strategy example

dcousens:
This pull request adds a custom session strategy example.
The example shows how to implement a simple session strategy for authentication, using a cookie.

The session strategy, `sillySessionStrategy`, is a demonstrative example acting as a starting point for implementing your own session strategy. The `Cookie:user={user_id}` format used in this example should outline the moving parts of an implementation, but it is by no means actual authentication. In practice, you should use actual authentication, such as OAuth or JWT.

You can pass a cookie with the format `Cookie:user={id}` in your HTTP requests, and the strategy will extract the `id` from the cookie and look up the corresponding user in the `User` list. If the user is found, the strategy will return a session object with an id field and an `admin` flag.


dcousens:
Separate to this, which is a rather arbitrary sandbox example that is fun to play with to learn how custom session strategies can work, we should add a `custom-session-jwt` as a more concrete example.

packages/core/src/types/config/index.ts
```diff
@@ -99,7 +99,7 @@ export type KeystoneConfig<TypeInfo extends BaseKeystoneTypeInfo = BaseKeystoneT
   lists: ListSchemaConfig;
   ui?: AdminUIConfig<TypeInfo>;
   server?: ServerConfig<TypeInfo>;
-  session?: SessionStrategy<any>;
+  session?: SessionStrategy<any, any, any>;
```
borisno2, line 102: Should this stay the way it was?

examples/custom-session/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "dev": "keystone dev",
+    "start": "keystone start",
+    "build": "keystone build",
+    "postinstall": "keystone postinstall"
+  },
+  "dependencies": {
+    "@keystone-6/auth": "^7.0.0",
+    "@keystone-6/core": "^5.0.0",
+    "@prisma/client": "^4.13.0",
+    "next": "^13.3.0",
```
borisno2, line 16: Is `next` needed here?

dcousens: Nope, thanks

borisno2 (approved):

Issue #8553, opened 2023-05-05
Problem launching new Keystone project due to issue with Next.js version


unbywyd:
**New installation is not working**, there is a problem with Next.js. It is necessary to manually downgrade to version 13.3.4, as the default installation installs the latest version ^13.4 which does not work

1. new installation with npm / npx
2. cd app and npm run dev
3. everything is fine until you try to open a running project in the browser:
<img width="863" alt="image" src="https://user-images.githubusercontent.com/24291513/236559408-26bebfbe-166f-40b4-adb0-3cac35dce45c.png">
<img width="513" alt="image" src="https://user-images.githubusercontent.com/24291513/236559466-7114791f-4952-43e9-afc6-00b0f609f42b.png">

By the way, I've noticed that if you install a version of Next.js lower than 13.3.4, the build process gets stuck, which can cause issues. To fix this, you can try changing the package dependency to a specific version of Next.js, such as "next:": "13.3.4", instead of using the default "^13.3.0". This should help mitigate the problem and allow the build process to complete successfully.

molotow11:
I have the same issue with latest version, checked with default project setup from examples

gautamsi:
use resolution to pin nextjs version to `13.3.4`

molotow11:
@gautamsi 
Thanks, works with resolutions (need to be in top level package.json if monorepo).
```
  "resolutions": {
    "next": "13.3.4"
  }
```

mmachatschek:
I've also bumped into this issue. Pinning to 13.3.4 works for me too

MohammadKurjieh:
Pinning to version 13.3.4 works. @gautamsi Any idea what might have caused this issue?

Thinkscape:
👍 would love to start using 13.4

MohammadKurjieh:
I was able to resolve the issue, it was caused by the appDir being set to true in next v13.4 and above.
Reference on Nextjs: https://github.com/vercel/next.js/issues/49261 and https://github.com/vercel/next.js/discussions/49251

Check the following pull request for more details: #8568

beerdedfellow:
Hello, this fix is not working for me:


```
rhutchison@Ryans-M2M.local:/Users/rhutchison $ mkdir keystone_test
rhutchison@Ryans-M2M.local:/Users/rhutchison $ cd keystone_test 
rhutchison@Ryans-M2M.local:/Users/rhutchison/keystone_test $ npm init keystone-app@latest

[...]

PR #8568, opened 2023-05-10
Fix `keystone dev` errors for `next@^13.4.0`

MohammadKurjieh:
This resolves: #8553 

On next version 13.4.0 and higher setting the appDir to false is mandatory in the next config file.
Ref: https://github.com/vercel/next.js/issues/49261 and https://github.com/vercel/next.js/discussions/49251

### DIFF
--- a/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/next-config.ts
+++ b/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/next-config.ts
@@ -3,6 +3,9 @@ import Path from 'path';
 import withPreconstruct from '@preconstruct/next';
 
 export const config = withPreconstruct({
+  experimental: {
+    appDir: false,
+  },
   typescript: {
     ignoreBuildErrors: true,
   },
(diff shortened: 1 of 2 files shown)


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/next-config.ts
```diff
@@ -3,6 +3,9 @@ import Path from 'path';
 import withPreconstruct from '@preconstruct/next';
 
 export const config = withPreconstruct({
+  experimental: {
+    appDir: false,
```
dcousens, line 7: I wonder, won't this flag disappear in the next release? :thinking:

dcousens (approved):
LGTM, but if `experimental.appDir` drops in an upcoming `next` version, we need to resolve this differently.
Users might actually want `appDir: true` too.

Thinkscape:
> LGTM, but if `experimental.appDir` drops in an upcoming `next` version, we need to resolve this differently. Users might actually want `appDir: true` too (though, this might not affect them if they are doing their own next app)

But this is only for the admin ui, which means that it only affects users that want to use `appDir` inside the `/admin/` folder. Probably the complete fix would be converting the current `/admin/pages*` generators to app router.

dcousens:
Until I have an alternative solution, this is good to go, thanks @MohammadKurjieh

PR #8570, opened 2023-05-11
Add `Session` type parameter for `TypeInfo`, `Lists` and `Context`

dcousens:
This pull request amends the types exported by Keystone to support a new `Session` type parameter for `TypeInfo`, written as `TypeInfo<Session>`.

The following type parameters have been added:
- `TypeInfo<Session>`
- `Lists<Session>`
- `Context<Session>`
- `Config<Session>`

The new `Session` type parameter will propagate to the underlying functions and schema in the same manner as `Context`, ensuring that `context.session` is changed from the default of `any`. 

This pull request additionally adds or fixes a number of session related examples to adhere to this new format, and actually resolved a few type errors that had existed with respect to the previous `Session` types used in the examples. :tada:


packages/core/src/session/index.ts
```diff
@@ -31,6 +23,12 @@ type StatelessSessionsOptions = {
    * @default 60 * 60 * 8 // 8 hours
    */
   maxAge?: number;
+  /**
+   * The name of the cookie used by `Set-Cookie`.
+   *
+   * @default keystonejs-session
+   */
+  cookieName?: string;
```
dcousens, line 31: I added this, as `keystonejs-session` as cookie name is wildly opinionated

borisno2 (approved):
Nice Work!!!

PR #8572, opened 2023-05-12
Updated MySQL code example default port

ciruz:
Updated MySQL code examples to use the correct MySQL port `3306` instead of PostgreSQL port `5432`, to prevent configuration errors for beginners.


dcousens (approved):
Thanks @ciruz

PR #8582, opened 2023-05-19
Add next-auth using a custom session strategy

borisno2:
This new example shows what it might look like to implement next-auth with Keystone without the use of an external package. Given https://github.com/OpenSaasAU/keystone-nextjs-auth is no longer working with later versions of next-auth I wanted to see/get feedback on using next-auth directly instead of potentially rearchitecting the package for the upcoming `next-auth@5` release.


dcousens (changes requested):
Awesome!

I think we should show how you might use the session returned, for example with GitHub my `session` object was the following:
```
session: {
  user: {
    name: 'Daniel Cousens',
    email: undefined,
    image: 'https://avatars.githubusercontent.com/u/413395?v=4'
  },
  expires: '2023-06-30T00:12:34.567Z'
}
```

gautamsi:
would it make sense to merge this session info with actual user in keystone and present compatible keystone session? This session is also useful and may represent fresh from provider without matching user in database then we can redirect them to complete signup page and either link account or complete signup, for example having user provide missing email.

borisno2:
Thanks for the feedback @dcousens and @gautamsi 
I have just pushed two commits:
- The first (`Add user email and id to session`) adds the user's email and subject id to the next-auth session
- The Second (`Add User to DB on Signin`) checks the database for the user and creates a new `Author` if they are not found, then adds the `itemId` to the Keystone session. (This is what [@opensaasau/keystone-nextjs-auth](https://github.com/OpenSaasAU/keystone-nextjs-auth) currently does)

The reason I started with just a `hasSession`is  I didn't want to overcomplicate what the minimum requirement is to get next-auth working with Keystone. However maybe given how `getContext` needs to be imported, this might be useful in the example.

examples/custom-session-next-auth/schema.ts
```diff
@@ -0,0 +1,43 @@
...
+      // the document field can be used for making rich editable content
+      //   learn more at https://keystonejs.com/docs/guides/document-fields
+      content: text(),
+      author: relationship({ ref: 'Author.posts', many: false }),
+    },
+  }),
+  Author: list({
+    access: {
+      operation: {
+        ...allOperations<Lists.Author.TypeInfo<Session>>(denyAll),
```
dcousens, line 33:
```suggestion
        ...allOperations(denyAll),
```

dcousens (approved):

dcousens (approved):

PR #8612, opened 2023-06-05
Fix db usage in onConnect when combined with migrations

### LINKED ISSUE #8611: Cannot interact with database in db.onConnect function using `--with-migrations`
## Reproduction Steps:

1. Generate db migrations (`npx keystone prisma migrate dev --frozen`);
2. Add an onConnect handler that interacts with the database in some way (i.e. using `await context.prisma.post.count()`)
3. Run your app against a new database with `npx keystone start --with-migrations`);
4. See an error thrown due to a database table not existing.

Reproduction repo here: https://github.com/acburdine/ks-migration-bug-repro

This is likely caused by [these lines](https://github.com/keystonejs/keystone/blob/main/packages/core/src/scripts/start.ts#L38-L42) - connect is called before deployMigrations in keystone's start script.

## Expectation

`onConnect` should be able to interact with the database.

## Screenshot

image

acburdine:
closes #8611
- reorders database connection to occur after migrations, to ensure that onConnect hooks are run after the database migrations are run

this seemed to work fine when I tested it locally with the reproduction repo from #8611, but I'm not sure if there are other consequences of reordering these statements

### DIFF
--- a/.changeset/flat-spoons-learn.md
+++ b/.changeset/flat-spoons-learn.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Fixes onConnect database usage with database migrations
+Fixes `db.onConnect` for `keystone start` to run _after_ migrations are completed 
--- a/.changeset/flat-spoons-learn.md
+++ b/.changeset/flat-spoons-learn.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Fixes `db.onConnect` for `keystone start` to run _after_ migrations are completed 
+Fixes `db.onConnect`, with `keystone start --with-migrations`, to run  _after_ migrations complete
--- a/.changeset/flat-spoons-learn.md
+++ b/.changeset/flat-spoons-learn.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fixes onConnect database usage with database migrations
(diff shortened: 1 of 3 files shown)


acburdine:
looks like the integration test failure may just need a re-run?

dcousens:
Interesting point, but I don't know if this feature is something we want to encourage.
We don't think that migrations should automatically run on a deployment, but should be specifically called and handled as a separate step.

This is why in recent work, we are considering deprecating `withMigrations`.

acburdine:
@dcousens I recall seeing the note about deprecating useMigrations in the keystone config, but it was my understanding that `--with-migrations` was added to `keystone start` as an alternative to that 🤔

dcousens:
@acburdine excellent point, maybe we should move this call into the `withMigrations` then? We probably shouldn't connect to the database unless it is required (to support short lived calls).

dcousens:
Actually, another consequence of this, is that `db.onConnect` is called. We might have to change that to `server.onStart` in a breaking change to change this behaviour...

acburdine:
@dcousens I looked at the implementation of `keystone start` vs `keystone dev`, and for what it's worth - the change in this PR aligns the behavior of start with dev - in the `keystone dev` command the db sync stuff (i.e. the prisma db push call) is done before the `keystone.connect()` call.

I don't _think_ anything else needs to be done to fix things apart from what's been done in ths PR? at least, if we want the dev behavior to more or less match what start does.

dcousens (approved):

PR #8619, opened 2023-06-08
Adds `itemValue` prop for custom field view components to allow value-driven field conditions

jim-lake:
Allows custom field views to access entire item in the editor so you can do cross field dependancies, visibility, etc.


dcousens:
Thanks @jim-lake for the foundational work on this, we have added a changeset and an example to ensure the functionality is working.  A type was missing which has been added too. :yellow_heart:

dcousens (approved):

PR #8626, opened 2023-06-13
Removes redirect functionality from `/signin` page to prevent open redirection CVE

dcousens:
Thanks to @scgajge12 for reporting https://github.com/keystonejs/keystone/security/advisories/GHSA-jqxr-vjvv-899m, for now there is no straight forward resolution.

It is my understanding that the best approach we could take here would be to only allow redirecting to routes accepted by the AdminUI, but that functionality did not appear to be easy to implement quickly.

Although this is a breaking change, I have opted to mark this as a `patch`, as users will likely prefer the security resolution in a patch compared to an upgrade.

@borisno2 and I intend to fix this in newer versions of this package.

### DIFF
--- a/docs/pages/404.tsx
+++ b/docs/pages/404.tsx
@@ -30,13 +30,13 @@ function ConstructionIllustration() {
   );
 }
 
-// Modifying this code may have security implications
-// See.. https://github.com/keystonejs/keystone/pull/6411#issuecomment-906085389
+// modifying this code may have security implications
+//   see https://github.com/keystonejs/keystone/pull/6411#issuecomment-906085389
 const v5PathList = ['/tutorials', '/guides', '/keystonejs', '/api', '/discussions'];
 
 export default function NotFoundPage() {
   const { asPath } = useRouter();
-  const tryV5Link = asPath.startsWith('/') && v5PathList.some(i => asPath.startsWith(i));
+  const tryV5Link = v5PathList.some(x => asPath.startsWith(x));
   return (
     <Page title={'Page Not Found (404)'} description={'Page Not Found (404)'}>
       <div
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -138,38 +138,26 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
    *
    * Validates the provided auth config; optional step when integrating auth
    */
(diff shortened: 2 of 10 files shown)


docs/pages/404.tsx
```diff
@@ -30,13 +30,13 @@ function ConstructionIllustration() {
...
-// Modifying this code may have security implications
-// See.. https://github.com/keystonejs/keystone/pull/6411#issuecomment-906085389
+// modifying this code may have security implications
+//   see https://github.com/keystonejs/keystone/pull/6411#issuecomment-906085389
 const v5PathList = ['/tutorials', '/guides', '/keystonejs', '/api', '/discussions'];
 
 export default function NotFoundPage() {
   const { asPath } = useRouter();
-  const tryV5Link = asPath.startsWith('/') && v5PathList.some(i => asPath.startsWith(i));
+  const tryV5Link = v5PathList.some(x => asPath.startsWith(x));
```
dcousens, line 39: This would be the preferred approach that we would want for the AdminUI, to be honest

packages/auth/src/index.ts
```diff
@@ -72,16 +72,23 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
...
+        }
+      : null),
+
+    ...(magicAuthLink
+      ? {
+          magicAuthToken: password({ ...fieldConfig }),
+          magicAuthIssuedAt: timestamp({ ...fieldConfig }),
+          magicAuthRedeemedAt: timestamp({ ...fieldConfig }),
+        }
+      : null),
```
dcousens, line 91: Verbose definitions are preferred to procedural, personally

packages/auth/src/index.ts
```diff
@@ -184,34 +176,31 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
...
-          return;
-        }
+        if (!session) return;
+        if (!session.itemId) return;
+        if (session.listKey !== listKey) return;
 
         try {
           const data = await sudoContext.query[listKey].findOne({
-            where: { id: session.itemId as any }, // TODO: fix this
+            where: { id: session.itemId },
```
dcousens, line 185: `string | number | null` is OK

borisno2 (approved):

PR #8640, opened 2023-06-19
Add example for custom session strategy using JWTs

dcousens:
This pull request adds an example for a `jwtSessionStrategy`, using the [`jsonwebtoken`](https://github.com/auth0/node-jsonwebtoken) library.

This has been a popular request from a number of developers, and although this example will need to tailored to whatever purpose it is used for, this should demonstrate the principles of how this might work in a way that is less abstract than the `sillySessionStrategy` or some of the other `custom-session-*` examples.


borisno2 (approved):
Nice one

PR #8641, opened 2023-06-19
Fixes redirect to signin page in `@keystone-6/auth` when using `basePath`

### LINKED ISSUE #8637: Issue with wrong  `/signin` redirection when basePath is changed to `/admin`
# Description
I encountered an issue when attempting to change the path for the admin UI to `/admin` in the `keystone.ts` file. 
Here's the code in the `keystone.js`: 

```
export default withAuth(
  config({
    db: {
      provider: 'sqlite',
      url: 'file:./keystone.db',
    },
    ui: {
      basePath: '/admin',
      isAccessAllowed: (context) => !!context.session?.data,
    },
    lists,
    session,
    server: {
      extendExpressApp: (app) => {
        app.use('/', express.static('pogiClient'));

        app.get(/^(?!\/admin(\/|$)).*/, (req, res) => {
          res.sendFile(path.resolve('pogiClient/index.html'));
        });
    
      }
    },
  })
)
```

When I access `/admin` without authentication, it redirects to `/signin?from=%2Fadmin`, resulting in a 404 error.

image

# Expected Behavior
When attempting to access `/admin` without authentication, it should properly redirect to `/admin/signin`.

I tried to remove the `extendExpressApp` configuration, but the result was still the same. 

# Contextual Information 

Environment: 
- NodeJS v18.16.0
- WSL

"@keystone-6/auth": "^7.0.0"
"@keystone-6/core": "^5.3.1"
"@keystone-6/fields-document": "^8.0.0"
"typescript": "^5.1.3"

borisno2:
- Fixes #8637

### DIFF
--- a/.changeset/no-redirect.md
+++ b/.changeset/no-redirect.md
@@ -1,5 +1,5 @@
 ---
-'@keystone-6/core': patch
+'@keystone-6/auth': patch
 ---
 
 Removes `?from` redirect from `/signin` page to prevent open redirection.
--- a/.changeset/twelve-toes-punch.md
+++ b/.changeset/twelve-toes-punch.md
@@ -1,0 +1,6 @@
+---
+'@keystone-6/core': patch
+'@keystone-6/auth': patch
+---
+
+Adds `basePath` with a default of `'/'` to `pageMiddleware` args to help with redirecting when using `ui.basePath`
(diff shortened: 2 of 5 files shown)


dcousens (approved):

pldvd:
I upgraded to "@keystone-6/core": "^5.3.0" last Friday because I wanted to use ui.basePath, I noticed that something was wrong with the redirects but couldn't investigate much to understand so I left it there. Today I had a closer look and found this MR. It fixed my app, and just in the right moment! Thank you!

PR #8648, opened 2023-06-23
Adds `cuid2` support and fixes types for custom-id example (from #8645)

dcousens:
This pull request adds `cuid2` identifier support as `db: { idField: { kind: 'cuid2' }`, and as part of that change removes null errors for unique filter values, allowing unique filters with more than 1 unique value.  This is to say that you can now uniquely query by _multiple_ unique values, instead of only one.

As part of extending support for `cuid2`, we have removed `idField` validation beyond the `dbField.type` (that is, whether it is a `String` or an `Int` or a `BigInt`).  This will allow developers to upgrade their `idField` types for new data - from `cuid` to `cuid2` in a non-breaking way - by not rejecting unknown identifier formats when querying.
Since validation has been removed, we can remove the default value for `kind: 'string'` (added in #8645), leaving users to always needing to provide that themselves. 

Before this pull request, the return type of `resolveInput.create` did not accept `id: string`, only `id: undefined`, this updates our types to allow that, as it was used in https://github.com/keystonejs/keystone/pull/8645.  However, as https://github.com/keystonejs/keystone/pull/8645 was missing the `: Lists` type for the example, that type error wasn't previously throwing. That has been fixed.

We can now drop the `cuid` dependency, but it is still being used by Prisma, so until we change our default, https://github.com/keystonejs/keystone/security/advisories/GHSA-5fp6-4xw3-xqq3 remains unresolved.


tests/api-tests/fields/types/calendarDay.test.ts
```diff
@@ -1,12 +1,20 @@
...
       orderableFilterTests(match, values, isNullable);
     });
-    uniqueEqualityFilterTest(calendarDay({ db: { isNullable }, isIndexed: 'unique' }), values);
+
+    uniqueEqualityFilterTest(
+      calendarDay({ db: { isNullable }, isIndexed: 'unique' }),
+      values,
+
+      // TODO: isNullable, failing for POSTGRES and MySQL
+      dbProvider === 'sqlite' ? isNullable : false
```
dcousens, line 17: Don't know why this is, but, I'll fix it in another pull request

borisno2 (approved):
LGTM

Issue #8663, opened 2023-07-03
Example pair custom fields are reset if not changed during update 


gathanase:
Add a 3-pair-json example field and a text field in a list. Update only the text field with the admin ui, the update will reset the pair field.

The pair field is not reset if not changed.

firefox
node v16.20.0

PR #8664, opened 2023-07-03
Fix custom pair fields examples (#8663)

gathanase:
- Fixes https://github.com/keystonejs/keystone/issues/8663

### DIFF
--- a/examples/custom-field/3-pair-field-json/index.ts
+++ b/examples/custom-field/3-pair-field-json/index.ts
@@ -44,6 +44,7 @@ export function pair<ListTypeInfo extends BaseListTypeInfo>(
   config: PairFieldConfig<ListTypeInfo> = {}
 ): FieldTypeFunc<ListTypeInfo> {
   function resolveInput(value: PairInput | null | undefined) {
+    if (value === undefined) return undefined;
     const { left = null, right = null } = value ?? {};
     return JSON.stringify({ left, right });
   }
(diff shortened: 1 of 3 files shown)


examples/custom-field/3-pair-field-json/index.ts
```diff
@@ -93,7 +93,9 @@ export function pair<ListTypeInfo extends BaseListTypeInfo>(
           arg: graphql.arg({ type: PairInput }),
           // eslint-disable-next-line @typescript-eslint/no-unused-vars
           resolve(value, context) {
-            return resolveInput(value);
+            if (value !== undefined) {
+              return resolveInput(value);
+            }
```
dcousens, line 98: This should probably be in the `resolveInput` function

gathanase: I applied your suggestion and it works great (I thought initially this would break the create operation).

examples/custom-field/3-pair-field-json/index.ts
```diff
@@ -44,6 +44,7 @@ export function pair<ListTypeInfo extends BaseListTypeInfo>(
   config: PairFieldConfig<ListTypeInfo> = {}
 ): FieldTypeFunc<ListTypeInfo> {
   function resolveInput(value: PairInput | null | undefined) {
+    if (!value) return undefined;
```
dcousens, line 47: @gathanase thanks for making the changes, but I think this might actually want to be specifically `if (value === undefined)`?
`null` is actually a value we might want to represent, whereas `undefined` means "unchanged".

gathanase: If the user want to put null in database then we should return null if value is null. In any case, I think you are right that we should only consider undefined.

dcousens (approved):

dcousens:
Thanks @gathanase! :yellow_heart:

PR #8698, opened 2023-07-15
fix(cli): regenerate prisma schema during `keystone prisma ..` commands

pahaz:
We also want to disable GrpahQL validations because it's not so important for the prisma command.

### DIFF
--- a/packages/core/src/scripts/prisma.ts
+++ b/packages/core/src/scripts/prisma.ts
@@ -19,7 +19,6 @@ export async function prisma(cwd: string, args: string[], frozen: boolean) {
   // TODO: this cannot be changed for now, circular dependency with getSystemPaths, getEsbuildConfig
   const config = getBuiltKeystoneConfiguration(cwd);
   const { graphQLSchema } = createSystem(config);
-  await validatePrismaAndGraphQLSchemas(cwd, config, graphQLSchema);
   await generateTypescriptTypesAndPrisma(cwd, config, graphQLSchema);
 
   const result = await execa('node', [require.resolve('prisma'), ...args], {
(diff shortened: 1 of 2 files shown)


pahaz:
Details here: https://github.com/keystonejs/keystone/discussions/8697

packages/core/src/scripts/prisma.ts
```diff
@@ -19,8 +20,15 @@ export async function prisma(cwd: string, args: string[], frozen: boolean) {
   // TODO: this cannot be changed for now, circular dependency with getSystemPaths, getEsbuildConfig
   const config = getBuiltKeystoneConfiguration(cwd);
   const { graphQLSchema } = createSystem(config);
-  await validatePrismaAndGraphQLSchemas(cwd, config, graphQLSchema);
-  await generateTypescriptTypesAndPrisma(cwd, config, graphQLSchema);
```
dcousens, line 23: Thinking about this with @borisno2, it is kind of weird that we did `generateTypescriptTypesAndPrisma` here at all. 

The `validatePrismaAndGraphQLSchemas` protects the developer in production by respecting the existing schemas, but then we go ahead and re-generate the types and Prisma, which is weird for production.

I don't think we should have run `generateTypescriptTypesAndPrisma` at all, and in many ways maybe this command shouldn't run that, and users should either just run `keystone dev` or `keystone build --no-ui` before they use `keystone prisma ...`.

Unfortunately, however you look at this, it is a breaking change, we do want to fix it @pahaz, but maybe the best workaround for you in the interim is to use `keystone build --no-ui` before your `migrate dev` command.

dcousens (approved):

PR #8706, opened 2023-07-20
Remove default description of timestamp field

wysher:
before:

before

---

after:
after

---

I find this additional info message in timestamp redundant, so I introduce way to disable it.
With createdAt it's shown only on create, but for updatedAt it's ALWAYS visible in ItemView.

I'm not sure about option name, please let me know if there's better one. :)

EDIT:
here's how it looks for updatedAt - message is always visible. In my case on sidebar.
updatedat

### DIFF
--- a/.changeset/remove-ui-message.md
+++ b/.changeset/remove-ui-message.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": minor
+---
+
+Removes default descrption for `timestamp` field when using `db.updatedAt`, use `ui.description` instead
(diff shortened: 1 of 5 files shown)


dcousens:
I understand the motivation behind this, as I don't like it either.

However, won't this be misleading if the field is manually set? Maybe if `db.updatedAt` is configured, the field should default to read-only too?

dcousens:
I honestly think, we should probably remove the message completely.
We shouldn't be differentiating the behaviour of this field for `db.updatedAt`, when the usual workflow for these considerations would be for the developer to add a `ui.description`.

borisno2 (approved):

PR #8707, opened 2023-07-20
Don't prevent navigation for `NODE_ENV === 'development'`

wysher:
During development I found prevent navigation really annoying. There are cases when I change some files many times before getting back to browser and then I need to confirm many times that I want to reload page (over a dozen or more).

So, here's a little proposal for disabling prevent navigation.

I'd rather like to set this with `config.ui.disablePreventNavigation`, but as far I could find, config is not passed to KeystoneProvider and it would probably need many changes. Therefore here's quick example using AdminConfig (which seems reasonable for me as well).

And if you like this feature, but need to set this somewhere else, I'm gonna need some guidance.

### DIFF
--- a/examples/custom-admin-ui-navigation/admin/config.ts
+++ b/examples/custom-admin-ui-navigation/admin/config.ts
@@ -1,5 +1,7 @@
 import { AdminConfig } from '@keystone-6/core/types';
 
+export const disablePreventNavigation = process.env.NODE_ENV === 'development';
+
 import { CustomNavigation } from './components/CustomNavigation';
 export const components: AdminConfig['components'] = {
   Navigation: CustomNavigation,
--- a/packages/core/src/admin-ui/utils/usePreventNavigation.tsx
+++ b/packages/core/src/admin-ui/utils/usePreventNavigation.tsx
@@ -1,11 +1,17 @@
 import { useEffect } from 'react';
 import { useRouter } from '../router';
+import { useKeystone } from '../context';
 
 export function usePreventNavigation(shouldPreventNavigationRef: { current: boolean }) {
   const router = useRouter();
+  const keystone = useKeystone();
+
+  const disablePreventNavigation = keystone.adminConfig?.disablePreventNavigation || false;
 
   useEffect(() => {
     const clientSideRouteChangeHandler = () => {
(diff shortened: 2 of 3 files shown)


borisno2:
Hi @wysher , thanks for taking to time to put together this PR. 

I am trying to replicate the issue you are talking about locally using `keystone dev` and am not seeing the same result. If I - Run `keystone dev`
- Open the Admin UI, open an item and make some edits without saving
- Make some changes in code, forcing a hot reload

My unsaved changes disappear without the dialogue box popping up, so no multiple popups happening.

I think if there is an issue we should look to solve that first.

wysher:
> Hi @wysher , thanks for taking to time to put together this PR.
> 
> I am trying to replicate the issue you are talking about locally using `keystone dev` and am not seeing the same result. If I - Run `keystone dev`
> 
> * Open the Admin UI, open an item and make some edits without saving
> * Make some changes in code, forcing a hot reload
> 
> My unsaved changes disappear without the dialogue box popping up, so no multiple popups happening.
> 
> I think if there is an issue with the popup we should look to solve that first.

Did you try to change any data field in item view? Changes on the code don't trigger prevent navigation hook.

Some examples I may find this useful:
- when I work with hooks.validateInput, which most of the times takes a while. During this, I need to do some changes to check item, inputData, etc. I heavily use cmd+s for saving. Then, when I made changes in code I'm happy with, my browser shows multiple prevent navigation message.
- I made some changes in item view, and forgot about them. Now I'm back at code, do many changes. And when I switch to browser again there's multiple alerts.

I must add than in some browsers, like mine there's **no way to close current tab if there's an alert**. Alert covers whole browser window. I cannot close or reload tab. I need to close every alert manually. If there's system animation (like in my case) it took a while.

After some further thought I think we can try to show alert only once. Because even there are further changes, clicking on first alert will reload whole page.
In this case, we don't even need additional setting.

dcousens:
@wysher what browser are you using? I cannot reproduce the "multiple prompts" aspect of your problem in Firefox or Chrome.
Only one prompt is shown, irrespective of how many fields are modified.

dcousens:
I don't think we want to support this, the problem seems to be highly specific to one browser's UX

Issue #8714, opened 2023-07-24
Field visible when `graphql.omit: true`


pc-erin:
### Steps to reproduce the problem

Create a text field.
Set `ui.itemView.fieldMode` to "hidden".
Set `graphql.omit.update` to `true`.

### What you expected to happen

The field is not visible.

### What actually happens

The field is still visible.

dcousens:
@pc-erin can you please attach a screenshot and your schema?
I cannot reproduce this behaviour.

tv42:
It seems the fix is wrong. See #9216.

PR #8715, opened 2023-07-25
Fix field visilibity in AdminUI when omitted

dcousens:
Fixes https://github.com/keystonejs/keystone/issues/8714, and switches to verbose snapshot testing of the possibility space in https://github.com/keystonejs/keystone/pull/8715/files#diff-c56d1cc3cff659fd4ab13d4370524a79120c964e3279953eeb0782fc79a48ca5, that way we can track visibility changes easily moving forward.


packages/core/src/lib/core/types-for-lists.ts
```diff
@@ -286,6 +286,31 @@ function getListsWithInitialisedFields(
...
+              : 'hidden', // fallback to read only if not omitted
+          },
+
+          listView: {
+            ...f.ui?.listView,
+            fieldMode: _isEnabled.read
+              ? f.ui?.listView?.fieldMode ?? f.ui?.listView?.defaultFieldMode
+              : 'hidden',
+          },
+        },
```
dcousens, line 313: I have moved this behaviour to `types-for-lists.ts`, so we can logically group this near to `isEnabled` - `types-for-lists` is not an appropriate name tbh

packages/core/src/lib/createSystem.ts
```diff
@@ -62,6 +62,8 @@ function getSudoGraphQLSchema(config: KeystoneConfig) {
   const lists = initialiseLists(transformedConfig);
   const adminMeta = createAdminMeta(transformedConfig, lists);
   return createGraphQLSchema(transformedConfig, lists, adminMeta, true);
+  // TODO: adminMeta not required
+  // return createGraphQLSchema(transformedConfig, lists, null, true);
```
dcousens, line 66: I can't think of a reason why a `sudo` context would have an `adminMeta`, but this would be a breaking change if someone was querying that using `getContext` or similar

borisno2:
Looks good, looks like a test is failing though

PR #8718, opened 2023-07-29
Fix field width of grouped fields

wysher:
Fields within groups have incorrect width, And may be different for each group depends on fields .

This PR fixes this issue.

Example:
groups-width

---
And here's another example, showing two groups with different widths. Before and after changes.

**Before:**
before

---
**After**
after


dcousens (approved):

dcousens:
Thanks @wysher, I have added a changeset too :yellow_heart:

PR #8721, opened 2023-07-31
Deprecate `@keystone-6/system` and `server.healthCheck`

dcousens:
This pull request deprecates `@keystone-6/system`, including each of the exports `createSystem`, `createExpressServer` and `initConfig`. Neither of these exports should be used by developers in new code, as we are slowly changing the way users interact with Keystone generally.  If you have a usecase for these functions, please comment on this pull request!

This pull request additionally deprecates `config.server.healthCheck`, with the new recommendation for developers to use `extendExpressApp` with a custom route that fits their use-case.   This configuration was highly opinionated and often unhelpful in different environments when you needed a very specific response format.
An implementation that wants to upgrade early could use the following:

```typescript
import { config } from '@keystone-6/core';
import { TypeInfo } from '.keystone/types';

  ...
  server: {
    extendExpressApp: (app, commonContext) => {
      app.get('/status', async (req, res) => {
        res.status(200).json({ status: 'up' })
      })
    },
  },
})
```

### DIFF
--- a/.changeset/less-systems-now.md
+++ b/.changeset/less-systems-now.md
@@ -2,4 +2,4 @@
 '@keystone-6/core': patch
 ---
 
-Deprecates `@keystone-6/system`, including `createSystem`, `createExpressServer` and `initConfig`
+Deprecates `@keystone-6/core/system`, including `createSystem`, `createExpressServer` and `initConfig`
--- a/.changeset/less-systems-now.md
+++ b/.changeset/less-systems-now.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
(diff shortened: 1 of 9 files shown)


.changeset/less-systems-now.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Deprecates `@keystone-6/system`, including `createSystem`, `createExpressServer` and `initConfig`
```
borisno2, line 5: Should this be `@keystone-6/core/system`?

borisno2 (approved):
Looks good!

PR #8722, opened 2023-07-31
Use console.error for Prisma errors

dcousens:
This pull request changes `KS_PRISMA_ERRORS` to now be logged using `console.error` on the server.
This isn't a breaking change in any sense, but it might be a change of scope for what users expect from these errors if they output sensitive information.

I am open  to feedback on if this is acceptable or not as a `minor`.

### DIFF
--- a/.changeset/more-errors-please.md
+++ b/.changeset/more-errors-please.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+`KS_PRISMA_ERRORS` are now logged with `console.error` on the server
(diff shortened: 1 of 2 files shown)


borisno2 (approved):

molomby:
Hmm.. I don't have strong opinions on whether this should be breaking or not, I'm more concerned with direct use of `console`. Can the `prismaError` function be replaced by devs somehow? If my Prisma errors do contain sensitive info, how do I prevent it being logged after this change? Do I need to redirect and filter the output of the entire process?

Most apps I've deployed prefer to always log JSON blobs (as Keystone 5 did by default with `pino`). When Keystone refs `console` directly, such has here, in the `start` and `migrate` scripts, etc. it makes this hard. I know it's a larger point but I'd love KS to accept a `logger` object, supporting the standard `log()`, `info()`, `error()`, etc. interface. It could default to `console`, just gives the dev control.

Basically, I say we should address #7668 before adding more output of any kind.

dcousens:
This is my thoughts too, and @borisno2 and I are working on that in the background, so I'm happy to park this for now.

dcousens:
Merging, but we need to address #7668 before release

PR #8725, opened 2023-07-31
Prefer `node:crypto` `randomBytes` with `base64url`

dcousens:
This pull request updates `@keystone-6/auth` and `@keystone-6/core` to use `node:crypto` and the `base64url` encoding instead of encoding to `base64`, or using the `safe-uid` dependency.

This is the recommended best practice and reduces our need on dependencies.

### DIFF
--- a/packages/auth/src/lib/createAuthToken.ts
+++ b/packages/auth/src/lib/createAuthToken.ts
@@ -1,15 +1,6 @@
 import { randomBytes } from 'crypto';
 import type { KeystoneDbAPI } from '@keystone-6/core/types';
 
-function generateToken(length: number): string {
-  return randomBytes(length) // Generates N*8 bits of data
-    .toString('base64') // Groups by 6-bits and encodes as ascii chars in [A-Za-z0-9+/] and '=' for padding (~8/6 * N chars)
-    .replace(/[^a-zA-Z0-9]/g, '') // Removes any '+', '/' (62, 63) and '=' chars as often require escaping (eg. in urls)
-    .slice(0, length); // Shortens the string, so we now have ~6*N bits of data (it's actually log2(62)*N = 5.954*N)
-}
-
-// TODO: Auth token mutations may leak user identities due to timing attacks :(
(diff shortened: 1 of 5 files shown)


packages/auth/src/lib/createAuthToken.ts
```diff
@@ -1,26 +1,20 @@
...
-    return { success: true, itemId: item.id, token: generateToken(20) };
-  } else {
-    return { success: false };
-  }
+  if (!item) return { success: false };
+
+  return {
+    success: true,
+    itemId: item.id,
+    token: randomBytes(16).toString('base64url').slice(0, 20) // (128 / Math.log2(64)) < 20
```
dcousens, line 18: Unlike the previous `generateToken`, this should now always be a string of length `20`, not variable length

packages/core/src/session/index.ts
```diff
@@ -142,7 +143,7 @@ export function storedSessions<Session>({
       return store.get(sessionId);
     },
     async start({ context, data }) {
-      const sessionId = uid(24);
+      const sessionId = randomBytes(24).toString('base64url'); // 192-bit
```
dcousens, line 146: I didn't want to change this, but it does seem arbitrary and I might up this to 256-bit in future

borisno2 (approved):

packages/auth/src/lib/createAuthToken.ts
```diff
@@ -1,26 +1,20 @@
 import { randomBytes } from 'crypto';
```
gautamsi, line 1: should this be `node:crypto` instead similar to https://github.com/keystonejs/keystone/pull/8725/files#diff-26c287c98ecdad9f09205169e711de7a0e467c876d50deb5f89fa9e58ac15005R1

dcousens: Yes, but that shouldn't be a problem either

gautamsi: just sanity check, no complain

PR #8726, opened 2023-07-31
Adds new `random` identifier type

dcousens:
This pull request adds a new random identifier type, with the following syntax:
```typescript
{
  kind: 'random'
  type?: 'String'
  bytes?: number // defaults to 32
  encoding?: 'hex' | 'base64url' // defaults to base64url
}
```

This can be used for your `db.idField` configuration, similar to the following:

```typescript
export const lists: Lists = {
  Person: list({
    access: allowAll,
    db: {
      idField: { kind: 'random', bytes: 31 },
    },
    ...
  })
}
```

Or as your new schema-wide default
```typescript
export default config({
  db: {
    ...
    idField: { kind: 'random', encoding: 'hex' },
  },
  ...
})
```

This new identifier type replaces the `cuid2` that was merged in https://github.com/keystonejs/keystone/pull/8648 (but not released). 
With this approach, we can easily add additional encodings in the future, including a `cuid2` encoding.

This new identifier takes advantage of [Prisma Client extensions](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions) to ensure that this identifier is automatically defaulted on `create`, and will work with Keystone's `context.prisma` calls.


packages/core/src/lib/createSystem.ts
```diff
@@ -66,6 +67,34 @@ function getSudoGraphQLSchema(config: KeystoneConfig) {
...
+      const { bytes, encoding } = dbField.default;
+      prismaClient = prismaClient.$extends({
+        query: {
+          [list.prisma.listKey]: {
+            async create({ model, args, query }: any) {
+              return query({
+                ...args,
+                data: {
+                  ...args.data,
+                  id: randomBytes(bytes).toString(encoding),
```
borisno2, line 85: Will this override any value for `id` that is given? Should we have the same behaviour as the other `default` types - if a value is provided use that?

packages/core/src/lib/createSystem.ts
```diff
@@ -82,7 +84,7 @@ function injectNewDefaults(prismaClient: any, lists: Record<string, InitialisedL
                 ...args,
                 data: {
                   ...args.data,
-                  id: randomBytes(bytes).toString(encoding),
+                  id: args.id ?? randomBytes(bytes).toString(encoding),
```
dcousens, line 87: `id` is provided as `undefined` by this call, thus

packages/core/src/lib/createSystem.ts
```diff
@@ -66,6 +67,36 @@ function getSudoGraphQLSchema(config: KeystoneConfig) {
...
+      const { bytes, encoding } = dbField.default;
+      prismaClient = prismaClient.$extends({
+        query: {
+          [list.prisma.listKey]: {
+            async create({ model, args, query }: any) {
+              return query({
+                ...args,
+                data: {
+                  ...args.data,
+                  id: args.data.id ?? randomBytes(bytes).toString(encoding),
```
dcousens, line 87: `id: undefined` by this point, thus

Issue #8732, opened 2023-08-02
`ui.isDisabled: true` is ignored


molomby:
According to [the docs](https://keystonejs.com/docs/config/config#ui), I can set `ui.isDisabled: true` to disable the Admin UI:

> `isDisabled` (default: `false`): If `isDisabled` is set to `true` then the Admin UI will be completely disabled.

But when `keystone start` us run, the CLI code passes `{ ui: true, ... }` in the second arg, [here](https://github.com/keystonejs/keystone/blob/main/packages/core/src/scripts/cli.ts#L92), which overrides the configured `isDisabled` value [here](https://github.com/keystonejs/keystone/blob/main/packages/core/src/scripts/start.ts#L53).

Related to #8046?

## Reproduction

* Create a project
* Set `ui: { isDisabled: true }`
* `keystone build`
* `keystone start`
* Get error:

```
$ keystone start
✨ Starting Keystone
✨ Connecting to the database
✨ Creating server
✅ GraphQL API ready
✨ Preparing Admin UI Next.js app
Error: Could not find a production build in the '/my-keystone-app/.keystone/admin/.next' directory. Try building your app with 'next build' before starting the production server. https://nextjs.org/docs/messages/production-start-no-build-id
    at NextNodeServer.getBuildId (/my-keystone-app/node_modules/next/dist/server/next-server.js:345:23)
    at new Server (/my-keystone-app/node_modules/next/dist/server/base-server.js:151:29)
    at new NextNodeServer (/my-keystone-app/node_modules/next/dist/server/next-server.js:172:9)
    at NextServer.createServer (/my-keystone-app/node_modules/next/dist/server/next.js:179:24)
    at async /my-keystone-app/node_modules/next/dist/server/next.js:200:31
    at async NextServer.prepare (/my-keystone-app/node_modules/next/dist/server/next.js:160:24)
    at async start (/my-keystone-app/node_modules/@keystone-6/core/scripts/cli/dist/keystone-6-core-scripts-cli.cjs.dev.js:1277:5)
```

## Expected

App should start without the Admin UI.

PR #8733, opened 2023-08-02
Fix `ui.isDisabled: true`

molomby:
Fix incorrect condition around in `keystone start` when looking to start the admin UI. See #8732.

Pretty sure this is what was intended.


dcousens (approved):
Thanks @molomby :yellow_heart:

PR #8738, opened 2023-08-04
Add sourcemap: true to esbuild configuration

TweededBadger:
Debugging typescript files does not currently work as sourcemap are not generated.


dcousens (approved):

dcousens:
Thanks @TweededBadger! :yellow_heart:

PR #8741, opened 2023-08-06
Fix broken context type narrowing for virtual field resolvers

### LINKED ISSUE #8740: Broken Virtual Field Type Narrowing
I'm not sure if this is a regression from the same source, but I _think_ this is related to #8616 and #8617.

TL;DR, using the `virtual` field with a narrowed type from `.keystone/types` results in a tsc compilation error when typescript strict mode is enabled

[Reproduction Repo](https://github.com/acburdine/keystone-virtualfield-brokentypes).

Example Code:

```ts
      notMyPostsCount: virtual<Lists.User.TypeInfo>({
        // field fails strict typing :(
        field: graphql.field({
          type: graphql.nonNull(graphql.Int),
          // without the context: Context, typing, context is untyped and not as useful
          // async resolve(item, args, context) {
          async resolve(item, _, context: Context) {
            return context.prisma.post.count({
              where: {
                authorId: { not: item.id }
              }
            });
          }
        })
      })
```

Typescript Error:

```
   schema.ts  1 
 │     Type 'Field<GetResult<{ id: string; name: string; email: string; password: string; createdAt: Date | null; }, unknown, never> & {}, {}, NonNullType<ScalarType<number>>, string, Context<...>>' is not assignable to type 'VirtualFieldGraphQLField<GetResult<{ id: string; name: string; email: string; password: string; createdAt: Date | null; }, unknown, never> & {}> | ((lists: Record<...>) => VirtualFieldGraphQLField<...>)'. typescript (2322) [63, 9]
 │        Type 'Field<GetResult<{ id: string; name: string; email: string; password: string; createdAt: Date | null; }, unknown, never> & {}, {}, NonNullType<ScalarType<number>>, string, Context<...>>' is not assignable to type 'VirtualFieldGraphQLField<GetResult<{ id: string; name: string; email: string; password: string; createdAt: Date | null; }, unknown, never> & {}>'. 
 │          Types of property '__context' are incompatible. 
 │            Type '(context: Context<any>) => void' is not assignable to type '(context: KeystoneContext<BaseKeystoneTypeInfo>) => void'. 
 │              Types of parameters 'context' and 'context' are incompatible. 
 │                Type 'KeystoneContext<BaseKeystoneTypeInfo>' is not assignable to type 'Context<any>'. 
 │                  Types of property 'db' are incompatible. 
 │                    Type 'KeystoneDbAPI<Record<string, BaseListTypeInfo>>' is missing the following properties from type 'KeystoneDbAPI<{ readonly User: TypeInfo<any>; readonly Post: TypeInfo<any>; readonly Tag: TypeInfo<any>; }>': User, Post, Tag
```

Ideally, specifying the type in `virtual<Lists.User.TypeInfo>` should be enough to allow `context` in the resolve function to have the correct typings.

Looking through the code for the `virtual` field type, I think this is fixable just by passing through another type - will have a PR shortly.

Version Info:
Node - v18.17.0
Typescript - v5.1.6
Keystone - latest (5.4.0)

acburdine:
closes #8740

`graphql.Field` supports the context type as a passthrough, so this PR passes through the inferred Context type.

Tested this via patch-package, it fixed the issue I had in #8740, though I'm not sure if there are other side-effects this type passthrough would have 🤔

### DIFF
--- a/.changeset/perfect-rats-cough.md
+++ b/.changeset/perfect-rats-cough.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fixes broken context type narrowing for virtual field resolvers
--- a/examples/virtual-field/README.md
+++ b/examples/virtual-field/README.md
@@ -1,6 +1,6 @@
 ## Feature Example - virtual field
 
-This project demonstrates how to add virtual fields to a Keystone list
+This project demonstrates how to add virtual fields to a Keystone list.
 It builds on the [Blog](../blog) starter project.
 
 ## Instructions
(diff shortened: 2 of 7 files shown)


dcousens (approved):

dcousens:
Thanks @acburdine, I added a broken example that should help prevent a regression too

PR #8758, opened 2023-08-11
Upgrade CI to node 20

molomby:
Node.js v20 is due to become the LTS release in October 2023, at which point we'll want to switch the KS project over. This kicks off that change over process, gives us an early indication of how the CI and tests go, etc.

I've only bumped the node version itself (where I could find it). Not sure if I've missed any refs or other packages also need to be upgraded.


.codesandbox/ci.json
```diff
@@ -1,5 +1,5 @@
 {
   "packages": ["packages/*", "design-system/packages/*"],
   "sandboxes": ["/tests/sandbox"],
-  "node": "18"
+  "node": "20"
```
borisno2, line 4: Code sandbox does not yet support node 20 https://codesandbox.io/docs/learn/sandboxes/ci should we keep this at 18 for now?

dcousens:
```suggestion
  "node": "18"
```

dcousens (approved):

PR #8771, opened 2023-08-15
Fix `ui.isAccessAllowed` when `undefined` to prevent access

dcousens:
This pull request fixes `ui.isAccessAllowed`'s behaviour when `undefined` to be the same as the default AdminUI middleware.

The default AdminUI middleware prevents access to the AdminUI if a `sessionStrategy` has been defined, and a `.session` does not exist.

This is not what happens for the `adminMeta` GraphQL query, which falls back on public access if `isAccessAllowed` is undefined.
This would not impact users of the `auth` package, or any users that have defined `isAccessAllowed`.

This may impact users who are relying on the behaviour of providing **only** a `sessionStrategy`, and expecting `adminMeta` to be inaccessible by the public.

### DIFF
--- a/.changeset/fix-admin-meta-access.md
+++ b/.changeset/fix-admin-meta-access.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fixes `ui.isAccessAllowed` when `undefined`, to prevent access to the `adminMeta` GraphQL query, akin to the behaviour for the default AdminUI `pageMiddleware`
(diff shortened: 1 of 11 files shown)


borisno2 (approved):

dcousens:
Maybe in the near future, we should drop the `isAccessAllowed` default, and instead update the types (or validation errors) to ensure that `isAccessAllowed` is _always_ defined if a `session` is defined.

PR #8788, opened 2023-08-29
Fix `keystone build` not exiting with versions of NextJS > 13.4.12

### LINKED ISSUE #8581: CLI build and prisma migrate commands never exit
Just create new Keystone app with `yarn create-keystone-app` then try to execute command `yarn build` -> 'keystone build'

The expected behaviour is to build production ready output in `.keystone/admin` and **exit process** when this will be successfully done. But now the process is never exited even when build is done successful which is very inconvenient because basically there is no chance to go to another command when we want to deploy application - build hang out forever without exit after success.

Screenshot 2023-05-18 at 08 17 29

node version 16.19.1 but I also tried with newest LTS 18
next version - I tried with 13.2.2, 13.2.3, 13.3, 13.4.2 

But I believe that this is no related to other similar issues with hang out build command with:` info  - Creating an optimized production build ...` because I saw that this is related to some next versions above 13.2.3 and I also had that issue but figured it out changing next version, so the message disappeared after build but still build process is not exited.

When I simply added `process.exit()` in `keystone-6-core-scripts-cli.cjs.prod.js` at the end of function build which is responsible for that, everything went as expected.

Also there is the same issue with `keystone prisma migrate` CLI commands.

### LINKED ISSUE #8760: Build command gets stuck
## Bug description
The `npm run build` command does not terminate and hangs indefinitely.

## Steps to reproduce
1. Run `npm run build` in the terminal.
2. Observe that the command does not finish and no error is thrown.

## Expected behavior
The `npm run build` command should complete successfully and produce an optimized production build.

## Screenshots
image

## Environment
- Node version: v18.17.1
- OS: Debian on WSL

borisno2:
The PR adds a `process.exit(0)` after the next build in the `keystone build` cli.
Fixes #8760
Fixes #8581

This is required due to the use of the undocumented `next/dist/build` see - https://github.com/vercel/next.js/issues/54494

### DIFF
--- a/packages/core/src/scripts/cli.ts
+++ b/packages/core/src/scripts/cli.ts
@@ -85,7 +85,8 @@ export async function cli(cwd: string, argv: string[]) {
   }
 
   if (command === 'build') {
-    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true }));
+    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true }))
+      .then(() => process.exit(0))
   }
 
   if (command === 'start') {
--- a/packages/core/src/scripts/cli.ts
+++ b/packages/core/src/scripts/cli.ts
@@ -85,8 +85,9 @@ export async function cli(cwd: string, argv: string[]) {
   }
 
   if (command === 'build') {
-    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true }))
-      .then(() => process.exit(0))
+    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true })).then(() =>
+      process.exit(0)
+    );
   }
 
   if (command === 'start') {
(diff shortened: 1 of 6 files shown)


dcousens (approved):

packages/core/src/scripts/cli.ts
```diff
@@ -85,7 +85,9 @@ export async function cli(cwd: string, argv: string[]) {
   }
 
   if (command === 'build') {
-    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true }));
+    return build(cwd, defaultFlags(flags, { frozen: false, prisma: true, ui: true })).then(() =>
+      process.exit(0)
+    );
```
dcousens, line 90: The `cli` function shouldn't call `process.exit(0)`, as that will impact on downstream users of the `cli` export.
This needs to be in `scripts/index.ts`.

I have fixed this in fca0f5ce9 as part of https://github.com/keystonejs/keystone/pull/8790 as it was causing my tests to fail.

Seyid-cmd:
When will be this release? I cannot wait to see my keystone app in production.

ttbarnes:
Nice work, can't wait for the next release 👍

PR #8790, opened 2023-08-30
Reduce number of dependencies

dcousens:
As the title suggests, this pull request reduces the number of dependencies used by `packages/core/` and our `tests/`. 
A number of listed dependencies were unused, only used for development, or otherwise probably unnecessary.


packages/core/src/lib/core/utils.ts
```diff
@@ -167,26 +166,6 @@ export function getDBFieldKeyForFieldOnMultiField(fieldKey: string, subField: st
   return `${fieldKey}_${subField}`;
 }
 
-// this whole thing exists because Prisma doesn't handle doing multiple writes on SQLite well
-// https://github.com/prisma/prisma/issues/2955
```
dcousens, line 171: Resolved https://github.com/prisma/prisma/issues/2955

borisno2 (approved):

packages/core/src/scripts/index.ts
```diff
@@ -1,7 +1,17 @@
 import { cli } from './cli';
 import { ExitError } from './utils';
 
-cli(process.cwd(), process.argv.slice(2)).catch(err => {
+async function main() {
+  const argv = process.argv.slice(2);
+
+  await cli(process.cwd(), argv);
+
+  if (argv[0] === 'build') {
```
dcousens, line 9:
```suggestion
  // WARNING: see https://github.com/keystonejs/keystone/pull/8788
  if (argv[0] === 'build') {
```

packages/core/src/lib/core/utils.ts
dcousens: I had to drop this from this pull request, the change isn't a straight forward for the other databases

PR #8801, opened 2023-09-08
Reword `sudo()` documentation to remove ambiguity

dcousens:
The existing documentation mentions that `sudo()` has a side effect of

> all filters enabled for subsequent API calls

It is my opinion that this is ambiguous as to what `enabled` means, it is referring to an `allowAll` policy, but that is not what a reasonable expectation might be, despite being contradictory to the leading text:

> with all access control disabled

This removes that ambiguity to minimize any confusion around this.


borisno2 (approved):
Looks good

PR #8808, opened 2023-09-18
Use required types internally for hooks

dcousens:
As the title suggests, this pull requests changes the types of the resolved hooks to always be `required` and easily and consistently differentiable.  The type verbosity has increased a little, but that has thankfully decreased the complexity of reading the hook types.

### DIFF
--- a/packages/core/src/lib/core/mutations/create-update.ts
+++ b/packages/core/src/lib/core/mutations/create-update.ts
@@ -380,7 +380,7 @@ async function resolveInputForCreateOrUpdate(
         // `hookArgs` based on the `operation` which will make `hookArgs.item`
         // be the right type for `originalItem` for the operation
         hookArgs.operation === 'create'
-          ? { ...hookArgs, item: updatedItem, originalItem: hookArgs.item }
+          ? { ...hookArgs, item: updatedItem, originalItem: undefined }
           : { ...hookArgs, item: updatedItem, originalItem: hookArgs.item }
       );
     },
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/core/mutations/create-update.ts
```diff
@@ -299,13 +299,10 @@ async function getResolvedData(
   resolvedData = Object.fromEntries(
     await Promise.all(
       Object.entries(list.fields).map(async ([fieldKey, field]) => {
-        if (field.hooks.resolveInput === undefined) return [fieldKey, resolvedData[fieldKey]];
```
dcousens, line 302: Replaced by `defaultFieldHooksResolveInput`

packages/core/src/lib/core/mutations/hooks.ts
```diff
@@ -1,43 +1,28 @@
...
 
-  // Field hooks
+  // field hooks
   const fieldsErrors: { error: Error; tag: string }[] = [];
   await Promise.all(
     Object.entries(list.fields).map(async ([fieldKey, field]) => {
       if (shouldRunFieldLevelHook(fieldKey)) {
         try {
-          await field.hooks[hookName]?.({ fieldKey, ...args });
+          await field.hooks[hookName]({ fieldKey, ...args } as any); // TODO: FIXME any
```
dcousens, line 25: I think `runSideEffectOnlyHook` will be removed soon, so this is less of a concern

borisno2 (approved):

Issue #8784, opened 2023-08-25
UI relationship search error for `id` fields of type `Int`


linus-ha:
0.  `nvm use 18` Using node 18 as required by some dependency
1. `yarn create keystone-app test-ks` as recommended on your website
2. `cd test-ks`
3. edit `keystone.ts`, so that the database config looks like this:
```js
    db: {
      provider: 'postgresql',
      url: "postgres://postgres:password@localhost/testDb?schema=public",
      enableLogging: false,
      useMigrations: false,
      idField: {
        type: "Int",
        kind: "autoincrement",
      },
    },
```
4. start a clean local PostgreSQL DB via docker  
    `docker run -it --rm -p 5432:5432 -e POSTGRES_PASSWORD=password postgres`
5. `yarn dev` start keystone
6. Open http://localhost:3000/ , create your user, create a post and try to search in the author relationship field by any non-numeric character, for example, be the user's name.
7. → Error, as searching for string values on an integer ID field makes no sense (see UI's GraphQL query in the browser dev tools)

Searching by label field when connecting items in relationship fields (for example, the `name` field of users in create-keystone-app) should be possible in the UI (i.e. `Linus`). When searching by int-convertible search teams, everything works as expected (i.e. `1`).

![grafik](https://github.com/keystonejs/keystone/assets/30652341/87ec6342-a065-4beb-8bde-2e03eda8f82c)
![grafik](https://github.com/keystonejs/keystone/assets/30652341/c12190b3-4dd9-4c6c-bf25-f38ddae68432)

- Firefox
- node -v → v18.16.0
- yarn -v → 1.22.19
- keystone → 5.5.1

`package.json`
```json
{
  "name": "keystone-app",
  "version": "1.0.2",
  "private": true,
  "scripts": {
    "dev": "keystone dev",
    "start": "keystone start",
    "build": "keystone build",
    "postinstall": "keystone build --no-ui --frozen"
  },
  "dependencies": {
    "@keystone-6/auth": "^7.0.0",
    "@keystone-6/core": "^5.0.0",
    "@keystone-6/fields-document": "^7.0.0",
    "typescript": "^4.9.5"
  }
}
```

```
yarn keystone prisma -v
yarn run v1.22.19
$ /home/linus/temp/test-ks/node_modules/.bin/keystone prisma -v
prisma                  : 4.16.2
@prisma/client          : 4.16.2
Current platform        : rhel-openssl-3.0.x
[...]

PR #8810, opened 2023-09-18
Fix `Input error: only a int can be passed to id filters` for AdminUI

dcousens:
- Fixes https://github.com/keystonejs/keystone/issues/8784

This pull request adds the `.type` to the `fieldMeta` in an `id-field`, such that we can determine what kind of type is applicable for the `equals` filter for an ID field.

### DIFF
--- a/.changeset/add-error-message.md
+++ b/.changeset/add-error-message.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": minor
+---
+
+Adds error messages for GraphQL errors on the List view in the AdminUI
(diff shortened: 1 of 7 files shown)


tests/api-tests/__snapshots__/admin-meta.test.ts.snap
```diff
@@ -0,0 +1,101 @@
...
+            },
+            {
+              "__typename": "KeystoneAdminUIFieldMeta",
+              "customViewsIndex": null,
+              "description": null,
+              "fieldMeta": {
+                "defaultValue": "",
+                "displayMode": "input",
+                "isNullable": false,
+                "shouldUseModeInsensitive": false,
```
borisno2, line 39: This will be `true` for Postgres

borisno2 (approved):

PR #8819, opened 2023-09-19
Fix types for validateInput hook arguments on update operations

acburdine:
looks like these were updated incorrectly in #8808. Checked the other types and these appeared to be the only incorrect ones.

didn't create an issue/repro case since it seems to be a pretty straightforward bug/typo, though if I need to create an issue/repro let me know and I can do that 😄 I did vet out this change in my project via patch-package.

### DIFF
--- a/examples/hooks/schema.ts
+++ b/examples/hooks/schema.ts
@@ -39,11 +39,30 @@ export const lists: Lists = {
       // we use isNonNull for these fields to enforce that they are always provided, and validated against a content filter
       title: text({
         validation: { isRequired: true },
-        graphql: { isNonNull: { create: true, update: true } },
+        graphql: {
+          isNonNull: {
+            create: true,
+            update: true
+          }
+        },
+
       }),
       content: text({
         validation: { isRequired: true },
-        graphql: { isNonNull: { create: true, update: true } },
+        graphql: {
+          isNonNull: {
+            create: true,
+            update: true
+          }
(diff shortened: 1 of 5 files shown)


dcousens:
Thanks @acburdine!
I have extended the `hooks` example to ensure we get an error for this next time.

dcousens (approved):

examples/hooks/schema.ts
```diff
@@ -106,9 +131,14 @@ export const lists: Lists = {
           return resolvedData;
         },
       },
-      validateInput: ({ context, inputData, addValidationError }) => {
+      validateInput: ({ context, operation, inputData, addValidationError }) => {
```
dcousens, line 134: My intent is to replace this soon with per-operation typed equivalent, equivalent to `validate.create`, `validate.update` and `validate.delete`, or similar.

examples/hooks/schema.ts
```diff
@@ -33,11 +39,30 @@ export const lists: Lists = {
...
+          isNonNull: {
+            create: true,
+            update: true
+          }
+        },
+      }),
+      feedback: text({
+        validation: { isRequired: true },
+        graphql: {
+          omit: {
```
dcousens, line 62:
```suggestion
          // we use omit for this field to enforce that validateInput needs to branch on the two different types of input
          omit: {
```

PR #8820, opened 2023-09-20
Fix Image field tests

dcousens:
These tests were disabled in https://github.com/keystonejs/keystone/pull/8304 as jest was exploding with a `Segmentation Fault`, this should help resolve that.


packages/core/src/lib/assets/createImagesContext.ts
```diff
@@ -1,13 +1,20 @@
-import { v4 as uuid } from 'uuid';
+import { randomBytes } from 'crypto';
+
+import { fileTypeFromBuffer } from 'file-type';
```
dcousens, line 3: `file-type` is ESM, which is problematic

`jest` is now passing, but we need the Keystone usage of `esbuild` to work now

package.json
```diff
@@ -97,6 +97,7 @@
           "target": "esnext"
         }
       ]
-    }
+    },
+    "transformIgnorePatterns": []
```
dcousens, line 101: This probably isn't the best pathway for this

packages/core/src/lib/assets/createImagesContext.ts
mmachatschek: Vercel/remix e.g. read the package.json of the project to check if `"type": "module"` is set. I guess we can read that and tell the esbuild pipeline to either export the `.keystone/config.js` in CJS or ESM style? 🤔 

https://github.com/keystonejs/keystone/blob/52337e3aa6a57c23c2f0756199a90058a6e5a2ac/packages/core/src/lib/esbuild.ts#L12

emmatown:
Solved in #9529

PR #8826, opened 2023-09-25
Add `beforeOperation.[create|update|delete]` and `afterOperation.[create|update|delete]` operation routing for list hooks

dcousens:
This pull request adds the `.[create|update|delete]` hooks for the `*Operation` list hooks, and adds the same for field hooks internally (but not externally, as that is a breaking change for fields).

This routing is preferred for the refined type guarantees it can offer your application without needing to branch the types yourself.

### DIFF
--- a/packages/core/src/lib/core/initialise-lists.ts
+++ b/packages/core/src/lib/core/initialise-lists.ts
@@ -48,9 +48,9 @@ export type InitialisedField = {
     cacheHint: CacheHint | undefined;
   };
   ui: {
-    label: string;
-    description: string;
-    views: string;
+    label: string | null;
+    description: string | null;
+    views: string | null;
     createView: {
       fieldMode: MaybeSessionFunction<'edit' | 'hidden', any>;
     };
@@ -348,9 +348,9 @@ function getListsWithInitialisedFields(
           },
         },
         ui: {
-          label: f.label ?? '',
-          description: f.ui?.description ?? '',
-          views: f.ui?.views ?? '',
+          label: f.label ?? null,
+          description: f.ui?.description ?? null,
+          views: f.ui?.views ?? null,
           createView: {
             fieldMode: _isEnabled.create ? fieldModes.create : 'hidden',
           },
--- a/packages/core/src/lib/core/initialise-lists.ts
+++ b/packages/core/src/lib/core/initialise-lists.ts
@@ -247,11 +247,22 @@ function parseFieldHooks(
   hooks: FieldHooks<BaseListTypeInfo>
 ): ResolvedFieldHooks<BaseListTypeInfo> {
   return {
-    resolveInput: hooks.resolveInput ?? defaultFieldHooksResolveInput,
+    resolveInput: {
+      create: hooks.resolveInput ?? defaultFieldHooksResolveInput,
+      update: hooks.resolveInput ?? defaultFieldHooksResolveInput,
+    },
     validateInput: hooks.validateInput ?? defaultOperationHook,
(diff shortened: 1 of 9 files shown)


packages/core/src/lib/core/initialise-lists.ts
```diff
@@ -48,17 +48,30 @@ export type InitialisedField = Omit<NextFieldType, 'dbField' | 'access' | 'graph
...
-};
+} & Pick<
+  NextFieldType,
+  | 'input'
+  | 'output'
+  | 'getAdminMeta'
+  | 'views'
+  | '__ksTelemetryFieldTypeName'
+  | 'extraOutputFields'
+  | 'unreferencedConcreteInterfaceImplementations'
```
dcousens, line 73: This will minimize the number of breaking changes moving forward and keeps us keenly aware of what else is being copied to this type.

packages/core/src/lib/core/initialise-lists.ts
```diff
@@ -48,17 +48,30 @@ export type InitialisedField = Omit<NextFieldType, 'dbField' | 'access' | 'graph
     cacheHint: CacheHint | undefined;
   };
   ui: {
+    label: string; // TODO: move to ui
```
dcousens, line 51:
```suggestion
    label: string;
```

packages/core/src/types/config/fields.ts
```diff
@@ -20,7 +20,7 @@ export type FilterOrderArgs<ListTypeInfo extends BaseListTypeInfo> = {
 export type CommonFieldConfig<ListTypeInfo extends BaseListTypeInfo> = {
   access?: FieldAccessControl<ListTypeInfo>;
   hooks?: FieldHooks<ListTypeInfo, ListTypeInfo['fields']>;
-  label?: string;
+  label?: string; // TODO: move to ui?
```
dcousens, line 23: Apparently `text({ label: 'a text field' })` is what we use for fields, but not for lists.
Strange

gautamsi:
what about `validateInput` and `resolveInput` for list and field. I hope the list and field both will have same/similar syntax.

dcousens:
@gautamsi `resolveInput` already has this syntax for lists, but not for fields.
Unfortunately for fields, it is a breaking change to add any of this to their types seeing as fields often wrap user configuration.

dcousens:
`validateInput` and `validateDelete`, is something I am playing with in https://github.com/keystonejs/keystone/tree/validate-hooks-routing - but it is not ready, and I am unclear how the types should work for users if you had:

```
validate: {
  create: ...
  update: ...
}
validateDelete: ...
```

I think probably only one version should be allowed, `validate*`, or `validate: { ... }`, not both.

dcousens:
Alternatively, I add `validateInput.[create | update]`, and `validateDelete.delete`, and then break that in the next major to merge into `validate.*`.

gautamsi:
why not `validate.create|update|delete`

borisno2 (approved):
Nice!

dcousens:
@gautamsi what do you do if users provide both?

```typescript
validate: {
  create: ...
  update: ...
  delete: ...
}
validateInput: ...
validateDelete: ...
```

That is the question I don't know the answer to, and thus it seems like it might be a breaking change.

gautamsi:
I guess throw error and have them provide only one, this is breaking changes so they should be able to work through

gautamsi:
Can you do that in upcoming breaking change ? It will streamline `[validateInput | resolveInput | beforeOperation | afterOperation].[create | update | delete]`

dcousens:
@gautamsi that is my intent, yes

PR #8816, opened 2023-09-19
Add Cloudinary example

dcousens:
This pull request is a prelude to https://github.com/keystonejs/keystone/pull/8735, which hopefully will help @jason-c-child extend the example to demonstrate how the transforms are used.

PR #8836, opened 2023-09-26
Tidy up Cloudinary package and add missing changeset

dcousens:
https://github.com/keystonejs/keystone/pull/8816 made a patch change to the `@keystone-6/cloudinary` package, but a `.changeset/` entry was not added.

This pull request adds the missing entry and additionally inlines the `cloudinary.ts` adapter into the field, reducing the complexity of the package.


packages/cloudinary/src/index.ts
```diff
@@ -106,21 +105,20 @@ export const outputType: graphql.ObjectType<CloudinaryImage_File> =
     },
   });
 
-export const cloudinaryImage =
-  <ListTypeInfo extends BaseListTypeInfo>({
-    cloudinary,
-    ...config
-  }: CloudinaryImageFieldConfig<ListTypeInfo>): FieldTypeFunc<ListTypeInfo> =>
-  meta => {
+// TODO: no delete support
```
dcousens, line 108: This field implicitly preserves the images that are uploaded, as mentioned in https://github.com/keystonejs/keystone/pull/8816#issuecomment-1724786029, we should align this field with the storage approaches we have elsewhere - or at least revisit that design

borisno2 (approved):

PR #8839, opened 2023-09-26
Remove unsafe types from `examples/extend-express-app` example

dcousens:
This pull request removes the `(req as any).context` types from the `extend-express-app` example, and instead prefers to pass a refined `Context` directly to the route functions.

This new pattern reduces the number of `req as` casts, while still keeping your root express router definition as minimal as possible.

### DIFF
--- a/examples/extend-express-app/keystone.ts
+++ b/examples/extend-express-app/keystone.ts
@@ -1,8 +1,19 @@
 import { config } from '@keystone-6/core';
+import type { Request, Response } from 'express';
+
 import { fixPrismaPath } from '../example-utils';
 import { lists } from './schema';
 import { getTasks } from './routes/tasks';
-import { TypeInfo } from '.keystone/types';
+import { TypeInfo, Context } from '.keystone/types';
+
+function withContext<F extends (req: Request, res: Response, context: Context) => void>(
+  commonContext: Context,
+  f: F
+) {
+  return async (req: Request, res: Response) => {
+    return f(req, res, await commonContext.withRequest(req, res));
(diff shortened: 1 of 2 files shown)


examples/extend-express-app/keystone.ts
```diff
@@ -23,18 +34,8 @@ export default config<TypeInfo>({
         Keystone schema and return the results as JSON
     */
     extendExpressApp: (app, commonContext) => {
-      app.use('/rest', async (req, res, next) => {
```
dcousens, line 26: We don't want to call `withRequest` for other HTTP requests, except for `GET /rest/tasks`, thereby we shouldn't use `app.use`.

examples/extend-express-app/routes/tasks.ts
```diff
@@ -11,10 +11,7 @@ import type { Context } from '.keystone/types';
   We're also demonstrating how you can query related data through the schema.
 */
 
-export async function getTasks(req: Request, res: Response) {
-  // This was added by the context middleware in ../keystone.ts
-  const { context } = req as typeof req & { context: Context };
```
dcousens, line 16: We don't want to encourage `as` in user code, the preferred approach is to use a `Context` parameter directly

dcousens:
In the future, we might recommend removal of `extendExpressApp`, and instead, for anyone who wants to use `express` themselves, they could opt for an approach similar to how we recommend using `astro`, or `nextjs` or or or.

That needs feedback from the community though.

borisno2 (approved):

PR #8847, opened 2023-10-04
WIP/POC support prisma 5.4 serverless driver adapters

mmachatschek:
This is a WIP and more like a POC.


mmachatschek:
@dcousens I'll rebase this onto the `major` branch when that one has the latest changes merged

packages/core/src/lib/createSystem.ts
```diff
@@ -108,6 +108,8 @@ export function createSystem(config: KeystoneConfig) {
     adminMeta,
     getKeystone: (prismaModule: PrismaModule) => {
       const prePrismaClient = new prismaModule.PrismaClient({
+        adapter:
+          typeof config.db.prismaAdapter === 'function' ? config.db.prismaAdapter() : undefined,
```
dcousens, line 112: Maybe we should support a custom `new prismaModule.PrismaClient` instead

dcousens: If users did that, we could either pass the `datasources` and `log` parameters to them, or alternatively prevent them through a sum type.

mmachatschek: @dcousens yes that makes totally sense 👍 will do that.

That also makes it possible to update the prisma client so users can use their own prisma client extensions (already tried to do that on a project but only could make it work with the not recommended Prisma Middleware approach)

Updating the PR asap 👍

mmachatschek: @dcousens could you check if this is going the desired direction?

https://github.com/keystonejs/keystone/pull/8847/commits/2d12d6bd182def439c9d8bde9d6c79ea8a4fbb98

dcousens: LGTM @mmachatschek, I'll probably refine and bike shed, but if you wanted to ensure this is working for your usecases, I'm happy to help push it across the line for the next major.

mmachatschek: @dcousens would be great if you could push that across the line. I can assist in any way. I haven't dived into the testing stuff of this repo. Happy to help in any direction

examples/custom-prisma-driver-adapter/README.md
```diff
@@ -0,0 +1,28 @@
...
+This will start the Admin UI at [localhost:3000](http://localhost:3000).
+You can use the Admin UI to create items in your database.
+
+You can also access a GraphQL Playground at [localhost:3000/api/graphql](http://localhost:3000/api/graphql), which allows you to directly run GraphQL queries and mutations.
+
+Congratulations, you're now up and running with Keystone! 🚀
+
+## Try it out in CodeSandbox 🧪
+
+You can play with this example online in a web browser using the free [codesandbox.io](https://codesandbox.io/) service. To launch this example, open the URL <https://githubbox.com/keystonejs/keystone/tree/main/examples/extend-prisma-schema>. You can also fork this sandbox to make your own changes.
```
mmachatschek, line 28: This whole directory (custom-prisma-driver-adapter) is the copy of the `extend-prisma-schema` example.

This could also be the start of showing an example with prisma client extensions like https://github.com/prisma/extension-read-replicas

would be super beneficial for users

packages/core/src/types/config/index.ts
```diff
@@ -139,18 +139,32 @@ type PrismaLogDefinition = {
...
   shadowDatabaseUrl?: string;
   onConnect?: (args: KeystoneContext<TypeInfo>) => Promise<void>;
-  enableLogging?: boolean | PrismaLogLevel | Array<PrismaLogLevel | PrismaLogDefinition>;
+  enableLogging?: boolean | Array<PrismaLogLevel | PrismaLogDefinition>;
   idField?: IdFieldConfig;
   prismaClientPath?: string;
   prismaSchemaPath?: string;
 
   extendPrismaSchema?: (schema: string) => string;
+  prismaClient?: (config: PrismaClientConfig) => any;
```
mmachatschek, line 167: I know that the core doesn't pull in the `@prisma/client` for various reasons, I don't know if its desired to have this type here.

packages/core/src/lib/createSystem.ts
```diff
@@ -107,15 +107,19 @@ export function createSystem(config: KeystoneConfig) {
...
           config.db.enableLogging === true
             ? ['query']
             : config.db.enableLogging === false
             ? undefined
             : config.db.enableLogging,
-      });
+      };
+
+      const prePrismaClient = config.db.prismaClient
+        ? config.db.prismaClient(clientConfig)
```
mmachatschek, line 121: theoretically we could do `instanceof prismaModule.PrismaClient` 🤔 this should probably have the correct instance of the prisma client which the end user is using

examples/custom-prisma-driver-adapter/README.md
dcousens: I am thinking we might use Prisma extensions for hooks in the near future, so it does have me thinking about how this functionality might impact the order of operations there.

As is, if we supported the custom `PrismaClient`, Keystone would be `.extend`ing the custom client, which I think is ok?

packages/core/src/types/config/index.ts
dcousens: The problem is that some code bases might have many different `@prisma/client`'s, so you can't rely on that type.

packages/core/src/lib/createSystem.ts
dcousens: I don't think we need to do that, we are relying on Typescript in many scenarios, I don't think we need to introduce a run-time burden for this.

I am open to adding something like `zod` for the configuration at some point, but, Typescript is the priority.

examples/custom-prisma-driver-adapter/README.md
mmachatschek: > I am thinking we might use Prisma extensions for hooks in the near future, so it does have me thinking about how this functionality might impact the order of operations there.

I love this idea would make the usage of `context.prisma` over `context.db` much easier if you have custom logic in list hooks

packages/core/src/lib/createSystem.ts
```diff
@@ -107,15 +107,19 @@ export function createSystem(config: KeystoneConfig) {
     graphQLSchema,
     adminMeta,
     getKeystone: (prismaModule: PrismaModule) => {
```
borisno2, line 109: This is great to see @mmachatschek thanks!!

I have been playing around with the new Prisma Planetscale Serverless driver with Keystone `getContext`. I have got it working here https://github.com/borisno2/on-the-hill-drama-club/pull/189 I wonder is we can also make this usecase easier or less confusing as to which "Prisma Client" would actually be used in that instance. I wonder if we can work towards maybe `prismaModule` optional here in `getKeystone` and by extension `getContext`? 

From a DX perspective it feels strange to pass in the Prisma Client twice, and I also found it unnecessarily complex to pass in an customised Prisma Client into `getContext`.

dcousens: Oh that's a good point, `getContext` requiring a `PrismaModule` is not reasonable with this change

mmachatschek: @borisno2 @dcousens  I pushed a commit to attempt to remove passing in the prismaModule into the `getKeystone` and `getContext` functions

https://github.com/keystonejs/keystone/pull/8847/commits/9fe430c2a8fbfc308b8860352684d4ca733685db

We now accept a `prismaClientModule` function that returns a `PrismaClient` instance and the `Prisma` object. 
If that function (`prismaClientModule`) exists, we can assume that the user passes in a prisma client and don't need to instantiate and import it anymore.

dcousens:
@mmachatschek are you interested in rebasing this?

mmachatschek:
@dcousens I'm currently unable to polish/rebase this PR to a state where it would be mergeable.

Prisma works with 5.11 in our huge codebase. We have tested the prisma serverless adapter which also works. We only needed to add the workaround that @borisno2 added to his project https://github.com/keystonejs/keystone/pull/8847#discussion_r1349602479

You can close the PR if you want to or pick up the changes I made. Feel free to decide how you want to move forward. I'm happy with any decision

PR #8864, opened 2023-10-21
chore: update workspace version constraints

mmachatschek:
Updates the version constraints of all workspaces (except examples folder). The version constraints with workspaces are supported

When publishing packages that contain the `workspace:` key in the version constraint, pnpm will automatically inject the correct version of the workspace package.
https://pnpm.io/workspaces#publishing-workspace-packages

I didn't update the examples with the `workspace:` prefix as they should be easily copiable.


packages/core/package.json
```diff
@@ -248,7 +248,6 @@
     "uuid": "^9.0.0"
   },
   "devDependencies": {
-    "@keystone-6/core": "5.7.0",
```
mmachatschek, line 251: The package.json was self referencing the `@keystone-6/core` package. I remove this one as it seems this was mistakenly added.

dcousens:
We use the following to publish

https://github.com/keystonejs/keystone/blob/41276b6877c0c1e9332d8c3e37ccd772a5b2c48b/.github/workflows/publish.yml#L26

Are we guaranteed that `changesets` will use `pnpm`? 🤔

mmachatschek:
@dcousens changesets is still necessary to detect which packages need a patch/minor/major bump. So that functionality stays.

https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish

the `changeset` cli will automatically detect that its running pnpm (same as it does now in the publishing process) and calls the `pnpm publish` command under the hood.  That `pnpm publish` command will automatically replace the `workspace:` version constraint from the package.json while publishing.

The `workspace:` version constraint just makes local development and versioning easier.

For publishing e.g. using `workspace:*` will not lead to updates like this: https://github.com/keystonejs/keystone/pull/8524/files#diff-3124b4b520bd0b5947ccd314a4b1f28ff07b2813f76af8ba2efdbf54631ccd94

mmachatschek:
marking this as draft as tests are failing that I need to debug

packages/core/package.json
mmachatschek: this is needed for the tests to run through. The scripts are self referencing `@keystone-6/core/package.json`. readding it

examples/cloudinary/package.json
```diff
@@ -10,7 +10,7 @@
     "postinstall": "keystone postinstall"
   },
   "dependencies": {
-    "@keystone-6/cloudinary": "workspace:^",
```
dcousens, line 13: For posterity
> I didn't update the examples with the workspace: prefix as they should be easily copiable.

Agreed 💙

dcousens (approved):

PR #8876, opened 2023-10-26
Update db-items.md

PaulAroo:
I am in the process of seeding data and using the syntax specified in the docs for the `createOne` method of the DB API  throws an error
Here is what my code looks like
```ts
await keystoneContext.db.Product.createOne({
    data: {
        ...product,
        photo: { create: [{ altText: product.description }] },
    },
})
```
Error ```GraphQLError: Field "0" is not defined by type "<certain type generated in the graphql schema>"```

but this 👇🏽 fixes it
```ts
// ........
        photo: { create: { altText: product.description } },
// ..........
```

### DIFF
--- a/docs/pages/docs/context/db-items.md
+++ b/docs/pages/docs/context/db-items.md
@@ -61,7 +61,7 @@ const count = await context.db.User.count({
 const user = await context.db.User.createOne({
   data: {
     name: 'Alice',
-    posts: { create: [{ title: 'My first post' }] },
+    posts: { create: { title: 'My first post' } },
   },
 });
 ```


mmachatschek:
@PaulAroo The docs are correct, what you are missing here is that in the docs, the posts relation is a multi relation, therefore you pass in an array of create objects whereas in your example (assumption made by looking at your relation key `photo`) you have a single relation on that model so instead you only pass in a single create object.

The graphql schema and the generated TypeScript types should reflect that

mmachatschek:
I just tested this with the `./examples/usecase-todo` example which didn't pass in the generated types to the `getContext`

This is what it looks like when the `getContext()` doesn't know of the generated types by keystone:

image


on multi relations both work (passing in a create as array or as normal object)
Single relations don't allow passing in arrays:

image

PaulAroo:
@mmachatschek , interesting... is `TypeInfo` generated automatically??... I don't have that on my end
image

mmachatschek:
its automatically generated in `node_modules/.keystone/types` by default. You can change the output path:

https://github.com/keystonejs/keystone/blob/ac35547c1cfc396838c613a56bc4573fb7f71622/examples/custom-output-paths/keystone.ts#L11

PaulAroo:
Oh... I see
Thanks @mmachatschek for pointing that out, the type checking works now

here is what it looks like in my own use-case
```ts
export default withAuth(
  config<TypeInfo>({
    db: {
       provider: "postgresql",
       url: process.env.DATABASE_URL || "",
       async onConnect(keystoneContext) {
	   if (process.argv.includes("--seed-data")) {
	      await insertSeedData(keystoneContext)
	    }
        }
    },
    lists
  })
)
// ...................................  //

async function insertSeedData(keystoneContext: KeystoneContext<TypeInfo>) {}
```

> It may be worth creating a pull request that adds the `TypeInfo` type

Yeah...  that should go into the [System Configuration](https://keystonejs.com/docs/config/config#title) part of the docs I guess

dcousens:
See https://github.com/keystonejs/keystone/pull/9016 for resolution of the examples
@PaulAroo if you wanted to update the documentation, that would really be appreciated :yellow_heart:

Issue #8888, opened 2023-11-06
keystone prisma migrate dev broken in latest version


yorkeJohn:
Keystone version: 5.7.0

Steps to reproduce:
1. Run `npx keystone prisma migrate dev` from a local terminal
2. See error message "Error: Prisma Migrate has detected that the environment is non-interactive, which is not supported."
3. Run `npx keystone prisma migrate reset --force` to reset the database
4. Run `npx keystone prisma migrate dev` again. When prompted "Enter a name for the new migration", no input can be made
5. The only way to exit the command is to press ctrl+C

Expected behaviour:
1. Run `npx keystone prisma migrate dev` from a local terminal
2. Get prompted to enter a name for the migration
3. A new migration is generated

Workaround:
Downgrade to Keystone 5.5.1 (might be able to go newer, but this was the working version that I had in another project)

stuible:
I imagine this has the same cause as my issue: https://github.com/keystonejs/keystone/issues/8858

dcousens:
Fixed in #8898, will be published soon

yorkeJohn:
When is this targeted for release?

Issue #8858, opened 2023-10-12
Unable to run "keystone prisma migrate dev" from Turborepo monorepo


stuible:
# How to reproduce the bug
### 1. I created this repo with Turbo Repo's starter project by running:

```sh
npx create-turbo -e basic
```

### 2. I then added Keystone CMS to `~/apps/keystone` using

```sh
cd apps
npm init keystone-app@latest keystone
```

### 3. I then modified the keystone package.json to have migrate script:

```json
"migrate": "NODE_ENV=development keystone prisma migrate dev"
```

### 4. Move to keystone directory and run keystone command
```sh
cd apps/keystone # move to keystone directory
npm run migrate # run our new migrate npm command
```

### 5. Receive the following output:
```
Error: Prisma Migrate has detected that the environment is non-interactive, which is not supported.
```

# Expected Result
I expect the migrations to be created through an interactive terminal

# Context
- Node v18.18.0
- [Turborepo](https://turbo.build/)

# Repro Repo
https://github.com/stuible/keystone-interactive-cli-bug

Here is a MVP repo that reproduces the issue described above.

austinbravodev:
I'm not using Turbo and the `keystone prisma migrate dev` command fails with the same error when run as an NPM script or via `npx`.

flippinjoe:
This is actually a bug in turborepo.  They swallow stdin (used for automatic migrations in keystone) https://github.com/vercel/turbo/issues/1235

austinbravodev:
Could be both - I think the Keystone CLI creates a non-interactive environment in which the Prisma command runs.

I'm just running it in my terminal, which should be interactive.

This would also explain why `npx prisma migrate dev` works.

stuible:
> This is actually a bug in turborepo. They swallow stdin (used for automatic migrations in keystone) [vercel/turbo#1235](https://github.com/vercel/turbo/issues/1235)

I have my doubts that this is a Turborepo issue because in the reproduction steps above I am bypassing the use of Turborepo completely (as far as I know).

SagnikPradhan:
Weird, `yarn keystone prisma migrate dev` freezes but `yarn dlx prisma@4.x migrate dev` worked flawlessly

stuible:
> Weird, `yarn keystone prisma migrate dev` freezes but `yarn dlx prisma@4.x migrate dev` worked flawlessly

Exactly my experience as well (with the equivalent  npx commands)

dcousens:
[...]

PR #8898, opened 2023-11-10
Fix prisma migrate cli regression error 

marekryb:
PR https://github.com/keystonejs/keystone/pull/8790 among many things replaced `execa` with `execFile` as a way to spawn child process for prisma cli commands.

Unfortunately `execFile` does not allow inheriting stdin from parent process like before https://github.com/keystonejs/keystone/blob/f92aa4010995d18ac15df83b70cbe9140bb72a7f/packages/core/src/scripts/prisma.ts#L27

Prisma commands that require user input (eg. `prisma migrate dev` or `prisma migrate reset`) now throw `Error: Prisma Migrate has detected that the environment is non-interactive, which is not supported.` and there is no flag or option to turn it off (https://github.com/prisma/prisma/issues/4669 https://github.com/prisma/prisma/issues/7113)

Related issues: https://github.com/keystonejs/keystone/issues/8888 https://github.com/keystonejs/keystone/issues/8858

There are at least 3 ways to fix this issue:
1) Revert code to old version (`execa`)
2) Switch to `execFileSync`
3) Switch to `spawn`

I chose the last one. In my limited testing everything seems to work fine. Let me know if other way is preferred instead.

### DIFF
--- a/.changeset/polite-llamas-sparkle.md
+++ b/.changeset/polite-llamas-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix prisma migrate non-interactive environment error
--- a/packages/core/src/scripts/prisma.ts
+++ b/packages/core/src/scripts/prisma.ts
@@ -1,4 +1,4 @@
-import { execFile } from 'node:child_process';
+import { spawn } from 'node:child_process';
 import esbuild from 'esbuild';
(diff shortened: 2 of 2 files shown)


dcousens:
Thanks for this @marekryb, I'm wondering what we can do to ensure tests capture this regression

dcousens (approved):

PR #8910, opened 2023-11-14
Drop `eslint` and `prettier` from CI, update Contributor Covenant Code of Conduct

dcousens:
The linting tooling is not helping the productivity of this project at this time, and as the primary maintainer, I am finding myself spending more time fighting the "errors" than I want to be. As part of this pull request, I am removing any linting tooling as part of our continuous integration pipeline.

This pull request additionally bumps the Contributor Covenant Code of Conduct to version 2.1.

Lastly, this pull request uses `eslint --fix` to apply a number of project-wide rules.
The rules have been applied atomically in their commits, to allow for easy commit verification and review if needed.

I'll still be using linting locally, so we can expect a return of an `.eslintrc` or similar in time, but I'll probably opt for something less opinionated.


.github/workflows/tests_ci_windows.yml
```diff
@@ -8,16 +8,6 @@ concurrency:
...
 jobs:
-  linting:
-    name: Linting
-    runs-on: windows-latest
-    steps:
-      - uses: actions/checkout@main
-      - uses: ./.github/actions/ci-setup
-
-      - name: TypeScript
-        run: pnpm lint:types
```
dcousens, line 19: We don't need to check typescript on Windows

dcousens:
The root `babel.config.js` is used by preconstruct, that has been restored

.gitignore
```diff
@@ -1,15 +1,11 @@
-.env
-.next
```
borisno2, line 2: Might want to keep `.next` for `docs`?

borisno2 (approved):

PR #8941, opened 2023-12-03
fix: fix readme

duidae:
there are incorrect instructions `pnpm` in some of those examples, so this PR aims to fix that. (should be `pnpm install`)


dcousens:
Thanks @duidae

PR #8966, opened 2023-12-15
Add search by relationship fields

pnodet:
Allow admin users inside a list page to search by relations text fields


packages/core/src/fields/types/relationship/views/RelationshipSelect.tsx
```diff
@@ -88,6 +88,21 @@ export function useFilter (search: string, list: ListMeta, searchFields: string[
       })
     }
 
+		for (const fieldKey of relationsSearchFields) {
+      const [relation, relationsSearchField] = fieldKey.split('.')
+      const field = list.fields[relation]
+
+      conditions.push({
+        [field.path]: {
+          some:{
```
gautamsi, line 97: `some` is only for one to many relationship, I believe this does not work for 1:1 relationship.

pnodet: Indeed, will think of something

pnodet:
Hi @dcousens I see some tests failing but some are due to CI setup failure I think, not sure what to do tbh

dcousens:
Thanks @pnodet, I'll check that out as soon as I can

pnodet:
HNY @dcousens  thanks for the formatting commit!
I this PR something you are interested in? do you think this will get merged or I better publish a npm fork?

pnodet:
@dcousens from the 58 tests the failing one seem to be related to tsx, not sure how to fix tbh

```
> @keystone-6/example-testing@0.0.8 test /home/runner/work/keystone/keystone/examples/testing
> node --loader tsx example-test.ts


node:internal/process/esm_loader:40
      internalBinding('errors').triggerUncaughtException(
                                ^
Error: tsx must be loaded with --import instead of --loader
The --loader flag was deprecated in Node v20.6.0
```

dcousens:
@pnodet happy new years!
I'm not convinced this will be intuitive and could result in unexpected performance costs in some usecases.
Maybe the user can opt-in to the behaviour?

pnodet:
@dcousens

By user you mean end-user (ie: a checkbox with label Activate search by relation fields) or keystone user the developers in the schema for example?

dcousens:
sorry, I mean developer

pnodet:
Sure, for the developer we could have something in the schema / list. Inside the relationship field there is already a `isFilterable` and `isOrderable` properties if I'm not mistaken. I'm not an expert and I don't quite remember how we use them but I imagine we could have another field like `isSearchable` for exemple, defaults to `false` for backward compatibility.

pnodet:
Hi @dcousens What do you have in mind? anything I can do to get this PR merged?

packages/core/src/fields/types/relationship/views/RelationshipSelect.tsx
```diff
@@ -107,6 +112,31 @@ export function useFilter (value: string, list: ListMeta, searchFields: string[]
...
 
+    for (const { field, refSearchFields, many } of relationsSearchFields) {
+      conditions.push(
+        ...refSearchFields.map((refSearchField) =>
+          many
+            ? {
+                [field]: {
+                  some: {
+                    [refSearchField]: {
+                      contains: trimmedSearch,
```
gautamsi, line 123: these fields having `contains` only works with the `text` field. This should be documented.

pnodet:
@dcousens rebased main to keep the branch up to date

packages/core/src/fields/types/relationship/views/RelationshipSelect.tsx
dcousens: @pnodet can you update the documentation to mention that fields as listed in `searchFields` need to support the `contains` filter?

dcousens: Actually @gautamsi, we should add that as a different pull request, this applies to `searchFields` generally

packages/core/src/fields/types/relationship/views/RelationshipSelect.tsx
```diff
@@ -99,6 +98,40 @@ export function useFilter (value: string, list: ListMeta, searchFields: string[]
 
     for (const fieldKey of searchFields) {
       const field = list.fields[fieldKey]
+
+      // @ts-expect-error TODO: fix fieldMeta type for relationship fields
+      if (field.fieldMeta?.refSearchFields) {
```
dcousens, line 103: @pnodet I merged this into the pre-exitsing `searchFields` query building code, less type passing

henribruvier:
This would be a nice feature !!

dcousens:
Replaced by https://github.com/keystonejs/keystone/pull/8966

pnodet:
Edit: Replaced by https://github.com/keystonejs/keystone/pull/9401

PR #8992, opened 2024-01-22
docs: Added twitter link in initPage

ahmedreza1:
Not a big one but the init page was missing Keystone twitter url, the icon was there but not clickable like the github icon, so added it.


dcousens (approved):

dcousens:
Thanks @ahmedreza1

PR #8995, opened 2024-01-25
feat(core/scripts): expose express and http server from start command

iamandrewluca:
This PR exposes the internal `express.Express` and `http.Server` instances to be handled imperatively.

Sometimes, there is a need to start/stop (production) Keystone manually using `@keystone-6/core/scripts/cli`

The reasons for this would be a Graceful Shutdown or maybe adding a Kubernetes management server.

Even if Graceful Shutdown is possible in [`server.extendHttpServer`](https://keystonejs.com/docs/config/config#extend-http-server), for K8 Management becomes more tricky.

**custom-start.js**

```ts
import { cli } from "@keystone-6/core/scripts/cli";
import express from "express";

// Kubernetes management
const kubernetesManagementExpress = express();
kubernetesManagementExpress.use(configureLivenessReadinessAndStartupProbes);
const kubernetesManagementServer = kubernetesManagementExpress.listen(3001);

kubernetesManagementServer.on("listening", async () => {
  const start = await cli(process.cwd(), ["start"]);

  // This will always run
  if (start && typeof start !== "function") {
    const { expressServer, httpServer } = start;

    // Graceful shutdown
    process.on("SIGTERM", () =>
      httpServer.close(() => {
        kubernetesManagementServer.close();
      })
    );
  }
});

// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
function configureLivenessReadinessAndStartupProbes() { /* ... */ }
```

### DIFF
--- a/packages/core/src/scripts/start.ts
+++ b/packages/core/src/scripts/start.ts
@@ -90,4 +90,6 @@ export const start = async (
       }/)`
     )
   })
+
+  return { expressServer, httpServer }
 }


iamandrewluca:
Hey, @dcousens, or maybe the best approach is to expose each CLI command separately 🤔

dcousens:
The change is uncontroversial, but, our CLI being exported is a bit weird.

dcousens:
I don't know, I understand your plight, but if you're diving this deep, why not fork https://github.com/keystonejs/keystone/blob/main/packages/core/src/scripts/start.ts ?

dcousens:
@iamandrewluca if we're missing internals for your use-case, we can look to exporting what you need - I think that is preferable to this

Issue #7255, opened 2022-02-01
Decimal field type does always fall back to precision of 16


evangeloszotos:
Steps to reproduce:
define a decimal field eg: 

```ts 
export const Trade = list({
  fields: {
    invest: decimal({
      precision: 32,
      scale: 8,
    }),
  },
});
```
Expected:
Values like: 
```ts
'1234567890123456789012.12345678' 
```
should be valid (works if directly typed in postgres with adminer, see last row in the screenshot)

<img width="739" alt="Screenshot 2022-02-02 at 00 08 41" src="https://user-images.githubusercontent.com/33234890/152067039-23ef8a2b-0be8-4647-84d1-cccd4d1a8d70.png">
But does not work when the entity is created with the keystone admin UI or with the graphql endpoint, see other rows.

see screenshot using graphql playground:

<img width="1067" alt="Screenshot 2022-02-02 at 00 14 25" src="https://user-images.githubusercontent.com/33234890/152067560-7e902cf1-9ad3-41c5-8d04-9258b7dcac1f.png">

I used: 
macOS 2021 M1 Max
keystone-6/core: 1.0.1
postgres: 13-buster (docker image) 
node: 14.18.1

Hope that helps.

dcousens:
I cannot reproduce this, using string `1234567890123456789012.12345678` and POSTGRES 15

PR #9011, opened 2024-02-08
Add decimal field to test sandbox, prefer POSTGRES for sandbox

dcousens:
Useful when trying to reproduce issues like https://github.com/keystonejs/keystone/issues/7255


tests/sandbox/schema.prisma
```diff
@@ -1,24 +1,23 @@
 // This file is automatically generated by Keystone, do not modify it manually.
 // Modify your Keystone config when you want to change this.
 
-datasource sqlite {
+datasource postgresql {
   url               = env("DATABASE_URL")
   shadowDatabaseUrl = env("SHADOW_DATABASE_URL")
-  provider          = "sqlite"
+  provider          = "postgresql"
```
dcousens, line 7: The sandbox isn't used by testing, so `postgresql` is OK

PR #9016, opened 2024-02-12
Add explicit TypeInfo types for examples with getContext

dcousens:
Resolves the comment from @mmachatschek in https://github.com/keystonejs/keystone/pull/8876#issuecomment-1783321292 that our `usecase-` examples didn't use `TypeInfo`, which left the `seed-data` examples using the unrefined context.


dcousens:
We should probably use `config<TypeInfo>` in every example...

PR #9038, opened 2024-02-25
Update auth-and-access-control.md

dagrinchi:
just a little correction


dcousens:
Prefer https://github.com/keystonejs/keystone/pull/9039 instead

Issue #8717, opened 2023-07-28
Attempted to call validateAndNormalizeDocument() from the server [fields-document]


Thinkscape:
## Summary

I believe this is a regression after #8403 @borisno2 
After adding `"use client"` to [packages/fields-document/src/validation.ts](packages/fields-document/src/validation.ts), things break server-side when validation is triggered by admin hooks.

```
    "@keystone-6/core": "^5.3.2",
    "@keystone-6/fields-document": "^8.0.0",
    "next": "^13.4.12",
```

## Steps

1. `next@13.4.0+` app using App Router (default `next.config.ts`).
2. A schema, including a list, with a field `content: document({})`
3. A [vanila /api/graphql route, such as this one](https://github.com/borisno2/on-the-hill-drama-club/blob/main/src/app/api/graphql/route.ts).
4. Go to admin
5. Try to add an item to the list

## Expected

I can add a list with document field.

## Actual
```
GraphQLError: An error occured while resolving input fields.
  - Article.content: Attempted to call validateAndNormalizeDocument() from the server but validateAndNormalizeDocument is on the client. It's not possible to invoke a client function from the server, it can only be rendered as a Component or passed to props of a Client Component.
    at resolverError (webpack-internal:///(sc_server)/./node_modules/.pnpm/@keystone-6+core@5.3.2_@aws-sdk+signature-v4-crt@3.363.0_@babel+core@7.22.8_@prisma+generator_w4vjmt6rg75wwrvq3lhneegeza/node_modules/@keystone-6/core/dist/graphql-errors-473725b1.esm.js:83:12)
    at getResolvedData (webpack-internal:///(sc_server)/./node_modules/.pnpm/@keystone-6+core@5.3.2_@aws-sdk+signature-v4-crt@3.363.0_@babel+core@7.22.8_@prisma+generator_w4vjmt6rg75wwrvq3lhneegeza/node_modules/@keystone-6/core/dist/createSystem-095bd0da.esm.js:987:105)
    at async resolveInputForCreateOrUpdate (webpack-internal:///(sc_server)/./node_modules/.pnpm/@keystone-6+core@5.3.2_@aws-sdk+signature-v4-crt@3.363.0_@babel+core@7.22.8_@prisma+generator_w4vjmt6rg75wwrvq3lhneegeza/node_modules/@keystone-6/core/dist/createSystem-095bd0da.esm.js:1112:29)
[...]

PR #9041, opened 2024-02-25
Fix `@keystone-6/fields-document` package breaking when compiling in SSR environments

marekryb:
This attempts to fix https://github.com/keystonejs/keystone/issues/8717 (Attempted to call `validateAndNormalizeDocument()` from the server).

As stated in https://github.com/keystonejs/keystone/pull/8403#issuecomment-1478851798 that made a workaround by adding `use client` clauses:

> The root cause of this is the crossover of React/Next code

In principle, this PR splits bunch of `.tsx` files into two separate ones. One for react-related (client-side) code and second one for platform-agnostic code. I am keeping the original file for backwards compatibility.
For example file:
`packages/fields-document/src/DocumentEditor/insert-menu.tsx`
has been split into:
`packages/fields-document/src/DocumentEditor/insert-menu-model.ts`
`packages/fields-document/src/DocumentEditor/insert-menu-view.tsx`
and the original file is now exporting above two:
```typescript 
export * from './insert-menu-model'
export * from './insert-menu-view'
```
External exports do not change, however internally, files that take part in validation process now import `-model.ts` files that are not importing disallowed hooks (`useState` etc).

There are no functional changes in the code apart from one in `packages/fields-document/src/DocumentEditor/utils-model.ts::insertNodesButReplaceIfSelectionIsAtEmptyParagraphOrHeading`. I will write about this one in the comment below.

It seems to work fine with my limited testing and all unit tests are passing. However I did not test this in some more advanced scenario and with custom components yet.

Please let me know whether this is something that you are willing to work on and eventually accept.


packages/fields-document/src/DocumentEditor/utils-model.ts
```diff
@@ -0,0 +1,155 @@
...
+  Transforms.insertNodes(editor, nodes)
+  let path = pathRefForEmptyNodeAtCursor?.unref()
+  if (path) {
+    Transforms.removeNodes(editor, { at: path })
+    // even though the selection is in the right place after the removeNodes
+    // for some reason the editor blurs so we need to focus it again
+
+    // TODO: this causes 'TypeError: ([...].createContext) is not a function' in SSR
+    // not sure how to reproduce state in which this is neccessary
+    // ReactEditor.focus(editor)
```
marekryb, line 93: As stated in the description, this is the only functional change. Importing anything from `slate-react` causes
```

 ⨯ TypeError: (0 , react__WEBPACK_IMPORTED_MODULE_3__.createContext) is not a function
    at eval (webpack-internal:///(rsc)/../../node_modules/.pnpm/slate-react@0.102.0_react-dom@18.2.0_react@18.2.0_slate@0.102.0/node_modules/slate-react/dist/index.es.js:108:91)
    at (rsc)/../../node_modules/.pnpm/slate-react@0.102.0_react-dom@18.2.0_react@18.2.0_slate@0.102.0/node_modules/slate-react/dist/index.es.js (/home/exar/os/keystone/examples/framework-nextjs-app-directory/.next/server/vendor-chunks/slate-react@0.102.0_react-dom@18.2.0_react@18.2.0_slate@0.102.0.js:20:1)
```
because it uses `createContext`. I alleviated this problem in other place by having two versions of `createDocumentEditor`:  `packages/fields-document/src/DocumentEditor/editor-create-display.ts` and `packages/fields-document/src/DocumentEditor/editor-create-validation.ts`.

However in this case I am kind of clueless what the original problem was and how to workaround it.

examples/framework-nextjs-app-directory/package.json
```diff
@@ -14,16 +14,16 @@
...
     "@keystone-6/document-renderer": "^1.1.0",
-    "@keystone-6/fields-document": "^8.0.0",
+    "@keystone-6/fields-document": "workspace:^",
     "@preconstruct/next": "^4.0.0",
     "@prisma/client": "^4.16.2",
     "graphql": "^16.8.1",
     "graphql-request": "^5.0.0",
     "graphql-yoga": "^3.1.0",
-    "next": "^13.3.0",
+    "next": "^13.4.9",
```
marekryb, line 26: `next:build` builds fine with latest `nextjs` (14.1.0), however `keystone:build` has some unrelated problem with >=14
```
Error: <Html> should not be imported outside of pages/_document.
```

dcousens:
I am really open to accepting this work, my only concern is that we're trying to solve a number of problems in one pull request.
If we can break this into a few smaller pull requests, I can review and approve more quickly.

examples/framework-nextjs-app-directory/package.json
dcousens: Can we break this out into it's own pull request, and why do we need to bump?

dcousens:
Can you please put any code moving operations into their own commits, so I can easily review those commits and then focus only on the diff?

marekryb:
> I am really open to accepting this work, my only concern is that we're trying to solve a number of problems in one pull request.
If we can break this into a few smaller pull requests, I can review and approve more quickly.

Unfortunately this is not possible. Everything here is related and do not function without other pieces.

> Can you please put any code moving operations into their own commits, so I can easily review those commits and then focus only on the diff?

This is already kind of the case. The initial ~18 "wip" labelled commits are splitting one file and then changing related imports in other places.

My workflow was like this:
1) remove existing 'use client' to trigger error
2) see the error when running `examples/framework-nextjs-app-directory`
3) split the file that has the error and fix imports related to it
4) run `pnpm jest packages/fields-document/`
5) `git commit -am "wip"`
6) goto 2 until working

examples/framework-nextjs-app-directory/package.json
marekryb: There is no real reason for this change. Originally I wanted to put `^14.1.0` here, but because I cannot we can just leave it as it was.

marekryb:
Removed all changes to `examples/framework-nextjs-app-directory`. Can be split to separate PR when this is done.

dcousens:
@marekryb OK, if you can get everything into a working order, as atomic as reasonable, I could jump in towards the end

dcousens:
I have updated this pull request and rebased each of the changes.
I effectively did the changes again from HEAD, to ensure I understood everything that was changed, but the credit and work will still be attributed to you @marekryb when merged.

Thanks again for this @marekryb, and I have the same conclusion, this may not resolve _every_ situation for now, but, we should be nearer to that with this pull request.

packages/fields-document/src/DocumentEditor/utils-model.ts
dcousens: When did this happen for you?

dcousens: I have resolved this with b85894f, by using a dependency passing pattern to side-step the problem without duplicating the code.

dcousens: The problem @marekryb is that `ReactEditor` has a call to `createContext`, which isn't allowed on the server.

PR #9053, opened 2024-03-06
Tidy up internal Prisma types

dcousens:
This pull request tidies up a number of internal Prisma types that were duplicated, and often out-of-date.  An explicit `as any` is replacing the two locations where the type was used, and the remaining boilerplate removed.  

An internal `new WeakMap` mapping the Prisma namespaces to Keystone contexts has been removed too.

### DIFF
--- a/packages/core/src/lib/core/initialise-lists.ts
+++ b/packages/core/src/lib/core/initialise-lists.ts
@@ -154,8 +154,8 @@ function getIsEnabled (listKey: string, listConfig: KeystoneConfig['lists'][stri
       create: notOmit,
       update: notOmit,
       delete: notOmit,
-      filter: notOmit && defaultIsFilterable,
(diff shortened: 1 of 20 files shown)


packages/core/src/lib/core/initialise-lists.ts
```diff
@@ -154,8 +154,8 @@ function getIsEnabled (listKey: string, listConfig: KeystoneConfig['lists'][stri
       create: notOmit,
       update: notOmit,
       delete: notOmit,
-      filter: notOmit && defaultIsFilterable,
-      orderBy: notOmit && defaultIsOrderable,
+      filter: notOmit ? defaultIsFilterable : false,
+      orderBy: notOmit ? defaultIsOrderable : false,
```
dcousens, line 158: This was confusing, `&&` technically worked, but, the point is that it is `false` or a function

Issue #8846, opened 2023-10-04
`db.idField` settings are not reflected in generated types


pc-erin:
To reproduce, create the following schema:
```ts
export const lists: Lists<Session> = {
	testfield: list({
		access: allowAll,
		db: {
			idField: {
				kind: 'autoincrement',
				type: 'Int',
			},
		},
		fields: {},
	}),
	singleton: list({
		access: allowAll,
		isSingleton: true,
		fields: {},
	}),
}
```

This should result in the `.keystone/types.ts` file containing a type called `testfieldWhereUniqueInput`.

It does, but the type of the id field is incorrect:

```ts
export type testfieldWhereUniqueInput = {
  readonly id?: Scalars['ID'] | null;
};
```

`Scalars['ID']` is a string, but the id should be a number. I would expect it to be `Scalars['Int']` instead.

This also happens for singletons, because `isSingleton: true` forces `db.idField.type` to be `Int`.

It's not super noticeable normally, but if you start using the generated types in your schema, you'll notice that there is a red squiggly on all your list definitions. The error you get will usually contain the string `The types of 'context.db' are incompatible between these types.`.

dcousens:
Thanks for reporting this @pc-erin, I have seen this too

dcousens:
@pc-erin I think the only reasonable option here is to stop using `ID` from GraphQL

PR #9054, opened 2024-03-06
Fix the type of `id` on type `BaseItem`, to be type `unknown`, not `string`

dcousens:
This pull request updates the type of `id` on type `BaseItem` to be of type `unknown` instead of `string`.  This helps reduce a number of type errors where `string | number | null` is possible, but only `string` was accepted by the `BaseItem`.

This is not strictly related to https://github.com/keystonejs/keystone/issues/8846, but will likely be required as part of fixing that problem in any event.

### DIFF
--- a/packages/auth/src/gql/getInitFirstItemSchema.ts
+++ b/packages/auth/src/gql/getInitFirstItemSchema.ts
@@ -61,7 +61,10 @@ export function getInitFirstItemSchema ({
           // the input value can't round-trip like the Upload scalar here is quite low)
           const item = await sudoContext.db[listKey].createOne({ data: { ...data, ...itemData } })
           const sessionToken = (await context.sessionStrategy.start({
-            data: { listKey, itemId: item.id.toString() },
+            data: {
+              listKey,
+              itemId: item.id.toString()
+            },
             context,
           }))
 
--- a/packages/core/src/types/next-fields.ts
+++ b/packages/core/src/types/next-fields.ts
@@ -12,7 +12,12 @@ import {
 
 export { Decimal }
 
-export type BaseItem = { id: { toString(): string }, [key: string]: unknown }
+export type BaseItem = {
+  id: {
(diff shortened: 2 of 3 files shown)


packages/fields-document/src/structure-graphql-input.tsx
```diff
@@ -442,7 +442,7 @@ export async function checkUniqueItemExists (
     throw missingItem(operation, uniqueInput)
   }
 
-  return { id: item.id.toString() }
+  return { id: item.id }
```
emmatown, line 445: Removing this `toString` will break if you have a bigint since a bigint wouldn't be accepted by the `ID` type

packages/core/src/types/type-info.ts
```diff
@@ -14,7 +14,7 @@ export type BaseListTypeInfo<Session = any> = {
     create: GraphQLInput
     update: GraphQLInput
     where: GraphQLInput
-    uniqueWhere: { readonly id?: string | number | null } & GraphQLInput
+    uniqueWhere: { readonly id?: unknown } & GraphQLInput
```
emmatown, line 17: `string | number | null` covers all that the `ID` GraphQL type could be, what does widening this help?

packages/auth/src/gql/getInitFirstItemSchema.ts
```diff
@@ -61,7 +61,10 @@ export function getInitFirstItemSchema ({
           // the input value can't round-trip like the Upload scalar here is quite low)
           const item = await sudoContext.db[listKey].createOne({ data: { ...data, ...itemData } })
           const sessionToken = (await context.sessionStrategy.start({
-            data: { listKey, itemId: item.id.toString() },
+            data: {
+              listKey,
+              itemId: item.id
```
emmatown, line 66: If you had a bigint id, this would fail since `JSON.stringify` with a bigint throws

PR #9055, opened 2024-03-06
Increase verbosity of `omit` tests 

dcousens:
When modifying code near to the `omit` functionality, I quickly found myself wanting amount of detail as to when and where the tests were failing.

Unlike `tape` and other testing suites, `jest` only really offers you insight into your tests from the `test(description, () => ...`, which you often might group a number of `expect`s statements within.  This isn't helpful when the tests are verifying a number of expectations at the same time.

This pull request changes that so we many `test`s, with 1 expectation each. This pattern is actually a huge performance win too, and I'll be using in my next pull request for hooks.

### DIFF
--- a/tests/api-tests/omit.test.ts
+++ b/tests/api-tests/omit.test.ts
@@ -2,39 +2,36 @@ import { setupTestSuite } from '@keystone-6/api-tests/test-runner'
 import { relationship, text } from '@keystone-6/core/fields'
 import { list } from '@keystone-6/core'
 import { allowAll } from '@keystone-6/core/access'
+import { type KeystoneContext } from '@keystone-6/core/types'
 import { dbProvider } from './utils'
-
-function yesNo (x: boolean | undefined) {
-  if (x === true) return 'T'
-  if (x === false) return 'F'
-  return 'U'
-}
-
-const UFT = [false, true] as const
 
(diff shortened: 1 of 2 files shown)


tests/api-tests/omit.test.ts
```diff
@@ -139,181 +136,181 @@ const listsMatrix = [...function* () {
...
+            ref: l.__name,
             many: true,
           })] as const
         }
       }()),
     },
-    defaultIsFilterable: undefined,
-    defaultIsOrderable: undefined,
+    defaultIsFilterable: true,
+    defaultIsOrderable: true,
```
dcousens, line 158: This code doesn't need to be testing defaults, we can do that elsewhere

tests/api-tests/omit.test.ts
```diff
@@ -139,181 +136,181 @@ const listsMatrix = [...function* () {
...
-            ref: l.name,
+          yield [`R${l.__name}_many`, relationship({
+            ref: l.__name,
             many: true,
           })] as const
         }
       }()),
     },
-    defaultIsFilterable: undefined,
-    defaultIsOrderable: undefined,
```
dcousens, line 161: This code doesn't need to be testing defaults, we can do that elsewhere

PR #9057, opened 2024-03-07
Adds `{field}.hooks.validate.[create|update|delete]` hooks

dcousens:
Similar to https://github.com/keystonejs/keystone/pull/9056

This pull request needed to adopt a slightly different approach, since we can't reasonably change the types of field hooks when field types are part of our public API, and they are not statically determined by nature as the field types themselves return functions, not an object like `list`.

This is only a slight inconvenience though, since we can throw a runtime error to prevent incompatibilities at the point of creating the GraphQL resolvers (`parseFieldHooks`).  I do want to flatten this in the future, but, for optimistic usecases you could start using `.validation.*` field hooks now.  This won't be compatible with many of the native field types, but that compatibility can come later.

Additionally, as part of this pull request (and part of the `validateInput` hooks), the `text` now accepts a default value type of `null`.  This was mentioned as a pain point in https://github.com/keystonejs/keystone/discussions/7158#discussioncomment-7584310, but has remained unresolved for some time.

If `db.isNullable: false`, and you pass `defaultValue: null`, you will end up with a Prisma error if your text input is undefined (this behavior is typical of other fields). If `db.isNullable: true`, then `defaultValue: null` will be exactly that.


packages/core/src/fields/types/checkbox/index.ts
```diff
@@ -19,13 +19,16 @@ export type CheckboxFieldConfig<ListTypeInfo extends BaseListTypeInfo> =
...
+  config: CheckboxFieldConfig<ListTypeInfo> = {}
+): FieldTypeFunc<ListTypeInfo> {
+  const {
+    defaultValue = false,
+  } = config
+
+  return (meta) => {
     if ((config as any).isIndexed === 'unique') {
-      throw Error("isIndexed: 'unique' is not a supported option for field type checkbox")
+      throw TypeError("isIndexed: 'unique' is not a supported option for field type checkbox")
```
dcousens, line 31: This is a type error

dcousens:
I think the distinctions between our nulls is poorly defined. We have too many nulls.

- GraphQL is `String | String!`, throws type errors
- Hooks for validation when `isRequired: true`, throws validation errors
- Prisma `String? | String`, throws prisma errors

Arguably we can unify some parts of this, or restructure it for progressive enhancement with less "conflicting" configuration.
At the least, `validation: { isRequired: true }` is unhelpful when you can use GraphQL to ensure the input is required.
We should leverage that instead.

Leaves some questions around `context.db`, but, that has many questions already...

packages/core/src/fields/types/text/index.ts
```diff
@@ -24,7 +27,7 @@ export type TextFieldConfig<ListTypeInfo extends BaseListTypeInfo> =
       match?: { regex: RegExp, explanation?: string }
       length?: { min?: number, max?: number }
     }
-    defaultValue?: string
+    defaultValue?: string | null
```
dcousens, line 30: Unlike other fields, the ambiguity of `?:` representing `''` or `null` is why this needs to exist.
We should allow `null` as a `defaultValue` typically, and maybe check it statically too.

packages/core/src/fields/types/multiselect/index.ts
```diff
@@ -42,12 +42,14 @@ export type MultiselectFieldConfig<ListTypeInfo extends BaseListTypeInfo> =
...
+  const {
     defaultValue = [],
-    ...config
-  }: MultiselectFieldConfig<ListTypeInfo>): FieldTypeFunc<ListTypeInfo> =>
-  meta => {
+  } = config
+
+  return (meta) => {
     if ((config as any).isIndexed === 'unique') {
       throw Error("isIndexed: 'unique' is not a supported option for field type multiselect")
```
dcousens, line 54:
```suggestion
      throw TypeError("isIndexed: 'unique' is not a supported option for field type multiselect")
```

PR #9090, opened 2024-04-11
Remove deprecated `config.db.useMigrations`, change keystone prisma behaviour

dcousens:
This pull request removes `config.db.useMigrations`, which was deprecated in https://github.com/keystonejs/keystone/pull/8046, but perhaps a better description of why is needed to help with updating your workflow.

I'll add a draft of that workflow here into our documentation soon.
The short of it could probably be summarised in the following text (work in progress):

```bash
# development lifecycle
$ keystone dev
  1. Keystone updates your schemas (GraphQL, Prisma, Typescript)
  2. Prisma pushes any updates to your database (unless you use --no-db-push)
  3. Go to 1, until ready

$ keystone prisma migrate dev --create-only
  4. Generate any database migrations, editing as required
   
$ keystone start --with-migrations
  5. Test your migrations
  6. If happy, continue, otherwise go back, using `keystone prisma migrate reset` as needed
   
# ...
  
# deploying
$ keystone build
 
$ keystone start --with-migrations
```
 
Any feedback is appreciated, as we move to integrate this workflow into the documentation.
 
**References**
- https://www.prisma.io/docs/orm/prisma-migrate/workflows/development-and-production
- https://www.prisma.io/docs/orm/reference/prisma-cli-reference#options-6


packages/core/src/scripts/cli.ts
```diff
@@ -89,7 +89,7 @@ export async function cli (cwd: string, argv: string[]) {
   }
 
   if (command === 'start') {
-    return start(cwd, defaultFlags(flags, { ui: true, withMigrations: false }))
+    return start(cwd, defaultFlags(flags, { server: true, ui: true, withMigrations: false }))
```
dcousens, line 92: `start --with-migrations --no-server` might be redundant by comparison to `keystone prisma migrate deploy`, but, I think `start --no-server` being ignored is worse

packages/core/src/lib/migrations.ts
```diff
@@ -174,155 +174,6 @@ function logWarnings (warnings: string[]) {
...
-    const migration = await runMigrateWithDbUrl(dbUrl, undefined, () => migrate.applyMigrations())
-    if (migration.appliedMigrationNames.length === 0) {
-      console.info(`✨ The database is already in sync with your migrations`)
-    } else {
-      console.info(`✨ Your database is now in sync with your migrations`)
-    }
-  })
-}
-
-export async function devMigrations (
```
dcousens, line 188: Effectively, use `keystone prisma migrate dev` instead

PR #9104, opened 2024-04-19
updated devcontainer for vscode

ggpwnkthx:


.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,32 @@
...
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "git submodule update --init --recursive",
+
+  // Configure tool-specific properties.
+  "customizations": {
+    "vscode": {
+      "settings": {},
+      "extensions": [
+        "dbaeumer.vscode-eslint",
+        "esbenp.prettier-vscode",
```
dcousens, line 23:
```suggestion
```

The repository isn't using `prettier` now

.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,32 @@
...
-  "extensions": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode", "prisma.prisma"],
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  // "features": {},
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   // "forwardPorts": [],
 
-  "postCreateCommand": "pnpm install",
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "git submodule update --init --recursive",
```
dcousens, line 15: Do we need this? This should probably be `pnpm install` still

.devcontainer/devcontainer.json
```diff
@@ -1,20 +1,32 @@
...
-  "name": "Keystone",
-  "build": {
-    "dockerfile": "Dockerfile",
-    "args": {
-      "VARIANT": "14"
-    }
-  },
-
-  "settings": {},
+  "name": "Node.js & TypeScript",
```
dcousens, line 4:
```suggestion
  "name": "Keystone",
```

.devcontainer/devcontainer.json
ggpwnkthx: I think your change makes more sense for Keystone in general. My though here was to allow packages to be added via submodules, but I think that breaks the spirit of a monorepo.

.devcontainer/devcontainer.json
dcousens: Not at the moment, it's a TODO

dcousens:
Thanks @ggpwnkthx

PR #9105, opened 2024-04-19
Custom next.config.js support for admin-ui

ggpwnkthx:
If `/.keystone/admin/../../next.config.js` exists, it will get merged into keystone's admin-ui's NextJS config at runtime.

### DIFF
--- a/packages/core/src/admin-ui/templates/next-config.ts
+++ b/packages/core/src/admin-ui/templates/next-config.ts
@@ -1,15 +1,32 @@
 export const nextConfigTemplate = (basePath?: string) =>
-  `const nextConfig = {
-    typescript: {
-      ignoreBuildErrors: true,
-    },
-    eslint: {
-      ignoreDuringBuilds: true,
-    },
-    // We use transpilePackages for the custom admin-ui pages in the ./admin folder
-    // as they import ts files into nextjs
-    transpilePackages: ['../../admin'],
-    ${basePath ? `basePath: '${basePath}',` : ''} 
+  `const fs = require('fs');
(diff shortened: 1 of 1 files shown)


dcousens:
Might it not be better to conditionally _not_ write to `.keystone/admin/next.config.js`?
Then users can simply use their own?

gautamsi:
when #9186 is merged, user should be able to do this

dcousens:
OK, this pull request is superseded by #9186

dcousens:
Actually @gautamsi, #9186 is a breaking change, maybe we can land this earlier?

gautamsi:
@dcousens 
do you plan to have any patch/minor release soon? I might make use of that if it comes before end of this month.

dcousens:
Yes, but I don't have time to test this particular pull request before then - can you?

gautamsi:
I tried testing this, I had to install lodash as dependency which is missing in PR.

I see that there is workaround available to allow using a custom nextjs config. Instead of releasing this I would wait for #9186

this is how to achieve same thing without needing this PR. 

1. extend the base config generated by keystone. you can run `keystone dev` once to generate one. Copy the content of `.keystone/admin/next.config.js` to custom next.config.js file (it can be named anything)
2. extend your config
3.  use following code snippet of `keystone.ts` to overwrite the generated next.config.js
```ts
export default config({
// ....other config
  ui: {
    getAdditionalFiles: [
      () => [{ mode: 'copy', inputPath: path.resolve('next.config.js'), outputPath: 'next.config.js' }]
    ],
  }
})
```

PR #9114, opened 2024-04-24
feat(core/db): allow to extend PrismaClient

iamandrewluca:
Using this feature, one can extend PrismaClient, and add for example hooks at prisma level

```ts
export default withAuth(
  config<TypeInfo>({
    db: {
      extendPrismaClient(client: PrismaClient) {
        return client.$extends({
          query: {
            async $allOperations({ operation, model, args, query }) {
              // Raw Query is executed
              if (model === undefined) {
                const result: unknown = await query(args);
                return result;
              }

              const hooks = getPrismaHooks(model, operation);

              await hooks?.before?.(client, args);
              const result: unknown = await query(args);
              await hooks?.after?.(client, args, result);

              return result;
            },
          },
        });
      },
    },
  }),
);
```

### DIFF
--- a/.changeset/extend-prisma-client.md
+++ b/.changeset/extend-prisma-client.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': minor
+---
+
+Add `config.db.extendPrismaClient` to support extending the `PrismaClient`
--- a/packages/core/src/lib/createSystem.ts
+++ b/packages/core/src/lib/createSystem.ts
@@ -224,7 +224,10 @@ export function createSystem (config_: KeystoneConfig) {
         log: config.db.enableLogging
       })
 
-      const prismaClient = injectNewDefaults(prePrismaClient, lists)
+      const prismaClient = config.db.extendPrismaClient(
+        injectNewDefaults(prePrismaClient, lists)
+      )
+
       const context = createContext({
         config,
         lists,
--- a/packages/core/src/lib/defaults.ts
+++ b/packages/core/src/lib/defaults.ts
@@ -74,6 +74,7 @@ function defaultIsAccessAllowed ({ session, sessionStrategy }: KeystoneContext) 
 }
 
 async function noop () {}
+function identity<T> (x: T) { return x }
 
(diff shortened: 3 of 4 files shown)


gautamsi:
I was looking to integrate ` @prisma/extension-read-replicas` with this yesterday, right on time!
hope this gets into the release.

iamandrewluca:
We are already using this using a patch on top of KS

dcousens:
LGTM, doesn't resolve the issues raised by https://github.com/keystonejs/keystone/pull/8847, but I think we'll look at that another  day

dcousens:
Thanks @iamandrewluca! I'll try and release this asap

iamandrewluca:
Thanks! That is great!

gautamsi:
@iamandrewluca  can you share the code of `getPrismaHooks` or is this exported by prisma?

iamandrewluca:
I have a big object mapper for all models/operations/lifecycles, and getPrismaHooks returns me the functions based on model/operation

```ts
import { type PrismaClient, type Prisma } from '@prisma/client';
import type { TypeInfo } from '.keystone/types';

type PrismaHook = (
	prismaClient: PrismaClient,
	args: unknown,
	operationResults?: unknown,
) => Promise<void> | Promise<unknown>;

type PrismaExtendedOperation = {
	[ListType in keyof TypeInfo['lists']]?: {
		[OperationType in Prisma.DMMF.ModelAction]?: Partial<
			Record<'before' | 'after', PrismaHook>
		>;
	};
};

const PrismaHooks: PrismaExtendedOperation = {
	Order: {
		create: {
			before: async (prismaClient: PrismaClient, args: unknown) => {
				// ...
			},
		},
		update: {
			before: async (prismaClient: PrismaClient, args: unknown) => {
				// ...
			},
		},
	},
};

export function getPrismaHooks(
	model: keyof TypeInfo['lists'],
	operation: string,
): Partial<Record<'before' | 'after', PrismaHook>> | undefined {
	const modelHooks = PrismaHooks[model];
	return modelHooks?.[operation as Prisma.DMMF.ModelAction];
}
```

dcousens:
@iamandrewluca out of interest, why use these by comparison to the Keystone hooks?

iamandrewluca:
@dcousens, we had to create custom mutations that use Prisma transactions.

PR #9125, opened 2024-05-01
docs(examples): add custom session with passport

iamandrewluca:
https://github.com/keystonejs/keystone/assets/1881266/dfc1a65f-13d1-4cbf-84e7-8dd64b3faca7


dcousens:
I can't review this this week @iamandrewluca, but this looks amazing! :yellow_heart:

iamandrewluca:
Great! I will also do a review myself, probably at the end of the week or next week. I Need to make some more adjustments.

examples/usecase-passport/auth.ts
```diff
@@ -0,0 +1,134 @@
...
+  createdAt: string;
+};
+
+export type Session = BaseSession<SessionData>;
+
+export const { withAuth } = createAuth<Lists.User.TypeInfo<Session>>({
+  listKey: "User",
+  identityField: "email",
+  // Strictly typed sessionData (see Join at the end)
+  sessionData: "name createdAt" satisfies Join<keyof SessionData>,
```
iamandrewluca, line 31: Strict typing won't work. I forgot this is a GQL query.

Note: Remove satisfies

examples/usecase-passport/keystone.ts
```diff
@@ -0,0 +1,25 @@
+import "dotenv/config";
+import { config } from "@keystone-6/core";
+import { lists } from "./schema";
+import { withAuth, session, createAuthenticationMiddleware } from "./auth";
+import { fixPrismaPath } from "../example-utils";
+
+export default withAuth(
```
borisno2, line 7: Any reason for wrapping `withAuth` here instead of just adding the config directly - seeing as it is an example, the `withAuth` abstraction feels confusing?

examples/usecase-passport/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "keystone-app",
```
borisno2, line 2: this might need to be updated to "@keystone-6/example-custom-session-passportjs" - I also wonder if this is better placed in a `custom-seesion-passportjs` folder as it better fits with custom-session than it does with `usecase` - but @dcousens might have more thoughts on that.

examples/usecase-passport/auth.ts
```diff
@@ -0,0 +1,134 @@
+import { randomBytes } from "crypto";
+import { Router } from "express";
+import { Passport, Profile } from "passport";
+import { Strategy, StrategyOptions } from "passport-github2";
+import { hashSync } from "bcryptjs";
+import { createAuth } from "@keystone-6/auth";
```
borisno2, line 6: Why does this need the auth package?

examples/usecase-passport/keystone.ts
borisno2, line 7: You shouldn't need to wrap `withAuth` here, I don't see what the keystone auth package is providing here?

examples/usecase-passport/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "dependencies": {
+    "@keystone-6/auth": "^7.0.0",
+    "@keystone-6/core": "^5.0.0",
+    "@keystone-6/fields-document": "^7.0.0",
+    "@prisma/client": "^5.13.0",
+    "@types/bcryptjs": "^2.4.2",
+    "@types/express": "^4.17.14",
+    "@types/passport": "^1.0.16",
+    "@types/passport-github2": "^1.2.9",
+    "@types/passport-oauth2": "^1.4.15",
```
borisno2, line 20: should these live in `devDependencies`?

examples/usecase-passport/schema.ts
```diff
@@ -0,0 +1,17 @@
+import { list } from "@keystone-6/core";
+import { allowAll } from "@keystone-6/core/access";
+import { text, password, timestamp } from "@keystone-6/core/fields";
+import type { Lists } from ".keystone/types";
+import { Session } from "./auth";
+
+export const lists = {
```
borisno2, line 7: Is it better to make this schema similar to the next-auth example https://github.com/keystonejs/keystone/blob/main/examples/custom-session-next-auth/schema.ts and use `authId` as `p.id` returned from passport?

examples/usecase-passport/auth.ts
```diff
@@ -0,0 +1,134 @@
...
+  return async (_a: string, _r: string, p: Profile, done: VerifyCallback) => {
+    const name = p.displayName;
+    const email = p.emails?.map((e) => e.value).at(0);
+
+    if (!email || !name) {
+      return done(new Error("No email or name"));
+    }
+
+    // Find or create user with this email
+    const user = await context.prisma.user.upsert({
```
borisno2, line 110: not sure if you need to actually save the user here if you store `p.id` in the session, and then use that in list access

borisno2:
Nice work @iamandrewluca ! 💙

examples/usecase-passport/schema.ts
iamandrewluca: Tried to replicate the schema of next-auth example. Not sure if did it right.

iamandrewluca:
@borisno2 @dcousens is ready to be reviewed again!

LabLamb:
Nice! May I ask what is the password set to?

iamandrewluca:
@LabLamb what do you mean by that?

LabLamb:
> @LabLamb, what do you mean by that? Trolling? Spam soon?

For the native Keystone auth method there is a password, if we use SSO for creating the account, what method is used to craete an password for that the newly created account? Is it a random long string or blank?

iamandrewluca:
In this case, there is no password. You always log in with SSO. SSO is responsible for user authentication

dcousens:
Squashed and rebased on `main`

dcousens (approved):
LGTM, nice work @iamandrewluca!

PR #9129, opened 2024-05-02
feat(core/types): export hooks function types

iamandrewluca:
It would be better to have the hooks function types exported from KS.

PR is in 2 commits. One formats the changed code, one does the actual changes.

Today, I had to write some code and reuse some KS types that are not exported directly but through the API. It was complicated, but I did it.

<details>
<summary>See the code 🙈</summary>

```ts
import { type BaseListTypeInfo, type ListHooks } from '@keystone-6/core/types';

export type ExtractObject<T> = T extends Record<string, unknown> ? T : never;
export type ApplicableOperation = 'create' | 'update' | 'delete';

// 😱
type ValidateArgs<
	ListTypeInfo extends BaseListTypeInfo,
	Operation extends ApplicableOperation,
> = Omit<
	Parameters<
		NonNullable<ExtractObject<ListHooks<ListTypeInfo>['validate']>[Operation]>
	>[0],
	'addValidationError'
>;

export type ValidateOptions<ListTypeInfo extends BaseListTypeInfo> = (
	| ValidateArgs<ListTypeInfo, 'create'>
	| ValidateArgs<ListTypeInfo, 'update'>
	| ValidateArgs<ListTypeInfo, 'delete'>
) & {
	field: string;
};

export interface Validator<
	ListTypeInfo extends BaseListTypeInfo = BaseListTypeInfo,
> {
	validate: (
		options: ValidateOptions<ListTypeInfo>,
	) => Promise<string | undefined> | string | undefined;
}
```
</details>

### DIFF
--- a/packages/core/src/types/config/hooks.ts
+++ b/packages/core/src/types/config/hooks.ts
@@ -9,7 +9,7 @@ type CommonArgs<ListTypeInfo extends BaseListTypeInfo> = {
   listKey: ListTypeInfo['key']
 }
 
-type ResolveInputListHook<
+export type ResolveInputListHook<
   ListTypeInfo extends BaseListTypeInfo,
   Operation extends 'create' | 'update',
 > = (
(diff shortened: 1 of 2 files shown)


packages/core/src/types/config/hooks.ts
```diff
@@ -14,9 +9,9 @@ type CommonArgs<ListTypeInfo extends BaseListTypeInfo> = {
   listKey: ListTypeInfo['key']
 }
 
-type ResolveInputListHook<
+export type ResolveInputListHook<
```
dcousens, line 12: Why did you need the `Resolve*` types?

packages/core/src/types/config/hooks.ts
```diff
@@ -316,12 +313,12 @@ type ValidateFieldHook<
     CommonArgs<ListTypeInfo> & {
       fieldKey: FieldKey
       addValidationError: (error: string) => void
-    }
+    },
 ) => MaybePromise<void>
 
-type BeforeOperationListHook<
+export type BeforeOperationListHook<
```
dcousens, line 319: Why not `Lists.Post['hooks']['beforeOperation']` as the type?

dcousens:
I am personally hesitant about this, as I think we should be promoting the usage of the refined types, not these abstract types.

iamandrewluca:
Hey, @dcousens, I understand your point. In my case, I need only the args type of the `validate` function; all other exports I made are just for the sake of being exported. But I get your point.

Will close this PR for now (while rethinking how to use it the right way)

PS: Our project requirements started diverging from KS's base functionalities.
We might drop it in favour of NestJS. Thanks a lot for all the effort!

dcousens:
@iamandrewluca no problem, and thanks for the contributions!

> while rethinking how to use it the right way

Let me know if any examples could better help demonstrate how you could use the refined types in an abstract way

PR #9131, opened 2024-05-04
Attempt to add a logger to context

iamandrewluca:


iamandrewluca:
I just like to waste time like this ))

No point in passing something through from config to context. Just import it directly.

dcousens:
@iamandrewluca this isn't a waste of time! I have personally written many different types of "logging" for Keystone 6, but nothing has landed as of yet in respect to an abstraction I am happy to push upstream. The requirements vary significantly between projects.

Maybe you could document what your needs are and we can collaborate on that?

iamandrewluca:
We mainly needed to log different sensitive actions in custom mutations, and we created a global Winston logger instance.

PR #9139, opened 2024-05-10
Virtual Input

ggpwnkthx:
Core changes made so that `virtual` field types can now handle generic JSON input values that can be used by their respective hooks.

The `virtual-field` example has also been updated to show the use case of where Tag records, as they relate to Post records, are orderable. The `tags` field of the Post list will output the Tag records in the order specified by first querying the explicit relationship list, PostTags. If an array of Tag records are provided in the `tags` field of a Post mutation, whatever order they are in is how they will be reordered for that Post.


dcousens:
I don't know if `virtual` is the right abstraction for this, and `JSON` is too opinionated.
I don't think I can accept this pull request.

Issue #8884, opened 2023-11-03
Invalid `uuid` is breaking relationship filters when using postgresql 


skindc:
I am currently trying to upgrade including core@5.6.0
There is now a bug in RelationshipSelect where on entering text to filter the related items the UI breaks and just displays 'Error'

To replicate:
As far as I can see all my relationships that use the 'select' ui option, therefore I am presuming this affects all relationship fields.

Expected:
When I type any value into the input field of RelationShip select I expect it to filter the dropdown list by filtering the `search fields` defined in the relationship or the related item.

<img width="447" alt="image" src="https://github.com/keystonejs/keystone/assets/73968/3048225c-0b98-4523-9391-eca2ce2f82fb">
<img width="477" alt="image" src="https://github.com/keystonejs/keystone/assets/73968/8a0a24a5-76cc-4a9e-b007-3d0d7a00f606">

I have done some investigation into this.

When attempting to debug this I watched the network panel of developer tools in chrome to see if there was a Graphql Error reported. In fact there was.

The response from the RelationshipQuery when entering the filter reported the error below.

```json
        {
            "message": "Prisma error: Inconsistent column data: Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `g` at 1",
            "locations": [
                {
                    "line": 7,
                    "column": 3
                }
            ],
            "path": [
                "count"
            ],
            "extensions": {
                "code": "KS_PRISMA_ERROR",
                "prisma": {
                    "name": "PrismaClientKnownRequestError",
                    "code": "P2023",
                    "clientVersion": "4.16.2",
                    "meta": {
                        "message": "Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `g` at 1"
                    }
                },
                "stacktrace": [
                    "GraphQLError: Prisma error: Inconsistent column data: Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `g` at 1",
                    "    at Object.prismaError (/Users/skin/Projects/activeProjects/Lokalliving/lokalliving-stack/apps/keystone6/node_modules/@keystone-6/core/dist/graphql-errors-f1c235e2.cjs.dev.js:43:10)",
[...]

PR #9155, opened 2024-05-23
Fix UUID's breaking relationship filters with POSTGRES

dcousens:
As described in https://github.com/keystonejs/keystone/issues/8884, this pull request prevents malformed UUIDs from being added to the search query, which then results in Prisma throwing an error breaking the search filters.


dcousens:
@fkrauthan

pahaz (approved):

dcousens:
Only thing blocking this is some form of tests, whether manually verified or automated

PR #9179, opened 2024-06-23
Tidy up formatting

gautamsi:
Have tried to cleanup eslint errors which I was trying to fight with vscode, I believe we should nto have dropped the prettier but I have not changed that

not suer if I need to add changeset for this.

### DIFF
--- a/examples/extend-graphql-schema-nexus/nexus-types.ts
+++ b/examples/extend-graphql-schema-nexus/nexus-types.ts
@@ -4,7 +4,7 @@
  */
 
 
-import type { Context } from "./keystone-types"
+import type { Context } from './keystone-types'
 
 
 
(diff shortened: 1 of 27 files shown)


examples/extend-graphql-schema-nexus/nexus-types.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 
-import type { Context } from "./keystone-types"
+import type { Context } from './keystone-types'
```
dcousens, line 7: See ^
```
* This file was generated by Nexus Schema
* Do not make changes to this file directly
```

This should have been ignored

dcousens:
Thanks for this @gautamsi, interested in why a few of them were missed by previous `--fix` changes - in any event, I'm still happy this isn't part of our CI

PR #9181, opened 2024-06-24
[WIP/RFC] Full Next.JS experience with App Router

gautamsi:
This is my attempt to modernize the template with App Router and server side actions. This also brings the fullNext.JS experience without using isolated nextjs folder.

changes:

* use NextJS app router
* Use this in client only rendering for now. `use client` for quick POC
* Work with real files instead of hidden `.keystone/admin`  instance of Next.JS
* generates `<src|optional>/app/admin` or `app/(admin)` folder to make use of `basePath` config

Need to fix
* dashboard/list paths (have fixed it one point but broken again)

Roadmap

- [ ] upgrade Next.JS and React to 15 and 19
- [ ] Use server actions for forms
- [ ] use server code with keystone context instead of graphql
- [ ] still use graphql for certain part of the portal
- [ ] Add support for composable UI elements, which will pave the way for UI Extension (aka Hooks in v5)
- [ ] upgrade `@keystone-ui` to be ssr friendly
- [ ] _...TBD_

@dcousens This PR is for feedback and collaboration to make it next gen, I have plans to add more features like nested lists. Wanted to get your take on this if ThinkMill is willing to work in this direction or not.

> this is using part of eslint cleanup branch, should be less clutter once the other PR merges


dcousens:
> Wanted to get your take on this if ThinkMill is willing to work in this direction or not.

@gautamsi I think it was a mistake for Keystone to use NextJS without leveraging this experience, and I'm happy to see you think we should be moving in this direction too!

dcousens:
In the interests of merging early, let's try and keep each each pull request small and as atomic as possible.  Some of the features you have listed will need more attention than others, and we need not hold everything back while reviewing one particular feature.

gautamsi:
Lets move this into an RFC which can be contributed towards the feature improvements. I will then create atomic PRs.

gautamsi:
@dcousens what is your opinion on React 19 and Next 15? If all goes well, the release could be timed with the final release of the two with some beta/rc/nightly release along the way. or do you want to focus with Next 14 only? IMO Next 15 will keep it going for long.

PR #9189, opened 2024-06-29
Return ResolvedKeystoneConfig instead of KeystoneConfig

gautamsi:
* renamed `__ResolvedKeystoneConfig` to `ResolvedKeystoneConfig`;
* Return default config from `config` function so that any config wrapper can make use of defaults. 
* `config()` now returns `ResolvedKeystoneConfig` instead of `KeystoneConfig`
* `withAuth` input now needs `ResolvedKeystoneConfig`

Breaking changes due to type change, There is no migration needed in most cases instead of type casting in some case.

> Changes only needed in advanced use case like manually running the dev/build process instead of `cli`


dcousens:
@gautamsi 
> Breaking changes due to type change, There is no migration needed in most cases instead of type casting in some case.

Do we actually need to break? The resolved type is only a type refinement, hopefully

gautamsi:
we can do minor change for this, as no one was using until now.

gautamsi:
@dcousens I have updated this to be minor instead of major change. 
Also changed base to main. let me know if you need any other changes

gautamsi:
@dcousens seems like you spent quiet a bit on this, still need more work to finish this? (and #9204)

PR #9191, opened 2024-06-29
Removed deprecated storedSession strategy

gautamsi:


packages/core/src/session/index.ts
```diff
@@ -123,36 +123,3 @@ export function statelessSessions<Session> ({
     },
   }
 }
-
-/** @deprecated */
```
gautamsi, line 127: @dcousens was this intentional? I can remove the documentation, for now I have moved this into redis example

Issue #9216, opened 2024-07-16
ui.hideCreate and ui.hideDelete don't work when graphql.omit.create and graphql.omit.delete are set


tv42:
```console
$ yarn create keystone-app
$ cd keystonejs-bug-hidecreate-hidedelete
$ npm run dev
```

Open browser. Create first user, navigate to `/posts`. Observe create button.

Edit `schema.ts`, add to `lists.Post` a section `ui: { hideCreate: true },`.
Observe browser page hide create button.

Add section `graphql: {omit: {create: true}},`.
Observe browser page start showing create button again.

Same is true for delete, with `ui: { hideDelete: true},` and `graphql: {omit: {delete: true}},`.

Create/delete should stay hidden whenever ui.hideCreate/ui.hideDelete is true. `graphql.omit.*` should only be a fallback, and not affect the hiding when the `ui.*` settings are explicitly set.

Current dependencies from `yarn create-app`:

```json
  "dependencies": {
    "@keystone-6/auth": "^7.0.0",
    "@keystone-6/core": "^5.0.0",
    "@keystone-6/fields-document": "^7.0.0",
    "typescript": "^4.9.5"
  }
```

Earlier related issue: #8714 (PR #8715).

tv42:
This logic is hilariously wrong. It says do NOT hide the button if graphql API is disabled. Ternary expressions claim another victim.

```javascript
      hideCreate: normalizeMaybeSessionFunction(
        list.graphql.isEnabled.create ? listConfig.ui?.hideCreate ?? false : false
      ),
```

Proposed fix:

```javascript
      hideCreate: normalizeMaybeSessionFunction(
        listConfig.ui?.hideCreate ?? !list.graphql.isEnabled.create,
      ),
```

That is, use whatever developer said explicitly; if no explicit configuration, hide when GraphQL API is not enabled.

Same for delete.

dcousens:
@tv42 nice. I'll make a pull request.

I don't blame the ternary expression though, I think our use of negative booleans is a pattern that repeatedly causes downstream errors.

PR #9217, opened 2024-07-17
Fix `list.ui.hide*` functionality

dcousens:
Resolves https://github.com/keystonejs/keystone/issues/9216
Thanks to @tv42 for reporting this and proposing a fix.

### DIFF
--- a/.changeset/fix-hide-ui.md
+++ b/.changeset/fix-hide-ui.md
@@ -1,0 +1,5 @@
+---
+'@keystone-6/core': patch
+---
+
+Fix `list.ui.hide*` defaulting to false when GraphQL is omitted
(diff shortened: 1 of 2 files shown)


packages/core/package.json
```diff
@@ -205,8 +205,8 @@
     "@keystone-ui/tooltip": "workspace:^",
     "@nodelib/fs.walk": "^2.0.0",
     "@prisma/client": "catalog:",
-    "@prisma/internals": "5.17.0",
-    "@prisma/migrate": "5.17.0",
```
dcousens, line 209: This is a follow-up to https://github.com/keystonejs/keystone/pull/9215, which missed some dependencies

dcousens:
Arguably having `ui.hideCreate` _and_ `graphql.omit.create` is silly.
Maybe we should have `graphql.omit.create = true | false | 'ui'` or something.

dcousens:
Thoughts @tv42?

dcousens:
Arguably having `ui.hideCreate` _and_ `graphql.omit.create` is silly.
As is, you can force `ui.hideCreate === false` and `graphql.omit.create: true`, and you'll have a broken create button in the admin.

Maybe we should have `graphql.omit.create = true | false | 'ui'` or something.

Thoughts @tv42?

PR #9204, opened 2024-07-08
Remove deprecated hooks

gautamsi:
This replaces all the hooks from original combined form with the granular one inline with #9056 #9057 #8826 

this cleans up the code and is breaking changes. 

- [x] Remove types for the combined form
- [x] Cleanup code of hooks
- [x] Update all Field code to use proper hooks
- [x] Fix tests
- [x] Update documentation #9231

PR #9231, opened 2024-07-25
Update hooks documentation to prefer new syntax

gautamsi:
for #9204 

@dcousens I am not very good at writing documentation, perhaps you can improve this if needed.


dcousens (approved):
Thanks @gautamsi, kick-starting these initiatives is really helpful and appreciated.
I think I have addressed a few nits, and a small fix ca289be.

Looks good to me!

PR #9186, opened 2024-06-25
Use app router template

gautamsi:
ref #9183

### Blocked by
- [x] #9189

This PR updates the template to generate the App router, default to typescript version which can be turned off by setting `config.ui.tsx = false`

## What is completed/Verified:
- [x] generate app routes in `app/(admin)` folder (creates admin route specific layout page. if `src` page exist, it generates it in `src/app/(admin)
- [x] Stop generating `next.config.js` file, this should be generated from `create-keystone-app` going forward (TODO)
- [x] ~~set default `ui.basePath` to `/admin`~~ can set `ui.basePath` to `/admin` or any other sub path
- [x] No longer copies admin files or generates pages export copy as they are no longer needed
- [x] Does not clear the admin folder files if exist, only overwrites the special config files which contains up to date admin meta and view cache, this should be removed before GA
- [x] Verified custom pages
- [x] Verified custom fields and field views
- [x] can use relative path or even "paths" from typescript (`@fields/text/myView`)
- [x] Added `--reset-admin` flag to `dev` command to force clean `(admin)` folder, useful in regenerate the admin template if updated in future. should never be needed except braking upgrade 
- [x] fix build and start script
- [x] `@keystone-6/auth` templates
- [x] page middleware

## Pending:

- [ ] Documentation
- [ ] fixing final set of failing tests

---

@dcousens  can you create a `v-next` branch which should be actual target for these PRs instead of `main`

---
For anyone interested in testing this out, I have published temporary packages (use at your own risk) I have been using it with very large project of mine.
`@k6js-next/core-next@0.0.0-20240905-01`
`@k6js-next/auth-next@0.0.0-20240905-01`

to use this you have to specify resolutions, for yarn I do it like this in (root package.json in monorepo)

```
"resolutions": {
    "graphql": "16.8.1",
    "next": "14.2.5",
    "react": "^18.3.1",
    "react-dom": "^18.3.1",
    "@keystone-6/core": "npm:@k6js-next/core-next@0.0.0-20240905-01",
    "@keystone-6/auth": "npm:@k6js-next/auth-next@0.0.0-20240905-01"
  }
```

PR #9235, opened 2024-07-25
Add support for developers to replace the default esbuild configuration

dcousens:
This pull request addresses the problem in https://github.com/keystonejs/keystone/discussions/8406 where the default `esbuild` configuration simply cannot account for how developers are integrating `keystone-6` into their projects, and thus, rather than relying on patching a package, maybe we can provide an appropriate escape hatch.

Feedback welcome before merge


acburdine:
this would definitely work for the use case I mentioned in #8406 - though I wonder if there isn't also a simpler way to solve certain use-cases without the need for an escape hatch. I've got another separate project that doesn't involve a monorepo per se - but it does use the tsconfig paths configuration pretty heavily to avoid `../../../../`-type imports, and that project also has to have a similar workaround to ensure those aliased local files get included.

What we're doing in our workaround in both projects is effectively making the keystone esbuild options respect the typescript config's `paths` setting, and ensuring any paths listed there are also included by esbuild. Something like [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) might be usable here - any paths in the list would be included.

The ability to completely customize the esbuild configuration would be useful, but it seems like it would introduce several new ways for devs to trip themselves up, especially if the default esbuild configuration were to be changed in any way in the future (i.e. if the output path/format/etc changed at all).

acburdine:
did some experimenting locally in our monorepo with `tsconfig-paths` - this setup seemed to work fairly well:

```ts
// packages/core/src/lib/esbuild.ts
import { loadConfig } from "tsconfig-paths";
import { type BuildOptions } from 'esbuild'

function getPathsMatcher(cwd: string): RegExp | null {
  try {
    const pathsResult = loadConfig(cwd);
    if (pathsResult.resultType !== 'success') return null;

    const pathMatchers = Object.keys(tsConfigPaths.paths).map((path) => path.replace(/\*/, '.*'));
    if (!pathMatchers.length) return null;

    return new RegExp(`^(?:${pathMatchers.join('|')})$`);
  } catch {}
}

export function getEsbuildConfig (cwd: string): BuildOptions {
  const pathMatcher = getPathMatcher(cwd);

  return {
    entryPoints: ['./keystone'],
    // ...rest of fields
    plugins: [
      {
        name: 'external-node_modules',
        setup (build) {
          build.onResolve(
            {
              // don't bundle anything that is NOT a relative import
              //   WARNING: we can't use a negative lookahead/lookbehind because esbuild uses Go
              filter: /(?:^[^.])|(?:^\.[^/.])|(?:^\.\.[^/])/,
            },
            ({ path }) => {
              f (pathMatcher?.test(path)) return;
              return { external: true, path }
            }
          )
        },
      },
    ],
  }
}

**Comment by gautamsi** [comment]
Does it help use path Aliases in custom admin pages as well?

`tsconfig-paths` may not work on custom admin pages as they are compiled in isolated context (`.keystone/.admin`). Once #9186 land it should work. I have tested that working without any esbuild changes. 

if https://github.com/keystonejs/keystone/pull/9235#issuecomment-2250205197 gets included, we may not even need the custom esbuild config, but we can still provide the support to load custom config

**Comment by acburdine** [comment]
> Does it help use path Aliases in custom admin pages as well?

since admin pages are compliled directly with next.js, I _think_ path aliases in admin pages should work out of the box already since Next.js supports it.

**Comment by marekryb** [comment]
> The ability to completely customize the esbuild configuration would be useful, but it seems like it would introduce several new ways for devs to trip themselves up, especially if the default esbuild configuration were to be changed in any way in the future (i.e. if the output path/format/etc changed at all).

I tend to agree. However, apart from import paths, I would like to be able to change `entryPoints` (eg. `src/keystone.ts` instead of default) and potentially `jsx` or `tsconfig` too. So how about providing the defaults through function like:
```typescript
// esbuild.config.js
export default function (defaultValues) {
  return {
    ...defaultValues,
    entryPoints: ['./src/keystone'],
  }
}
```
The `tsconfig-paths` thing seems great and can be added separately as an improvement. Should it be opt-in though? Just in case `tsconfig.json` has some strange things - eg. would the common monorepo pattern `"extends": "../../tsconfig.base.json",` work?

dcousens:
@marekryb that's solid suggestion, I'll update `esbuild.keystone.js` to be a function that accepts a `defaultValues`

dcousens:
@marekryb @acburdine @gautamsi I have added support based on @marekryb suggested, but additionally added support for `esbuild.keystone.ts`.  This was easy to add, and I couldn't foresee any reason why we shouldn't.

acburdine:
@dcousens i think this function approach makes sense - can always make an example showing how to do something like tsconfig-paths with the custom file 🙂

dcousens:
I'm happy for the scope of that example to be expanded on separately to this pull request, as the example already added shows how to use the functionality generally.

asgeo1:
I'm trying to get absolute paths working with Keystone. I came across this PR, and was hopeful it would work.

I've tried setting up `tsconfig-paths` with `esbuild.keystone.ts` here:

```
// esbuild.keystone.ts

import { loadConfig } from 'tsconfig-paths'
import { type BuildOptions } from 'esbuild'

function getPathsMatcher(cwd?: string): RegExp | null {
  try {
    const pathsResult = loadConfig(cwd)
    if (pathsResult.resultType !== 'success') return null

    const pathMatchers = Object.keys(pathsResult.paths).map(path => path.replace(/\*/, '.*'))
    if (!pathMatchers.length) return null

    return new RegExp(`^(?:${pathMatchers.join('|')})$`)
  } catch {
    return null
  }
}

export default function (defaults: BuildOptions) {
  const pathMatcher = getPathsMatcher(defaults.absWorkingDir)
  return {
    ...defaults,
    plugins: [
      {
        name: 'external-node_modules',
        setup(build: any) {
          build.onResolve(
            {
              // don't bundle anything that is NOT a relative import
              //   WARNING: we can't use a negative lookahead/lookbehind because esbuild uses Go
              filter: /(?:^[^.])|(?:^\.[^/.])|(?:^\.\.[^/])/
            },
            ({ path }: { path: string }) => {
              if (pathMatcher?.test(path)) return
              return { external: true, path }
            }
          )
        }
      }
    ]
  }
}
```

It does work for most of my code, but not for the custom Keystone admin pages, which seems explained by @gautamsi 's [comment](https://github.com/keystonejs/keystone/pull/9235#issuecomment-2250378352) above.

Given [PR#9186](https://github.com/keystonejs/keystone/pull/9186) is not yet merged, is there any other known work around  to get absolute paths working with custom Keystone admin pages?

gautamsi:
@asgeo1 I added info about temp package while we wait for merging the app router part. I am already using with a large project which also includes heavily customized pages, all works good.

PR #9236, opened 2024-07-25
Revert "Update hooks documentation to prefer new syntax"

gautamsi:
Reverts keystonejs/keystone#9231

@dcousens I accidently had this PR for `main` instead of `v-next`, have created another PR to `v-next` so that it lives there instead of main website


dcousens:
No, I think this is OK, remember that `main` has this functionality too, and the other hook types are _deprecated_.

We have often removed documentation for deprecated functionality, for better or worse.

PR #9247, opened 2024-08-01
Fix Astro and Remix examples not building

dcousens:
This pull request fixes the `examples/framework-astro` and `examples/framework-remix` examples which both had been suffering from `vite` build issues outside of our normal `keystone postinstall` checks.


examples/framework-astro/astro.config.mjs
```diff
@@ -1,12 +1,22 @@
 import { defineConfig } from 'astro/config'
+import node from '@astrojs/node'
 
 // https://astro.build/config
 export default defineConfig({
   output: 'server',
+  adapter: node({
+    mode: 'standalone',
+  }),
```
dcousens, line 9: An `adapter` is required for `output: 'server'`

examples/framework-astro/keystone.ts
```diff
@@ -18,7 +19,7 @@ export default config({
 
     // WARNING: this is only needed for our monorepo examples, dont do this
     //   we use using myprisma, not .myprisma, because vite
-    prismaClientPath: 'node_modules/myprisma/client',
```
dcousens, line 21: The nested nature of `/client` is not compatible with `vite`, we should probably roll-out this change to our `fixPrismaPath` utility generally

examples/framework-remix/schema.prisma
```diff
@@ -9,7 +9,7 @@ datasource sqlite {
 
 generator client {
   provider = "prisma-client-js"
-  output   = "node_modules/.myprisma/client"
+  output   = "node_modules/myprisma"
```
dcousens, line 12: Incompatible here too

PR #9248, opened 2024-08-01
Bump `@braintree/sanitize-url`

dcousens:
`jest` is currently failing when bumping `sanitize-url` as a result of https://github.com/braintree/sanitize-url/issues/78 - this pull request [will] fix that issue and moves us back to a caret range.

The problem is `jsdom` doesn't support `URL.canParse` for now


dcousens:
Replaced by https://github.com/keystonejs/keystone/pull/9259

dcousens:
May be resolved soon by https://github.com/braintree/sanitize-url/pull/80/

PR #9251, opened 2024-08-02
Update website roadmap

simonswiss:
CleanShot 2024-08-05 at 17 22 57@2x


kennedybaird:
That looks good - but one question, why wouldn't you highlight that there have been completion of features? I think just removing cards doesn't make the most sense.

I was thinking about different styling (eg, green label - maybe with a tick), and maybe linking to the release/commit?

Or another section could be even better, to highlight even more clearly?

Happy to help with either option

dcousens:
> why wouldn't you highlight that there have been completion of features

This is a good callout, and something we have neglected to do.
We probably need a features page.

kennedybaird:
@simonswiss - I did a [quick attempt](https://github.com/kennedybaird/keystonejs/commit/cebd644e6bd7b15692d1dbeebda6b44a40d7985e) at a completed indicator, tried to use all internal concepts

simonswiss:
@dcousens — the "Why Keystone" page has a ["what's in the box" section](https://keystonejs.com/why-keystone#features) that lists all the main features. Maybe new added features could be added there? 

CleanShot 2024-08-06 at 10 24 37@2x


The section is linked to from the homepage with a "See all Features" button:

CleanShot 2024-08-06 at 10 25 04@2x

kennedybaird:
Agree with your sentiment @simonswiss - I'm trying to consider / suggest how to make it clearer all the great energy that is being put into Keystone. It has been mentioned a lot across discussions, Slack, etc.

However I do disagree re no existing features on a roadmap / future page. A roadmap including recent "stops" / progress makes a lot of sense to me, as it makes it clear that progress is happening.

Anything that achieves people seeing the great work happening is good in my books, as long as it simplifies maintainer time - which is why I thought a simple completed flag could be a nice short-term solve

simonswiss:
Those are very valid points — thanks for pushing back!

Maybe a "Recently added" section that has the green ticks you prototyped would work well! I'll work on something.

simonswiss:
CleanShot 2024-08-06 at 14 30 12@2x

I think this would work pretty well? Implemented in [1a831da](https://github.com/keystonejs/keystone/pull/9251/commits/1a831daccb78147c391ba02e85807cceb0b90511)

kennedybaird:
> I think this would work pretty well? Implemented in [1a831da](https://github.com/keystonejs/keystone/pull/9251/commits/1a831daccb78147c391ba02e85807cceb0b90511)

Nice, I think that's great 👍

dcousens:
Thanks to each of you! :yellow_heart:

PR #9252, opened 2024-08-04
Fix edit links in docs/

kennedybaird:
These links haven't been working, I think I simplified a little by making sure we only set `isIndexPage` on the section homepages.


kennedybaird:
Realised this doesn't solve for updates / blog routes.

Thinking about this more, I would suggest removing the edit button from updates/roadmap/blog posts.

Happy to go either direction however.

simonswiss:
Nice one — thanks for picking that up!

I think we can fix the remaining ones fairly easily:

1) For the Blog, you could change the `editPath` in the `<BlogPage>` component in `app/(site)/blog/[post]/page-client.tsx` (line 24) from `docs/${params?.post}.md` to `blog/${params?.post}.md`.

2) `updates/roadmap/page.tsx` can take an `isIndexFile` to resolve to the correct edit path.

3) In the same vein of the changes you made in the PR, the `docs/guides/document-field-demo/page.tsx` needs to be updated with `isIndexPage` as well.

Would you like to do these changes well to get this PR merged?

If you don't have time, I can make those changes!

kennedybaird:
Great, LGTM @simonswiss :+1:

PR #9257, opened 2024-08-05
Update yarn commands with npm

simonswiss:
Yarn isn't the most common package manager these days, and I think it makes sense to not use it as the "go-to" in our CLI examples, including the main marketing landing pages.

Of course, folks are free to use yarn if they wish — but we shouldn't make it sound like you have to, nor that it is our recommendation.


dcousens (approved):
We use `pnpm` typically, but I concur that `npm` is what our documentation should refer to.

Issue #2480, opened 2020-03-04
Why the AdminUI supports only email/pass auth strategy?


chomamateusz:
I've added 2 auth strategies: Google and FB, with redirects URL `/auth/google` and `/auth/fb`. 

Next I've forked the AdminUI, and I've added 2 regular buttons with href to URLs mentioned above.

![FBandGoogleLogin](https://user-images.githubusercontent.com/20150525/75907281-53836e00-5e48-11ea-9f19-b51874653ac1.png)

Why this is not allowed by `new AdminUIApp` constructor? Only thing that UI needs in OAuth auth is the redirect URL for each method (and maybe button props) and it could simply generate this buttons for me.

I've not battle tested it yet, but for now as I'm testing Keystone possibilities, it works like a charm - I can log is through FB and Google.

Is that more complicated than I think? Or simply no one implemented it yet?

jesstelford:
> Why this is not allowed by new AdminUIApp constructor?

We literally just haven't gotten to it yet 😉

Would love to see a PR with some of these enhancements.

One thing to consider is; not every app will want every authentication method. And there may be apps which use a non-core authentication method (for example; I have a custom library which does a hacky passwordless thing that's WIP right now, maybe I want to be able to plug that straight into the Admin UI too).

dimaip:
Does anyone know if this is still the issue for Keystone-next or if its admin supports more auth methods now?

dcousens:
Please see our discussions on this topic, https://github.com/keystonejs/keystone/discussions?discussions_q=is%3Aopen+authentication and for specific feature requests

PR #9258, opened 2024-08-05
Diff graphql prisma types

dcousens:
This pull request changes some of the ordering in our GraphQL types, and begins to introduces a differentiation between GraphQL types and resolved (or Prisma) types in our generated Typescript types.

This change is important as begins to address a future where these types may not always match.
The reason for that will be evident in #2480, but additionally, it refactors the Typescript type generation to start accounting separate Typescript types for a `sudo()` context.

Primarily for now though, this pull request inlines parts of the `printGeneratedTypes`  and orders the outputs more consistently.


examples/document-field-customisation/keystone-server/schema.graphql
```diff
@@ -211,11 +211,11 @@ type Mutation {
 }
 
 type Query {
-  users(where: UserWhereInput! = {}, orderBy: [UserOrderByInput!]! = [], take: Int, skip: Int! = 0, cursor: UserWhereUniqueInput): [User!]
   user(where: UserWhereUniqueInput!): User
+  users(where: UserWhereInput! = {}, orderBy: [UserOrderByInput!]! = [], take: Int, skip: Int! = 0, cursor: UserWhereUniqueInput): [User!]
```
dcousens, line 215: This will be a noisy update for developers, but, this has been an inconsistency in our code for many moons, and is annoying to uphold.

packages/core/src/lib/typescript-schema-printer.ts
```diff
@@ -1,106 +1,60 @@
...
-      type.ofType,
-      scalars
-    )}`
-  }
+  if (type instanceof GraphQLList) return `ReadonlyArray<${printTypeReference(type.ofType)}> | ${printTypeReference(type.ofType)}`
 
   const name = type.name
   if (type instanceof GraphQLScalarType) {
-    if (scalars[name] === undefined) return 'any'
-    return `Scalars['${stringify(name)}']` // TODO: inline?
```
dcousens, line 51: I don't see why we needed to refer to this `Scalars` object continually for our resolved types; this isn't helpful to developers.

tests/cli-tests/__snapshots__/artifacts.test.ts.snap
```diff
@@ -84,7 +74,7 @@ export type TodoUpdateArgs = {
 }
 
 export type TodoCreateInput = {
-  readonly title?: Scalars['String'] | null
+  readonly title?: string | null
```
dcousens, line 77: This is easier on the eyes from a developer documentation point of view

gautamsi: I have been annoyed by this scalar in generating types to be used with separate frontend app.

Issue #9113, opened 2024-04-24
Cannot find module 'fp-ts/lib/Either'


ttbarnes:
Hiya, i've bumped the following packages to the [latest versions](https://github.com/keystonejs/keystone/releases/tag/2024-04-16), with node version `21.6.2`, NPM version `10.2.4`:

```js
"@keystone-6/auth": "^8.0.0",
"@keystone-6/core": "^6.0.0",
"@keystone-6/fields-document": "^9.0.0",
```

When running `npm run dev`, which has the following:

```js
"keystone dev --no-ui",
```

I get the following error, Keystone cannot start:

```js
> keystone dev  --no-ui

✨ Starting Keystone
(node:87386) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: Cannot find module 'fp-ts/lib/Either'
Require stack:
- /Users/tonyb/project/src/api/node_modules/io-ts/lib/index.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/fields-document/dist/validation-d47c8b80.cjs.dev.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/fields-document/dist/keystone-6-fields-document.cjs.dev.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/fields-document/dist/keystone-6-fields-document.cjs.js
- /Users/tonyb/project/src/api/.keystone/config.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/core/dist/createSystem-b211ce31.cjs.dev.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/core/scripts/cli/dist/keystone-6-core-scripts-cli.cjs.dev.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/core/scripts/dist/keystone-6-core-scripts.cjs.dev.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/core/scripts/dist/keystone-6-core-scripts.cjs.js
- /Users/tonyb/project/src/api/node_modules/@keystone-6/core/bin/cli.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1149:15)
    at /Users/tonyb/project/src/api/node_modules/next/dist/server/require-hook.js:54:36
    at Module._load (node:internal/modules/cjs/loader:990:27)
[...]

PR #9261, opened 2024-08-06
Fix missing `fp-ts` for `@keystone-6/field-document`

kennedybaird:
Adding peer dep for `fp-ts` for #9113

The schema for `examples/custom-esbuild` was out of date for some reason, so `pnpm i` was failing.


dcousens:
See https://github.com/keystonejs/keystone/pull/9263

PR #9263, opened 2024-08-06
Remove io-ts, use zod

dcousens:
Alternative to https://github.com/keystonejs/keystone/pull/9261

This pull request places the use of `io-ts` with `zod`, something I am planning to use internally elsewhere in the project.

### DIFF
--- a/packages/fields-document/src/DocumentEditor/component-blocks/api-shared.ts
+++ b/packages/fields-document/src/DocumentEditor/component-blocks/api-shared.ts
@@ -304,8 +304,8 @@ export type HydratedRelationshipData = {
 
 export type RelationshipData = {
   id: string
-  label: string | undefined
-  data: Record<string, any> | undefined
+  label?: string
+  data?: Record<string, any>
 }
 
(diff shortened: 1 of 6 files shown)


dcousens:
@kennedybaird thanks for your initiative on #9261, I will add you as credit for this pull request :yellow_heart:

kennedybaird:
Great work, I was going to suggest moving to zod!

packages/fields-document/src/DocumentEditor/component-blocks/api-shared.ts
```diff
@@ -304,8 +304,8 @@ export type HydratedRelationshipData = {
 
 export type RelationshipData = {
   id: string
-  label: string | undefined
-  data: Record<string, any> | undefined
+  label?: string
+  data?: Record<string, any>
```
dcousens, line 308: This is required as a result of `zod` assuming that `| undefined` is equivalent to `?:`, which it isn't.

Reference https://github.com/colinhacks/zod/issues/2464 and https://github.com/colinhacks/zod/issues/2598

Issue #8462, opened 2023-04-07
None of the codesandbox examples work 


JonasGruenwald:
This applies to all of the examples, when clicking the Codesandbox link it fails to build with an error.
Error from [this example](https://github.com/keystonejs/keystone/tree/main/examples/usecase-blog) is the following:

```
yarn run v1.22.19
$ keystone dev
✨ Starting Keystone
✘ [ERROR] Could not resolve "../example-utils"

    keystone.ts:2:30:
      2 │ import { fixPrismaPath } from '../example-utils';
        ╵                               ~~~~~~~~~~~~~~~~~~

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentationabout this command.
```

Codesandbox link:
https://githubbox.com/keystonejs/keystone/tree/main/examples/usecase-blog

husinzz:
in keystone.ts delete line 2, 11,12 in your sandbox, refresh the browser after the change

Thinkscape:
Keystone is currently broken with latest version of next #8575

kisankumavat:
@JonasGruenwald You have to copy this (https://github.com/keystonejs/keystone/blob/main/examples/example-utils.ts) file's content and create new file with the same name and put that file inside of root dir of that sandbox.

sadakura:
Still not fixed :(

dcousens:
@sadakura unfortunately we don't know of an easy fix for this at this time - open to any suggestions :)

sadakura:
> @sadakura unfortunately we don't know of an easy fix for this at this time - open to any suggestions :)

sorry if I have insulted someone, no I'm not similar with these things to help, just looked up for demo :) great project anyway.

JonasGruenwald:
The workaround suggested by @kisankumavat works fine btw.

I don't really care but I guess since `example-utils.ts` just exports a single line function it could just be added or in-lined into all the example repos?

dcousens:
@kennedybaird indeed, the trade-off is helping make my life easier as a maintainer by ensuring the examples stay up to date.

I'm happy to inline the `fixPrismaPath` code, but the version updates are a pain and I'm trying to simplify our release process.

PR #9266, opened 2024-08-07
wip - examples management script

kennedybaird:
Following up on #8462, I did a simple POC for a script that updates all the dependencies in the examples directories.

The intention is this would be triggered alongside `prepare-release.js`. It requires that the package versions are updated to work correctly.

`scripts/update-example-dependencies.js`:

1. iterates through the `packages/*` directories, getting the current package version for each
2. extracts all catalog versions from `pnpm-workspace.yaml`
3. iterates (recursively) through `examples/*` looking for package.json, and updating with the values from 1 / 2
   - if `@prisma/client` is updated, it triggers rebuilding GraphQL and Prisma schemas, which will avoid issues like 53112038c5249d34ec44201e570ce8ca9a78a7b0

To test:

1. checkout this branch
2. delete `examples/assets-local/schema.graphql`
3. run `pnpm i` - this will fail due to missing file above
4. run `node scripts/update-example-dependencies.js`
   - _note, this will trigger generating prisma for all examples, on consequent runs it would only trigger if prisma client is updated_
5. run `pnpm i` - all should be fine

To improve:

- I would solve for `fixPrismaPath` and `example-data` so that CodeSandbox works
- I would probably look at this making its own commit

Any feedback appreciated, I like these kinds of automations, but I know they're not popular with everyone - in my books they fall inside the reducing complexity category.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "eslint": "^9.2.0",
     "jest": "^29.0.0",
     "jest-environment-jsdom": "^29.0.0",
+    "js-yaml": "^4.1.0",
     "typescript": "^5.5.0",
     "typescript-eslint": "^7.8.0"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -122,6 +122,9 @@ importers:
       jest-environment-jsdom:
         specifier: ^29.0.0
         version: 29.7.0
+      js-yaml:
+        specifier: ^4.1.0
+        version: 4.1.0
       typescript:
         specifier: ^5.5.0
         version: 5.5.4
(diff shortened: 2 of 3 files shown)


dcousens:
We do have `@changeset/cli` in the repository, which manages this same automation when configured to - the problem has been misconfiguration.

> Any feedback appreciated, I like these kinds of automations, but I know they're not popular with everyone - in my books they fall inside the reducing complexity category.

I think they fall in the adding of complexity category :sweat_smile: , but I see `@changesets/cli` in the same boat

dcousens:
I'll see if I can't try and help this by verifying the current behaviour of `pnpm changeset version`

dcousens:
See https://github.com/keystonejs/keystone/pull/9269 for reverting everything for now, and _maybe_ fixing this for `pnpm changeset version`, but we'll see how that plays out

PR #9269, opened 2024-08-07
Revert packages to use nominal package versions

dcousens:
Alternative to https://github.com/keystonejs/keystone/pull/9266
Reverting https://github.com/keystonejs/keystone/pull/9112 and other pull requests.


.changeset/config.json
```diff
@@ -5,6 +5,9 @@
   "baseBranch": "main",
   "linked": [],
   "access": "public",
+  "privatePackages": {
+    "version": false
```
dcousens, line 9: This stops `examples/` from generating `CHANGELOG.md` and version bumps

package.json
```diff
@@ -53,6 +53,28 @@
...
+      "@keystone-ui/modals": "workspace:^",
+      "@keystone-ui/notice": "workspace:^",
+      "@keystone-ui/options": "workspace:^",
+      "@keystone-ui/pill": "workspace:^",
+      "@keystone-ui/popover": "workspace:^",
+      "@keystone-ui/segmented-control": "workspace:^",
+      "@keystone-ui/toast": "workspace:^",
+      "@keystone-ui/tooltip": "workspace:^"
+    }
+  },
```
dcousens, line 77: This is the bulk of what helps this work for development, while allowing the examples to be indepedent

examples/document-field-customisation/package.json
```diff
@@ -1,9 +1,8 @@
 {
   "name": "@keystone-6/monorepo-example-document-field-customisation",
-  "description": "Example to demonstrate customising document field with custom components",
-  "version": "0.0.1",
+  "version": null,
   "private": true,
-  "author": "Dinesh Pandiyan <flexdinesh@gmail.com>",
```
dcousens, line 6: @flexdinesh :yellow_heart: but we don't add this field for any of the examples, that is what `git` is for

dcousens:
@kennedybaird if you can follow up this work with a pull request that resolves the `fixPrismaPath` and bump CodeSandbox to Node 20, that would be really appreciated.

I think this configuration works well with `pnpm changeset version`, which is what we are using for release.

PR #9272, opened 2024-08-09
Docs: example links fixing + table of contents fixes

kennedybaird:
- removed loose `"` on deployment example links
- fixed the github example links
- removed some invalid examples / duplicates

- moved `extractHeadings` to a new `headings.ts` utils file to fix the table of contents for blog and docs

---

After looking through all the examples - there are a lot more than are documented. I didn't want to go through adding more and having the examples page get a lot more ungainly, but I did wonder about creating more sections?

Currently we have 3:

1. Standalone
2. End-to-End
3. Deployment

I think either having sub-sections in standalone, or just adding new sections could be beneficial:

- Fields
- Admin UI
- Session
- Database

I do at least like the idea of adding a `fields-` prefix to all the field examples, so viewing the examples dir is easier.

Happy to copy the above to a discussion and suggest some prefixes or organisation as well


kennedybaird:
@simonswiss - you may want to have a quick review as 1cbe903 touched on your recent work

dcousens:
@kennedybaird @simonswiss for examples, maybe there is an opportunity to have the README in each folder feed into Keystatic?

kennedybaird:
@dcousens - can you clarify why?

So that each example readme can be viewed without leaving docs? It seems like an unnecessary double up.

Or are you wanting some automated mirroring of the examples folder, so that we're not maintaining the docs side as well?

With prefixes in the folder names, and `displayName` + `description` set, the individual example package.json could be the single source of truth.

dcousens:
@kennedybaird the latter, I'm hoping we can effectively move the guides into markdown, in each example, and then we can use them for the documentation too.

The matter of maintaining `docs/` and `examples/` for each example is not pragmatic, and as shown, it will fall behind

simonswiss (approved):
LGTM 🚀

simonswiss:
Thanks for this work, @kennedybaird 🙏

Agreed that manually playing "catch up / keep up" for the Keystatic examples list is not ideal. As a first step I tried to re-create the existing setup (which was hardcoding the examples links in the template), but this is indeed probably not the final solution.

Probably outside the scope of this PR though — and I'll approve that as this fixes a whole lot of broken links.

kennedybaird:
Do you want me to have a look at a keystatic approach @simonswiss?

PR #9275, opened 2024-08-09
Fix schemas for `create-keystone-app`

dcousens:
Fixes the following error when running `npm create keystone-app`

> [!CAUTION]
> Your GraphQL schema is not up to date

We should try and prevent this problem from happening in the future


dcousens:
@iamandrewluca 

> Wouldn't it be better to remove the postinstall script in starter, or at least not run it on app creation?

Maybe this is a better first step than trying to put this into our CI for now

PR #9166, opened 2024-06-05
Update built-in fields to new validate hook syntax

acburdine:
closes #9165
- this updates the built-in fields using validateInput to use the new unified validation syntax, and handles merging with the old syntax to ensure user-land old syntax still works

PR #9278, opened 2024-08-09
Fix validation hook merge function passthrough

acburdine:
In #9166 a merge function was added to ensure that the deprecated valdiateInput/validateDelete syntax could be used by users alongside the newer validate syntax used by builtin fields.

Because we're passing through the user-specified hooks and only overriding the `validate` one, however, the deprecated hooks are still being passed through which leads to an error in the parseFieldHooks check later on.

This sets the deprecated fields to undefined in the merge function so that they don't get passed through from user hooks.

### DIFF
--- a/.changeset/fix-hooks-errors.md
+++ b/.changeset/fix-hooks-errors.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": patch
+---
+
+Fix blended usage of `{field}.hooks.validateInput` and `{field}.validate.*` resulting in an error
--- a/.changeset/long-bees-thank.md
+++ b/.changeset/long-bees-thank.md
@@ -1,5 +1,0 @@
----
-"@keystone-6/core": patch
----
-
-Handle combined use of old and new validate syntax
--- a/.changeset/long-bees-thank.md
+++ b/.changeset/long-bees-thank.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": patch
+---
+
+Handle combined use of old and new validate syntax
--- a/.changeset/spicy-shrimps-thank.md
+++ b/.changeset/spicy-shrimps-thank.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": patch
+---
+
+Fix validation hook merge function to not pass through deprecated hooks
--- a/.changeset/spicy-shrimps-thank.md
+++ b/.changeset/spicy-shrimps-thank.md
@@ -1,5 +1,0 @@
----
-"@keystone-6/core": patch
----
-
-Fix validation hook merge function to not pass through deprecated hooks
--- a/packages/core/src/fields/resolve-hooks.ts
+++ b/packages/core/src/fields/resolve-hooks.ts
@@ -28,13 +28,10 @@ function resolveValidateHooks <ListTypeInfo extends BaseListTypeInfo> ({
   validateDelete
 }: FieldHooks<ListTypeInfo>): Exclude<FieldHooks<ListTypeInfo>["validate"], Function> | undefined {
   if (!validate && !validateInput && !validateDelete) return
-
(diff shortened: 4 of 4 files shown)


acburdine:
@dcousens without this fix anyone who updates to 6.2.0 and uses the deprecated validate field hook syntax in conjunction with built-in field validation will get an error (kind of the inverse of the problem I had before trying to use new syntax 😅

kennedybaird:
@acburdine - That change LGTM - but I'm wondering if L#55 could also be an issue? Don't really have my head around it fully, but seems like that lets the old hooks through

acburdine:
@kennedybaird I don't believe L#55 should be an issue - the problem occurs when you mix the old validateInput syntax with the new unified validate syntax. In the case of L#55 (user hooks using deprecated syntax, no builtin hooks), there would only be the user hooks in the final hooks output (meaning there isn't a mix and thus no issue)

kennedybaird:
> @kennedybaird I don't believe L#55 should be an issue - the problem occurs when you mix the old validateInput syntax with the new unified validate syntax. In the case of L#55 (user hooks using deprecated syntax, no builtin hooks), there would only be the user hooks in the final hooks output (meaning there isn't a mix and thus no issue)

Ok, I'm wondering if we want to handle this for users during the transition? If I change the Person list in `examples/usecase-todo/schema.ts` to:

```js
  Person: list({
    access: allowAll,
    fields: {
      name: text({
        validation: { isRequired: true }, isIndexed: 'unique',
        hooks: {
          validateInput: ({ resolvedData, operation, addValidationError }) => {
            console.log('validateInput', operation, resolvedData)
            if (resolvedData.name === 'forbidden') {
              addValidationError('name')
            }
          },
          validate: {
            create: ({ resolvedData, addValidationError }) => {
              console.log('validate create', resolvedData)
              if (resolvedData.name === 'forbidden') {
                addValidationError('name_second_create')
              }
            }
          }
        },
      }),

      tasks: relationship({ ref: 'Task.assignedTo', many: true }),
    },
  }),
```

then run `pnpm dev`, and create a new person with name `forbidden` - it only ever uses the `validateInput` method (even if the order is swapped around).

Again, not sure if it's an issue - just noting in case it is. To me it seems a bit odd, I get no indication in my editor that this won't behave normally:

image

acburdine:
@kennedybaird hmm - that's a good point. That might be solvable with typings though - I can see if there's an easy typing solution to make it clear it's an either-or

acburdine:
hmm - so I added the type changes to make the different new/old syntaxes exclusive, but it does cause some issues for people defining custom fields (it currently breaks the custom field example)

there are a couple of options I can think of, depending on how long it'll be until there's a planned major release and the deprecated syntax will be removed:
1. leave the types as-is for now, maybe update the documentation to note that the different syntaxes are mutually exclusive
2. export the `mergeFieldHooks` or `resolveValidateHooks` function from packages/core/fields/resolve-hooks, allowing custom fields to more easily compose hooks (might necessitate moving the function to `core/src/types`, idk)

dcousens:
If we assume that fields are responsible [for now] to ensure only one syntax prevails, we can apply that to our fields in this function and move forward.
Inevitably they are "changed" internally to use the new syntax anyway, we can simply move that change higher.

acburdine:
Went ahead and removed the type change in favor of modifying the resolveValidateHook function to just merge the old and new syntax together - seems to me like that's the best approach until there's a breaking release

kennedybaird:
LGTM :rocket: nice @acburdine

kennedybaird (approved):

gautamsi:
I have some work done in #9204  which would completely remove the dual syntax and use one only, possibly we can add a check if people are using older syntax we throw error at run time instead of just type error.

dcousens (approved):

PR #9280, opened 2024-08-12
Add support for providing prisma transaction options

### LINKED ISSUE #9265: Transaction timeout
1. Init new project
2. Try to use transaction with slow database connection
3. It will timeout in 5 seconds which is default, to simulate this, add a sleep of few seconds.

We should be able to add a timeout option to `context.transaction` which is passed to `prisma.$transaction`
preferably allow all `prisma.$transaction` options which is passed directly to `prisma.$transaction` call

acburdine:
closes #9265
- add options support to context.transaction
- add prismaTypes field to TypeInfo for passing through the Prisma namespace of the generated client

I added the `prismaTypes` passthrough so that I could access the TransactionIsolationLevel enum, but i'm not sure if that's the best approach to do it. Happy to remove that bit if it doesn't make sense.


gautamsi (approved):

examples/custom-output-paths/my-types.ts
```diff
@@ -167,6 +167,7 @@ export type TypeInfo<Session = any> = {
     readonly Post: Lists.Post.TypeInfo<Session>
   }
   prisma: import('./node_modules/.myprisma/client').PrismaClient
+  prismaTypes: import('./node_modules/.myprisma/client').Prisma
```
dcousens, line 170: Couldn't the developer import this themselves?

dcousens (approved):
LGTM, I want to think about the `prismaTypes` situ, but I doubt there is a better proposal than this

PR #9285, opened 2024-08-12
wip - playing with a keystatic-based examples explorer

kennedybaird:
This is an idea for managing part of the examples with keystatic (just a rough quick example).

It would use/manage the README.md file for each example and adds the ability to define a preview file, and select related docs.

It has some important notes:

1. It needs to add a `.yaml` file in each dir - which could be confusing for end users (side note: maybe those files should have a commented `#keystatic file` at the start of them?)
2. The examples would need to move to `docs/examples` as keystatic can't access the dir above (or keystatic needs to move to the project root, but I don't think that's appropriate)

What I like about this:

- allows to customise the examples viewing experience on the website, whilst respecting the example files as the main source of truth
- selecting related docs seems useful
- using tags could allow for building a more robust explorer

image


kennedybaird:
@simonswiss - I'd love to keep playing with this idea, but will save putting time in until there's some feedback. Hope you can take a glance this week :pray:

kennedybaird:
I'd still be interested in following through with this - if it's a no go then I would go back to transitioning all the examples to the new codesandbox approach so that they're at least working.

PR #9286, opened 2024-08-13
Fix CodeSandbox examples

kennedybaird:
Using `examples/usecase-todo` to change the examples approach to solve for codesandbox issues.

1. Moving to devcontainers as codesandbox is stopping supporting sandbox.config.json
2. Hardcoding seed data in the example, whilst keeping the `seed-data` command
3. Setting prisma to inherit versions
4. Removing `postinstall` to avoid possible conflicts with prisma minor patch changes to their generator

To solve for the examples snapshot tests:

- leaving the schema's in the example's folder
- adding a new `test:monorepo-snapshot-test` command to replace `postinstall`
- added a new `test:examples-snapshot-tests` to root package.json which targets the above command recursively in all examples folders
- added a new `examples_snapshot_tests` test to ci to run the above command once

This is going to significantly reduce the time for actions as each of the smoke tests currently runs the postinstall for all the examples each time (~5m).

This approach also allows us to handle `framework-nextjs-two-servers` as well.


dcousens:
> Seed data - I think it's best if we just stop ignoring the example.db files

Honestly, I think this is a good approach.
We do this already, and it's solid.

The primary downside I can see is the lack of binary human readability.
We have `examples/script` for demonstrating how to seed data otherwise.

dcousens:
Maybe we can "generate" the `keystone.db` files deterministically from our root `scripts/` folder? Anything to give us meaningful diffs.

examples/usecase-todo/package.json
```diff
@@ -7,16 +7,15 @@
...
   },
   "dependencies": {
     "@keystone-6/core": "^6.2.0",
-    "@prisma/client": "5.17.0"
+    "@prisma/client": "*"
   },
   "devDependencies": {
-    "prisma": "5.17.0",
-    "tsx": "^4.0.0",
+    "prisma": "*",
```
dcousens, line 17: Won't this result in potentially incompatible versions running on CodeSandbox?

examples/usecase-todo/.codesandbox/tasks.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://codesandbox.io/schemas/tasks.json",
+  "setupTasks": [
+    {
+      "name": "Installing Dependencies",
+      "command": "yarn install"
```
dcousens, line 6: Can we prefer `pnpm` for our examples?

examples/usecase-todo/keystone.ts
```diff
@@ -1,15 +1,14 @@
 import { config } from '@keystone-6/core'
-import { fixPrismaPath } from '../example-utils'
 import { lists } from './schema'
 import type { TypeInfo } from '.keystone/types'
 
 export default config<TypeInfo>({
   db: {
     provider: 'sqlite',
-    url: process.env.DATABASE_URL || 'file:./keystone-example.db',
+    url: process.env.DATABASE_URL || 'file:./keystone.example.db',
```
dcousens, line 8:
```suggestion
    url: process.env.DATABASE_URL ?? 'file:./example.db',
```

Maybe this is a better name than `keystone.db`, and shorter too

examples/usecase-todo/package.json
```diff
@@ -7,16 +7,15 @@
     "dev": "keystone dev",
     "start": "keystone start",
     "build": "keystone build",
-    "postinstall": "keystone postinstall",
```
dcousens, line 10: > add the schema files to .gitignore so that they are always recreated on the respective machine, using the relevant prisma version from keystone

The downside to this, is we lose the nature of these examples capturing changes to schemas in situations of higher complexity than our unit tests can easily account for

examples/usecase-todo/package.json
kennedybaird: This seems to solve a couple of issues, especially when combining with removing and untracking the schema files.

It means that it's always respecting the version from `keystone/core` - this is one of the reasons code sandbox isn't working at the moment.

After fixing the prisma URL issue, the next issue was schemas out of sync, because the current version of keystone/core that gets installed (6.1), uses prisma 5.13 - since 5.13 the generator has changed a little.

kennedybaird:
> Maybe we can build the `keystone.db` files deterministically from our root `scripts/` folder?

A script to do this would be fine, if you were interested in that then I would be considering resurfacing my previous script from #9266 but removing the versioning.

An iterative script would solve for a few problems, but I went the other route based on the previous feedback.

examples/usecase-todo/package.json
kennedybaird: Could you elaborate? Doesn't generating them on the fly ensure that they're always in the correct "condition" and therefore true to what we're testing?

Unless I'm missing something, it definitely seems it solves for a couple of issues, minimises management overhead - the only caveat being that users can't browse the files until they've run dev or build.

dcousens: @kennedybaird when I'm developing new features or fixing bugs, changes in our the schemas generated by our examples are quickly detected from the CI errors produced - this has been a highly reliable canary in the coal-mine that what I am changing has behaviour impacts beyond the limited scope of the QA & testing I am applying.

Our unit tests often capture this change in behavior too, but, the expanded scope of the examples have captured nuanced situations which I don't often think about.

dcousens: Another way of framing this, our `examples/` are acting as snapshot tests

kennedybaird: Copy that - it makes sense. It's given me some other thoughts that I've noted, which I'll put somewhere else.

kennedybaird:
(original PR content)

> This would be my suggestion to solving CodeSandbox - and versioning with prisma in particular.
> 
> CodeSandbox was supposed to end support for `sandbox.config.json` at the end of '23, so moving to devcontainer is needed. Removing the `postinstall` command is needed to fix for the prisma versioning issue.
> 
> Seed data - I think it's best if we just stop ignoring the example.db files
> 
> To solve for schemas potentially being different based on the prisma version we:
> - set the prisma versions to `*` to inherit from `keystone/core`
> - add the schema files to `.gitignore` so that they are always recreated on the respective machine, using the relevant prisma version from keystone
> 
> If we're happy with those conditions, I can go through and update all the examples to this new approach.

kennedybaird:
@dcousens - I think I've found a reasonable solution, have updated the PR.

PR #9290, opened 2024-08-14
Fix telemetry disable command

DavidMulder0:


DavidMulder0:
Minor thought, the right thing to do might be to discard any telemetry data collected from the period this command was broken.

dcousens:
The `telemetry disable` command was broken in e69def247d76462c568695f6581ff4fda6ad2880 (by me, sigh) - but this is the first report we have had about this problem, which is interesting.

Affected versions are `>=6.1.0`, which was published on 30th of April 2024.

```bash
$ keystone telemetry disable
telemetry disable is an unknown command
```

As we want to err on the side of consent being missed, I think we could do the following:

- Delete project events where `@keystone-6/core === 6.1.0` and `@keystone-6/core === 6.2.0`
- Delete _ALL_ device events between `2024-04-29` (N-1) and `2024-??-??`

The nature of device events being unversioned however results in an ongoing problem that we would not know if device events are coming from a device with `@keystone-6/core === 6.1.0` or `@keystone-6/core === 6.2.0`. In light of that, I think the easiest conclusion is to shutdown the current `telemetry.keystonejs.com` endpoint and activate a **new, different endpoint** in a future pull request.

- [x] Shutdown [telemetry.keystonejs.com](telemetry.keystonejs.com/)
- [x] Retrieve final snapshot of the database, and destroy upstream copy
- [ ] Delete project events where `@keystone-6/core === 6.1.0` and `@keystone-6/core === 6.2.0`
- [ ] Delete device events >= `2024-04-29`
- [ ] Delete any events older than 1 year from the snapshot in [accordance with our telemetry privacy policy](https://keystonejs.com/docs/reference/telemetry#how-long-will-we-keep-this-information)
- [ ] Aggregate remaining data that has consent, and delete database

I'm open to feedback on this from the community, and thanks @DavidMulder0 for reporting this

packages/core/src/lib/telemetry.ts
```diff
@@ -51,38 +41,46 @@ function getTelemetryConfig () {
...
-        if (!existing) return
+      '^2.0.0': (store) => {
+        const existing = store.get('telemetry') as TelemetryVersion1
+        if (!existing) return // skip non-configured or known opt-outs
 
-        const replacement: Telemetry = {
-          // every informedAt was a copy of device.informedAt, it was copied everywhere
-          informedAt: existing.device.informedAt,
+        const replacement: TelemetryVersion2and3 = {
+          informedAt: null, // re-inform
```
dcousens, line 51: If someone has unknowingly missed an opt-out, this will re-inform them

packages/core/src/lib/telemetry.ts
```diff
@@ -51,38 +41,46 @@ function getTelemetryConfig () {
...
 
         store.set('telemetry', replacement)
       },
+      '^3.0.0': (store) => {
+        const existing = store.get('telemetry') as TelemetryVersion2and3
+        if (!existing) return // skip non-configured or known opt-outs
+
+        store.set('telemetry', {
+          ...existing,
+          informedAt: null, // re-inform
```
dcousens, line 81: If someone has unknowingly missed an opt-out, this will re-inform them

dcousens:
Merged as 830d46dcb

PR #9302, opened 2024-08-20
Update CodeSandbox versions, and fix prisma client paths for `examples/` (continuation of #9300)

dcousens:
#9300 missed a few paths


examples/custom-output-paths/schema.graphql
```diff
@@ -1,232 +0,0 @@
-# This file is automatically generated by Keystone, do not modify it manually.
```
dcousens, line 1: Unrelated, but `schema.*` is not used by the `custom-output-paths` example

PR #9318, opened 2024-08-31
Initial support for mongodb

arihantdaga:
It's basically working, need to test more. 
Few more things that we have to do 
- [ ] Support embedded fields in mongodb
- [ ] Clean mongodb fields in packages
- [ ] Test more complex scenarios with apis


Mod5ied:
@arihantdaga How is this approach working out for you.
Any significant progress yet?

PR #9326, opened 2024-09-12
Fix URL validation by preserving percent-encoded characters

kidneyweakx:
This issue where the `@braintree/sanitize-url` package automatically decodes certain percent-encoded characters (e.g., %E5%BD%93%E5%89%8D into 当前). This causes the sanitized URL to differ from the original, which can result in false validation failures for URLs containing encoded Unicode characters.

To resolve this:

- Updated the isValidURL function to use encodeURI, ensuring that percent-encoded characters are properly maintained.
- The function now checks both the original and encoded sanitized URLs, improving validation accuracy.

This fix ensures that URLs with percent-encoded Unicode characters are considered valid if they match either the original or the encoded version of the sanitized URL.

Test Cases:
- Verified that both valid and invalid URLs are correctly handled after this update.

Before
Before
After
After

Let me know if you need any further changes!

### DIFF
--- a/.changeset/old-dolphins-obey.md
+++ b/.changeset/old-dolphins-obey.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/fields-document": patch
+---
+
+Fixes URL validation bug by using `encodeURI` to preserve percent-encoded characters during validation.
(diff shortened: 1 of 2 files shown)


dcousens:
@kidneyweakx can you please add a changeset?

kidneyweakx:
> @kidneyweakx can you please add a changeset?

Added.

dcousens (approved):

Issue #9347, opened 2024-10-02
maxFileSize not working


jj-matos:
I have set maxFileSize to 20GB. 
![image](https://github.com/user-attachments/assets/5df693a1-de23-49bf-821f-4f3d1a88b465)

Got this error on trying to upload a 1.5GB file
![image](https://github.com/user-attachments/assets/854680d5-ae95-4701-b231-e951cf5cf6fc)

Seems the limit stays at 200MB.

Our expectation is to be able to set a higher upload limit. Specifically 20GB.

I also tried to set a lower limit of 20MB and upload a 30MB file to check if it was just an upper limit issue, but it allowed to upload the 30MB file. So it seems that maxFileSize is not working at all. No matter the value is set, the limit is always 200MB.

Keystone 6
node v18.19.0

Any workaround this issue?

gautamsi:
is there a proxy in between? then you have to change this on proxy as well. if not, try this
``` ts
  server: {
      maxFileSize: 100 * 1024 * 1024,
      extendExpressApp(app) {

        app.use(bodyParser.json({ limit: '20gb' }));
      },
    },
```

dcousens:
Fixed in https://github.com/keystonejs/keystone/pull/9348

jj-matos:
Thanks for the feedback.
No proxy. It doesn't work either in the dev environment in my pc nor in production.
Tried your solution. Not working either.
I am not using any of the suggested storage. Either local nor S3.
I am running a script in a resolveInput hook.
Before it just gave an error reaching the 200MB and wouldn't upload.
I went into node_modules and commented out the error trigger:
node_modules/graphql-upload/processRequest.js - line 249
It downloaded the file, but still only up to 200MB.
Unfortunately, your solution didn't change anything.
Still the same.
Being so, I request to reopen this issue, since it is not solved.

jj-matos:
I was using keystone 6.1.0
Update to 6.3.0 and also fields-document and auth to the latest versions.
Deleted node_modules, yarn.lock and yarn cash.
Installed everything.
Keystone installation failed with graphql version conflict.
Tried resolution to 15.9.0, failed with QUERY error.
Installation worked with 16.9.0.
However, maxFileSize continues not working.

jj-matos:
Again, I believe this is not solved.
I looked into the patch code.
Changed node_modules core file default.ts and nothing changed.
Then changed this on node_modules/graphql-upload/processRequest.js:
fileSize: maxFileSize,
[...]

PR #9348, opened 2024-10-02
Fix `maxFileSize` parameter being ignored

dcousens:
This pull request fixes https://github.com/keystonejs/keystone/issues/9347, the problem is as simple as the value not being passed through


packages/core/src/lib/defaults.ts
```diff
@@ -130,7 +130,7 @@ export function resolveDefaults <TypeInfo extends BaseKeystoneTypeInfo> (config:
     },
     lists: injectDefaults(config, defaultIdField),
     server: {
```
dcousens, line 132:
```suggestion
    server: {
      ...config.server,
```

In the event other fields are added but not accounted for here

PR #9395, opened 2024-11-11
fix: fix default open for field group

artemSr:


dcousens:
@artemSr if you want to submit this as an option for the `group` function, maybe as `ui.openByDefault` or similar, please do

PR #9401, opened 2024-11-19
Fix search filters and add support to search by relationships in the list view

dcousens:
This pull request fixes the search functionality in the AdminUI by ensuring that the `list` configuration for `ui.listView.searchFields` is used when searching relationship fields within the list view, *and* in the relationship field field view. 

This work builds on https://github.com/keystonejs/keystone/pull/8966 which was started by @pnodet and is therefore jointly attributed.


.changeset/fix-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-6/core": patch
+---
+
+Fix AdminUI list view ignoring `.ui.searchFields` when searching
```
dcousens, line 5:
```suggestion
Fix list view ignoring `.ui.listView.searchFields`
```

PR #9402, opened 2024-11-21
FIX: DB API should use the types from graphQLSchemaSudo

mikehazell:
[The keystone List config docs say this](https://keystonejs.com/docs/config/lists#graphql:~:text=omit,your%20GraphQL%20API):
omit (default: undefined): Allows you to configure which parts of the CRUD API are autogenerated for your GraphQL API …
.. but this also impacts the CRUD operations of the DB API.

Given list config like this:
```ts
const SomeList = list({
  graphql: {
    omit: {
      create: true
    },
  },
  fields: {
    someFieldName: text({
      label: '...',
    }),
  },
});
```

I get Type errors when I try to call this

```
context.db.SomeList.create({
  data: { ... }
});
```

The same limitation applies on a field level.

### DIFF
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -83,7 +83,7 @@ export async function generateArtifacts (cwd: string, system: System) {
 
 export async function generateTypes (cwd: string, system: System) {
   const paths = getSystemPaths(cwd, system.config)
-  const schema = printGeneratedTypes(paths.types.relativePrismaPath, system.graphQLSchema, system.lists)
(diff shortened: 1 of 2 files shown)


mikehazell:
@dcousens for the record. The type error is not  "the method is 'undefined'".

I get this:
```
Type '{ ... input data  ... }' is not assignable to type 'never'.ts(2322)
context.d.ts(107, 18): The expected type comes from property 'data' which is declared here on type '{ readonly data: never; }'
```

PR #9452, opened 2025-01-21
Fix long identityField values from pushing out SignoutButton component

Octopixell:
Currently if the `identityField` contains a long value; it will push the SignoutButton out of its wrapping `<div>` element and the user can no longer see it. In order to fix this I've added some styles to the wrapping and containing elements to provide a better user experience.

Old situation:

image

New situation:

image

This issue was initially reported on Slack by Joakim Trulsson: https://keystonejs.slack.com/archives/C01STDMEW3S/p1737446375987869


dcousens:
Should be resolved with #9253

PR #9460, opened 2025-02-03
Rename `graphql` export to `g`

emmatown:
This renames the `graphql` export to `g` for two reasons:
- More concise reading/writing schemas
- To avoid conflicts with exports named `graphql` from libraries such as Relay, GraphQL Code Generator and gql.tada that use the variable name `graphql` to declare GraphQL operations

The `graphql` export still exists but is now deprecated and may be removed in a future release.

To quickly update usages of the `graphql` export in your project to use `g`:

1. Navigate to `node_modules/@keystone-6/core/dist/declarations/src/types/schema/legacy-alias.d.ts` in your editor ("Go to Definition" will not take you to the correct file)
2. Use "Rename Symbol" to rename `graphql` to `g`, this will update usages of `graphql` to `g`
3. Change this deprecated alias back from `g` to `graphql` (avoid using "Rename Symbol" again because you want to preserve the updates you made in step 2)

---

- Related graphql-ts PR: https://github.com/Thinkmill/graphql-ts/pull/19


.changeset/lemon-oranges-promise.md
```diff
@@ -0,0 +1,7 @@
+---
+"@keystone-6/fields-document": major
+"@keystone-6/cloudinary": major
+"@keystone-6/auth": major
+---
+
+Use `g` export instead of `graphql` from `@keystone-6/core`. The only breakage is that you must be on the version of `@keystone-6/core` that includes the `g` export or newer
```
emmatown, line 7: Note that this changeset has major changes but the `@keystone-6/core` changeset is only a minor since it's not breaking to add the new name and keep the old one around but _using_ the new name is breaking since people _must_ update `@keystone-6/core` otherwise these packages will not work. While this would be easily avoidable by using the old version, I don't think that complexity is worth it since people will already be forced to use all the new versions together because of the changes for `@keystar/ui`.

packages/core/src/types/schema/legacy-alias.d.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @deprecated
+ */
+// this is what you should run "Rename Symbol" on
+// when that is done, you should change it back to `export import graphql = _graphql;`
+// avoid using "Rename Symbol" to revert it because you want to
+// preserve the updates you made to the usages of `graphql`
+//            |||||||
+//            vvvvvvv
+export import graphql = _graphql
```
emmatown, line 28: If you're curious why I'm using this weird syntax, see the latter part of the PR description of https://github.com/Thinkmill/graphql-ts/pull/19

PR #9474, opened 2025-02-13
Typo in Astro integration example

emilshr:
Hello maintainers,
I observed a small typo while I was browsing through the readme of the astro integration example. Fixed it.

Thanks.


dcousens:
Thanks @emilshr!

PR #9486, opened 2025-02-17
Error on passwords that would be truncated by bcrypt

emmatown:
This pull request adds length validation to the `password` field in `@keystone-6/core` to ensure passwords do not exceed 72 bytes when using the default `bcrypt` key derivation function. 

This prevents silent truncation by the bcrypt algorithm. 
Tests have been added to verify the behavior.

This update follows concerns raised by a recent Okta incident where improper handling of the input length to bcrypt allowed an authentication bypass in particular situations ([source](https://www.theverge.com/2024/11/1/24285874/okta-52-character-login-password-authentication-bypass)). 

This flaw is additionally present in `bcryptjs`, highlighting similar concerns for downstream users of our `password` field when using the default key derivation function. 

We are not currently opting to publish a CVE, as whether this is a problem strongly depends on how the `password` field is being used downstream. 

If you feel differently, please comment.

### DIFF
--- a/tests/api-tests/fields/types/password.test.ts
+++ b/tests/api-tests/fields/types/password.test.ts
@@ -42,7 +42,7 @@ test('password longer than 72 characters is rejected', runner(async ({context}) 
   }
 }))
 
-test('password that is 72 utf-16 code units but 76 utf-8 code units is rejected', runner(async ({context}) => {
+test('password that is 72 utf-16 code units but >72 utf-8 code units is rejected', runner(async ({context}) => {
   const password = 'a'.repeat(70) + '❤️'
   expect(password.length).toBe(72)
   expect(new TextEncoder().encode(password).length).toBe(76)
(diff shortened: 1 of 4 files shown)


tests/api-tests/fields/types/password.test.ts
```diff
@@ -9,3 +12,45 @@ filterTests(password(), match => {
...
+  try {
+    await context.db.User.createOne({data: { password: 'a'.repeat(73) }})
+    expect(false).toBe(true)
+  } catch (error: any) {
+    expect(error.message).toEqual(`An error occurred while resolving input fields.
+  - User.password: value must be no longer than 72 characters`)
+  }
+}))
+
+test('password that is 72 utf-16 code units but 76 utf-8 code units is rejected', runner(async ({context}) => {
```
dcousens, line 45:
```suggestion
test('password that is 72 utf-16 code units but >72 utf-8 code units is rejected', runner(async ({context}) => {
```

PR #9525, opened 2025-03-03
Fix afterOperation hooks in one-to-one relationships with foreign keys

lushkovsky-s:
## Description
This PR fixes an issue where afterOperation hooks weren't properly triggered on both sides of one-to-one relationships with `db: { foreignKey: true }`. Previously, when updating a relationship from the side without the foreign key, only the hooks for that list were triggered, while the actual database change occurred on the list with the foreign key.

## Related Issues
Issue #9433

### DIFF
--- a/packages/core/src/lib/core/mutations/index.ts
+++ b/packages/core/src/lib/core/mutations/index.ts
@@ -193,7 +193,17 @@ async function updateSingle(
   })
 
   // after operation
-  await afterOperation(updatedItem)
+  if (updatedItem) {
+    await runSideEffectOnlyHook(list, 'afterOperation', {
+      context,
+      listKey: list.listKey,
+      operation: 'update' as const,
+      originalItem: item || updatedItem,
+      item: updatedItem,
+      inputData: inputData || {},
+      resolvedData: {},
+    })
+  }
 
   return updatedItem
 }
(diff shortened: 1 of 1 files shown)


emmatown (commented):
I'm not sure this makes sense to do. While the behaviour of relationships only running hooks on the local list (with the exception of nested creates going through the normal create flow) can certainly be a bit confusing, I think it's clearer for the behaviour to be exactly that rather than adding more exceptions like running only `afterOperation` and only for one-to-one relationships. Also, the performance impact of fetching the related items to pass into the hooks when it will often be unnecessary doesn't seem worth it.

stuible:
The performance concerns are very valid but I, for one, would love some way to enable this behaviour for related lists.  Opt-in?   I just had this issue come up in my own codebase today.

dcousens:
@stuible I concur this can be a foot-gun if you're not thinking about it from day zero.

Maybe we can write up an `example/relationship-hooks` that shows a common usecase that is confusing, and then look into how that could be improved?

plap979:
If you look at my example, the hook fires (telling us something changed) but `item` and `originalItem` are actually equal — so it doesn't actually tell us what changed.

On the official reference about Configuration > Hooks > [afterOperation](https://keystonejs.com/docs/config/hooks#after-operation), speaking about `item` and `originalItem`, I can read:

> This object is an internal database item. [DB API](https://keystonejs.com/docs/context/db-items) for more details on internal database items.

So my understanding of docs is that they are about internal database items, not action triggered from the API... To me this fix doesn't look like one more exception, but a step toward consistency with the already-public documentation. 

I discovered this while I was working on hooks to feed a Change Data Capture (CDC) topic on a message broker system. The day after Dec 12, 2024, I had to workaround with very inefficient code to allow reliable CDC. My workaround need to track changes from `beforeOperation`. It needs a query during `beforeOperation` to fetch the `originalItem`s and another later from `afterOperation` to capture the data change. But if Keystone already has this data internally, couldn't it do it in a single query?

Thank you so much to have spent time on it.

dcousens:
@plap979 `inputData` is typically the information that changes, is that what you wanting?
`item` and `originalItem` will generally not be equal if something changed.

plap979:
> @plap979 `inputData` is typically the information that changes, is that what you wanting?

I just need to produce side effects when data changes, namely when records has been created/updated/deleted to the database. I think `afterOperation` is there for this. [Docs](https://keystonejs.com/docs/config/hooks#after-operation) say:

> The afterOperation hook is used to perform side effects after the data has been saved to the database (for a create or update operation), or deleted from the database (for delete operations).

Hence `afterOperation` are about operations on tables? I suspect yes, since the docs tells about operations that affect the database... 

If `afterOperation` are for tables: then the wrong hook is called (half the cases). I expect only one invocation on the hook of the affected table (that with `foreingKey: true`).

If `afterOperation` are for lists: then only one hook is invoked. I expect invocations on both hooks since the change is effecting both lists.

But, never mind, I'm ok with my work around. Thanks so far, I'm really happy I had chance to do something.

Issue #9419, opened 2024-12-08
S3 file upload bug when same file name re-uploaded


gautamsi:
* configure file upload with S3
* set preserve  flag to false (default)
* set `transformName` to `(filename) => filename.replace(/[^a-zA-Z0-9.]/g, '-')`  - essentially just return original file name
* upload file twice
* notice the file is actually deleted on s3

File should be replaced and not deleted

### Workaround
set `preserver` to true so that file is kept, but that could be unintended consequences.

gautamsi:
possible improvement should be to add config which throws error on same file being uploaded again, silently upload if the flag is false (default)

dcousens:
alternatively, the delete could happen before the upload

dcousens:
> possible improvement should be to add config which throws error on same file being uploaded again, silently upload if the flag is false (default)

that could probably be a validate hook, and left to developers

gautamsi:
if you delete before and the update failed then you are left with nothing :)

I would run delete if the file name was not same

dcousens:
this would introduce a HEAD request for every upload, but I don't think we have any other option

gautamsi:
We can check file name locally from existing record before deleting.

willypuzzle:
Hello, I wish to work in this issue. What's the procedure? Can you assign this task to me?

Thanks!

dcousens:
@willypuzzle pull requests accepted 💛💙 - other than that, if the pull request resolves the problem in a way that reaches consensus, I'll do everything I can to help the pull request land

willypuzzle:
@dcousens according to [this](https://github.com/keystonejs/keystone/blob/main/packages/core/src/types/config/index.ts#L59-L95) `transformName` was removed from s3 config type storage. In fact it doesn't work anymore in my local machine. How can I workaround?

dcousens:
@willypuzzle keep following the type, see `& FileOrName` https://github.com/keystonejs/keystone/blob/main/packages/core/src/types/config/index.ts#L26

willypuzzle:
> [@willypuzzle](https://github.com/willypuzzle) keep following the type, see `& FileOrImage` on the type you referenced.
> 
> https://github.com/keystonejs/keystone/blob/main/packages/core/src/types/config/index.ts#L26

@dcousens yes, you're right. My fault. I reproduced the bug. I'm going to work on it.

willypuzzle:
[...]

PR #9529, opened 2025-03-04
Refactor image and file fields

emmatown:


gautamsi:
@emmatown since we are here again refactoring it, can you help us with something to create custom storage easily? Like if I want to have Azure storage or Google cloud storage?

It would be good to be able to register custom storage and use it with the file/image field. 

I have been using my own field for s3 which generates 3-4 copy of image processed with sharp to help reduce any compute cost in dynamic image optimization. - https://github.com/keystonejs-contrib/k6-contrib/tree/main/packages/fields-s3images

do you see a way to do that with this new refactoring?

dcousens:
This pull request will additionally add an example for doing a few of the common ones

gautamsi:
> > It would be good to be able to register custom storage and use it with the file/image field.
> 
> This PR is doing exactly that

are we still keeping multiple fields in database? (<field>_id, <field>.filename, <field>.extension) etc? I was thinking more like that there is no filtering in this field, we should convert into JSON field internally. This will help adding more data to the field like I can add multiple image size in my custom implementation.
 
At least add a meta field `<field>.meta` so that custom implementation can have flexibility.

dcousens:
Interesting idea @gautamsi, but as a breaking change that would be a pretty significant hurdle for many migrations compared to this thus far.

gautamsi:
@dcousens I understand that migrating such data structure to json will be cumbersome. Adding `_meta` field should help with custom implementation

dcousens:
How are you thinking the `_meta` field would work for now?

gautamsi:
I can save additional information specific to provider or save more information like image metadata (orientation, geocoding etc.) and use that in my frontend/workflow.

dcousens:
OK, so you want a `meta` JSON sub-field on `file` and `image` that you could interact with arbitrarily in `resolveInput` hooks?

marekryb:
These are all great changes! Also the "node_modules black hole" gets smaller by removing the direct dependency on aws-sdk!

Since we're already making breaking changes, could we also address the update problem with these fields? This issue is related to https://github.com/keystonejs/keystone/issues/9419, but the problem is more general.

Currently, the order of operations is:

    update → beforeOperation hook (delete old) → transformName → put (upload) → update DB

However, uploading is a fragile operation and can fail, leaving us with nothing.
A simple solution would be to change the order to:

    update → transformName → put (upload) → update DB → afterOperation hook (delete old)

With this approach, we would need to enforce the limitation that transformName must produce a unique string. This is already the case unless the user provides a custom implementation.

If this limitation is not practical, the only alternative I see, would be to change the interface. For example, we could introduce a differentiating argument for `transformName` and/or `storage.delete` like version, operationType or some unique identifier.

gautamsi:
> OK, so you want a `meta` JSON sub-field on `file` and `image` that you could interact with arbitrarily in `resolveInput` hooks?

exactly.

marekryb:
Thank you @emmatown :heart: Simple and elegant solution.

dcousens:
@gautamsi I don't think I have an answer to your meta field issue as part of this pull request

gautamsi:
would you accept a pull request?

dcousens (approved):

caronbletzer:
For future versions, it would be good to consider adding an option to field config that allows defining which file types are accepted for both file and image fields. During penetration testing of our keystone app, one of the things that was flagged was files of any type could be uploaded.

dcousens:
@caronbletzer we opted to do that for images, but honestly I think it might be too prescriptive - maybe our examples could show it as a best practice?

rburgst:
@dcousens when will this be released?

dcousens:
@rburgst I will keep my answer updated in https://github.com/keystonejs/keystone/discussions/9142#discussioncomment-15932609 rather than following up here

PR #9541, opened 2025-03-11
Remove JSON polyfill since Prisma now supports the Json scalar for SQLite

emmatown:
This pull request removes the JSON polyfill for SQLite and moves to Prisma's native support for the `Json` scalar type universally.

Unfortunately, as a result of https://github.com/prisma/prisma/issues/26571 we can't rely on Prisma to support default values for SQLite for now.

However, rather than continue the string-based polyfill only for native `defaultValue`s, we are happy to defer that functionality upstream while maintaining `defaultValue` support through Keystone (not Prisma).


dcousens (approved):

dcousens (approved):

emmatown:
Needs https://github.com/prisma/prisma/issues/26571

PR #9568, opened 2025-03-26
Fix `initFirstItem` fields not applying on InitPage

dcousens:
No changeset required, as this hasn't been publicly released


packages/auth/src/pages/InitPage.tsx
```diff
@@ -20,12 +20,10 @@ function InitPage({
   authGqlNames,
   listKey,
   fieldPaths,
-  enableWelcome,
```
dcousens, line 23: Unused

PR #9579, opened 2025-04-02
Remove session end/start from core and provide generic request and response Headers on context instead of node-specific req/res

emmatown:
The current Keystone session API that is provided to config looks like this

```tsx
export type SessionStrategy<
  Session,
  TypeInfo extends BaseKeystoneTypeInfo = BaseKeystoneTypeInfo,
> = {
  get: (args: { context: KeystoneContext<TypeInfo> }) => Promise<Session | undefined>
  start: (args: { context: KeystoneContext<TypeInfo>; data: Session }) => Promise<unknown>
  end: (args: { context: KeystoneContext<TypeInfo> }) => Promise<unknown>
}
```

and then it's provided on the `KeystoneContext` as `sessionStrategy`. `@keystone-6/core` only uses the `get` function though. The `start` and `end` functions are only either used by `@keystone-6/auth` or they're called directly by users in their resolvers or etc. If people are using next-auth, passport or etc. these `start` and `end` functions go unused since the session is started/ended elsewhere.

So this PR replaces the `SessionStrategy` with a `getSession` function in `@keystone-6/core` and the `SessionStrategy` API and `statelessSession` and `storedSessions` implementations have been moved into `@keystone-6/auth`.

Combined with #9578, this also improves the type-safety & flexibility around sessions with `@keystone-6/auth` accepting a `getSession` function like `@keystone-6/core` does except that it's provided `data` with the item id that the session is for.
```tsx
  getSession: ({ context, data }) =>
    context.query.User.findOne({
      where: { id: data.itemId },
    }),
```
---
Supersedes #8097 

Stacked on #9578


emmatown:
Replaced with #10028 and #10029

PR #9582, opened 2025-04-02
Fix multiple filter values for the same property and pass them as an AND query to GraphQL

kamilbiela:
The current implementation doesn't work with multiple filters for the same property. 
It's possible to add ">" and  "<"filters for the value on UI, but only the last filter is sent to the backend (see screenshots).
This PR is rough code that fixes the issue.

image
*Notice that viewCount in DevTools has only one filter value*

Screenshot From 2025-04-02 21-41-23
*After applying this PR changes*

### DIFF
--- a/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/useFilters.tsx
+++ b/packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/useFilters.tsx
@@ -30,16 +30,13 @@ export function useFilters(list: ListMeta) {
     const filters: Filter[] = []
     for (const key in query) {
       const filter = possibleFilters[key]
+      if (!filter) continue
       const val = query[key]
-      if (filter && typeof val === 'string') {
-        let value
-        try {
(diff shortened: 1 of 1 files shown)


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/useFilters.tsx
```diff
@@ -42,15 +42,16 @@ export function useFilters(list: ListMeta) {
...
         list.fields[filter.field].controller.filter!.graphql({
           type: filter.type,
           value: filter.value,
         })
       )
-    }, {})
+      return acc;
+    }, {AND: []})
+
     if (list.isSingleton) return { filters, where: { id: { equals: 1 }, AND: [where] } }
```
dcousens, line 55:
```suggestion
    if (list.isSingleton) return { filters, where: { id: { equals: 1 }, AND: where.AND } }
```

dcousens:
No changeset required

dcousens:
Thanks for finding, reporting and fixing this @kamilbiela 🩵

PR #9600, opened 2025-05-01
Refine field value types in field hooks

dcousens:
This pull request updates field hook types throughout Keystone to appropriately refine the field-level type information without requiring the `ListTypeInfo` for that particular list.

Hook arguments now include:

- `itemField`: the value of `item[fieldKey]`
- `inputFieldData`: the raw input value for the field
- `resolvedFieldData`: the post-resolution value passed to Prisma

Additionally `ui.fieldMode` and `ui.fieldPosition` functions are now passed `itemField` and `fieldKey`. 

Using these parameters should now be preferred compared to patterns like `resolvedData[fieldKey]`, which could not readily infer the type of the value as the exact `fieldKey` type is a sum type.

The only caveat for `itemField` is that it may not always be defined for complex field types like relationships and fields with no direct mapping between the prisma type and the field key.

### DIFF
--- a/.changeset/add-itemfield-to-hooks.md
+++ b/.changeset/add-itemfield-to-hooks.md
@@ -1,5 +1,5 @@
 ---
-"@keystone-6/core": minor
+"@keystone-6/core": major
 ---
 
 Add `itemField`, `originalItemField`, `inputFieldData` and `resolvedFieldData` parameters to field hooks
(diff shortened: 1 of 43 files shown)


.changeset/add-itemfield-to-hooks.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-6/core": minor
+---
+
+Add `itemField`, `inputFieldData` and `resolvedFieldData` parameters to field hooks
```
emmatown, line 5:
```suggestion
Add `itemField`, `originalItemField`, `inputFieldData` and `resolvedFieldData` parameters to field hooks
```

emmatown (approved):

Issue #9601, opened 2025-05-05
TypeScript Error TS2305: Cannot find exported member 'Context' in generated types after clean install with latest packages


rostikowb:
Description / Steps to Reproduce:

Start with a Keystone 6 project using @keystone-6/core, @keystone-6/auth, and @keystone-6/fields-document.

Ensure tsconfig.json has "strict": true.

Update Keystone packages to their latest versions:

`npm install @keystone-6/core@latest @keystone-6/auth@latest @keystone-6/fields-document@latest`

Perform a clean installation:

```
rm -rf node_modules package-lock.json .keystone
npm install
```

Generate Keystone types and build the project:

`npm run build`

(Note: This build command might complete successfully, even with the underlying type error present).

Run TypeScript type checking:

`tsc --noEmit`

Expected Behavior:

tsc --noEmit should complete successfully without reporting errors originating from within @keystone-6/core's type declaration files. The specific Context type should be correctly generated within the .keystone directory and usable in custom code (like access control functions) without causing type mismatches with the base Keystone context/db types.

Actual Behavior:

tsc --noEmit consistently fails with the following primary error:

node_modules/@keystone-6/core/dist/declarations/src/types/schema/schema-api-with-context.d.ts:2:15 - error TS2305: Module '"./graphql-ts-schema.js"' has no exported member 'Context'.

import { type Context } from "./graphql-ts-schema.js"
                ~~~~~~~

This initial error leads to numerous cascading TS2322 errors in code, particularly in src/core/schema.ts when defining lists that use custom access control functions. For example:

// src/core/schema.ts shows errors like:
// error TS2322: Type '{ ... access: ListAccessControl<BaseListTypeInfo>; ... }' is not assignable to type 'ListConfig<TypeInfo<any>>'.
// ... (long type incompatibility message, often pointing to context.db mismatch) ...


The root cause appears to be that the internal type declaration file (schema-api-with-context.d.ts) attempts to import a Context type from graphql-ts-schema.js, but that type is not actually exported from that file within the installed node_modules package.


Screenshots:

Not applicable, as the primary issue is a TypeScript compilation error message shown above.

Contextual Information:

- Node.js Version: node -v -> v22.13.0
- npm Version: npm -v -> 10.9.2
- Operating System: Windows 11
- @keystone-6/core Version: 6.5.0
- @keystone-6/auth Version: 8.1.0
- @keystone-6/fields-document Version: 9.1.1
- typescript Version: 5.8.3
[...]

PR #9602, opened 2025-05-05
Fix schema-api-with-context on v6

emmatown:


dcousens:
Fixes https://github.com/keystonejs/keystone/issues/9601

PR #9606, opened 2025-05-15
Fix error when accessing Apollo Sandbox

emmatown:
`body-parser@^2` only sets `req.body` [if a `req.body` is parsed](https://github.com/expressjs/body-parser/blob/master/HISTORY.md#200-beta1--2021-12-17), Apollo always wants `req.body` to be set.


dcousens:
Regression from https://github.com/keystonejs/keystone/pull/9577, no changeset required as this hasn't been published in `latest`

PR #9609, opened 2025-05-17
Sync Examples GraphQL Schemas

unrevised6419:
I've run this using npm, but it might be wrong because npm installed deps not from lock file because does not exist, and installed instead based on package.json

I could use pnpm, but it won't do any schema updates, probably because `@keystone-6/*` are not updated.

We need first to update the `@keystone-6/*` dependencies in all examples folders, then generate again the schemas. (will try to create a different PR for that)

```bash
for dir in $(find examples/ -name "package.json" -not -path "*/node_modules/*" -exec dirname {} \;); do
  echo "\nChecking $dir..."
  cd "$dir"
  if grep -q '"build":' package.json 2>/dev/null; then
    npm i --cache ~/.npm_keystone --prefer-offline --legacy-peer-deps --ignore-scripts
    npm run build
  else
    echo "No build script found in $dir"
  fi
  cd - > /dev/null
done
```

### DIFF
--- a/examples/custom-admin-ui-pages/schema.graphql
+++ b/examples/custom-admin-ui-pages/schema.graphql
@@ -241,22 +241,24 @@ type KeystoneAdminMeta {
 type KeystoneAdminUIListMeta {
   key: String!
   path: String!
-  description: String
   label: String!
-  labelField: String!
   singular: String!
   plural: String!
+  description: String
+  pageSize: Int!
+  labelField: String!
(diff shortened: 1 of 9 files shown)


dcousens:
These are out of date with `main`, which should be using the workspace versions, not `npm`

unrevised6419:
@dcousens shouldn't the examples have `workspace` in the `package.json`? 🤔

PR #9644, opened 2025-07-02
Add `{list}.graphql.singular` and fix the computed GraphQL list plural to retain the original casing

dcousens:
This pull request fixes a list with the name `My_User` to compute `my_user` and `my_users` as the GraphQL singular and plural names respectively.

Previously to this change the computed names were `my_user` and `myUsers`, which has lost the developers preferred casing.


.changeset/poor-months-wear.md
```diff
@@ -2,4 +2,4 @@
 "@keystone-6/fields-document": patch
 ---
 
-Handle adding extra form fields to component blocks for existing data
```
dcousens, line 5: Unrelated, but updated to use the same changeset format as our others

packages/core/src/types/utils.ts
```diff
@@ -70,29 +70,30 @@ export function __getNames(listKey: string, list: KeystoneConfig['lists'][string
...
+        plural: graphqlPlural
+      }),
     },
     ui: {
       labels: {
-        label: ui?.label || computedLabel,
-        singular: ui?.singular || computedSingular,
-        plural: ui?.plural || computedPlural,
-        path,
+        lable: ui?.label || humanize(isSingleton ? singular : plural),
```
emmatown, line 93:
```suggestion
        label: ui?.label || humanize(isSingleton ? singular : plural),
```

PR #9658, opened 2025-07-16
Add example for document field hooks, and fix Tags showing removability when read only

dcousens:
This example shows how you might setup hooks on a document field to map or validate the content in hooks.

> fix Tags showing removability when read only

To my knowledge this is a regression with the new design system, thereby no changeset is required.

> fix missing resolvedFieldData for {field}.hooks.validate.create

This functionality has not been released since merging https://github.com/keystonejs/keystone/pull/9600, thereby no changeset is required.

### DIFF
--- a/examples/document-field-hooks/schema.ts
+++ b/examples/document-field-hooks/schema.ts
@@ -44,13 +44,14 @@ export const lists = {
           validate({ resolvedFieldData, addValidationError }) {
             mapNodes(resolvedFieldData as Node[], (node, __parents) => {
               if ('text' in node) {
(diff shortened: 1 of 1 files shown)


packages/core/src/types/config/hooks.ts
```diff
@@ -285,6 +285,7 @@ export type ValidateFieldHook<
        * The GraphQL input **after** being resolved by the field type's input resolver
        */
       resolvedData: ListTypeInfo['prisma']['create']
+      resolvedFieldData: FieldTypeInfo['prisma']['create']
```
dcousens, line 288: Missed this in https://github.com/keystonejs/keystone/pull/9600

PR #9664, opened 2025-07-22
Revert https://github.com/keystonejs/keystone/pull/9663, keeping fixes only

dcousens:
https://github.com/keystonejs/keystone/pull/9663 opted to always use the provided values, but in retrospect, this will only break things downstream.

This pull request reverts that breaking change, but maintains the fixes to ensure the provided value is used when that value won't break downstream.

### DIFF
--- a/packages/core/src/lib/core/initialise-lists.ts
+++ b/packages/core/src/lib/core/initialise-lists.ts
@@ -663,37 +663,34 @@ function getListsWithInitialisedFields(
           views: f.ui?.views ?? null,
           createView: {
             isRequired: f.ui?.createView?.isRequired ?? false,
-            fieldMode:
-              f.ui?.createView?.fieldMode ??
-              (isEnabledField.create
-                ? (group?.ui?.createView?.defaultFieldMode ??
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/core/initialise-lists.ts
```diff
@@ -663,37 +663,34 @@ function getListsWithInitialisedFields(
...
-                : isEnabledField.read
-                  ? 'read'
-                  : 'hidden'),
+            fieldMode: isEnabledField.update
+              ? (f.ui?.itemView?.fieldMode ??
+                group?.ui?.itemView?.defaultFieldMode ??
+                listConfig.ui?.itemView?.defaultFieldMode ??
+                'edit')
+              : isEnabledField.read
+                ? (f.ui?.itemView?.fieldMode ?? 'read')
```
dcousens, line 683: @emmatown a developer could opt for `edit`, but, that is OK IMHO?

dcousens:
This won't protect developers who return `ui.itemView.fieldMode: 'edit'` when `graphql.omit.update: true` - we probably need a wider solution to this as resolving this with filters is quite complex now.

PR #9666, opened 2025-07-23
Update relationship fields to use `searchFields`, `initialSort` and `initialFilter` from the foreign list by default

dcousens:
This pull request updates a few behaviours related to relationship fields, including their document field counterparts.

`initialFilter` and `initialSort` defaults are now used if provided on the foreign list.
The normal relationship field retains the support to override this as desired.

Support to do that on the document relationship field will likely be added at a later date.

### DIFF
--- a/examples/reuse/schema.ts
+++ b/examples/reuse/schema.ts
@@ -51,7 +51,7 @@ function trackingFields<ListTypeInfo extends CompatibleLists>() {
           },
           async update() {
             return undefined
-          }
+          },
         },
       },
     }),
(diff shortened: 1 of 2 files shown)


emmatown (approved):

.changeset/filter-relationships-too.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-6/fields-document": patch
+---
+
+Fix the relationship fields to default to `initialFilter` and `initialSort` from the foreign list (if provided)
```
dcousens, line 5: This is primarily a patch for `initialSort`, as `initialFilter` is only added in https://github.com/keystonejs/keystone/pull/9629

dcousens:
I am reverting the default use of `list.ui.listView.initialFilter` for relationships in https://github.com/keystonejs/keystone/pull/9688

PR #9669, opened 2025-07-24
Fix ListView sort not adhering to GraphQL schema for *OrderBy types

dcousens:
The sorting bug has not been released, thereby a changeset is not required.


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/PaginationControls.tsx
```diff
@@ -73,10 +72,6 @@ export function PaginationControls(props: {
               props.onChangePageSize(Number(key))
             }}
             selectedKey={pageSize}
-            // disable sizes greater than the total, allowing the next page to be the last
-            disabledKeys={PAGE_SIZES.filter(n => {
-              return n > snapValueToNextAvailable(total)
-            })}
```
dcousens, line 79: This isn't ideal when a user switches their preference then can't switch back

PR #9673, opened 2025-07-29
Fix regressions from https://github.com/keystonejs/keystone/pull/9669

dcousens:
https://github.com/keystonejs/keystone/pull/9669 had quite a few structural changes, and it seems many were difficult to test for.

This pull request fixes the known regressions from that pull request and adds an integration test that ensures.


tests/admin-ui-tests/relations.test.ts
```diff
@@ -30,27 +30,54 @@ adminUITests('./tests/test-projects/basic', browserType => {
       headers: { 'Content-Type': 'application/json' },
       body: JSON.stringify({
         query: gql`
-          mutation ($people: [PersonCreateInput!]!) {
-            createPeople(data: $people) {
-              id
-            }
```
dcousens, line 36: `createPeople` had no useful outcome in this test

packages/core/src/types/admin-meta.ts
```diff
@@ -133,10 +133,6 @@ export type ListMeta = {
   hideDelete: boolean
 }
 
-export type AdminMeta = {
-  lists: { [list: string]: ListMeta }
-}
```
dcousens, line 138: This type isn't helpful

PR #9689, opened 2025-08-26
Add new `{list}.actions` feature 

dcousens:
https://github.com/user-attachments/assets/bea05507-db8d-4de0-8e4f-0a51e3afec03

This pull request adds a new approach to adding actions to your Keystone project.
Previously developers may have used a mix of `extendGraphqlSchema`, virtual fields and or customisations to the AdminUI Nextjs application to add action-like behaviours. The Actions feature captures feedback from the many developers who made that journey and happily passed on their experience, so that we can now confidently deliver on one of our [longest-standing community feature requests](https://github.com/keystonejs/keystone/discussions/5315).

This pull request is not inclusive of documentation or tests for now, as I want to seek feedback from the community who is using our `rc` snapshots. Documentation and tests will come soon, but please try this out if you can! 💙


gautamsi:
I like this, I have done similar things already in my custom ui work, good to see this.

gautamsi:
there are still some place you can have custom actions, Some of them which we worked in v5:

ListAction - action on the page without selecting the any items on List page
ListHeaderAction - customize the header component.

for Item page, two more actions I have used extensively - Sidebar and Header

I am going to see how I can keep my fork up to date.

packages/core/src/lib/core/mutations/index.ts
```diff
@@ -304,6 +306,54 @@ export async function deleteMany(
...
 }
 
+export async function actionOne(
+  where: UniqueInputFilter,
+  list: InitialisedList,
+  context: KeystoneContext,
+  action: InitialisedAction
+) {
+  const operationAccess = await action.access({
+    context,
```
dcousens, line 316: Missing `session: context.session`, added in https://github.com/keystonejs/keystone/commit/51e767059c78ba1c0b1114c4e24aa8d6a2aa6e69

packages/core/src/lib/core/mutations/index.ts
```diff
@@ -304,6 +306,54 @@ export async function deleteMany(
...
+}
+
+export async function actionMany(
+  wheres: UniqueInputFilter[],
+  list: InitialisedList,
+  context: KeystoneContext,
+  action: InitialisedAction
+) {
+  const operationAccess = await action.access({
+    context,
```
dcousens, line 339: Missing `session: context.session`, added in https://github.com/keystonejs/keystone/commit/51e767059c78ba1c0b1114c4e24aa8d6a2aa6e69

dcousens:
> ListAction - action on the page without selecting the any items on List page

Did you have any examples you could share?

gautamsi:
This one is header action.  Something I use for related to products nit not exactly an action with products
image

another example where I have these buttons matching the style

image


Not sure if the action you have with list definition would allow UI or not, in my case I have created in UI instead of list. advantage is that you can have full UI features, like you can create dialog with custom content, show list of related items or pretty much anything you can do with react. 

for item actions I have this
image

gautamsi:
here is the admin ui fork I have to work in my project, it is created on top of my app router PR. 
https://github.com/k6js/k6js/blob/main/examples/basic/app/(admin)/posts/page.tsx

gautamsi:
I see this is slick.
I have future idea. We should allow adding a view just like the custom field view, this can be rendered as component which is activated by clicking the button, that view will take over from there. View gets the selected items and on success, on failure or even setItems to decide when to clear the selection after action is done. 
this would allow creating the ui they need. 


My approach is to provide this to List page and Item page as components and that renders at the place.

packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ListPage/index.tsx
```diff
@@ -632,96 +676,113 @@ function ListPageHeader({ listKey, showCreate }: { listKey: string; showCreate?:
...
+      console.error(error)
     }
-
-    return refetch()
   }
 
   return (
     <AlertDialog
-      title="Delete items"
+      tone={'neutral'}
```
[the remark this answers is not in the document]
dcousens: @gautamsi yes! I hit a block on naming ha, but the documentation and types (commented out) are sitting there ready to add

PR #9691, opened 2025-08-28
Add error details for actions on the List page and fix action modes

dcousens:
This pull request expands on the list view to show any errors that may have occurred when executing a particular action, as added in https://github.com/keystonejs/keystone/pull/9689.

Additionally, this pull request fixes the `{action}.ui.itemView.actionMode` not propagating to the item view context when it is different than the static variant (`item === null`).

As neither of these have been released, no additional changesets are required.


dcousens:
Cannot reproduce https://github.com/keystonejs/keystone/actions/runs/17288819242/job/49071232827?pr=9691 locally, I will continue for now and resolve that later

The text is expected but it seems playwright varies as to whether it includes the text in the `innerText`.

PR #9694, opened 2025-08-31
fixed jsx error

gautamsi:
this fixes error in console
\client\…console-error.ts:40 A props object containing a "key" prop is being spread into JSX

this is coming from doing `...action` on component, action has key prop which it does not like.


dcousens (approved):

dcousens:
Thanks @gautamsi!

PR #9696, opened 2025-09-01
Add Open Telemetry spans and relevant example

dcousens:
This pull request adds `@opentelemetry/api` to the Keystone project for observing internal queries and mutations.

A minimal example is added that shows how you can view these spans using trivial `console.log` calls, without diving into the complexity that open telemetry can fall into.

**This feature is experimental**, but seeing as `@opentelemetry/api` is near to a no-op without a provider in-place, this should be OK to keep around.


examples/logging/keystone.ts
```diff
@@ -72,7 +72,7 @@ export default config<TypeInfo>({
     getAdditionalFiles: async () => [
       {
         mode: 'write',
-        src: `export default ${JSON.stringify({
+        src: `module.exports = ${JSON.stringify({
```
dcousens, line 75: this fixes Node showing a warning

PR #9698, opened 2025-09-02
Fix list page intrinsic Delete action tone and bump dependencies

dcousens:
Incorporates https://github.com/keystonejs/keystone/pull/9576 and https://github.com/keystonejs/keystone/pull/9695

This pull request is a follow up to https://github.com/keystonejs/keystone/pull/9689, fixing a few small issues, some made more apparent when bumping the dependencies in https://github.com/keystonejs/keystone/pull/9576.

Fixes
- The action tone for the Delete action on the list page is now critical, a regression
- The order of the toast messages when executing an action is now Fail, and then Success, another regression

This pull request additionally fixes an issue with recent next versions, which may need to be back-ported to `v6`.


dcousens:
Some flaky issues with live reloading, @emmatown you might know why

PR #9700, opened 2025-09-03
Add additional Open Telemetry attributes (and adhere to canonical attribute naming)

dcousens:
Followup to https://github.com/keystonejs/keystone/pull/9696

This pull request adds additional Open Telemetry `id` attributes to internal mutations and adhere to canonical attribute naming in the examples.

Additionally, this pull request fixes `humanize` when using inputs like `Foo11_WithBar11` to fix the regression introduced in https://github.com/keystonejs/keystone/pull/9696.

Unfortunately, when using `inflection` previously, we were stripping the plural, and I am not reverting that out of preference.  I added the changeset for that in https://github.com/keystonejs/keystone/commit/c06ee774962a5cb54c3aecbf9b6f3d03732ccb20 (directly to `main`)


examples/logging-opentelemetry/keystone.ts
```diff
@@ -28,9 +28,12 @@ export default config<TypeInfo>({
...
             return {
               async willSendResponse({ operation, request }) {
-                span.setAttribute('type', operation?.operation || 'unknown')
-                span.setAttribute('name', request.operationName || 'unknown')
-                span.setAttribute('hash', request.query ? sha256(request.query) : 'empty')
+                span.setAttribute('gql.query.type', operation?.operation || 'unknown')
+                span.setAttribute('gql.query.name', request.operationName || 'unknown')
+                span.setAttribute(
+                  'qql.query.sha256',
+                  request.query ? sha256(request.query) : 'empty'
```
dcousens, line 35: This should probably be `graphql.query.*` if we want to be similar to what Apollo is using internally, but, for now this is OK

PR #9706, opened 2025-09-13
Update Prisma to 16.6

gautamsi:
This update is just enough to allow using latest Prisma. Still can not use rust free version. I am going to create another PR for allowing use of rust free version. I am leaning towards prisma cli based approach instead of using internal prisma methods. I see some other libraries are already going for cli approach instead of using Migrate or Internal packages.

### DIFF
--- a/packages/auth/src/pages/SigninPage.tsx
+++ b/packages/auth/src/pages/SigninPage.tsx
@@ -1,5 +1,5 @@
 import NextHead from 'next/head'
-import { type FormEvent, useState } from 'react'
+import { useState } from 'react'
 
 import { Button } from '@keystar/ui/button'
 import { Grid, HStack, VStack } from '@keystar/ui/layout'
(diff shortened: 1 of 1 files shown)


gautamsi:
@dcousens any traction on this?

gautamsi:
any update on this @dcousens

trickreich:
@gautamsi you got some info from the project owners? it feels like this project isn't under active development

dcousens:
@gautamsi looking to try and resolve this ASAP, especially with `v7` now landing

gautamsi:
we need to use prisma with config file and invoke prisma cli instead of the programmatic undocumented means to work with prisma IMO.

This will resolve further issues and allow using split schema in a folder which helps me keep multiple other tables out of keystone but in prisma easily. currently it is hectic to update the extendPrimaSchema method to excluded model by model.

gautamsi:
if anyone is working on this, I suggest go straight to latest instead of intermediate version.

dcousens:
@gautamsi I'm happy to follow up with another major pull request, but this was easier to rebase

PR #9719, opened 2025-10-10
Fix documentation for `graphql.extendGraphqlSchema`

OlleMattsson:
Fixed outdated docs:
extendGraphqlSchema should reside under the graphql key in order to be picked up by the config.


dcousens:
Thanks @OlleMattsson! 💛

PR #9730, opened 2025-11-12
Fix `item` and `itemField` is `null` for `ui.*.fieldPosition` and `ui.*.fieldMode`

dcousens:
https://github.com/keystonejs/keystone/pull/9689 moved the GraphQL resolving of these fields up to the `list` rather than resolving the `item` at the field.

This pull request fixes an unreleased regression that prevented `item` from flowing down to the `field` resolver unintentionally.

### DIFF
--- a/tests2/omit.test.ts
+++ b/tests2/omit.test.ts
@@ -89,7 +89,8 @@ function makeList({
         delete: boolean
       }
 }) {
-  const prefix = `List${fields.length}_Filt${yn(defaultIsFilterable)}_Ord${yn(defaultIsOrderable)}` as const
+  const prefix =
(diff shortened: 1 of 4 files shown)


tests2/omit.test.ts
```diff
@@ -73,13 +73,13 @@ const fieldsMatrix = [
 
 function makeList({
   fields,
-  isFilterable,
-  isOrderable,
```
dcousens: I edited this file as I was reviewing tests to understand how this was missed.

We aren't testing this functionality well.
For now I will move forward with the fixes as is, but I do have some new tests in the pipeline to add coverage for this regression and others.

PR #9731, opened 2025-11-13
Add `context.internal().*` for using omitted GraphQL queries

dcousens:
Following on from https://github.com/keystonejs/keystone/pull/9689, when you use actions, you may sometimes supplant the requirement to allow particular operations in your public GraphQL schema.

You may thereby start to opt to omit the different mutations from your schema using `{list/field}.graphql.omit: true`. Unfortunately that stops you from using the same operations with `context.db` and `context.query` in your hooks and actions, as each of them respect the underlying GraphQL schema.

This pull request adds support for you to bind to a new context with `context.internal()` that is, like `sudo()`, unbounded by `graphql.omit`, but still bound by access control.

*WARNING*: If you want to use this now, please be aware `context.internal()` bypasses `{field}.isFilterable` and `{field}.isOrderable` access control as a limitation of that approach, and that this should be resolved before a stable release.

### DIFF
--- a/packages/core/src/artifacts.ts
+++ b/packages/core/src/artifacts.ts
@@ -81,7 +81,7 @@ export async function generateTypes(cwd: string, system: System) {
     const paths = system.getPaths(cwd)
     const schema = printGeneratedTypes(
       paths.types.relativePrismaPath,
-      system.graphql.schemas.public,
+      system.graphql.schemas.internal,
       system.lists
     )
     await fs.mkdir(path.dirname(paths.schema.types), { recursive: true })
--- a/packages/core/src/lib/core/where-inputs.ts
+++ b/packages/core/src/lib/core/where-inputs.ts
@@ -36,7 +36,7 @@ export async function resolveUniqueWhereInput(
 export async function resolveWhereInput(
   inputFilter: InputFilter,
   list: InitialisedList,
-  context: KeystoneContext,
+  context: KeystoneContext
 ): Promise<PrismaFilter> {
   return {
     AND: await Promise.all(


packages/core/src/lib/system.ts
```diff
@@ -56,44 +55,26 @@ function getSystemPaths(cwd: string, config: KeystoneConfig) {
...
                 return [
                   fieldKey,
-                  (data: FieldData) => {
+                  data => {
                     const f = field(data)
                     return {
                       ...f,
-                      access: allowAll,
-                      isFilterable: true,
-                      isOrderable: true,
```
emmatown: Doesn't isFilterable/isOrderable: false do an omit so don't we need to override this in some way?

tests2/omit.test.ts
```diff
@@ -345,7 +345,10 @@ describe(`Omit (${dbProvider})`, () => {
   })
```
emmatown: It doesn't seem like we test the omitting/not behaviour around `isFilterable`/`isOrderable`, can we add that?

packages/core/src/lib/system.ts
dcousens: yes, I have added a temporary warning for now about this

emmatown (approved):

tests2/omit.test.ts
dcousens: this will be added in future work, before the next stable

ARCH-angel93:
Watch other merges it may leak your words in other formats from person or scope their data and inputs it gets to be a bit much if it starts to consume your data its not fun or gives you "content" feed back that isn't so great as in Scholarly or ones personalized "stuff" as in browser consumed topics or interests.

ARCH-angel93:
Frame works Help and redirects if at minimum a format structure or tree can simplify things as a basic overview to not over process and a completion point to finalize codes or builds or random things.

PR #9736, opened 2025-11-18
Fix the `module` condition for modules that are bundled into the `config.js`

emmatown:
Within a monorepo, we want the imports of other packages (within the monorepo) to import their own dependencies as `cjs` modules, not `esm`.

When we opted for `conditions: []`, we accidentally opted into always using the `cjs` exports of our monorepo packages too. If, as an example, you are using `preconstruct` in your own project, this then results in `@babel/core` hooks in your production build - not good.

This pull request resolves that by reverting https://github.com/keystonejs/keystone/pull/9699, and now opts to resolve external dependencies of packages within your own monorepo as `cjs`.


gautamsi:
what does this do exactly?

dcousens (approved):

PR #9740, opened 2025-11-25
Update to Apollo Client v4

emmatown:
This a breaking change as have we publicly re-exported `@apollo/client` in the past.

Of note, the newer version of `apollo-upload-client` is Node ESM-only, but as that is only used by the Admin UI it doesn't matter since the code is bundle.

### DIFF
--- a/packages/core/src/admin-ui/context.tsx
+++ b/packages/core/src/admin-ui/context.tsx
@@ -17,7 +17,14 @@ import type {
   ListMeta,
 } from '../types'
 import { type AdminMetaQuery, adminMetaQuery } from './admin-meta-graphql'
-import { ApolloClient, ApolloProvider, type ErrorLike, InMemoryCache, gql, useQuery } from './apollo'
+import {
+  ApolloClient,
+  ApolloProvider,
+  type ErrorLike,
+  InMemoryCache,
+  gql,
+  useQuery,
+} from './apollo'
 
 type KeystoneContextType = {
   adminConfig: AdminConfig | null
--- a/packages/auth/src/components/Navigation.tsx
+++ b/packages/auth/src/components/Navigation.tsx
@@ -5,7 +5,12 @@ import { Divider } from '@keystar/ui/layout'
 import { TooltipTrigger, Tooltip } from '@keystar/ui/tooltip'
 import { Text } from '@keystar/ui/typography'
 
-import { useQuery, useMutation, gql, type TypedDocumentNode } from '@keystone-6/core/admin-ui/apollo'
+import {
+  useQuery,
+  useMutation,
+  gql,
+  type TypedDocumentNode,
+} from '@keystone-6/core/admin-ui/apollo'
 import {
   DeveloperResourcesMenu,
   NavList,


packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/HomePage/index.tsx
```diff
@@ -9,7 +9,7 @@ import { css, FocusRing, tokenSchema, transition } from '@keystar/ui/style'
 import { Tooltip, TooltipTrigger } from '@keystar/ui/tooltip'
 import { Heading, Text } from '@keystar/ui/typography'
 
-import { gql, useQuery } from '../../../../admin-ui/apollo'
+import { gql, TypedDocumentNode, useQuery } from '../../../../admin-ui/apollo'
```
dcousens:
```suggestion
import { gql, type TypedDocumentNode, useQuery } from '../../../../admin-ui/apollo'
```

packages/core/src/___internal-do-not-use-will-break-in-patch/admin-ui/pages/ItemPage/common.tsx
```diff
@@ -11,7 +11,7 @@ import { Grid, HStack } from '@keystar/ui/layout'
 import { breakpointQueries, css, tokenSchema } from '@keystar/ui/style'
 import { toastQueue } from '@keystar/ui/toast'
 import { Heading, Text } from '@keystar/ui/typography'
-import { gql, useApolloClient } from '../../../../admin-ui/apollo'
+import { gql, TypedDocumentNode, useApolloClient } from '../../../../admin-ui/apollo'
```
dcousens:
```suggestion
import { gql, type TypedDocumentNode, useApolloClient } from '../../../../admin-ui/apollo'
```

packages/core/src/admin-ui/utils/useCreateItem.ts
```diff
@@ -11,13 +11,13 @@ import {
   useInvalidFields,
 } from '../../admin-ui/utils'
 import type { ListMeta } from '../../types'
-import { type ApolloError, gql, useMutation } from '../apollo'
+import { ErrorLike, gql, TypedDocumentNode, useMutation } from '../apollo'
```
dcousens:
```suggestion
import { type ErrorLike, gql, type TypedDocumentNode, useMutation } from '../apollo'
```

packages/core/src/admin-ui/context.tsx
```diff
@@ -17,13 +17,12 @@ import type {
   ListMeta,
 } from '../types'
 import { type AdminMetaQuery, adminMetaQuery } from './admin-meta-graphql'
-import type { ApolloError, QueryResult } from './apollo'
-import { ApolloClient, ApolloProvider, InMemoryCache, gql, useQuery } from './apollo'
+import { ApolloClient, ApolloProvider, ErrorLike, InMemoryCache, gql, useQuery } from './apollo'
```
dcousens:
```suggestion
import { ApolloClient, ApolloProvider, type ErrorLike, InMemoryCache, gql, useQuery } from './apollo'
```

Issue #9456, opened 2025-01-28
Integer field defaultValue {kind: "autoincrement"} triggers missing value validation


simonsagstetter:
Configure a list with an integer autoincrement field
```ts
import { list } from "@keystone-6/core";
import { integer } from "@keystone-6/core/fields";

list({
 fields: {
  no: integer({
        defaultValue: { kind: "autoincrement" },
        db: { isNullable: false }, // if this is not set I get a prisma error
    }),
 }
});
```

Prisma Error when db.isNullable is not set to false
`Error loading your Keystone config Error: Request.no specifies defaultValue: { kind: 'autoincrement' } but doesn't specify db.isNullable: false.
Having nullable autoincrements on Prisma currently incorrectly creates a non-nullable column so it is not allowed.
https://github.com/prisma/prisma/issues/8663`

I also tried `validation: { isRequired: false}`

When you try to create a new record for this list it will return an validation error for a missing value on that field:
`You provided invalid data for this operation. - Request.no: missing value`

<img width="755" alt="Image" src="https://github.com/user-attachments/assets/0712f462-88e2-45d6-923c-9b3e4c06cb3a" />

I setup the integer field like described in the docs:
`defaultValue (default: undefined): Can be either an integer value or { kind: 'autoincrement' }. This value will be used for the field when creating items if no explicit value is set. { kind: 'autoincrement' } is only supported on PostgreSQL.`

I should not need to put in a value for that field on any db operation because it is configured to auto-increment. When I put in a value into the integer field this value will be saved. What am I missing or is this a bug?

My db config 
```ts
...
db: {
            provider: "postgresql",
            url: `postgres://${process.env.DB_USER}:${process.env.DB_PASS}@localhost:5432/keystone`,
            enableLogging: false,
            idField: { kind: "uuid" }, // is it because of this??!
            shadowDatabaseUrl: `postgres://${process.env.DB_USER}:${process.env.DB_PASS}@localhost:5432/shadowdb`,
},
...
```

```
node version v20.18.1
brave browser v1.74.50
postgresql v17.2
keystone-6/core v6.3.1
```

lushkovsky-s:
@simonsagstetter, please try: 
```
no: integer({
  defaultValue: { kind: "autoincrement" },
  db: { isNullable: false }, // Required for autoincrement
  validation: { isRequired: false } // Explicitly set to false
})
```

Maybe it's worth also considering for the project: 
[...]

PR #9752, opened 2025-12-17
Fix `{ kind: 'autoincrement' }` behaviour for `integer` field

dcousens:
This pull request fixes the broken integration between Prisma and Keystone when a user defines `defaultValue: { kind: 'autoincrement' }` (as reported in https://github.com/keystonejs/keystone/issues/9456).

Fixes https://github.com/keystonejs/keystone/issues/9456

### DIFF
--- a/packages/core/src/fields/types/integer/index.ts
+++ b/packages/core/src/fields/types/integer/index.ts
@@ -94,7 +94,7 @@ export function integer<ListTypeInfo extends BaseListTypeInfo>(
             }
           }
         : undefined,
-      defaultValue !== null,
+      defaultValue !== null
     )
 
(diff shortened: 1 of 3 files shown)


packages/core/src/fields/types/integer/index.ts
```diff
@@ -144,7 +130,7 @@ export function integer<ListTypeInfo extends BaseListTypeInfo>(
           }),
           resolve(value) {
             if (value === undefined) {
-              if (defaultValue === 'autoincrement') return null
```
dcousens: If a user passed nothing, we gave Prisma `null`, which is wrong

Issue #9785, opened 2026-03-06
`statelessSessions` attempts to use unsupported `Authorization: Basic`  header rather than the cookie


rostikowb:
When deploying a Keystone app to a staging environment hidden behind a reverse proxy (like Nginx or Caddy) with HTTP Basic Authentication, Admin UI access breaks (`Access denied`), even if the user logs in correctly and has a valid `keystonejs-session` cookie.

**Steps to reproduce:**
1. Setup a Keystone app using `statelessSessions`.
2. Put the app behind a proxy that requires Basic Auth, passing the `Authorization: Basic ...` header down to the Node.js backend.
3. Log in to the Admin UI successfully (the cookie is set in the browser).
4. Refresh the page or try to access `adminMeta`.
5. Result: `Access denied` because `context.session` becomes `undefined`.

**Expected behaviour:**
Keystone should ignore `Authorization: Basic ...` headers and correctly fallback to parsing the `keystonejs-session` cookie.

---

Node.js - v22.13.0
keystone-6/auth - 8.1.0
keystone-6/core - 6.5.1

<img width="485" height="322" alt="Image" src="https://github.com/user-attachments/assets/a0b81729-2830-48dc-a9b1-896fb7c4a552" />

dcousens:
@rostikowb perhaps https://github.com/keystonejs/keystone/pull/9786 will help you if you don't have control of the proxy or otherwise cannot stop the headers passing through.

#9786 won't be a breaking change for developers who used the `Bearer ` prefix, and may continue to expect it to take priority > the cookie.

rostikowb:
Thank you for the quick response and the PR! It took me a few days to realize that the problem was not in my code, .env, or Docker 🥲
For now, I'm using a temporary Express middleware workaround that simply deletes the Authorization: Basic header before it reaches Keystone. I would be grateful if this fix prevents other developers from encountering the same issue

emmatown:
This is now fixed

PR #9786, opened 2026-03-06
Fix `statelessSessions` to only accept `Authorization: Bearer` prefixed tokens

dcousens:
This may help https://github.com/keystonejs/keystone/issues/9785, by restricting the types of `Authorization` headers that we accept.

`Authorization: Basic` is not supported by `statelessSessions`, but `Iron.unseal` would have been attempted for any `Authorization:` header.

This may break downstreams who inadvertently used `Authorization: {token}`, but that has not been intentionally supported.


dcousens:
Needs a changeset

PR #9787, opened 2026-03-11
Add `{action}.args.*` to actions

emmatown:
This adds an experimental option on actions which defines arguments that can be passed. 
No changeset since actions haven't been properly released.


packages/core/src/types/config/lists.ts
```diff
@@ -8,126 +9,162 @@ import type { ListAccessControl, ActionAccessControlFunction } from './access-co
...
    */
-  [key: string]: {
+  access: ActionAccessControlFunction<ListTypeInfo>
+  resolve: (
+    args: {
+      listKey: ListTypeInfo['key']
+      actionKey: ListTypeInfo['actions']
+      where: ListTypeInfo['inputs']['uniqueWhere']
+      args: ActionArgs<Args>
+    },
```
emmatown, line 50: Do we want to make this more in the style of a normal GraphQL resolver rather than the nested args when it's not nested in the actual GraphQL mutation?

packages/core/src/lib/core/mutations/index.ts
```diff
@@ -864,12 +869,13 @@ export function getMutationsForList(list: InitialisedList) {
                     type: g.nonNull(list.graphql.types.uniqueWhere),
                     defaultValue: defaultUniqueWhereInput,
                   }),
+                  ...action.graphql.types.arguments,
```
emmatown, line 872: This is interesting, nowhere else do we spread arguments like this e.g. we don't have `updateUser(where: {id: $id}, name: $name)`, we have `updateUser(where: { id: $id }, data: { name: $name })`. seems fine, though just interesting.

packages/core/src/types/config/lists.ts
emmatown: e.g. maybe 
```ts
(
  args: { where: ListTypeInfo['inputs']['uniqueWhere'] } & ActionArgs<Args>,
  context: KeystoneContext<ListTypeInfo['all']>,
  info: {
    listKey: ListTypeInfo['key']
    actionKey: ListTypeInfo['actions']
  }
) => MaybePromise<ListTypeInfo['item'] | null>
```

noting that it's more similar to the signature of normal resolvers: `(source, args, context, info) => result`

dcousens: yes, we should follow this up

PR #9822, opened 2026-04-21
feat: add startup telemetry status logs

### LINKED ISSUE #9299: Update starting logs to include telemetry status
#9292 revealed that we should consider adding to the initial start logs

See [thread](https://github.com/keystonejs/keystone/pull/9292#discussion_r1722534526)

abhijnyan-codes:
Closes #9299

Added log messages to runTelemetry() to inform users of telemetry status at startup:

- Keystone Telemetry is disabled (running in CI)
- Keystone Telemetry is disabled (NODE_ENV is production)
- Keystone Telemetry is disabled (via KEYSTONE_TELEMETRY_DISABLED env)
- Keystone Telemetry is disabled (via opt-out)
- Keystone Telemetry is enabled

### DIFF
--- a/packages/core/src/lib/telemetry.ts
+++ b/packages/core/src/lib/telemetry.ts
@@ -273,23 +273,33 @@ export async function runTelemetry(
   dbProviderName: DatabaseProvider
 ) {
   try {
-    if (
-      ci.isCI || // don't run in CI
-      process.env.NODE_ENV === 'production' || // don't run in production
-      process.env.KEYSTONE_TELEMETRY_DISABLED === '1' // don't run if the user has disabled it
-    ) {
+    if (ci.isCI) {
+      console.log(`Keystone Telemetry is ${r`disabled`} (running in CI)`)
+      return
+    }
+
+    if (process.env.NODE_ENV === 'production') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (NODE_ENV is production)`)
+      return
+    }
+
+    if (process.env.KEYSTONE_TELEMETRY_DISABLED === '1') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via KEYSTONE_TELEMETRY_DISABLED env)`)
       return
     }
 
     const { telemetry, userConfig } = getTelemetryConfig()
 
-    // don't run if the user has opted out
-    //   or if somehow our defaults are problematic, do nothing
-    if (telemetry === false) return
+    if (telemetry === false) {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via opt-out)`)
+      return
+    }
 
     // don't send telemetry before we inform the user, allowing opt-out
     const telemetryDefaulted = getDefault(telemetry)
     if (!telemetryDefaulted.informedAt) return inform(telemetryDefaulted, userConfig)
+
+    console.log(`Keystone Telemetry is ${g`enabled`}`)
 
     await sendProjectTelemetryEvent(cwd, lists, dbProviderName, telemetryDefaulted, userConfig)
     await sendDeviceTelemetryEvent(telemetryDefaulted, userConfig)
--- a/packages/core/src/lib/telemetry.ts
+++ b/packages/core/src/lib/telemetry.ts
@@ -273,6 +273,13 @@ export async function runTelemetry(
   dbProviderName: DatabaseProvider
 ) {
   try {
+    const { telemetry, userConfig } = getTelemetryConfig()
+
+    if (telemetry === false) {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via opt-out)`)
+      return
+    }
+
     if (ci.isCI) {
       console.log(`Keystone Telemetry is ${r`disabled`} (running in CI)`)
       return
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/telemetry.ts
```diff
@@ -273,24 +273,34 @@ export async function runTelemetry(
   dbProviderName: DatabaseProvider
 ) {
   try {
-    if (
-      ci.isCI || // don't run in CI
-      process.env.NODE_ENV === 'production' || // don't run in production
-      process.env.KEYSTONE_TELEMETRY_DISABLED === '1' // don't run if the user has disabled it
-    ) {
+    if (ci.isCI) {
+      console.log(`Keystone Telemetry is ${r`disabled`} (running in CI)`)
```
dcousens: please use the `log` function used by the rest of file, not `console.log`

please change the prefix, `s/Keystone Telemetry is/telemetry is`

packages/core/src/lib/telemetry.ts
```diff
@@ -273,24 +273,34 @@ export async function runTelemetry(
...
       return
     }
 
     const { telemetry, userConfig } = getTelemetryConfig()
 
-    // don't run if the user has opted out
-    //   or if somehow our defaults are problematic, do nothing
-    if (telemetry === false) return
+    if (telemetry === false) {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via opt-out)`)
```
dcousens: I kind of want this one to be first, so users don't think their opt out is being ignored

packages/core/src/lib/telemetry.ts
abhijnyan-codes: Thanks for the feedback! I noticed the existing log() function only outputs when KEYSTONE_TELEMETRY_DEBUG=1, which would hide these status messages from regular users. Should I use log() anyway, or is there another logging utility I should use instead?

dcousens: ah, true, sorry, I think we want the logging approach used by the rest of the codebase for this - not the `log` from this file

dcousens:
@abhijnyan-codes at the same time as this, did you want to add support to opt-out if the user has `DO_NOT_TRACK=1` in the environment too?

packages/core/src/lib/telemetry.ts
abhijnyan-codes: Done, moved opt-out check to be first.

abhijnyan-codes:
Added DO_NOT_TRACK=1 support as well.

dcousens:
Thanks @abhijnyan-codes, if you can add a changeset, I will resolve the last remaining tidbits

abhijnyan-codes:
Added the changeset!

abhijnyan-codes:
Hi @dcousens , just checking in — is there anything else needed from my side to get this merged? Happy to make any changes.

packages/core/src/lib/telemetry.ts
```diff
@@ -273,31 +273,45 @@ export async function runTelemetry(
...
-    if (telemetry === false) return
+    if (process.env.NODE_ENV === 'production') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (NODE_ENV is production)`)
+      return
+    }
+
+    if (process.env.KEYSTONE_TELEMETRY_DISABLED === '1') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via KEYSTONE_TELEMETRY_DISABLED env)`)
+      return
+    }
```
dcousens, line 296: move this to 2nd

packages/core/src/lib/telemetry.ts
```diff
@@ -273,31 +273,45 @@ export async function runTelemetry(
...
+
+    if (process.env.DO_NOT_TRACK === '1') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (DO_NOT_TRACK is set)`)
+      return
+    }
```
dcousens, line 301: move this to 3rd

dcousens:
two minor nits and then we can merge :heart:

packages/core/src/lib/telemetry.ts
```diff
@@ -273,31 +273,45 @@ export async function runTelemetry(
...
-    // don't run if the user has opted out
-    //   or if somehow our defaults are problematic, do nothing
-    if (telemetry === false) return
+    if (process.env.NODE_ENV === 'production') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (NODE_ENV is production)`)
+      return
+    }
+
+    if (process.env.KEYSTONE_TELEMETRY_DISABLED === '1') {
+      console.log(`Keystone Telemetry is ${r`disabled`} (via KEYSTONE_TELEMETRY_DISABLED env)`)
```
dcousens:
```suggestion
      console.log(`Keystone Telemetry is ${r`disabled`} (KEYSTONE_TELEMETRY_DISABLED is set)`)
```

abhijnyan-codes:
Done @dcousens , pushed the reordering and fixed the log message for KEYSTONE_TELEMETRY_DISABLED

dcousens:
thanks @abhijnyan-codes, sorry I missed this and thanks for the following up my requests

dcousens:
@abhijnyan-codes [I needed to move the `via opt-out` check to _after_](https://github.com/keystonejs/keystone/pull/9850) the `isCI` check, otherwise testing locally kept adhering to my testing environment, not what we are testing

PR #9830, opened 2026-05-02
Fix seeding errors in `exampless/usecase-blog`

envsecure:
## Description
This PR fixes the seeding errors in the `usecase-blog` example reported in #9782.

The issues were:
1. The `Post` schema was missing `status` and `publishDate` fields which are present in the example data.
2. The `content` field (Document type) was receiving a raw string from the example data, causing a GraphQL error.
3. Lack of safety checks when connecting authors during seeding.

## Changes
- Updated `examples/usecase-blog/schema.ts` to include `status` and `publishDate` fields.
- Added `select` to the imported fields in `schema.ts`.
- Updated `examples/usecase-blog/seed-data.ts` to transform the raw `content` string into a valid Document structure `[{ type: 'paragraph', children: [{ text: content }] }]`.
- Added a check in `seed-data.ts` to skip posts if the author is not found, preventing crashes.

## Verification
I have manually verified the schema and seed logic changes. Since this is a fix for an example's seed script, it ensures a better onboarding experience for new users.

Resolves #9782

PR #9835, opened 2026-05-07
fix: regenerate schema artifacts for usecase-blog after #9830

envsecure:
## Description

This is a follow-up fix for #9830.

PR #9830 added `status` (select) and `publishDate` (timestamp) fields to `examples/usecase-blog/schema.ts`, but the auto-generated `schema.prisma` and `schema.graphql` artifacts were not regenerated and committed. This caused the CI `keystone postinstall` check to fail with:

```
Error: Your Prisma and GraphQL schemas are not up to date
```

## Changes

- Updated `examples/usecase-blog/schema.prisma` — added `status String @default("draft")` and `publishDate DateTime?` to the `Post` model
- Updated `examples/usecase-blog/schema.graphql` — added `status` and `publishDate` fields to the `Post` type, `PostWhereInput`, `PostOrderByInput`, `PostUpdateInput`, and `PostCreateInput`

## Root Cause

The generated schema files (`schema.prisma`, `schema.graphql`) are committed to the repo and must always stay in sync with `schema.ts`. When `schema.ts` is modified, `keystone postinstall` must be run locally and the regenerated files committed alongside the change.

Fixes the broken `postinstall` CI check introduced by #9830.

### DIFF
--- a/examples/usecase-blog/schema.graphql
+++ b/examples/usecase-blog/schema.graphql
@@ -141,6 +141,8 @@ input PostRelateToManyForCreateInput {
 type Post {
   id: ID!
   title: String
+  status: String
+  publishDate: DateTime
   content: Post_content_Document
   author: Author
   tags(where: TagWhereInput! = {}, orderBy: [TagOrderByInput!]! = [], take: Int, skip: Int! = 0, cursor: TagWhereUniqueInput): [Tag!]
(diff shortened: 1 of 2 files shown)


dcousens:
Thanks @envsecure, we fixed this in https://github.com/keystonejs/keystone/pull/9848/changes/a404882efcd452729bf1ca65e696097033dee283

PR #9857, opened 2026-06-26
Update nextjs to v16

trickreich:


trickreich:
@dcousens Not sure what the package @preconstruct/next is doing, but next16 is now using turbopack in default mode.

emmatown:
This needed various other changes to what's in this PR, I've done this in #9827

trickreich:
@emmatown thanks for fixing this!

PR #9872, opened 2026-07-13
Prisma 7

emmatown:
This updates Keystone to Prisma 7. This involved a number of changes:

- Instead of accepting `db.url`, `db.shadowDatabaseUrl` and `db.enableLogging`, we just have `db.prismaClientOptions` which is a function that needs to return the Prisma adapter and any other Prisma client options you want to use.
- The `keystone migrate` commands and `--with-migrations` flag have been removed, you should use the Prisma CLI directly instead
- The `keystone prisma` command has been removed, you should use the Prisma CLI directly instead
- The Prisma client is generated into `generated/prisma` by default now. This can still be customised with `db.prismaClientPath`
- `prisma` and `@prisma/client` are now peer dependencies that must be installed alongside Keystone
- `@keystone-6/core/testing`'s `resetDatabase` has been replaced with new implementations in `@keystone-6/core/testing/postgresql`, `@keystone-6/core/testing/sqlite` and `@keystone-6/core/testing/mysql`. Instead of using `prisma db push`, it accepts a path to a migrations directory and runs the migrations
- We no longer use Prisma internals in Keystone itself or our tests, this should make things more reliable, concretely what this means now is besides the removed commands, we directly run the `prisma generate`, `prisma format` and `prisma db push`(this is just for dev) commands instead of using internal functions to do these 
  - There is still one last usage of Prisma internals in the repo for the script to generate GraphQL filter types, I haven't removed that since it hasn't broken in a while and I don't think it's worth rewriting to something like parsing the generated TypeScript types of a Prisma schema


trickreich:
@emmatown what's the plan with this feature? are you going to merge that in near future? will all this dependency updates included in a new major release v8?

emmatown:
Yes, it will be merged very soon and included in the next major

.changeset/remove-keystone-prisma.md
```diff
@@ -0,0 +1,5 @@
+---
+"@keystone-6/core": major
+---
+
+Removes the `keystone prisma` command, use the Prisma CLI directly instead.
```
dcousens:
```suggestion
Removes the `keystone prisma` command, you should use `prisma` directly.
```

.changeset/remove-migration-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Removes the `keystone migrate` commands and the `--with-migrations` option, use the Prisma CLI directly instead.
```
dcousens:
```suggestion
Removes the `keystone migrate` commands and the `--with-migrations` option, you should use `prisma` directly.
```

.changeset/testing-reset-database.md
```diff
@@ -0,0 +1,5 @@
+---
+'@keystone-6/core': major
+---
+
+Replaces `@keystone-6/core/testing`'s `resetDatabase` function with new implementations in `@keystone-6/core/testing/postgresql`, `@keystone-6/core/testing/sqlite` and `@keystone-6/core/testing/mysql`. Instead of using `prisma db push`, it accepts a path to your migrations directory and runs the migrations.
```
dcousens:
```suggestion
Replaces `@keystone-6/core/testing`'s `resetDatabase` function with new implementations in `@keystone-6/core/testing/postgresql`, `@keystone-6/core/testing/sqlite` and `@keystone-6/core/testing/mysql`
```

dcousens (approved):

PR #9883, opened 2026-07-15
Add separate `access` and `omit` controls for list query kinds `.count`, `.many`, `.one`, and field read kinds `.filter` and `.order`

emmatown:
This does two largely seperate changes around access control and `graphql.omit`:

First, you can specify an object in `access.operation.query` and `graphq.omit.query` with `one`, `many` and `count`, this restricts those specific operations. A value directly at `access.operation.query` and `graphq.omit.query` is still allowed. For `access.operation.query`, the function receives a `kind: 'one' | 'many' | 'count'` property in the argument.

Second, `isOrderable` and `isFilterable` have been replaced with `access.read` and `graphql.omit.read` now accepting objects with `item`, `filter` and `order` which restrict those actions. Similar to the `query` object, a value directly at `access.read` and `graphql.omit.read` is still allowed. For `access.read`, the function receives a `kind: 'item' | 'filter' | 'order'` property in the argument and note the `item` property is only available when `kind` is `'item'`. These values can also be defaulted in the list level `fieldDefaults`. This change should make it much harder to accidentally allow filtering/ordering when you don't intend to.


examples/empty-lists/schema.ts
```diff
@@ -92,7 +92,7 @@ export const lists = {
 
       createdAt: timestamp({
         access: {
-          read: allowAll,
+          read: { item: allowAll, filter: denyAll, order: denyAll },
```
dcousens: why?

packages/core/src/lib/core/mutations/index.ts
```diff
@@ -810,24 +810,25 @@ export function getMutationsForList(list: InitialisedList) {
 
   const collectedTypes: GraphQLNamedType[] = []
   const { isEnabled } = list.graphql
+  const hasItemQuery = isEnabled.query.one || isEnabled.query.many
   if (isEnabled.type) {
     // adding all of these types explicitly isn't strictly necessary but we do it to create a certain order in the schema
     collectedTypes.push(list.graphql.types.output)
-    if (isEnabled.query || isEnabled.update || isEnabled.delete) {
+    if (hasItemQuery || isEnabled.query.count || isEnabled.update || isEnabled.delete) {
```
dcousens: Why is `count` needing `uniqueWhere`?

tests/api-tests/auth.test.ts
```diff
@@ -35,8 +35,12 @@ const runner = setupTestRunner({
           name: text(),
           email: text({ validation: { isRequired: true }, isIndexed: 'unique' }),
           password: password(),
-          noRead: text({ access: { read: () => false } }),
-          yesRead: text({ access: { read: () => true } }),
+          noRead: text({
+            access: { read: { item: () => false, filter: denyAll, order: denyAll } },
+          }),
+          yesRead: text({
+            access: { read: { item: () => true, filter: allowAll, order: allowAll } },
```
dcousens:
```suggestion
            access: { read: allowAll },
```
?

dcousens: Do we need to change this?

packages/core/src/lib/system.ts
```diff
@@ -76,10 +76,6 @@ function getInternalGraphQLSchema(config: KeystoneConfig) {
                     return {
                       ...f,
                       graphql: { ...(f.graphql || {}), omit: false },
-
-                      // WARNING: this bypasses access control checks for filtering and ordering on the internal schema
```
dcousens, line 80: Nice to see this is now resolved

tests/api-tests/auth.test.ts
emmatown: fixed

examples/usecase-roles/schema.ts
```diff
@@ -136,13 +107,12 @@ export const lists: Lists<Session> = {
       //   should not be publicly visible
       password: password({
         access: {
-          read: denyAll, // TODO: is this required?
+          read: { item: denyAll, filter: denyAll, order: denyAll }, // TODO: is this required?
```
dcousens:
```suggestion
          read: denyAll,
```

examples/custom-session-invalidation/schema.ts
```diff
@@ -59,13 +63,11 @@ export const lists = {
...
           isRequired: true,
         },
-        // TODO: is anything else required
       }),
 
       // a passwordChangedAt field, invalidates a session if changed
       passwordChangedAt: timestamp({
         access: denyAll,
-        isFilterable: false,
         ui: {
```
dcousens, line 72: add `omit` for the filtering?

packages/auth/src/index.ts
```diff
@@ -45,7 +45,7 @@ export function createAuth<ListTypeInfo extends BaseListTypeInfo>({
   secretField,
   initFirstItem,
   identityField,
-  sessionData = 'id',
+  sessionData = 'id', // WARNING: uses sudo to retrieve this data
```
dcousens: can we add this warning to the Typescript type of `AuthConfig`

dcousens (approved):

dcousens (approved):

PR #9920, opened 2026-07-22
update graphql to 17

trickreich:


emmatown:
Just fyi, this won't happen in the upcoming major because of the number of dependencies we have that don't have peer deps allowing `graphql@17`

trickreich:
@emmatown which one? I think apollo-upload-client is the only one actually

emmatown:
`@apollo/server` mainly. Also `@graphql-ts/schema` but that's for me to resolve

trickreich:
Ah, I see
https://github.com/apollographql/apollo-server/issues/8220
https://github.com/jaydenseric/apollo-upload-client/issues/368

PR #9929, opened 2026-07-23
Move packages to Node ESM

emmatown:
This moves Keystone's packages to exclusively ESM which can be run in Node. They can still be used in CommonJS because of Node adding support for requiring ESM in CommonJS. This also doesn't change anything about what Keystone builds, we still build CommonJS in `.keystone` and currently using `"type": "module"` in a Keystone project will not work. I may fix this in a future PR though if we do support that, we will still compile to CommonJS because using ESM for dev live reloading is difficult to do without memory leaks.

The main benefit here is a faster CI and less complexity.

This also includes a bump to `meow`.


emmatown:
~~This is blocked on bumping to graphql@17 (which is itself blocked on a number of other things) because of cursed Vitest behaviour where because `graphql@16` has `"main": "index"` in it's `package.json`, when `graphql` is imported from ESM (like proper `"type": "module"`/`.mjs`) Vitest resolves to the `index.mjs` instead of the `index.js` that it should resolve to (and Node does) but other CJS deps require graphql and get the `index.js`.~~

~~Note while we could work around this in our Vitest setup, that workaround would also be necessary for consumers of Keystone which I'd rather avoid~~

See https://github.com/keystonejs/keystone/pull/9936

PR #9952, opened 2026-07-27
Serve llms.txt artifacts from the docs site

### LINKED ISSUE #9948: Serve llms.txt artifacts from the docs site
## Summary

Add [llmstxt.org](https://llmstxt.org) support to the docs site (`@keystone-6/website`) so LLM tools can discover and fetch Keystone docs as clean markdown. Three build-time artifacts:

- `/llms.txt` — an index of the docs, grouped by the docs sidebar nav
- `/docs/<slug>.md` — each docs page as raw markdown
- `/llms-full.txt` — every docs page concatenated in nav order

Scoped entirely to the `docs/` package — no changes to `packages/*`, Admin UI, or examples.

## Why it's cheap

- **Nav is already machine-readable:** the Keystatic singleton `content/navigation.yaml` (`navGroups: [{ groupName, items: [{ label, link, status }] }]`), flattened by `getNavigationMap()` in `components/docs/docs-navigation/index.tsx`. Groups → `##` sections, items → `- [label](url): description`.
- **Markdown already on disk:** `content/docs/**/*.md` (55 pages), each with required frontmatter `title` + `description`, read via `reader.collections.docs`.
- **Static-generation precedent:** the sitemap is already emitted at build via `next-sitemap`.

## Design decisions

| Decision | Choice |
|---|---|
| Scope | All three artifacts; **docs only** (blog & examples excluded) |
| Fidelity | **Gated raw Markdoc** — keep `{% hint %}`/`{% well %}`/etc. tags intact; no CommonMark conversion |
| Generation | Static at build (postbuild emitter into `public/`, mirroring `next-sitemap`) |
| URL shape | `/docs/<slug>.md` — append `.md` to the canonical page URL |
| Index links | Point at the `.md` URLs so agents fetch clean markdown directly |
| Full-text order | Nav order (mirrors the index) |

## ⚠️ The one correctness trap

Production hides unreleased content at **render time** (`nextRelease: false` when `SHOW_NEXT_RELEASE` is unset), **not on disk** — the on-disk `.md` still contains `{% if $nextRelease %}` blocks. The existing `removeNextReleaseConditions` helper (`scripts/replace-show-next-release`) does the *opposite*: it's the release-time *reveal* that unwraps and bakes the content in. So it **cannot be reused as-is**. The generator must strip these blocks itself (a small inverse transform, or an AST render with `nextRelease: false`) or it will publish unreleased docs. **This must be covered by a test.**

Related, acceptable for v1: a page whose body is entirely inside a next-release block renders empty in production; its `.md` will likewise be empty while still appearing in the index. Nav `status` (`new`/`updated`) is a badge only — hides nothing, needs no handling.

## Implementation phases

- **P1 — `/llms.txt` index:** read nav groups + docs frontmatter; emit grouped index with `.md` links; H1/blurb from the site's `defaultTitle`/`defaultDescription`.
- **P2 — per-page `/docs/<slug>.md`:** gated raw Markdoc; **includes the next-release stripping transform + test** (the load-bearing piece).
- **P3 — `/llms-full.txt`:** concatenate all gated pages in nav order.
- **Cross-cutting:** wire a `postbuild` emitter into `public/`; confirm the exact `.md` serving mechanism; add to `build`/`validate` as appropriate.

benderham:
Closes #9948.

Adds [llms.txt](https://llmstxt.org) support to the docs site (`@keystone-6/website`), generated statically at build time. Scoped entirely to `docs/` — no changes to `packages/*`, Admin UI, or examples.

## What it adds

A `build:llms` step (wired into `build`, after `sitemap`) that emits three artifacts into `public/`:

- **`/llms.txt`** — an index of the docs, grouped by the sidebar nav (`content/navigation.yaml`). Each item links to the `.md` URL with the page's frontmatter `description` as the blurb.
- **`/docs/<slug>.md`** — every docs page as raw markdown (all 55 pages).
- **`/llms-full.txt`** — every docs page concatenated in nav order.

The generated files are gitignored; they're produced fresh on each build (same model as `next-sitemap`).

## The one correctness detail worth reviewing

Production hides unreleased content at **render time** (the `nextRelease` Markdoc variable is `false` unless `SHOW_NEXT_RELEASE` is set) — it is **not** stripped from the files on disk. The existing `removeNextReleaseConditions` helper does the *opposite* (it unwraps/reveals at release time), so it can't be reused here.

`scripts/llms-txt/strip-next-release.ts` is the inverse: it removes `{% if $nextRelease %}` branches (keeping any `{% else %}`) so the artifacts match what production serves. Covered by unit tests in `strip-next-release.test.ts`.

## Notes

- `.md` files are served from `public/`, which takes precedence over the `/docs/[...rest]` HTML catch-all — verified locally that `/docs/<slug>.md` returns `text/markdown` while `/docs/<slug>` still returns the HTML page (incl. nested slugs).
- Docs not present in the navigation (e.g. `getting-started`, the walkthrough lessons) are listed under an "Additional pages" section rather than dropped, and logged during the build.
- Custom Markdoc tags (`{% hint %}`, `{% well %}`, etc.) are left intact — LLMs parse them fine, and this avoids a conversion layer that would need maintaining. Can revisit if desired.

## Out of scope (follow-up)

Noticed while implementing: `markdoc/load-all.ts` still globs `pages/docs/**` and `pages/blog/**`, which no longer exist after the Keystatic migration to `content/**` — so `remove-conditionals` is currently a no-op. Left untouched here; worth a separate issue.

## Test plan

- [x] `vitest run docs/scripts/llms-txt/strip-next-release.test.ts` (4 passing)
- [x] `tsx scripts/llms-txt/generate.ts` emits `llms.txt`, `llms-full.txt`, and 55 per-page files with no `nextRelease` leakage
- [x] `next dev` + curl: `/docs/<slug>.md` → 200 `text/markdown`, `/docs/<slug>` → 200 `text/html`, `/llms.txt` & `/llms-full.txt` → 200, missing `.md` → 404
- [x] prettier + oxlint clean


docs/scripts/llms-txt/strip-next-release.ts
```diff
@@ -0,0 +1,22 @@
...
+//
+//   {% if $nextRelease %}A{% else /%}B{% /if %}  ->  B
+//   {% if $nextRelease %}A{% /if %}              ->  (nothing)
+//
+// Nested `$nextRelease` conditionals are not supported, matching the existing
+// reveal script's regex approach.
+const pattern =
+  /{%\s+if\s+\$nextRelease\s+%}\s*[^]*?\s*(?:{%\s+else\s+\/%}\s*([^]*?)\s*)?{%\s+\/if\s+%}/g
+
+export function stripNextReleaseConditions(contents: string): string {
```
emmatown: Can you just remove this, we're not using the $nextRelease stuff anymore

emmatown (approved):

PR #10007, opened 2026-08-17
Fix Admin UI remaining editable when graphql.omit.update is set

### LINKED ISSUE #9665: Field can be editable when `graphql.omit.update` is set

JiuYue0820:
## Problem
When `graphql.omit.update` is set on a field, the Admin UI item view can still be editable.

`packages/core/src/lib/core/initialise-lists.ts` currently prefers the configured `ui.itemView.fieldMode` even after update is omitted. If a developer (or a leftover default) sets `fieldMode: 'edit'`, the field is shown as editable even though GraphQL has no update input for it. This is the remaining case of #9665 after #9663/#9664.

## Solution
If `graphql.omit.update` is true and the field is still readable, force `itemView.fieldMode` to `'read'`. An explicit `edit` configuration cannot override the omitted update operation.

## Verification
Added `graphql.omit.update forces itemView.fieldMode to read even if edit is configured` in `tests/api-tests/admin-meta.test.ts`.

Fixes #9665

### DIFF
--- a/.changeset/omit-update-itemview-read.md
+++ b/.changeset/omit-update-itemview-read.md
@@ -1,0 +1,5 @@
+---
+"@keystone-6/core": patch
+---
+
+Force Admin UI item view to read-only when `graphql.omit.update` is set, even if `ui.itemView.fieldMode` is configured as `edit`.
--- a/.changeset/omit-update-itemview-read.md
+++ b/.changeset/omit-update-itemview-read.md
@@ -2,4 +2,4 @@
 "@keystone-6/core": patch
 ---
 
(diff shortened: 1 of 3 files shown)


emmatown (commented):
This would break a dynamic fieldMode changing between read and hidden

JiuYue0820:
Thanks @emmatown. The previous change collapsed any non-hidden fieldMode to a static `read`, which would break a dynamic fieldMode that switches between `read` and `hidden`.

I updated this so `graphql.omit.update` only coerces a static `edit` to `read`. Functions and filter objects are left intact, and a function that resolves to `edit` is still demoted to `read`. Added coverage for both a function and a `{ hidden }` filter.

`tests/api-tests/admin-meta.test.ts` is green locally.

PR #10025, opened 2026-08-30
Feature: Admin meta hook for request-aware meta transformations

MauriceArikoglu:
## Summary
This PR adds the optional ui.hooks.resolveAdminMeta API, enabling request-aware Admin UI metadata transformations such as localization, tenant-specific labels, or branding.

## Functionality
When configured, the hook runs in the adminMeta root resolver after access checks and request-dependent metadata values have been resolved. It receives the full KeystoneContext and a request-local, serializable AdminMeta, supports synchronous and asynchronous returns, and does not expose internal AdminMetaSource fields. It is not invoked for ordinary data GraphQL operations.

## Disclaimer
Without the hook, the existing resolver path, access-control order, GraphQL schema, and Admin UI queries remain unchanged. Request-local metadata reconstruction prevents hook mutations from leaking between requests. Tests cover synchronous and asynchronous hooks, context and headers, session-specific results, invocation behavior, isolation, and error propagation.

### DIFF
--- a/packages/core/src/admin-ui/admin-meta-graphql.ts
+++ b/packages/core/src/admin-ui/admin-meta-graphql.ts
@@ -1,5 +1,4 @@
-import type { GraphQLNames, JSONValue } from '../types/utils.ts'
-import type { ListMeta, FieldMeta, FieldGroupMeta } from '../types/index.ts'
+import type { AdminMeta } from '../types/index.ts'
 import { gql } from './apollo.ts'
 
 export const adminMetaQuery = gql`
@@ -139,32 +138,8 @@ export const adminMetaQuery = gql`
   }
 `
 
-// TODO: duplicate, reference core/src/lib/admin-meta.ts
 export type AdminMetaQuery = {
   keystone: {
-    adminMeta: {
-      lists: (ListMeta & {
-        fields: ListMeta['fields'][string][]
-        actions: ListMeta['actions']
-        groups: (FieldGroupMeta & {
-          fields: FieldMeta[]
-        })[]
-        graphql: {
-          names: GraphQLNames
-        }
-
-        pageSize: number
-        initialColumns: string[]
-        initialSearchFields: string[]
-        initialSort: ListMeta['initialSort'] | null
-        initialFilter: JSONValue
-        hiddenFilter: JSONValue | null
-        isSingleton: boolean
-
-        hideNavigation: boolean
-        hideCreate: boolean
-        hideDelete: boolean
-      })[]
-    }
+    adminMeta: AdminMeta
   }
 }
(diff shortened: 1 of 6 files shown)


MauriceArikoglu:
@emmatown @dcousens Looking for a review. We support localization in our products and this simple hook extension enables ui translations and other request aware admin ui / meta extensions, which, I believe, greatly enhance Keystone's functionality and capabilities. This would enable us to also offer localization based on the same cookie inside keystone (next to tenant & config specific changes).

emmatown:
I think a much more straightforward way to do this would be accepting functions with the context in more places rather than this global API

MauriceArikoglu:
> I think a much more straightforward way to do this would be accepting functions with the context in more places rather than this global API

@emmatown @dcousens

Thanks for the suggestion. I chose a global hook because the main use case is applying request-aware changes across multiple lists and fields (for example localization, tenant-specific labels, or branding, which is currently impossible)

I wanted to keep this change lightweight. Adding context-aware functions to every metadata property would require changes to many public types and resolver paths. The hook adds one small, optional extension point instead. I also believe specific changes like localization benefit from a simple API like my suggested API instead of a fine-grained context-aware field-transformer.

Existing per-property callbacks still work well for local conditions such as field modes, filters, and visibility. The implementation is intentionally localized to the adminMeta resolution path. It does not change the GraphQL schema or existing configurations,

---
Edit: Its also a separation of concerns and deliberatly splits admin meta (presentation layer) and functional layer (field specifics). I dont want to clutter my field definitions with localization semantics if I have context aware logic running in a context specific field function. If your suggestion is a function per field property, I'd suggest not to do that, as one could end up creating a function monster. One "dynamic" function per field seems more suitable imo.

MauriceArikoglu:
@emmatown how do you suggest we move forward? I would love to see this landing - can offer to build extended functionality in a different PR as well

MauriceArikoglu:
Hey @emmatown are you gonna consider this PR for merge or not at all? we are unfortunately required to move forward with the functionality of a translated CMS UI and would prefer not to fork and release our own copy of keystone for this relatively small change.

emmatown:
We don't think this approach makes sense, it also breaks resolving of items in the `itemView` resolvers (we certainly need tests for this). If you want to make a PR where label, description and etc. accept context functions like other APIs and then it could be used like the code below, we'd be happy with that. 

```ts
const translations = {
  name: {
    en: 'Name',
  },
  user: {
    en: 'User',
  },
  users: {
    en: 'Users',
  },
}

function localised(key: keyof typeof translations) {
  return ({ context }: { context: Context }) => {
    const lang = context.session?.lang ?? 'en'
    return translations[key][lang]
  }
}

export const lists = {
  User: list({
    ui: {
      label: localised('user'),
      plural: localised('users'),
      singular: localised('user'),
    },
    fields: {
      name: text({
        label: localised('name'),
      }),
    },
    access: { operation: allowAll },
  }),
}
```

MauriceArikoglu:
@emmatown Reproduced the bug in [2003238](https://github.com/keystonejs/keystone/pull/10025/commits/2003238674b523366c13729b07bd3de9b1c76343) and fixed in [f040643](https://github.com/keystonejs/keystone/pull/10025/commits/f040643c31ea29a4ef57351c6ec1ce3d1da938f6) by changing the implementation as suggested.

PR #10040, opened 2026-09-17
fix(core): run field before/after hooks for defaulted values

### LINKED ISSUE #9382: Field hooks beforeOperation and afterOperation not running with default value
**To reproduce:**

1. Create new Keystone app with the following config:

`
import { config, list } from "@keystone-6/core";
import { allowAll } from "@keystone-6/core/access";
import { text, integer } from "@keystone-6/core/fields";
import { session } from "./auth";

export default config({
  db: {
    provider: "sqlite",
    url: "file:./keystone.db",
  },
  lists: {
    Material: list({
      access: allowAll,
      fields: {
        name: text({ validation: { isRequired: true }, isIndexed: "unique" }),
        order: integer({
          validation: {
            isRequired: false,
            min: 1,
          },
          defaultValue: 1,
          hooks: {
            resolveInput: async () => console.log("resolveInput"),
            beforeOperation: async () => console.log("beforeOperation"),
          },
        }),
      },
    }),
  },
  session,
  ui: {
    isAccessAllowed: () => true,
  },
});
`
3. Create new entity, without changing the default value
4. Observe the stdout having only "resolveInput" being logged

Note: If the value of order field is different from default value, the hooks are firing as expected

**Expectation:**

Fields with default values should not behave any different from fields with default values.

Using:
node  20.18.0
keystone-6/core 6.3.0

Tyagiquamar:
Fixes #9382

## Problem
Field-level \eforeOperation\ and \fterOperation\ hooks never run when a field is omitted from the create/update input but still resolves to a value, eg. an \integer\ field using its \defaultValue\. \
esolveInput\ and \alidate\ hooks run for every field unconditionally, so this is inconsistent: only \
esolveInput\ fires and the side-effect hooks are silently skipped.

## Root cause
In \
unSideEffectOnlyHook\ (\packages/core/src/lib/core/hooks.ts\), field hooks for create/update were gated solely on the field being present in the original \inputData\. A defaulted field is absent from \inputData\ but present in \
esolvedData\ (field input resolvers substitute the default, eg. \packages/core/src/fields/types/integer/index.ts\), so its hooks were skipped even though a value is written to the database.

## Fix
Also run the field hooks when the field resolved to a non-\undefined\ value. One guard in the shared function covers create/update $\times$ before/after. Fields with genuinely no input and no resolved value are still skipped, and delete behaviour (always run) is unchanged.

## Verification (Docker, node:24-bookworm, pnpm 11.21.0)
- New test \packages/core/tests/field-hooks-default-value.test.ts\: fails on pristine code (2 failed, 4 passed), passes with the fix (6/6).
- \pnpm vitest run packages/core/tests/conditional-filters.test.ts packages/core/tests/field-hooks-default-value.test.ts\: 17/17 passed.
- \pnpm lint\ (oxlint): 0 warnings, 0 errors.
- \pnpm test:types\ (tsc): exit 0.
- \prettier --check\ on changed files: clean.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

